PR #1, opened 2016-08-29
Added type method to connection generator stub

alexwhb:
Added type method to stub, which needs to be implemented sense you are using the Connection interface.

### DIFF
--- a/src/Support/Console/Commands/stubs/relay_connection.stub
+++ b/src/Support/Console/Commands/stubs/relay_connection.stub
@@ -39,4 +39,14 @@ class DummyClass implements Connection
     {
         // TODO: Resolve connection
     }
+    
+    /**
+     * Get name of connection.
+     *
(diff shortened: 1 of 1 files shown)


chrissm79:
Whoops, looks like I didn't cut a new release with my latest changes. Checkout v0.7.6 and it should be there. Thanks @alexwhb!

alexwhb:
Oh excellent. I'm loving the direction you are taking this by the way. Looking very clean.

chrissm79:
Thanks @alexwhb! Yeah, I'm much happier with this approach. It's tested, which will make PRs and new features MUCH easier to integrate. Going to implement something like a DataLoader as well, but I think I'll concentrate on docs, cut a 1.0 release them get back to it. Won't totally resolve the N+1 issue you can run into, but it should allow a lot more flexibility.

PR #2, opened 2016-09-07
added the ability to override the default naming of custom connections.

alexwhb:
I'm integrating this project into a work project, and I wanted to have clean connection names. I found that the default names where a bit unwieldy especially if you have your custom connections nested deep in some namespaces. No worries if you don't want to merge this in.

### DIFF
--- a/src/Schema/Field.php
+++ b/src/Schema/Field.php
@@ -53,7 +53,7 @@ class Field
     /**
      * Attach middleware to field.
      *
-     * @param array $attributes
+     * @param array $middleware
      */
     public function addMiddleware(array $middleware)
     {
--- a/src/Schema/Registrars/ConnectionRegistrar.php
+++ b/src/Schema/Registrars/ConnectionRegistrar.php
@@ -112,8 +112,11 @@ class ConnectionRegistrar extends BaseRegistrar
     protected function instanceName($name)
     {
         if ($name instanceof Connection) {
-            $class = (new ReflectionClass($name))->getName();
+            if($name->name() != null){
+                return $name->name();
+            }
 
+            $class = (new ReflectionClass($name))->getName();
             return strtolower(snake_case((str_replace('\\', '_', $class))));
         }
 
--- a/src/Schema/Registrars/ConnectionRegistrar.php
+++ b/src/Schema/Registrars/ConnectionRegistrar.php
@@ -112,12 +112,7 @@ class ConnectionRegistrar extends BaseRegistrar
     protected function instanceName($name)
     {
         if ($name instanceof Connection) {
-            if($name->name() != null){
-                return $name->name();
-            }
-
-            $class = (new ReflectionClass($name))->getName();
-            return strtolower(snake_case((str_replace('\\', '_', $class))));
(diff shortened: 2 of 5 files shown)


chrissm79:
@alexwhb I agree that an override here is a good idea. We could do something like this in the stub and interface:

```php
/**
 * Get the name of the connection.
 * Note: Connection names must be unique
 *
 * @return string
 */
public function name()
{
    return (new ReflectionClass(new static))->getName();
}
```

That way we have a sensible default that can be overwritten if needed and adheres to the interface. Then in the registrar, since we know we have an instance of the interface we can just do:

```php
protected function instanceName($name)
{
    if ($name instanceof Connection) {
        return strtolower(snake_case((str_replace('\\', '_', $name->name()))));
    }
    
    // ...
}
```

_*I need to change the variable name here at some point so it makes more sense... `$name->name()` seems a bit ridiculous :-)_

Update the PR with these changes and I'll merge it in! Thanks for contributing!!!

BTW: [Youshido/GraphQL](https://github.com/Youshido/GraphQL) looks like a great alternative to graphql-php. It's updated to GraphQL April 2016 revision and looks like it's pretty well maintained. I may switch over to this once I've had a chance to test it out and check out the performance differences.

alexwhb:
Excellent! I just added the changes you requested. That's defiantly a cleaner way to solve that problem. Also I took a look at that new GrapQL library. Looks awesome! I love the idea of using a well maintained library behind this especially sense GraphQL is most defiantly in its infancy. I'm excited to see where this project goes. I think it's going to be big. Eloquent and Laravel are way more simple in my opinion than most of the node based implementations. Let me know if you want any more changes made. Also I'd be happy to contribute more if you need help.

chrissm79:
@alexwhb Thanks again for the PR! I was overthinking the default name and just changed it so it copies the name of the connection class.

```bash
php artisan lighthouse:connection UserTasks
```

defaults to:

```php
class UserTasks implements Connection
{
    /**
     * Get the name of the connection.
     * Note: Connection names must be unique
     *
     * @return string
     */
    public function name()
    {
        return 'UserTasks';
    }
   
    // ...
}
```

If you're feeling up for a challenge, you can take a look at this implementation of [graphql-relay-subscription](https://github.com/taion/graphql-relay-subscription) which adds subscription support to GraphQL (works with [relay-subscriptions](https://github.com/edvinerikson/relay-subscriptions)). Just stumbled across this yesterday. Would be great if I could get some help on coming up with a plan to integrate this into Lighthouse! 

I'm going to work on including Unions and then potentially switching over to the Youshido/GraphQL package so it will be awhile before I can even dig into subscriptions. But definitely exciting to see it in the wild finally!!

alexwhb:
Oh ya no worries! Happy to help. Ya overthinking systems is most defiantly something I struggle with as well. I like that solution. It makes the most sense... and is consistent with how everything else works. 

I will take a look at that subscriptions system. I'm going to be on vacation for a week starting this Saturday, so I might not be too active on it for that time, but I'll look into it and come up with a plan. and we can discuss it. 

excellent! I'm excited to see how that turns out. 

what's the best way to have some conversations in regards to subscriptions when come up with an initial plan? should I open an issue?

PR #5, opened 2017-01-03
Update GraphQLType.php

darrenmerrett:
$this->fields() shouldn't be accessed directly, change $this->fields to $this->getFields()


chrissm79:
@darrenmerrett Spot on, thank you!

PR #6, opened 2017-01-09
Fix docs

hailwood:
Helps the ide-helper complete the full namespace for the SchemaBuilder otherwise it ends up saying `GraphQL::schema()` returns an instance of `Nuwave\Lighthouse\SchemaBuilder` instead of `Nuwave\Lighthouse\Schema\SchemaBuilder`


chrissm79:
@hailwood Thanks for this!

PR #19, opened 2017-03-23
Allow setting description on Connections

hailwood:
Currently the connection descriptions are hard coded to `A connection to a list of items.`.

This small update allows you to add a `description(){}` method to your `Connection` to override this.

This is useful in the instance where you have several relations named in a similar fashion where the name itself is not enough to distinguish the purpose.


chrissm79:
Sorry for the delay but thanks for submitting this!

PR #28, opened 2017-04-19
No need to export tests and assets.

lucasmichot:
No need to export tests and assets.


lucasmichot:
Rebased, but StyleCI still seems to be disabled...

PR #33, opened 2017-04-21
Let TravisCI run tests

lucasmichot:
Let TravisCI run tests

---

It would be good to enable TravisCI for this repo, and enable TravisCI PR status too, so that anyone can check that the builds pass

PR #34, opened 2017-04-21
No need to use app()

lucasmichot:
Avoid useless call to `app()`

---

Requires https://github.com/nuwave/lighthouse/pull/33

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -294,7 +294,7 @@ class GraphQL
     public function schema()
     {
         if (! $this->schema) {
-            $this->schema = app(SchemaBuilder::class);
+            $this->schema = new SchemaBuilder;
         }
 
         return $this->schema;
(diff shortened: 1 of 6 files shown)


lucasmichot:
@chrissm79 - I've juste rebased on master after #33 has been merged, but it seems that TravisCI is not running.

Can this be enabled ?

---

screen shot 2017-04-23 at 15 11 08

chrissm79:
@lucasmichot Just updated Travis CI so it should be good to go.

I'm currently working on another project to enable subscriptions (w/ Redis to start) then I intend to clean up some of this code and add more code coverage. i.e.,

```php
if (!$this->schema) {
    $this->schema = app(SchemaBuilder::class);
}

return $this->schema;

// would turn into
return app(SchemaBuilder::class);
```

I want to continue using the IoC container so I can covert most of these to singletons and I can also mock them while testing. I definitely appreciate the PR! But I think in this case I want to keep the code the way it is and just clean up the checks on local instances (and replace them with singletons).

lucasmichot:
Nice ! totally make sense - closing this PR

PR #37, opened 2017-04-23
Let lighthouse also support Laravel 5.3

lucasmichot:
Let lighthouse also support Laravel 5.3


lucasmichot:
Not sure that Codecov.io is has PR status notifications enabled...

lucasmichot:
@chrissm79 Not sure that Codecov.io is has PR status notifications enabled...

PR #42, opened 2017-04-23
Fix typo

lucasmichot:
Fix typo

---

Mostly to check that Codecov.io triggers a diff. status...

Strangely, no status is added for now...


chrissm79:
I've never used Codecov.io before but it looked like I just needed to add the repo (since it's public and utilizing Travis CI) and it would trigger automatically. Will go through the docs to make sure I didn't miss something.

PR #44, opened 2017-04-23
[WIP] Also support Illuminate 5.4

lucasmichot:
Also support Illuminate 5.4


spawnia:
We heavily use some PHP 7 features, so we will not be able to support 5.3

@lucasmichot Can you take a look at #145 and maybe build upon that?

spawnia:
Closing this because of inactivity. Definitely still open to improving our test configuration.

Issue #66, opened 2018-03-05
Lighthouse v2 Roadmap


chrissm79:
## Overview

I'll fill this section out when time permits but I'm currently working on v2.0 of Lighthouse. Spoiler alert, it's a large breaking change that (at the moment) doesn't have an upgrade path. That being said, with the new approach (schema first), the benefits that come with these breaking changes along w/ how much more the end-user can do with it (i.e., plugins) make this a requirement. I want Lighthouse to be approachable to anyone, doesn't require anything but a schema to get started, and provides the user w/ enough flexibility to extend Lighthouse to fit just about any project regardless of complexity.

<img width="601" alt="screen shot 2018-03-04 at 5 22 39 pm" src="https://user-images.githubusercontent.com/1976169/36952734-f6c71840-1fd0-11e8-9a58-1a8e38cf61e8.png">

By defining the schema above, Lighthouse can: 

- Create and register `User` and `Task` types
- Auto resolves the `User` hasMany task relationship
- Auto resolves the `Task` belongsTo User relationship
- Create a `me` query that requires and returns the authenticated user
- Generates a `createUser` mutation that will autofill and create a new `User` w/ rule validation on the `email` argument
- Generates a `createTask` mutation that will inject the authenticated user's `id` into the `user_id` argument (currently trying to figure out how to remove this argument from the schema the client sees). Mutation requires that the user `can` create tasks via a policy

That's a lot it can do for you w/ only 31 lines and no code!!!

### Road to 2.0

- [x] Directive system that the user can extend to resolve Nodes/Fields/Args
- [x] Context argument injector
- [x] create/update/delete autofill directives
- [x] Query/Mutation middleware
- [x] Ability to add scopes to filter relationships
- [x] Use `graphql-php` built in `Deferred` class to handle relationships and prevent N+1
- [x] Cache schema AST
- [x] Ability to attach single directive to Argument/Field/Node
- [x] Create Walkthrough video
- [x] Update Documentation

*For those who want to check out the example project that the Walkthrough will be based off of, [here it is](https://github.com/nuwave/lighthouse-todo-example)! I'm going to try and work in as many built-in directives as possible to highlight what Lighthouse can do "out of the box" as I prep for the v2.0 release.*

### After 2.0
- [ ] Stubs for GraphQL types
- [ ] Auto create/update/delete mutation generation for models (boolean in `@model` directive)
- [ ] Custom Context - Allow creation of custom context when request comes in
[...]

PR #68, opened 2018-03-20
Installation error fix.

kikoseijo:
Got an error after install, 

My suggestion its to publish a demo file if not found with the user model with a login and a create mutation, maybe also a relationship as you got in the v2 roadmap.

Have also added the package auto discovery functionality including the Alias, I think its cool feature.

Does project needs anything else than a initial .graphql file to make a working demo? 
Maybe 1 .graphql with a couple imports?

BTW: as far as I got, your code its crystal clear, congrats!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,14 @@
         "test" : "vendor/bin/phpunit --colors=always",
         "test:ci": "composer test -- --verbose --coverage-text --coverage-clover=coverage.xml"
     },
+    "extra": {
+      "laravel": {
+          "providers": ["Nuwave\\Lighthouse\\Providers\\LighthouseServiceProvider"],
+          "aliases": {
+            "graphql": "Nuwave\\Lighthouse\\GraphQL"
+        }
+      }
+    },
     "config": {
         "sort-packages": true
     }
--- a/src/Schema/Utils/SchemaStitcher.php
+++ b/src/Schema/Utils/SchemaStitcher.php
@@ -44,7 +44,12 @@ class SchemaStitcher
      */
     protected function appSchema($path)
     {
-        $schema = file_get_contents($path);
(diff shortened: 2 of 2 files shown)


kikoseijo:
Regarding initial installation, i suggest to have a publish method (im good at this) and provide use with a query (classical viewer could be good), couple mutations (login + create user), Custom directive (maybe they example regarding related task), missing something?

What you think? Sometimes a working example its better than documentation.

SimonErich:
I think that's an awesome idea.
A good example says a lot of already, without having to write 1000 lines of documentation. Especially if the code changes quickly.
If something doesn't work, I can just check if the example works and copy the approach used there.

chrissm79:
@kikoseijo Thanks for PR!! And an example is a great idea! I'll base it on the walkthrough I created for v1. I'll try to get it up and running tonight or tomorrow afternoon and I'll follow up w/ a link here (as well as #66)

kikoseijo:
I suggest creating the folder structure, based on whats needed now, I guess connections and types folder are not needed, because its tedious to have to create all this folders and if we get an example its easier to duplicate files than got to the docs and find the one. (Probably easier also to maintain them example updates if they belong to the actual project)

Also, if you can provide some working examples ill be able to advance in the implementation a bit faster.

Another thing I end up doing its having a middleware to log request so I can move app development into postman or the playground. Maybe this is something can be embedded as a plus, same thing with the coors, its a must do most of the times.

Doing a bit brainstorm, sorry for the mess of ideas.

BTW: I'm watching this repo, post things anywhere you want, ill get notified.

chrissm79:
@kikoseijo @SimonErich Here's the first iteration of the example project. I'll likely make some updates to it but this should get you started!

PR #71, opened 2018-03-27
add testcase for belongsTo directive with two different relations

latik:
Hey @chrissm79 ,
I found issue with belongsTo directive: if use more than one(I test with two) relation in query only last relation is queried and result used in other relation. 
In PR I add test with two different relations and belongsTodirective that failed at this moment.
 
Thank you for your awesome job.


chrissm79:
@latik Thanks for this! It's almost certainly because I'm resolving the BelongsTo directive as a singleton instead of giving each instance it's own unique name (I had the same issue w/ the HasManyDirective). I'll take a look tonight and push a fix but thanks for the feedback!!

chrissm79:
@latik Thanks again for catching this!! The latest `v2-beta.3.8` fixes the bug!

PR #75, opened 2018-04-10
Issues saver

kikoseijo:
I think its convenient to have this on users control.
Otherwise they will have to look inside your service provider or router file to find out what its meant to be happening.
If you have another reason to not do, just discard the pull request.


chrissm79:
@kikoseijo Thanks for this!!

PR #76, opened 2018-04-10
Lumen support

kikoseijo:
With this Lumen does not break on getting the context.
@auth directive can't be used.
instead of using `\Auth::user()` it could be obtained from the `$request->user()`.

Im going to add full auth with passport on the lumen demo, you can take from there what you need for your documentation, no need to ask.

NOTE: tried to split the commits but obviously I forgot to roll back before last!

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -12,10 +12,13 @@ return [
     | setup more endpoints, ej: pointing to the controller value inside your route file
     |
     */
+    'route_name' => 'graphql',
+
     'route' => [
-        'prefix' => 'graphql',
+        'prefix' => '',
         // 'middleware' => ['web','api'],    // [ 'loghttp']
     ],
+
 
     /*
     |--------------------------------------------------------------------------
--- a/config/config.php
+++ b/config/config.php
@@ -3,6 +3,22 @@
 return [
     /*
     |--------------------------------------------------------------------------
+    | LightHouse endpoint & middleware
+    |--------------------------------------------------------------------------
+    |
+    | Setup this values as required,
+    | default route endpoints its yourdomain.com/graphql
+    | setup middleware here for all request,
+    | setup more endpoints, ej: pointing to the controller value inside your route file
+    |
+    */
+    'route' => [
+        'prefix' => 'graphql',
+        // 'middleware' => ['web','api'],    // [ 'loghttp']
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | Directive registry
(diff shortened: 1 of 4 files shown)


kikoseijo:
BTW: 2 helper methods are missing in order to make Lighthouse to work out of the box with lumen, maybe its good if we have it directly on your helper methods.

Here there are:

```php
if (!function_exists('config_path')) {
    function config_path($path = '')
    {
        return app()->basePath() . '/config' . ($path ? '/' . $path : $path);
    }
}
```

```php
if (!function_exists('app_path')) {
    function app_path($path = '')
    {
        return app()->basePath() . '/app' . ($path ? '/' . $path : $path);
    }
}
```

chrissm79:
@kikoseijo Looks great, and I'll add those helper methods myself. Thanks for pointing that out!!

kikoseijo:
@chrissm79 sorry, I forgot how to use the Input types:

```
input CreateUserInput {
  name: String!
  email: String!
  password: String!
  password_confirmation: String!
}
```

Where we able to add validation to this also?

right now I'm using embedded on a field like this:

```graphql

input ContactInput {
  tipo: String!
  nombre: String!
  email: String!
  comentarios: String
  telefono: String
  carId: String
  fecha: String
  modelo: String
  matricula: String
  km: Int
  codigo_postal: Int
  fabs: [String]
  fuels: [String]
  precio: Int
  copyMe: Boolean
}

type Aviso @model {
  id: ID! @globalId
  name: String!
  email: String!
  comments: String!
}

type ContactPayload {
  success: Boolean
  msg: String
  payload: String
}

extend type Mutation {
  contactForm(payload: ContactInput!): ContactPayload
    @interaction(class: "App\\Interactions\\FormRequest")
}
```

kikoseijo:
JAJA,  @chrissm79 

My interaction does validation and resolve on the fly, you seen it:  
[https://github.com/kikoseijo/lumen-lighthouse-graphql/blob/master/app/GraphQL/Directives/InteractionDirective.php](https://github.com/kikoseijo/lumen-lighthouse-graphql/blob/master/app/GraphQL/Directives/InteractionDirective.php)

I was trying to substitute `contactForm(payload: ContactInput!)` for something that works in the way of `contactForm(name: String!,...etc)` from the Input Field, without embedded.

But you know, was thinking that wrapping into a field like payload its more convenient to write Relay queries...using `types` makes it faster and easy to implement.

We are all having fun with your v2 baby!!!!

PR #82, opened 2018-04-13
[Fix] Changed @paginate to use name relay instead of connection

olivernybroe:


chrissm79:
@olivernybroe didn't even realize that, thanks!!

PR #83, opened 2018-04-13
 [feat] Support for scopes in @paginate

olivernybroe:
Most of the work was already done for adding scopes in paginate, just needed an explode and trimming

### DIFF
--- a/src/Schema/Directives/Fields/PaginateDirective.php
+++ b/src/Schema/Directives/Fields/PaginateDirective.php
@@ -126,12 +126,15 @@ class PaginateDirective implements FieldResolver
      */
     protected function getScopes(FieldValue $value)
     {
-        return explode(
-            ',',
-           $this->directiveArgValue(
-                $this->fieldDirective($value->getField(), $this->name()),
-                'scopes',
-                []
+        return array_map(
+            'trim',
+            explode(
+                ',',
+                $this->directiveArgValue(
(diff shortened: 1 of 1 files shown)


chrissm79:
@olivernybroe Thanks for submitting this! However, the `scopes` arg should already provide an array. If you have the following:

```graphql
type Query {
  posts: [Post!]! @paginate(model: "App\\Post" scopes:["firstScope", "secondScope"])
}
```

Is that not working for you? If not then that's certainly a bug, but the [getScopes](https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/Fields/PaginateDirective.php#L129-L133) method should return the scopes in an array form already.

chrissm79:
Ah, I think I know what the problem is, are you setting the `scopes` argument to a string?

```graphql
type Query {
  posts: [Post!]! @paginate(model: "App\\Post" scopes:"firstScope, secondScope")
}
```

If so that would be the problem, that should be an array:

```graphql
type Query {
  posts: [Post!]! @paginate(model: "App\\Post" scopes:["firstScope", "secondScope"])
}
```

olivernybroe:
@chrissm79 Yep just realised that myself, my bad! :) 
I had tried doing it as an array, but did it like this `"[bla, bla]"`. Had no idea that we could define them like this `["firstScope", "secondScope"]`.

PR #86, opened 2018-04-18
Added resolver helper method.

kikoseijo:
Fix for Lumen.


kikoseijo:
And make it pretty avatars.


screen shot 2018-04-18 at 20 00 14

kikoseijo:
Im I doing right sending to master? you able to then move as you wish?, this being a fix should be straight to master, but on features.. ¿ as I only see master or v2.

Issue #93, opened 2018-04-27
[Feat] Scout support


olivernybroe:
Thanks for supplying query to args directives, really helps alot!

So I am creating an Arg directive called `@search` for adding support for Laravel scout (fulltext search).

Is this something you would want in your package or should I make an extension for it?

olivernybroe:
Yep, it's a single file. Just submitted a pull request for some changes which are needed for implementing this.

Hmm yeah we could add it in documentations, here is a gist with my directive.
https://gist.github.com/olivernybroe/2d2b164851b4fc7d4b62df8f37bfbb2a

chrissm79:
@olivernybroe, love the idea and yes that would be awesome to submit a PR for it!! let me know if you have any questions or if there's anything I can help with!

olivernybroe:
@chrissm79 I added a PR for some changes needed, however I didn't add the directive. 
The reason why I didn't add the directive is because Scout requires some Laravel features which we don't have in our package, and it would tie Lighthouse to Laravel and make it incompatible with Lumen.

I can add it without tests, or I can create a package so it can easily be added.
(not that great at composer, so it might be possible to add the tests with more knowledge there)

4levels:
Hi @olivernybroe,

I'll be looking into this since I have some experience with testing / porting Laravel libraries to Lumen.
I'm currently not very active with Lighthouse (I'm focussing on the app counterpart that actually consumes the Lighthouse features), but I expect to have time for this in the next couple of weeks..

Kind regards,

Erik

4levels:
Upon a first look, it pitty it that Scout seems to be configured with Algolia (a paid subscription service - only free for open source projects) only.  They claim it should be easy to integrate other search indexing services, just sad to see that ElasticSearch is not the preferred interface by default.

I do have some experience with ElasticSearch (and eg. Apache Solr) so I'll take it as a challenge to make Scout work with Lumen AND ElasticSearch instead of Algolia.

olivernybroe:
For the test in lighthouse, we could just use fakes or maybe use TNT search? Just didn't felt like adding the directive without tests for it.

chrissm79:
[...]

PR #94, opened 2018-04-27
More flexible directives

olivernybroe:
- Changed so pagination now uses `pageresolvers`, so we don't have to write default parameters in our code. (This also gives more flexibility when creating custom directives)
- Changed so query filters now uses the returned query as the new query. This gives us a lot of flexibility, now we can even return a whole other query, with a filter directive. 

These changed was also required before I could implement a Scout directive, or at least to implement it really easy. #93

### DIFF
--- a/src/Schema/Directives/Fields/PaginateDirective.php
+++ b/src/Schema/Directives/Fields/PaginateDirective.php
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Schema\Directives\Fields;
 
+use Illuminate\Pagination\Paginator;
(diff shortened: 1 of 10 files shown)


chrissm79:
@olivernybroe, this is a pretty good idea... however, does it cause any issues when you have different nested counts on different fields? i.e.

```graphql
{
    me {
        posts(count: 10, page: 2) { # page 2
            #...
            comments(count: 10) { # page 1
                # ....
            }
        }
    }
}
```

chrissm79:
Thanks @olivernybroe!!!

Issue #88, opened 2018-04-23
Subscriptions


chrissm79:
Currently, this is just a placeholder for discussions. But as a quick progress report, I have gotten started on subscriptions this weekend and it's coming along quite nicely! Right now I have 2 variations in mind:

1. Websockets (`lighthouse-subscriptions-ws`) - The main goal of this package is to keep everything in PHP land. The WS server will be built with [Ratchet](http://socketo.me/). I've tested some things out and I'm able to get Playground to connect/disconnect to a Ratchet server successfully (using the same methods as [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws)). The next step is to start pushing down messages from the server. Since a subscriber can subscribe to multiple subscriptions (lol), I need storage. So it will launch w/ InMemory and Redis storage out of the box, but it can also be extended (i.e., MySQL storage). Because of how far along I am, this will likely be the first package released.

2. Pusher (`lighthouse-subscriptions-pusher`) - Maintaining an websocket server for a massive project doesn't sound fun to me... so I want to include at least one third-party solution. [graphql-ruby](https://github.com/rmosolgo/graphql-ruby) has a great solution for this, however, it does rely on a custom Link for Apollo to work with it. Relay Modern is mentioned too, but I haven't dived into the JS parts of that yet.

The coolest thing (I think) is the fact that I can create these packages completely separate from the main `lighthouse` repo by leveraging directives! So hopefully, the community will be able to create some great alternatives if you use a different 3rd party WS provider or if you have a solution that relies on something other than Ratchet (i.e., Node). 

I currently don't have a timeline for this, it just being done after work hours and on the weekends but this weekend was pretty productive so I'm really hoping it won't be too long before I can get this in everyone's hands!

**Initial Beta Release of [lighthouse-ws](https://github.com/nuwave/lighthouse-ws)**

hailwood:
Hey @chrissm79 completely half baked thought here that is probably entirely down the wrong track, but is there any way to leverage Laravel echo as the websocket provider?

Also, what I'm most curious about here is what api you're looking at implementing for these in terms of the .graphql schema?

4levels:
Hi @chrissm79,

Subscriptions is indeed my last number one feature for lighthouse, great to hear you're making progress! Since I currently have already a Node ws server running I'd be happy to see where I can help to get this up and running with it.

Redis is also my preferred store for queues and websocket storage..

[...]

PR #95, opened 2018-04-30
Implement WebSockets Subscriptions

TGNThump:
Full Subscriptions Implementation #88 

The only code needed for a Subscription is:
```php
<?php

namespace App\GraphQL\Subscriptions;

use Illuminate\Queue\SerializesModels;
use Illuminate\Contracts\Broadcasting\ShouldBroadcast;
use Nuwave\Lighthouse\Support\Schema\GraphQLSubscription;
use Illuminate\Broadcasting\Channel;

class ChatMessageSubscription extends GraphQLSubscription
{
    public $message;

    public function __construct($message)
    {
        $this->message = $message;
    }

    public function broadcastAs()
    {
        return 'onChatMessage';
    }


  public function resolve(){
	return $this->message;
  }
}
```
with schema
```graphql
type Mutation{
	chatMessage(message: String): Message
		@create(model: "App\\Models\\ChatMessage")
		@inject(context: "user.id", name: "user_id")
		@event(fire: "App\\GraphQL\\Subscriptions\\ChatMessageSubscription")
}

type Subscription{
	onChatMessage: Message
}

type Message{
	id: ID!
	user: User
	message: String
	updated_at: DateTime
	created_at: DateTime
}
```
and the following broadcasting config. You also need to run `php artisan queue:work` and `php artisan websocket:init`
```php
'graphql' => [
    'driver' => 'lighthouse',
    'port' => 8080,
]
```

### DIFF
--- a/assets/introspection-4.12.txt
+++ b/assets/introspection-4.12.txt
@@ -2,9 +2,7 @@ query IntrospectionQuery {
     __schema {
       queryType { name }
       mutationType { name }
-#      Currently this package does not support subscriptions...
-#      but maybe in the future!
-#      subscriptionType { name }
+      subscriptionType { name }
       types {
         ...FullType
       }
--- a/composer.json
+++ b/composer.json
@@ -18,12 +18,14 @@
         "source": "https://github.com/nuwave/lighthouse"
     },
     "require": {
+        "cboden/ratchet": "^0.4.1",
         "illuminate/contracts": "^5.4",
         "illuminate/http": "^5.4",
         "illuminate/pagination": "^5.4",
         "illuminate/routing": "^5.4",
         "illuminate/support": "^5.4",
         "opis/closure": "^3.0",
+        "ratchet/pawl": "^0.3.1",
         "webonyx/graphql-php": "^0.11.5"
     },
     "require-dev": {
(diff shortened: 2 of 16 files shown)


TGNThump:
I don't know how to stop Laravel from throwing errors when Passport isn't installed, even though it's an optional argument of the service container.

chrissm79:
@TGNThump Thank you for submitting this, this is pretty huge!! I'll dig into it this afternoon, but just a head's up, I have a separate repo for subscriptions and I'm really close with a different implementation. The primary difference is allowing for different storage backends (in-memory, redis, or custom), and allowing for different auth guards (passport, token or custom).

I'll see how much of this I can reuse in my repo and I'll **try** to get it posted this afternoon and provide the link here so this PR can move over.

But what's great about this is the fact that we can have multiple solutions with  directives (I do need to make some tweaks to this repo to get subscriptions to work properly).

Thanks again @TGNThump!!

TGNThump:
Awesome, I wrote it for my project (ergo only supporting Passport and Queue), but if any of it is useful for the open source version, that's even better!

TGNThump:
Closed as of the beta release of https://github.com/nuwave/lighthouse-ws

PR #97, opened 2018-05-04
[Refactor|Feat] New directives and some cleanup and small improvements

olivernybroe:
While I was fiddling with trying to add Scout support I made some other improvements.

- Added Scout Directive
- Cleanup imports
- Fixed spelling errors
- Use Laravel route loader
- Changed GraphQL execute method to not deprecated version
- Moved macros for Collection to a separate class for easier maintenance.
- and some other stuff ;)

Resolves: #93

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,8 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.10",
         "illuminate/validation": "^5.4",
+        "laravel/scout": "^4.0",
+        "mockery/mockery": "dev-master",
         "orchestra/testbench": "^3.6",
         "phpunit/phpunit": "^7.0"
     },
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -114,7 +114,7 @@ class GraphQL
     {
         $schema = $this->graphqlSchema ?: $this->buildSchema();
 
-        return GraphQLBase::executeAndReturnResult(
+        return GraphQLBase::executeQuery(
             $schema,
             $query,
             $rootValue,
--- a/src/Providers/LighthouseServiceProvider.php
+++ b/src/Providers/LighthouseServiceProvider.php
@@ -36,9 +36,11 @@ class LighthouseServiceProvider extends ServiceProvider
 
         $this->app->alias('graphql', GraphQL::class);
 
-        $this->commands([
-            \Nuwave\Lighthouse\Support\Console\Commands\CacheCommand::class,
-        ]);
+        if ($this->app->runningInConsole()) {
+            $this->commands([
+                \Nuwave\Lighthouse\Support\Console\Commands\CacheCommand::class,
+            ]);
+        }
     }
 
     /**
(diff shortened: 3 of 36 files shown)


olivernybroe:
Sorry, messed up, forgot to make a new branch. Hope it's okay, but now scout directive is part of this pull request

chrissm79:
@olivernybroe no worries on the branch, I'll try to get this reviewed and merged this afternoon. Thanks!!!

chrissm79:
Thanks @4levels! I'll revert that back!

chrissm79:
@4levels If you have any easy way to test, could you confirm the latest from `master` fixes the issue (I just reverted back to `realpath`)? Thanks!

4levels:
Hi @chrissm79,

I've tested this locally by modding the file in my local vendor directory and indeed, restoring the realpath call does make the issue go away.
Since this is a confirmed bug in Lumen, I don't think this PR here should be reverted beacuse the changes in this PR are definitely worth it.  I'm still trying to work around this atm and I'm sure I can come up with a better approach so we can still keep the added value of the Laravel route loader and keep Lumen in play as well.
I'm thinking of extending the AppServiceProvider class in the LighthouseServiceProvider (instead of the Lumen/Support/ServiceProvider) so we can apply the workaround as suggested in the Lumen issue, but I'm still investigating if there's always an AppServiceProvider present and if this is a good aproach at all.

I'll keep you posted here, thanks for your quick response!

Kind regards,

Erik

4levels:
Ok, so when I do extend the App\Provider\AppServiceProvider class (after applying the workaround as discussed here https://github.com/laravel/lumen-framework/issues/746) it does work as well, but I'm still not sure if this is a good way to handle this.  Can you shed some light on this?

This is what I've changed in Lighthouse/Providers/LighhouseServiceProvider
```diff
<?php

namespace Nuwave\Lighthouse\Providers;

use Illuminate\Support\Collection;
--- use Illuminate\Support\ServiceProvider;
+++ use App\Providers\AppServiceProvider;
use Nuwave\Lighthouse\GraphQL;
use Nuwave\Lighthouse\Support\Collection as LighthouseCollection;

--- class LighthouseServiceProvider extends ServiceProvider
+++ class LighthouseServiceProvider extends AppServiceProvider
{


```

chrissm79:
@4levels Sorry, I think I caused come confusion when I mentioned "revert"... I just meant that I would revert that [single line of code](https://github.com/nuwave/lighthouse/commit/1dd7b386d0e1e943b16dd2f941e163e17b04687c), but the rest of the PR would stat intact. 

As for extending the `AppServiceProvider`, that's really odd that it would work. Does Lumen include some code in the provider? In a brand new Laravel project it's empty:

```php
namespace App\Providers;

use Illuminate\Support\ServiceProvider;

class AppServiceProvider extends ServiceProvider
{
    /**
     * Bootstrap any application services.
     *
     * @return void
     */
    public function boot()
    {
        //
    }

    /**
     * Register any application services.
     *
     * @return void
     */
    public function register()
    {
        //
    }
}
```

4levels:
Hi @chrissm79,

That looks identical to the service provider I currently have in Lumen. I'm still not sure though if extending this serviceprovider is a good idea from a library perspective.
Maybe a a simple method_exists and a fallback to the regular require if not could do the trick as well..

I'll be only able to further investigate as of tomorrow.

Kind regards and thanks again for the swift follow up!

Erik

olivernybroe:
Maybe we can use something like this https://github.com/irazasyed/larasupport for adding the support? (I haven't use this package before)

PR #104, opened 2018-05-07
Use $user from Context instead of app('auth')

TGNThump:


chrissm79:
@TGNThump I originally was going to go this way but the main issue is that the `Context` can be customized and the user may choose to set the authorized user as `$context->auth` instead of `$context->user`... I didn't really think about it until I created context drivers for the subscriptions package. I think the `app('auth')` should probably stay to keep things safe.

TGNThump:
@chrissm79 Then I'd suggest allowing the user to customize what guard lighthouse pulls the user from.

chrissm79:
@TGNThump That sounds like a good idea! I can add a `guard` argument to the `auth` directive

spawnia:
Seems like this PR is superseded by the changes in https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/Fields/AuthDirective.php

PR #107, opened 2018-05-09
Enable GET requests to the graphql endpoint when app is in debug mode

hailwood:
Being able to hit the graphql endpoint with a get request is extremely useful for debugging

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -7,12 +7,15 @@ return [
     |--------------------------------------------------------------------------
     |
     | Setup this values as required,
-    | default route endpoints its yourdomain.com/graphql
+    | default route endpoints is yourdomain.com/graphql
+    | get requests default to only being enabled in debug mode
     | setup middleware here for all request,
     | setup more endpoints, ej: pointing to the controller value inside your route file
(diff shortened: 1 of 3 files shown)


TGNThump:
I'd say make this an off-by-default configurable option, or maybe add other options. I run graphql playground as the get request for that endpoint, for example.

hailwood:
@TGNThump it's now off by default.

@olivernybroe I've added a test, unfortunately without creating an entire other test case for a different `getEnvironmentSetUp` there's no way for me to test that you *can't* do a get request if the config key is false and I don't believe it's a big enough issue to justify a new test case. As such we enable the GET option for this test alone and just test it works.

olivernybroe:
@hailwood You can do it with ´app('config')->set('lighthouse.route_enable_get', false);` in the test itself :)

hailwood:
@olivernybroe I actually tried that originally, unfortunately that's too late as by then the server provider that loads the route file has already booted.

chrissm79:
Thanks for submitting this @hailwood!!

PR #108, opened 2018-05-09
[Fix] Fixed a bug with Lumen support

olivernybroe:
Fixed a bug with lumen not having support for loadRoutesFrom method.


chrissm79:
Thanks for providing this @olivernybroe!!

Issue #112, opened 2018-05-12
Need a directive for a single model query


hailwood:
Right now if we're doing a query for multiple models we have the fantastic `@paginate` directive.
Unfortunately I cannot find anything similar for a single model.

Do we need to add a new `@find(model: "App\\User")` directive to allow the following

```gql
type Query {
    users(name: String @where(operator: "like"), id: ID @in): [User!]! @paginate(model: "App\\User")
    user(id: ID @eq, email: String @eq): User @find(model: "App\\User")
}
```

olivernybroe:
Can't you just use the node endpoint and then use the global ID?

olivernybroe:
Yep, got it. So should the directive be called find or should it be called first?

hailwood:
I think that depends on whether we want to throw an error if the query would return more than one result? To me I feel it would likely be an error if the built query wasn't specific enough to nail just one result?

olivernybroe:
Maybe we should have both? Both of them should be easy to make, and I can see a use case for both of them. So find requires the result to only return 1 row, while first just takes the first result.

kikoseijo:
Hummm,

Im a bit amazed, seems we want to be able to do all with a single schema definition?

Don't know about you, but when you can resolve a query using a repository, a class method, or ... isn't better to have a Eloquent repository and use one of your hand made methods to resolve this queries?

> keep it simple and elegant!

Just some thoughts.

 🤯

olivernybroe:
@kikoseijo What some of us is aiming for is basically defining our whole query with directives. And when we have more advanced queries, then have a custom scope to handle that. 
However this package does not limit us to one approach which I like a lot, so we can easily use a repository pattern or just use the query builder directly.

I got a big system which I have been able to easily translate to graphQL by using scopes and the directives, so I am a big fan of query builder manipulation approach. ( I have only translated the fetching of data, not inserting/updating)

robsontenorio:
Nice, but i cant see that (@find) on docs.

PR #114, opened 2018-05-13
Feat: Added support for find directive

olivernybroe:
Resolves: #112


chrissm79:
@olivernybroe This looks great! I'll try to review it and get it merged in this afternoon. 

It also looks like I could refactor the `@can` directive a bit to allow the user to lock down these queries (right now, it just get's the model class name from the `$root` argument, but it can be adjusted to pass in the model itself so users can possibly avoid hitting the DB).

Issue #53, opened 2017-05-10
[FEATURE REQUEST]: Add message support for validation.


joshmanders:
Be cool if we could have a `messages` array that can go with `rules` on mutation args so we can define what we want the message to be for that validation.

chrissm79:
Plan on allowing custom validators on mutation fields, just finding the best way to do so. I would love to re-use Form Requests but not sure how to pull that one off at the moment.

spawnia:
Implemented as of #176, will be available in v2.2

Issue #81, opened 2018-04-13
[Feat] Error message from exceptions


olivernybroe:
Could be nice with an interface to define custom rendering of exceptions just like Laravels `Illuminate\Contracts\Support\Responsable` interface.

olivernybroe:
@kikoseijo Oh If I can overwrite the error function then I can prob. just implement it myself. 

But what I mean is in Laravel when I implement the Responsable interface, Laravel will render my exception accordingly to the `toResponse($request)` method. 
I for example like throwing exceptions when something is wrong and then handling them in the exceptionshandler and render them to the front end with the `toResponse($request)` method. 

So my proposal is to have an interface like this, but just for rendering the exception for the errors array in graphql.

kikoseijo:
I wonder if this is posible, 
but i guess the Graphql has a try-catch and does its own thing.

Maybe this helps you:

https://github.com/nuwave/lighthouse/blob/882a1b21508f60fcadb21ec7b8bdc320a270a649/src/Support/Traits/CanFormatError.php

olivernybroe:
@spawnia that is another problem, but not the one I am trying to solve here. I merely won't to implement a interface on my exception class, and if this interface exists, then a specific method on the exception tells how to render the exception in the errors array on the output.

spawnia:
@olivernybroe can this be closed as of #235 ?

PR #118, opened 2018-05-16
[Feat] Changed Arg directives to use a pipe

olivernybroe:
I have changed Arg directive to use a Laravel Pipeline instead.
This give us the power to actually define if a arg directive should be executed last or if any directive after it should be run also, just like when using Laravel's middlewares.

- Arg middleware uses pipeline
- Node middleware uses pipeline
- Field middleware uses pipeline
- Custom exception handler
- Customisable exceptions
- Custom validation messages when using validate directive
- Validation errors are converted to graphql errors
- Changed controller to single action controller by using __invoke method.
- Added advanced filter method, works like this (https://www.graph.cool/docs/reference/graphql-api/query-api-nia9nushae/#filtering-by-field)

## Advanced Filter
When giving the following schema
```
type User @model @filter{
    name: String!
}

type Query {
   id: ID!
   users(filter: UserFilter @filter): [User!]! @paginate(model: "User", type: "relay")
}
```
which will autogenerate the following for UserFilter
```
input MetaFilter {
  # logical operators
  AND: [MetaFilter!] # combines all passed `MetaFilter` objects with logical AND
  OR: [MetaFilter!] # combines all passed `MetaFilter` objects with logical OR

  # ID filters
  id: ID # matches all nodes with exact value
  id_not: ID # matches all nodes with different value
  id_in: [ID!] # matches all nodes with value in the passed list
  id_not_in: [ID!] # matches all nodes with value not in the passed list
  id_lt: ID # matches all nodes with lesser value
  id_lte: ID # matches all nodes with lesser or equal value
  id_gt: ID # matches all nodes with greater value
  id_gte: ID # matches all nodes with greater or equal value
  id_contains: ID # matches all nodes with a value that contains given substring
  id_not_contains: ID # matches all nodes with a value that does not contain given substring
  id_starts_with: ID # matches all nodes with a value that starts with given substring
  id_not_starts_with: ID # matches all nodes with a value that does not start with given substring
  id_ends_with: ID # matches all nodes with a value that ends with given substring
  id_not_ends_with: ID # matches all nodes with a value that does not end with given substring

  # String filters
  title: String # matches all nodes with exact value
  title_not: String # matches all nodes with different value
  title_in: [String!] # matches all nodes with value in the passed list
  title_not_in: [String!] # matches all nodes with value not in the passed list
  title_lt: String # matches all nodes with lesser value
  title_lte: String # matches all nodes with lesser or equal value
  title_gt: String # matches all nodes with greater value
  title_gte: String # matches all nodes with greater or equal value
  title_contains: String # matches all nodes with a value that contains given substring
  title_not_contains: String # matches all nodes with a value that does not contain given substring
  title_starts_with: String # matches all nodes with a value that starts with given substring
  title_not_starts_with: String # matches all nodes with a value that does not start with given substring
  title_ends_with: String # matches all nodes with a value that ends with given substring
  title_not_ends_with: String # matches all nodes with a value that does not end with given substring
}
```

Resolves: #53, #81


chrissm79:
@olivernybroe LOVE this 😍Give me a bit to review this and I'll definitely merge this in. I'll cut a beta before merging in as I assume this will be a breaking change and I'll add some docs to the release to explain the necessary changes that will need to be made.

I didn't realize Laravel provided a Pipeline otherwise I would've used this pattern from the get go. Thanks again!!

olivernybroe:
@chrissm79 Laravel relies heavily on their pipeline behind the scene :)
It will be a breaking change in the way that it changes how the middleware interfaces work.
I'll fix the tests today so you can merge it in.

olivernybroe:
I have changed the way we handle errors, we know uses a custom exception handler class and this class works a lot like Laravel's Exception handler. The new error system now gives us the power to make any exception into a GraphQL compliant error.

Sorry I added this in the same pull request, but it depended on some other changes made in this pull request.

chrissm79:
@olivernybroe Thanks for all the work that went into this!! I think I'm going to merge this in w/ `v2.2` rather than merging it in w/ `v2.1` just so we have one more release w/ all the additions that have been made up to this point and I'll create an upgrade guide for `v2.2` that explains how to adjust custom directives to adhere to this new functionality. 

I've been out of pocket for awhile so I haven't had a chance to really dig into this PR in depth but I can say that I'm excited to get this merged in soon!!!

olivernybroe:
@chrissm79 Sounds great Christopher. Everything I am working on locally is using the changes from here. 
Right now I am working on an advanced filter like graphcool has. https://www.graph.cool/docs/reference/graphql-api/query-api-nia9nushae/#filtering-by-field

PR #122, opened 2018-05-18
Keep routes compatible with Lumen framework

4levels:
Lumen routing has no concept of ::match, therefor I just split the one-liner into two, adding the ::get if allowed, whilst keeping the default ::post


chrissm79:
@4levels Thank you for this!!!

PR #125, opened 2018-05-18
Fixed a bug with DeleteDirective when using a globalId

alexwhb:
When you use the delete directive with a global ID $id comes back as an array with `['modelName', 'id']`, so $model::find($id) returns a collection instead of the model instance. I'm simply access the first index of the global ID resolver, so we get the actual model instance back when doing the `$model::find($id)` operation.


chrissm79:
@alexwhb Thank you for adding this!!

alexwhb:
@chrissm79 Happy to help. :)

PR #130, opened 2018-05-24
Added fallback if no resolver is passed to the @union directive

4levels:


chrissm79:
Thanks for adding this @4levels!!

PR #132, opened 2018-05-26
Handle custom model primary key

SquareBeard:
Fixes relations not set when the model primary key is not 'id'.


chrissm79:
Thanks for this @SquareBeard!!!

Issue #131, opened 2018-05-25
[Feat] CRUD and Nested Mutations


spawnia:
I am currently trying to build an API on top of an existing Laravel backend, using [laravel-graphql](https://github.com/Folkloreatelier/laravel-graphql). Querying data seems to work fine, although the amount of boilerplate code is quite high.

However, i found myself struggling with implementing mutations.

We have 2 major requirements that make it quite hard:
1. We have quite a complex schema with about 100 inter-related models and a few thousand columns so we do need a way to either generate code or deduce basic CRUD operations from a schema.
2. We need to be able to both query and mutate data with deeply nested relationships.

Prisma does offer a sensible structure for how nested mutations can be dealt with and i feel it would play very nicely with Eloquent, have a look: https://www.prisma.io/docs/1.9/reference/prisma-api/mutations-ol0yuoz6go#nested-mutations

I think we might be able to deduce sensible defaults for CRUD operations simply by defining the types in the schema definition, what do you think?

olivernybroe:
For your question about generating code, you should look into making a custom directive. Then you can make that pretty much do whatever you want.

chrissm79:
Hey @spawnia, I was originally going to do something like this for v2.0 but it quickly spiraled into a much bigger issue than I originally thought. I've built on my initial ideas and I could see a nested implementation that would look something like this:

```graphql
# The `crud` argument tells Lighthouse to create crud mutations for this model.
#
# The `connect` argument allows you to specify which 
# nested relationships you can include in a auto-generated
# crud mutation.
# 
# The @hidden directive prevents these fields from being included in the final schema,
# the fields are just there for validation purposes.
# 
type User @model(
  class: "App\\User", 
  crud: true, 
  connect: ["posts"]
  ) {
  id: ID!
  name: String @validate(rules: ["required"])
  email: String @validate(rules: ["required", "emails", "unique:users,email"])
  password: String @hidden @validate(rules: ["required", "min:8", "confirmed"])
  password_confirmation: String @hidden
  posts: [Post] @hasMany
}

# The `validators` argument allows you to specify validator 
# classes for the `create`, `update`, or `delete` crud mutations.
# 
# The `policies` argument allows you to specify a Laravel policy for
# each crud mutation
# 
# The `transformers` argument allows you to transform the input before 
# it gets injected into the model.
# 
type Post @model(
  class: "App\\Post",
  crud: true,
  validators: { 
[...]

PR #133, opened 2018-05-29
[WIP] [Feat] Generate basic CRUD operations

spawnia:
Relates to #131

As mentioned in the discussion of the Issue, i implemented a first version of `@generate`. Right now it is very incomplete in terms of functionality, but the pieces to build it out further should be in place.

The directive works by modifying the schema definition before passing it along. This cleanly marks it as schema-modifying and enables it to leverage existing directives.

Let me know what you think of this approach and if i should continue going down this path.

### DIFF
--- a/src/Schema/Directives/Nodes/GenerateDirective.php
+++ b/src/Schema/Directives/Nodes/GenerateDirective.php
@@ -10,6 +10,7 @@ use Nuwave\Lighthouse\Support\Contracts\Directive;
 use Nuwave\Lighthouse\Support\Traits\CanParseTypes;
 use Nuwave\Lighthouse\Support\Traits\HandlesDirectives;
 
+// Extend or implement a super type that marks this as a schema-changing directive
 class GenerateDirective implements Directive
 {
     use HandlesDirectives, CanParseTypes;
@@ -31,16 +32,36 @@ class GenerateDirective implements Directive
 
         // Get config values and set defaults
         $model = $this->directiveArgValue($directive, 'model');
-        $crud = $this->directiveArgValue($directive, 'crud', false);
-        $read = $this->directiveArgValue($directive, 'read', false);
 
+        $crud = $this->directiveArgValue($directive, 'crud', false);
+        $create = $this->directiveArgValue($directive, 'create', false);
+        // Might have more sophisticated options here to control the generated queries
+        $read = $this->directiveArgValue($directive, 'read', false);
+        $update = $this->directiveArgValue($directive, 'update', false);
+        $delete = $this->directiveArgValue($directive, 'delete', false);
+
+        if($crud || $create) {
+            // TODO add in update mutations
+            // Look at the relations and possibly other type definitions in $originalDocument to figure out nested mutations
+        }
+
+        // TODO modify the type definition in the schema so that hidden fields are excluded
         if ($crud || $read) {
             $query = $this->getRootQueryDefinition($documentNode);
-
+            
             $query = $this->addPaginatedQuery($query, $nodeName, $model);
             $query = $this->addQueryById($query, $nodeName, $model);
-
+            
+            
             $documentNode = $this->setRootQueryDefinition($documentNode, $query);
+        }
+
+        if($crud || $update) {
+            // TODO add in update mutations
+        }
+
+        if($crud || $delete) {
+            // TODO add in delete mutations
         }
         
         return $documentNode;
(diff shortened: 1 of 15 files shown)


spawnia:
Sure thing! I added some additional comments which will hopefully make it easier to grasp what i am going for.

This PR does deal with two points actually:
1. Add structure for how directives can generate schema definitions in a general manner that allows for easy customization
2. The actual directive for generating CRUD operations, depends on 1.

Do you think this can be done in one PR or should i rather do 1. as a seperate PR first?

chrissm79:
@spawnia Thanks a TON for putting the work in on this! I like the idea of having a `@generate` directive that allows users/plugins to extend the document as needed.

However, I keep going back to using a plain directive primarily because I don't want future plugins to be forced into a naming convention (i.e., `@generateCrud` instead of just `@crud`). 

What would probably be most helpful (for those that are attempting to extend the schema) is to create some sort of helper to parse the new schema and automatically register the generated types. If you take a look at the [CreatesPaginator](https://github.com/nuwave/lighthouse/blob/master/src/Support/Traits/CreatesPaginators.php#L22-L34) trait you'll see that it extends the schema by injecting it's own, but it's pretty verbose at the moment.

So for something like `@crud` (or `@resource` to keep familiar to Laravel), I could see it producing the following schema on-the-fly:

```graphql
# original schema created by the user
type User @resource {
  id: ID!
  # ....
}

# schema created by the @resource directive
extend type Mutation {
  createUser(input: CreateUserInput!): User @resourceCreate
  updateUser(input: UpdateUserInput!): User @resourceUpdate
  deleteUser(input: DeleteUserInput!): User @resourceDelete
}

input CreateUserInput {
  # ...
}

input UpdateUserInput {
  # ...
}

input DeleteUserInput {
  # ...
}
```

Directives would also be created for the `resourceCreate`, `resourceUpdate` and `resourceDelete` but it's nothing the end-user would need to worry about.. they're just used to handle the field as necessary (i.e., adding validation, handling the model, transforming input, etc).

An example w/ all options on the `@resource` directive would look something like this (but wouldn't have an effect on the generated schema):

```graphql
type User @resource(
  model: "App\\Models\\User"
  except: ["update"],
  validators: {
    create: "App\\GraphQL\\Validators\\CreateUserValidator",
    delete: "App\\GraphQL\\Validators\\DeleteUserValidator"
  },
  policies: {
    create: "create",
    delete: "delete"
  },
  transformers: {
    create: "App\\GraphQL\\Transfomers\\UserTransformer@create",
    delete: "App\\GraphQL\\Transfomers\\UserTransformer@delete",
  }
) {
  id: ID!
  first_name: String @validate(rules: ["required"])
  last_name: String @validate(rules: ["required"])
  email: String @validate(rules: ["required", "email", "unique:users,email"])
}
```

Finally, I think nested mutations should be a separate PR after the `@resource` directive is created. I've been going through more use-cases and it's much more difficult than I originally thought (to the point where I think it may be better as an external plugin, but still debating that) and I want to keep this PR as straightforward as possible.

Thanks again @spawnia!!!

spawnia:
Hey, thank you for the feedback. I agree we should split up this work across multiple PRs, focus on the easier stuff at first.

I like your suggestions around naming, especially how we should aim to keep it close to what Laravel is doing. Doing basic CRUD the Laravel way seems like it is the most intuitive and obvious.

I did abandon the idea for having a naming convention for directives. It seems like it leads to some really verbose naming and might be too strict in the long run. However, i would still like to introduce a more clean structure around how Directives are built internally.

Right now we have Middleware and Resolver directives, which are marked as such by implementing an interface. I propose to introduce another interface which might look something like this:

````php
<?php

namespace Nuwave\Lighthouse\Support\Contracts;

use GraphQL\Language\AST\DefinitionNode;
use Nuwave\Lighthouse\Schema\Utils\DocumentAST;

interface SchemaGenerator extends Directive
{

    /**
     * @param $definitionNode The definition of the node where the directive was defined
     * @param DocumentAST $current Current state of the schema where some transformations may have occured already
     * @param DocumentAST $original The schema as it was originally defined by the user
     *
     * @return DocumentAST DocumentAST is basically a helper class which provides
     * convenient methods to reflect upon the schema and to modify it easily. Custom 
     * directives are able to implement SchemaGenerator and be able to do schema generation themselves.
     */
    public function handleSchemaGeneration($definitionNode, DocumentAST $current, DocumentAST $original);
}
````

Directives are allowed to implement more than one of those interfaces, e.g. @paginate does both schema generation and also provides a resolver. The methods defined in the directives are called in a specific order though: First, all Generators are called and produce a generated schema. Then, all directives on this generated schema are evaluated again and middlewares and resolvers are handled.

This approach ensures predictability for what directives actually do. I am in the process of converting the existing directives to work this way. If it works, it should make the process of schema generation more predictable and straight-forward.

spawnia:
I will close this one for now, as i have been reworking how schema generation works quite a bit. As soon as we get that merged, i will pick up on this again and implement it on top.

Issue #109, opened 2018-05-10
New GraphQL implementation


crisu83:
Hello Maintainers,

We have been working on a new PHP implementation of the GraphQL specification and we are closing in on a version 1.0 release. I just wanted to stop by and mention this to see if you're interested in teaming up with us. There is a Slack for the project where we can invite you.

The repository can be found here:
https://github.com/digiaonline/graphql-php

Best,
Christoffer

chrissm79:
@crisu83 Thanks for reaching out!!! Lighthouse is pretty heavily dependant upon `graphql-php` but I'm always open to considering alternatives! Quick question for you, do you have a way to handle N+1 issues in your project? For instance, `graphql-php` provides a [Deferred](http://webonyx.github.io/graphql-php/data-fetching/#solving-n1-problem) class that works similar to a [Dataloader](https://github.com/facebook/dataloader) and is a must have to resolve N+1 queries. Thanks @crisu83!!

crisu83:
Hello @chrissm79,

We have a similar Promise based approach built on top of `react/promise`. My colleague @hungneox can tell you more about it as he is the one who implemented it.

I've heard that people have had trouble implementing features like [Schema Decorators](https://dev-blog.apollodata.com/graphql-schema-decorators-3c01a445b628) on top of `webonyx/graphql-php`.  Our library has been built from the start with extendability in mind, so we have pretty good support for custom types, custom directives, and custom validators. You should also be able to override most of the library by injecting your own implementation.

We also encourage developers to use [Schema Definition Language](https://blog.graph.cool/graphql-sdl-schema-definition-language-6755bcb9ce51) over schemas programmatically, because we believe that SDL is the future of GraphQL.

I'd love to hear what you think about our implementation.

Best,
Christoffer

olivernybroe:
@crisu83 Sounds neat,
Have you guys done any performance testing compared to the two other graphql libraries out there?

crisu83:
@olivernybroe yes we have and our library performs well compared to the other implementations, in fact on PHP 7.2 we are a bit faster than the others.

We are also working on Relay support, see https://github.com/digiaonline/graphql-relay-php for more info.

fubhy:
I can confirm this. The code quality of https://github.com/digiaonline/graphql-php is outstanding and I also did quite a few performance comparisons. I am the maintainer of https://github.com/drupal-graphql/graphql and we are already in the process of switching to it now :).

crisu83:
[...]

PR #135, opened 2018-06-05
[WIP][Feat] Driver-based

olivernybroe:
This pull request will rewrite the system to a driver based model and implement DigiaOnline as the standard driver. 

There are still a lot of work needed before this is done. Will try and update this when new stuffs are added. No need to look through the code yet, it is not working enough for code reviews.

- [ ] Query building/manipulating
- [ ] Manipulating the Schema from directives
  - [ ] Node directive
  - [ ] Arg directive
  - [ ] Field directive
- [ ] Cache
- [ ] Error handler
- [ ] Node directive
- [ ] Client directive
- [ ] Default directives
  - [X] Event
  - [X] Auth
  - [ ] HasMany
  - [ ] BelongsTo
  - [ ] Field
  - [ ] Paginate
  - [ ] Validation
  - [ ] Filters (in, notIn, equals, notEquals, where, whereBetween, whereNotBetween)
  - [ ] Scout (search)
  - [ ] Complexity
  - [ ] Create
  - [ ] Update
  - [ ] Delete
  - [ ] Rename
  - [ ] Group
  - [ ] Model
  - [ ] Interface
- [X] Arg directive
- [X] Field directive

Resolves: #109


spawnia:
In general, i really like the idea of abstracting away the underlying implementation. 

However, it seems as though you had to add in a lot of code which is kind of re-implementing the basics of the GraphQL type system. When we can not rely on the underlying type definitions as well as the AST, it seems to me that we are halfway at writing our own GraphQL implementation.

At that point, what do we even gain by using different implementations? It seems as though we are depending on the underlying parser as well as the execution, but have to implement everything inbetween ourselves. Then, we have to maintain compatability with multiple implementations.

I would lean towards keeping it simple by just picking one implementation and relying on the features it provides. That way, the scope of Lighthouse shrinks - it is simply a relatively thin integration layer between the GraphQL implementation and Laravel.

What are your experiences with Digia so far? Do you think it is better? Is it worth it to make the switch over?

olivernybroe:
@spawnia This isn't actually that much different from what we did before. 
Our solution before was to make the underlying GraphQL implementation parse the type system file and then we would manipulate with the AST tree and make the implementation parse the AST tree again.
The way this driver based solution do it is by first make the GraphQL implementation parse the type system file to our Schema object. Then we parse our Schema object to the GraphQL implementations Schema object and run the query on it.

Until now I haven't had any reason to actually define the AST tree in our abstracting and I don't think we need to.

One of the biggest advantages I have found by our driver based system is that I can do a lot of stuff much simpler as I can define my own methods on the types. So now we don't have to rely on a lot of helper methods which looks through the AST tree.

Digia has better performance and I found their namings of functions and the amount of interfaces better. However they are lacking documentation atm., so right now I have to check tests for how to do stuff.

spawnia:
You make a good point: Schema manipulation is currently quite a large pain point. I am currently trying to do this in the way you mentioned, by manipulating the AST. As i started adding helper methods, i found it is still quite difficult to do. It seems like a hack - the AST is not meant to be manipulated.

I see two main points for how the internal schema should be designed:

1. The internal representation should be easy to manipulate. Directives should have a clear and straightforward way of reaching into the schema and manipulating it.

2. The resulting schema must be cacheable. Schema generation and validation are quite costly and should not be done on every request.

olivernybroe:
@spawnia 
That is exactly what I am trying to do :) I was in the same situation as you with manipulation the AST and it really feels hacky.
The way I want it to work is that we just change stuff in our Schema object. So for example changing the name of a type could be done like this `schema()->type('User')->setName('Employee')`. Something like this is how I want it to work.

Regarding the caching, I haven't looked into that yet. Need to get something up and working and then I'll add caching. Maybe integrate it with Laravel's cache system?

spawnia:
Just as food for thought, here is a random list of schema mutations that i found a need for in my experimentation so far:
- Add a new type
- Delete an existing type
- Add fields to a type
- Add arguments to a field
- Add directives to anything

However, i think a well-defined process needs to be in place that clearly defines how directives are able to change the schema. I would suggest the following:
1. Parse the schema files into our internal schema object
2. Iterate over all type definitions and apply directives that are marked as schema-manipulating (by implementing an Interface probably). The output of each such call is a new version of the Schema. The newly generated types/fields may have directives defined upon them.
3. Take the new Schema and iterate over all fields. Again, field directives can generate a new version of the schema.
4. Iterate over the arguments and apply generator directives.

The structure of the resulting schema is now final! Perhaps, we can even make it immutable at this point? We are now able to:
- Cache the result
- Validate if the generated schema is valid
- Output the schema in GraphQL-SDL and use it for static query analysis, mocking, etc.

It is generally possible to [serialize closures](https://github.com/jeremeamia/super_closure), so we might be able to build an executable schema particular to the used implementation and cache that as well. For this, we evaluate our internal schema again and apply resolver directives. Those may define resolver functions and add them to the schema, but are not allowed to modify the schema in any other way.

At runtime, we can look up the cached version of our internal schema (which still has the relevant directives on it). Now, the directives may register Middleware, do additional runtime checks - all kinds of side effects are allowed.

olivernybroe:
@spawnia Thanks for the thoughts.
What if the only directive which can manipulate the Schema is a Node directive?
The reason for this is lazy loading. Else we will have to evaluate the fields of a type we might not even use.

spawnia:
Some field directives have to make changes to the schema. For example, the `@paginate` directive creates a new output type which it then returns.

I think full schema generation has to be done before, considering that some types/fields might otherwise not even exist - how could we lazy load them in that case?

However, i agree that it is quite a costly operation as the schema grows larger. Because of this, caching the result of schema generation is crucial. If we can guarantee that the schema does not change during execution, this should be safe to do.

Other directives, such as Middleware or Resolvers, can be evaluated on-the-fly, as they only apply to their respective type/field.

olivernybroe:
You are right, `@paginate` is a great example on where we need this.
We can only lazy load fields, we cannot lazy load types, lazy loading of fields is done via a callback.
But yes, this essentially means we can't have lazy loading, but only caching.

spawnia:
Hey @olivernybroe 

Going to close this one for now, as a lot has changed in the meantime and you seem to have stopped working on it.

PR #139, opened 2018-06-06
Add docker-compose file for easily setting up a local dev environment for running tests

spawnia:
This should be helpful when developing with lighthouse.

Maybe we can put some additional info about contributing into the readme, or a seperate contributing.md?

I would be willing to write something up if you are interested.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,23 @@
+version: '2'
+
+services:
+  php:
+    image: spawnia/php7.1-fpm:dev
+    volumes:
+      - ./:/var/www
+    environment:
+      - APP_ENV=local
+      - APP_DEBUG=true
+      - APP_KEY=base64:w+nUc1ZQlOX4JI4PfIp1TZORaju1JfAxnsJ0SLT1A+4=
+
+      - DB_HOST=mysql
+      - DB_USERNAME=root
+
+  mysql:
+    image: mysql:5.7
+    command:
+      - --character-set-server=utf8
+      - --collation-server=utf8_general_ci
+    environment:
+      - MYSQL_DATABASE=lighthouse
+      - MYSQL_ALLOW_EMPTY_PASSWORD=true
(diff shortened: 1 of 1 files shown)


hailwood:
I may be looking at this wrong, but this seems strange to me,

I mean it's not like you can just download the lighthouse repository and run it and have a working application, other than the test, it's a package you have to add into an existing Laravel setup so what benefit do we gain by adding a docker-compose file to the package?

spawnia:
@hailwood Did you read my description? This is only meant for making development of lighthouse itself easier by providing a ready-to-use environment for running tests.

The database setup was a bit tricky so i figured this might help other contributors out.

hailwood:
@spawnia I did, and *developing with lighthouse* is different to *development of lighthouse* hence the question.

In that regard then this seems like a good idea to me.

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    image: spawnia/php7.1-fpm:dev
+    volumes:
+      - ./:/var/www
+    environment:
+      - APP_ENV=local
+      - APP_DEBUG=true
+      - APP_KEY=base64:w+nUc1ZQlOX4JI4PfIp1TZORaju1JfAxnsJ0SLT1A+4=
+
+      - DB_HOST=mysql
+      - DB_USERNAME=root
```
hailwood: Do we need to add `DB_NAME=lighthouse` or is that picked up somewhere else?

spawnia:
You are right, i was not all clear about that. Thanks for clearing it up!

docker-compose.yml
spawnia: The default test settings are made in DBTestCase, this just overwrites two values that are necessary for making this work with the mysql container.

spawnia: I chose to use the root user with an empty password, since that seems like the most convenient option if the need should arise to manually do something in the database.

Issue #134, opened 2018-05-30
Can't get @group directive to work


EslamNasser:
Lighthouse version: 2.1-beta.5 for lumen support

Hello, i'm using lumen with lighthouse and everything is working perfectly except for the @group directive as i want to apply an auth middleware to a group of queries. Every time i try to execute one of the queries in that group it goes straight to the resolver without passing by the middleware first.

I made sure that the middleware is working as when i apply the middleware in the lighthouse config it works perfectly and all the graphql requests pass by the middleware first before being handled by the graphql controller.

This is my **schema.graphql:**

```
scalar DateTime @scalar(class: "DateTime")

type User{
   id: ID!,
   username: String,
   age: Int,
   notes: String,
   created_at: DateTime!,
   email: String,
   mobileNumber: String
}

type LoginResult{
   userid: String!,
   token: String!,
   msg: String
}

input LoginInput{
   username: String!,
   password: String! @bcrypt
}

type Query{
   login(mobileNumber: String!, password: String!): LoginResult! @field(resolver: "App\\GraphQL\\Queries\\Query@Login")
}

extend type Query @group(middleware: "jwt"){
   test: String! @field(resolver: "App\\GraphQL\\Queries\\Query@testQuery")
   getUsers: [User!]! @field(resolver: "App\\GraphQL\\Queries\\Query@GetUsers")
   getUserInfo: User! @field(resolver: "App\\GraphQL\\Queries\\Query@getUserInfo")
}

type Mutation{
   addUser(username: String!, age: Int!, notes: String, password: String! @bcrypt, mobileNumber:String!, email: String!): User! @create(model: "App\\Models\\User")
}
```

What could be the problem here?

Thanks a lot for an amazing project btw

kikoseijo:
Try to add also the others, like web,  did you removed all middleware in the configuration?

EslamNasser:
Yes i stopped all the middlewares in the configurations 

This is my **lighthouse.php:**
```
<?php

return [
        /*
    |--------------------------------------------------------------------------
    | LightHouse endpoint & middleware
    |--------------------------------------------------------------------------
    |
    | Setup this values as required,
    | default route endpoints its yourdomain.com/graphql
    | setup middleware here for all request,
    | setup more endpoints, ej: pointing to the controller value inside your route file
    |
    */
    'route_name' => 'graphql',

    'route' => [
        'prefix' => '',
        //'middleware' => ['jwt'],    // [ 'loghttp']
    ],

[...]

Issue #121, opened 2018-05-18
Does group middleware works?


mits87:
Hi,

I tried use middleware in my queries and mutators like:

``` graphql
type Query @group(middleware: ["jwt.auth", "cors"]) {
  accounts: [Account!]! @paginate(model: "App\\Account")
}
```

unfortunately it doesn't work. 
How can I use the middleware in my Queries / Mutators  etc. ?

kikoseijo:
Should post your full schema. could be many things, and when you are working on middleware more,

Just a guess:: you probably missing the `extend type Query`. This is when you using more than 1 type query definition.

Another guess: make sure it works without graphql, you can always build the middleware in a route and just call the default one something like:

```php
// routes/api.php
/*
|--------------------------------------------------------------------------
| GraphQL Lighthouse Auth Route
|--------------------------------------------------------------------------
|
*/
$controller = config('lighthouse.controller');

Route::group(['middleware' => ['auth:api']], function () use ($controller) {
    Route::post('graphql', ['as' => 'graphql.secured', 'uses' => '\\'.$controller]);
});
```

mits87:
Hi, 

thank you for your response. 
First of all my middlewares works. 


My full schema:

**auth.graphql**
``` graphql
type AuthPayload {
    access_token: String!
}

extend type Mutation @group(namespace: "App\\Http\\GraphQL\\Mutations") {
  authenticate(
    email: String!
    password: String!
  ): AuthPayload @field(resolver: "AuthMutator@login")
}
```


**schema.graphql**
```graphql
#import ./auth.graphql

type Account {
  id: ID!
  identifier: String!
}

type User {
  id: ID!
  email: String!
  account: Account @belongsTo
}

type Query @group(middleware: ["jwt.auth", "cors"]) {
  accounts: [Account!]! @paginate(model: "App\\Account")
}

type Mutation @group(middleware: ["jwt.auth", "cors"], namespace: "App\\Http\\GraphQL\\Mutations") {
  createUser(
    email: String! @validate(rules: ["min:5"])
  ): User @field(resolver: "UserMutator@create")
}
```

After execute query: 
``` graphql
query {
  accounts(count: 1) {
    data {
      id
    }
  }
}
```

I got:

> Error
>     at file:///Applications/GraphQL%20Playground.app/Contents/Resources/app.asar/lib/dist/app.e59169620134e31c7561.js:33636:31
>     at <anonymous>
[...]

Issue #84, opened 2018-04-13
[Feat] Argument filters


olivernybroe:
I can't figure out how we can set a field to be filterable.
For example i have some users and I would like to make the name something i can filter by.

chrissm79:
Hey @olivernybroe,

If you trying to filter something that is using the `@belongsTo`, `@hasMany` or `@paginate` directives you could just use the `scopes` argument like so and it will pass the field's arguments to your scope.

```graphql
type User {
  posts(active: Boolean): [Post!]! @hasMany(scopes: ["active"])
}
```

```php
class Post extends Model
{
    public function scopeActive($query, array $args)
    {
        return $query->when(!empty($args['active']), function ($q) use ($active) {
            return $q->where('active', $args['active']);
        });
    }
}
```

Otherwise, you just create an argument on the field and use it in your resolver

```graphql
type User {
  avatar(size: String!): String @field(resolver: "App\\Http\\GraphQL\\UserType@avatar")
}
```

```php
class UserType
{
    public function avatar($root, array $args)
    {
        // Here the $root could be your User model (or whatever you assigned to the User type)
        return $root->avatar($args['size']);
    }
}
```

kikoseijo:
The `@field(resolver: "App\\Http\\GraphQL\\UserType@avatar")` can do the job also.

You just tell where to resolve and take the input from `$this->args` on resolve function.

Check examples here:
https://github.com/nuwave

olivernybroe:
Right now my query looks like this
```

```

olivernybroe:
Right now my query looks like this
```
agreements(id: ID): [Agreement!]! @paginate(model: "Agreement", type: "connection", scopes: ["byAuth"])
```
With the @field I have to do a lot of the work manually, so was thinking about adding a ArgDirective like this
```
agreements(id: ID @filterBy(field: "id")): [Agreement!]! @paginate(model: "Agreement", type: "connection", scopes: ["byAuth"])
```
Just trying to find out how to access the query from a `ArgMiddleware`. 
As I kinda like the syntax I made, as it doesn't bloat my scopes when just filtering by id.

kikoseijo:
What client you using?

I guess the magic its a bit limited, you trying to have all done from the schema, 
Question for magic its to know how you should reply, 

here its a pagination example on root fields. #70 
If you dont need pagination just build your own query resolve function, you can even use a repository function,

olivernybroe:
Setting it up with relay pagination. 

[...]

PR #140, opened 2018-06-08
Cacheable schema manipulation via directives

spawnia:
This PR rebuilds how Lighthouse deals with schema manipulation.

### Reasoning

Directives can now implement an Interface that marks them as schema-manipulating. This manipulation is executed directly on the AST. This has multiple benefits:
- Schema manipulation is very explicit, no more surprises!
- The final AST is completely cached when in production
- Directives are now able to easily add new types, fields or arguments - which can take full advantage of directives themselves!

In the spirit of 'schema-first', this also enables full utilization of the SDL.
The final schema can be validated and printed before using it in production.
Users can use the output for documentation, mocking or static query analysis.

This also opens the door for more complex schema manipulation. The features proposed in #131 and #84 would greatly benefit from this.

### Bugfixes

Coincidentally, i found and fixed a bug that prevented middlewares being applied in some cases, this might resolve the Issues #134 and #121 

### Backwards compatibility

I made sure to keep most of the tests as they were before to ensure we do not break existing use-cases. While the inner workings where changed quite a bit, the functionality of the directives should be the same as before.

Caching now uses Laravels Cache Facade. To enable caching, a new configuration option was added. Users that have the previous configuration will not get caching for their AST but can easily update their config to enable caching.

### Breaking changes

Custom directives will have to be changed. All directives have to implement a common Interface. Schema manipulation should only happen by implementing the respective Manipulator interface: `ManipulateType`, `ManipulateField` and `ManipulateArg`

Some internal classes were shifted around so users that reach directly into our source might have to change their code as well. This should not be a big issue as we haven't formally specified a public API yet.


src/Support/Console/Commands/PrintSchemaCommand.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Console\Commands;
+
+use GraphQL\Utils\SchemaPrinter;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Cache;
+use Nuwave\Lighthouse\GraphQL;
+
+class PrintSchemaCommand extends Command
```
olivernybroe: This is pretty nice.

src/Schema/Directives/Fields/FieldMiddleware.php
```diff
@@ -1,18 +1,12 @@
 <?php
 
-namespace Nuwave\Lighthouse\Support\Contracts;
+namespace Nuwave\Lighthouse\Schema\Directives\Fields;
```
chrissm79: Let's not mix the contracts with the classes, so let's move these back to the `Contracts` namespace (which should also help w/ breaking changes)

src/Schema/Factories/DirectiveFactory.php
```diff
@@ -2,255 +2,37 @@
...
+    {
+        return new Directive([
+            'name' => $directive->name->value,
+            'locations' => collect($directive->locations)->map(function ($location) {
+                return $location->value;
+            })->toArray(),
+            'args' => collect($directive->arguments)->map(function (InputValueDefinitionNode $argument) {
+                return new FieldArgument([
+                    'name' => $argument->name->value,
+                    'defaultValue' => $argument->defaultValue->value,
```
chrissm79: Still testing, but it looks like this will fail if no default value is provided so we may want to wrap this with:


```php
'defaultValue' => data_get($argument, 'defaultValue.value', null),
```

src/Support/helpers.php
```diff
@@ -24,27 +24,27 @@ function auth()
...
-if (! function_exists('schema')) {
+if (! function_exists('types')) {
     /**
      * Get instance of schema container.
      *
-     * @return \Nuwave\Lighthouse\Schema\SchemaBuilder
+     * @return \Nuwave\Lighthouse\Schema\TypeRegistry
      */
-    function schema()
+    function types()
```
chrissm79: I think `schema` may be a safer global function name (but that's just a personal preference)... let's add the `schema` function back and have it just call the `types` function to minimize breaking changes.

src/Schema/TypeRegistry.php
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @var Collection
+     */
+    protected $types;
+
+    public function __construct()
+    {
+        $this->types = new Collection();
+    }
+
```
chrissm79: Let's add a `instance` function that calls `get` to minimize breaking changes (goes along w/ adding the `schema` function back into the `helpers.php` file)

src/Schema/Directives/Fields/FieldMiddleware.php
spawnia: The reason why i did it was to make it clear which kind of directive each of the Interfaces belongs to. I am all for avoiding unnecessary breaking changes though, is there a way to get the best of both?

chrissm79: I think `FieldResolver` and `FieldMiddleware` clarify that they belong to a field and `ArgMiddlware` to and arg and so on. I wouldn't mind restructuring some of the namespaces but I'd prefer to do that a bit down the line as some users are getting used to the overhaul of `v2.0` and I'd hate to have them encounter any more breaking changes than they need to.

spawnia: I will make it so!

src/Support/helpers.php
spawnia: I can rename it to schema, no problem. The global functions do seem a bit dangerous though. Maybe, in the next major release, we can namespace them under graphql() ?

src/Schema/Factories/DirectiveFactory.php
spawnia: Good catch. Wrote a test - it failed - and fixed it!

src/Support/helpers.php
chrissm79: @spawnia totally agree w/ that, much safer

chrissm79:
@spawnia Other than the small changes this thing is just about ready to go!!! I noticed when testing that my fields w/ multiple directives (specifically a `@field` and `@validate` combo) always received `null` when calling `$value->getResolver()`. I haven't had too much time to investigate but if I can provide some insight I will as soon as possible!

Thanks for your work on this PR! Can't wait to get it merged in!!!

src/Schema/TypeRegistry.php
spawnia: Done!

spawnia:
https://github.com/nuwave/lighthouse/pull/140/commits/a728eefefeab9a50fd7c8064b835e8d7a1d2194a improves backwards compatibility

src/Schema/Directives/Types/GroupDirective.php
```diff
@@ -0,0 +1,116 @@
...
+
+        if (! $namespaceValue) {
+            return $objectType;
+        }
+
+        if (! is_string($namespaceValue)) {
+            throw new DirectiveException('The value of the namespace directive on has to be a string');
+        }
+
+        $namespaceDirective = PartialParser::directive('@namespace(value: "'.addslashes($namespaceValue).'")');
```
spawnia: Change to:

spawnia:
@chrissm79 Implemented namespaces in the way we discussed yesterday. I did not put in the new version of `@namespace` yet, but made it so the path is clear for adding it in a later PR. For now, i converted `@group` to work as before while still enabling schema caching and clean schema manipulation.

hailwood:
image

> @spawnia: Don't mind me, I'm just rewriting Lighthouse

Monumental effort, exciting stuff here!

PR #143, opened 2018-06-18
Quote table name alias

spawnia:
This resolves a bug i had where one of my table names was a MySQL keyword ("lines").

### DIFF
--- a/src/Support/DataLoader/QueryBuilder.php
+++ b/src/Support/DataLoader/QueryBuilder.php
@@ -105,7 +105,7 @@ class QueryBuilder
         })->implode(' UNION ALL ');
 
         $table = $related->getTable();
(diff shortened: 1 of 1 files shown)


chrissm79:
@spawnia I actually originally had the back ticks in there but apparently [Postgres doesn't like it](https://github.com/nuwave/lighthouse/commit/0660bdade4538e4c4cdf7c10bd14c61892427f81#diff-6b028005130428f85c23f5c32914edcd). I thought I could add a check but different models might use different connections. Haven't circled back to this, not sure what the best resolution is that will work for everyone, but back ticks totally breaks Postgres.

spawnia:
Going to close this one off for now, as i am starting to contribute from our companies fork from now on. I am on still on it though!

PR #145, opened 2018-06-19
Test multiple PHP versions in travis

spawnia:


spawnia:
Closing this as we will be moving to PHP 7.1 once webonyx/graphql-php does so too.

PR #158, opened 2018-06-24
Rule Factory

chrissm79:
This PR adds a new `@rules` directive that can be applied to both mutation arguments as well as nested input object fields.

The directive takes advantage of schema manipulation to flatten out nested rules onto the mutation field which will be run before processing a mutation and provide Laravel validation error(s) in the response meta.

If the `@rules` directive is placed on list, the path will automatically include the `*` notation used by Laravel's validator.

Example Usage:

```graphql
input AddressInput {
  street: String @rules(apply: ["required"])
}

input UserInput {
  name: String @rules(apply: ["required"])
  address: [AddressInput]
}

type Mutation {
  createUser(input: UserInput): String
}
```

Will automatically manipulate the schema and place the following directives on the mutation field:

```graphql
type Mutation {
  createUser(input: UserInput): String
    @rules(path: "input.name", rules: ["required"])
    @rules(path: "input.email", rules: ["required", "email"])
    @rules(path: "input.address.*.street", rules: ["required"])
}
```

### DIFF
--- a/src/Schema/AST/ASTBuilder.php
+++ b/src/Schema/AST/ASTBuilder.php
@@ -3,6 +3,7 @@
 namespace Nuwave\Lighthouse\Schema\AST;
 
 use GraphQL\Language\AST\FieldDefinitionNode;
+use GraphQL\Language\AST\InputObjectTypeDefinitionNode;
 use GraphQL\Language\AST\InputValueDefinitionNode;
 use GraphQL\Language\AST\NamedTypeNode;
 use GraphQL\Language\AST\Node;
@@ -11,6 +12,7 @@ use GraphQL\Language\AST\ObjectTypeDefinitionNode;
 use GraphQL\Language\AST\TypeExtensionDefinitionNode;
 use Nuwave\Lighthouse\Support\Contracts\ArgManipulator;
 use Nuwave\Lighthouse\Support\Contracts\FieldManipulator;
+use Nuwave\Lighthouse\Support\Contracts\InputValueManipulator;
 use Nuwave\Lighthouse\Support\Contracts\NodeManipulator;
 
 class ASTBuilder
@@ -30,6 +32,7 @@ class ASTBuilder
         $document = self::mergeTypeExtensions($document);
 
         $document = self::applyFieldManipulators($document);
+        $document = self::applyInputFieldManipulators($document);
         $document = self::applyArgManipulators($document);
 
         $document = self::addNodeSupport($document);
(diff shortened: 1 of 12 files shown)


spawnia:
This approach works fine for relatively simple Inputs. It does fall apart when dealing with Input Objects that are either really large and heavily nested, or contain circular references.

Because of this, i think this can be a temporary solution, but will have to be changed at a later point. I think the SDL part is great, so we could safely merge this for now and improve on it later.

spawnia:
Oh actually, the generated mutation is not valid. The SDL prohibits the same directive being applied multiple times to the same location.

This might execute fine but will be a problem when we try to validate the schema.

chrissm79:
@spawnia Still working on implementation but can you provide an example of a really large or heavily nested input object(s) that would fail? That way I can write a test for it. As for the circular reference I believe I have a resolution for that. Thanks!

spawnia:
@chrissm79 I will not spell out an example here, as it is pretty tedious to do. Basically, we have mutations in our schema that have ~10 possible child mutations. Each of those can have about the same number of nested mutations itself.

This leads to the rules array growing exponentially in size. This by itself does not guarantee failure, but might cause PHP to run out of memory. I managed to use up > 1GB at the 7th layer of nested mutations.

src/Schema/AST/ASTBuilder.php
```diff
@@ -120,6 +123,45 @@ protected static function applyFieldManipulators(DocumentAST $document)
         }, $document);
     }
 
+    /**
```
spawnia: Clean stuff, really like this!

src/Schema/AST/ASTHelper.php
```diff
@@ -29,4 +31,16 @@ public static function mergeNodeList($original, $addition)
...
     }
+
+    /**
+     * Clone definition node.
+     *
+     * @param Node $node
+     *
+     * @return Node
+     */
+    public static function cloneDefinition(Node $node)
```
spawnia: That is in #160 right?

src/Schema/AST/DocumentAST.php
```diff
@@ -134,6 +134,16 @@ public function objectTypes()
         return $this->definitionsByType(ObjectTypeDefinitionNode::class);
     }
 
+    /**
+     * Get all definitions for input types.
+     *
+     * @return Collection
+     */
+    public function inputTypes()
```
spawnia: Maybe we should clean up the naming with these functions. We always have `FooBarDefinitionNode` as the underlying type.

src/Schema/DirectiveRegistry.php
```diff
@@ -356,15 +373,15 @@ public function argMiddleware(InputValueDefinitionNode $arg)
...
     /**
-     * @param Directive           $directive
-     * @param FieldDefinitionNode $fieldDefinition
+     * @param Directive $directive
+     * @param Node      $node
      *
      * @return Directive
      */
-    protected function hydrate(Directive $directive, FieldDefinitionNode $fieldDefinition)
+    protected function hydrate(Directive $directive, Node $node)
```
spawnia: Hydration can be applied to all directives as of #154 so maybe rebase on that?

src/Schema/Directives/Args/RulesDirective.php
```diff
@@ -0,0 +1,82 @@
...
+use Nuwave\Lighthouse\Schema\Factories\RuleFactory;
+use Nuwave\Lighthouse\Schema\Values\ArgumentValue;
+use Nuwave\Lighthouse\Support\Contracts\ArgMiddleware;
+use Nuwave\Lighthouse\Support\Contracts\InputValueManipulator;
+use Nuwave\Lighthouse\Support\Traits\HandlesDirectives;
+
+class RulesDirective implements ArgMiddleware, InputValueManipulator
+{
+    // TODO: Remove this need for this
+    use HandlesDirectives;
```
spawnia: No longer needed once we have #154

src/Schema/Factories/RuleFactory.php
```diff
@@ -0,0 +1,220 @@
...
+                $field->arguments = new NodeList($arguments);
+
+                return $field;
+            })->toArray();
+
+        $mutation->fields = new NodeList($fields);
+
+        // NOTE: This is currently not needed because we've already
+        // mutated the fields on the original instance.
+        // $documentAST->setDefinition($mutation);
```
spawnia: Still true, now that you use `cloneDefinition()` ?

PR #162, opened 2018-06-25
Remove php_cs fixer and renew styleci

spawnia:


spawnia:
@chrissm79 As we discussed in chat, this change would mean that we would rely on `styleci` for all our formatting needs.

While `php_cs` works quite well, it is easy to forget running it, especially for other contributors. If we could just automate that entirely and have the styleci bot make automatic formatting commits, that would enable us to really focus on the important stuff.

spawnia:
@chrissm79 I think this would be nice to have. I have been using StyleCI for a few projects now, the automatic formatting is quite nice. For example [in this project](https://github.com/mll-lab/graphql-php-scalars/commits/master), StyleCI automatically commits correct formatting.

mfn:
Also a StyleCI user/customer here. Love it ❤️

PR #167, opened 2018-06-26
[WIP] Only generate minimal needed rules

spawnia:
The generated rules array should be as small as possible. Else, heavily nested and large Input Objects might cause the rules array to grow too large.

### DIFF
--- a/tests/Unit/Schema/Factories/RuleFactoryTest.php
+++ b/tests/Unit/Schema/Factories/RuleFactoryTest.php
@@ -179,4 +179,35 @@ class RuleFactoryTest extends TestCase
             'input.settings.*.setting.value' => ['required'],
         ], $rules);
     }
+
+    /**
+     * @test
+     */
+    public function itGeneratesOnlyMinimalNeededRules()
+    {
+        $documentAST = ASTBuilder::generate('
+        input FooInput {
+            self: FooInput
+            email: String @rules(apply: ["email"])
+        }
+        type Mutation {
+            createFoo(input: FooInput @rules(apply: ["required"])): String
+        }');
+
+        $variables = [
+            'input' => [
+                'self' => [
+                    'self' => [
+                        'email' => 'asdf'
+                    ],
+                ],
+            ],
+        ];
+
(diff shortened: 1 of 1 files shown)


spawnia:
@chrissm79 I just added a failing test so far which shows that more then the necessary rules are generated.

chrissm79:
What about if the user defined it this way?

```graphql
input FooInput {
    self: FooInput
}
```

chrissm79:
Let's keep it as is... Personally, if I define rules on inputs I want them to all be applied regardless of what input is provided so it can catch fields that are `required` in validation by not defined as `NonNull` in the schema so I can get consistent validation errors sent back in the response.

spawnia:
`required` fields do require special treatment, as you say they must be put up regardless of whether the input is given or not.

However, we must only follow the chain until we hit a nullable field. Children of nullable fields must not be in the rules array, even if they have required field, unless there is an actual input given up until that level.

chrissm79:
The factory doesn't go any deeper than the input provided. It only generates rules for the fields of any input type included in the provided variables which is what we want so we include rules that are "required" or that are custom. In other words, if you have the following:

```graphql
input FooInput {
    self: FooInput
    email: String @rules(apply: ["email"])
}
type Mutation {
    createFoo(input: FooInput @rules(apply: ["required"])): String
}
```

and the input provided is 

```php
$variables = [
    'input' => [
        'self' => [
            'email' => 'foo'
        ],
    ],
];
```

then the rules will be built as

```php
$rules = ['input.self.email' => ['email']];
```

and it doesn't go any deeper than that because that was the input provided. Let's close this one because it's too difficult to determine if a rule is `safe` to leave off since we can't verify custom validation rules. The rules array can grow in size but it's the safest way to ensure all "required" and custom rules are accounted for.

spawnia:
The example you provided is **exactly** what is wrong currently. Validation will fail, since no email is provided for `input.email`. It should be fine though, since the field is nullable (as all fields are per default in GraphQL).

chrissm79:
@spawnia If the `email` field does not have a `required` rule then it won't fail

PR #169, opened 2018-06-26
DocumentAST Lock

chrissm79:
Adding a lock to the DocumentAST allows us to bypass generating clones for nodes (performance). It also prevents accidental schema manipulation after node/field/arg manipulators have been run. A `unlock` method is provided if needed by the user.


spawnia:
Really nice!

PR #174, opened 2018-06-28
Custom Validator

chrissm79:
This PR introduces a new custom validator which allows you to grab the `root`, `context` and `resolve` info used on the field in your own custom validation classes. 

Additionally, a new `required_with_mutation` rule has been created to mark certain fields as required with certain mutation fields. For example, with the following schema:

```graphql
input UserInput {
  first_name: String @rules(apply: ["required_with_mutation:createUser"])
  # ...
}

extend type Mutation {
  createUser(input: UserInput): User
  updateUser(input: UserInput): User
}
```

the `first_name` is only required if the `createUser` mutation is being processed.

This PR also allows `@rules` to be placed on ObjectType field arguments so part of the query can still return a response if a nested field fails validation.

### DIFF
--- a/src/Schema/Directives/Args/RulesDirective.php
+++ b/src/Schema/Directives/Args/RulesDirective.php
@@ -2,9 +2,14 @@
 
 namespace Nuwave\Lighthouse\Schema\Directives\Args;
 
+use GraphQL\Language\AST\ArgumentNode;
+use GraphQL\Language\AST\DirectiveNode;
+use Nuwave\Lighthouse\Schema\Values\ArgumentValue;
 use Nuwave\Lighthouse\Support\Contracts\Directive;
+use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
+use Nuwave\Lighthouse\Support\Contracts\ArgMiddleware;
 
-class RulesDirective implements Directive
+class RulesDirective extends BaseDirective implements Directive, ArgMiddleware
 {
     /**
      * Name of the directive.
@@ -14,5 +19,30 @@ class RulesDirective implements Directive
     public function name()
     {
         return 'rules';
+    }
+
+    /**
+     * Resolve the field directive.
+     *
+     * @param ArgumentValue $value
+     *
+     * @return ArgumentValue
+     */
+    public function handleArgument(ArgumentValue $value)
+    {
+        // Mutation arguments are handled in the RuleFactory. This check
+        // should be unnecessary when additional interfaces are created for
+        // more fine-grain control.
+        if ('Mutation' === $value->getField()->getNodeName()) {
+            return $value;
+        }
+
+        $current = $value->getValue();
+        $current['rules'] = array_merge(
+            array_get($value->getArg(), 'rules', []),
+            $this->directiveArgValue('apply')
+        );
+
+        return $value->setValue($current);
     }
 }
--- a/src/Schema/Factories/FieldFactory.php
+++ b/src/Schema/Factories/FieldFactory.php
@@ -166,7 +166,7 @@ class FieldFactory
 
             if (sizeof($rules)) {
                 $validator = validator($inputArgs, $rules, [], [
-                    'root' => $root,
+                    'root' => $rootValue,
                     'context' => $context,
                     'resolveInfo' => $resolveInfo,
                 ]);
--- a/src/Schema/Factories/FieldFactory.php
+++ b/src/Schema/Factories/FieldFactory.php
@@ -157,6 +157,16 @@ class FieldFactory
             });
     }
 
+    /**
+     * Wrap field resolver function w/ validation logic.
+     *
+     * @param \Closure                       $resolver
+     * @param \Illuminate\Support\Collection $inputValueDefinitions
+     *
+     * @throws \Nuwave\Lighthouse\Support\Exceptions\ValidationError
+     *
+     * @return \Closure
+     */
     protected function wrapResolverWithValidation(\Closure $resolver, $inputValueDefinitions)
     {
         return function ($rootValue, $inputArgs, $context = null, $resolveInfo = null) use ($resolver, $inputValueDefinitions) {
@@ -229,10 +239,10 @@ class FieldFactory
             })
             ->filter();
 
-        // Rules are applied to the fields which are on the root operation types
-        // Nested fields are excluded because they are validated as part of the root field
+        // Rules are applied to the fields which are on the root Mutation type.
+        // Nested fields are excluded because they are validated as part of the root field.
         $parentOperationType = data_get($resolveInfo, 'parentType.name');
-        if ('Mutation' === $parentOperationType || 'Query' === $parentOperationType) {
+        if ('Mutation' === $parentOperationType) {
             $documentAST = graphql()->documentAST();
             $rules = $rules->merge((new RuleFactory())->build(
                 $documentAST,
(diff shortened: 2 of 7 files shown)


src/Schema/Directives/Args/RulesDirective.php
```diff
@@ -15,4 +20,29 @@ public function name()
...
+     * @param ArgumentValue $value
+     *
+     * @return ArgumentValue
+     */
+    public function handleArgument(ArgumentValue $value)
+    {
+        // Mutation arguments are handled in the RuleFactory. This check
+        // should be unnecessary when additional interfaces are created for
+        // more fine-grain control.
+        if ('Mutation' === $value->getField()->getNodeName()) {
```
spawnia: Why differentiate between Queries and Mutations here? They are actually handled the same in RuleFactory as of #166

src/Schema/Factories/FieldFactory.php
```diff
@@ -224,10 +239,10 @@ public function getRules($rootValue, $inputArgs, $context, ResolveInfo $resolveI
             })
             ->filter();
 
-        // Rules are applied to the fields which are on the root operation types
-        // Nested fields are excluded because they are validated as part of the root field
+        // Rules are applied to the fields which are on the root Mutation type.
+        // Nested fields are excluded because they are validated as part of the root field.
         $parentOperationType = data_get($resolveInfo, 'parentType.name');
-        if ('Mutation' === $parentOperationType || 'Query' === $parentOperationType) {
+        if ('Mutation' === $parentOperationType) {
```
spawnia: The difference between Query and Mutation is only a semantic one, i see no need to treat them different here.

spawnia:
The `required_with_mutation` rule seems like it might get real messy real fast. Consider that an InputObject may be used in multiple fields or as part of a nested InputObject.

Having a definition behave differently based on context makes it less predictable and makes further changes quite difficult. I would advocate towards just having multiple InputObjects for such a case.

One of the key strenghts of GraphQL is that you can look at a schema definition and know exactly how to operate on it, how it will behave, what it will return.

src/Schema/Factories/FieldFactory.php
chrissm79: The reason is because we don't want the entire query to fail if a nested field's rule fails. For instance

```graphql
{
  me {
    first_name
    posts(invalidArg: "foo") {
      title
    }
  }
}
```

This should return the `me` fields `first_name` field and `null` for the `posts` field w/ validation meta. If we ran rules on the entire query then we would get `null` back for `me`.

spawnia: That should not happen, definitely agree with that. However, i think excluding Queries from Validation is not the best way to achieve that.

chrissm79:
The use for `required_with_mutation` will be left up to the user. I'm using it in several places with some great success. If it manipulated the actual schema so you'd have different fields for different mutations I would agree, but rules and the schema are not the same thing.. it's just how the validation is run on the input provided.

src/Schema/Factories/FieldFactory.php
chrissm79: Not completely done w/ the PR, the Query's root fields will be validated... just not the nested fields like mutations.

chrissm79: But it needs to be handled differently, so there will be an `else if` statement here

spawnia: Mutations and Queries are absolutely the same thing. The only difference is that Queries carry the semantics of not changing stuff on the server.

spawnia: I would not even begin to handle them any different, because it might lead us down a path where they diverge more and more.

spawnia:
Passing additional args to the Validator seems interesting. What do you use that for?

chrissm79:
It's primarily for custom validation classes that need to pull from the context or resolve info. An example is the `required_with_mutation` rule which needs to confirm that the field is on the `Mutation` object so it uses the `ResolveInfo`.

src/Schema/Factories/FieldFactory.php
spawnia: Correct me if i am wrong, but to me it looks as if before this part of the code did handle fields the same whether they are defined on the parent type "Mutation" or "Query". Now, you removed the condition where it includes "Query".

chrissm79: Right, because if we run rules the same way for both Queries and Mutations then an entire query will fail if a field that is nested 10 levels deep didn't get the correct args which is not what we want. So in other words, we want to handle Queries and Mutations differently.

chrissm79: Another way to think about is Mutations have input types while queries have object types, and they need to be handled differently because an input type is not the same thing as an object type.

spawnia: That could easily happen with mutations too. There might be some args down below in the fields that are requested back from the server that might be invalid. You are right in that this should not cause the whole operation to fail.

chrissm79: Great use case, if the user has a valid mutation but an invalid arg in the fields requested back that shouldn't cause the mutation to fail.

I'm currently writing tests for this PR and I'll make sure that is one of the checks.

spawnia: While Queries and Mutations are used for different purposes, they are structurally the same. They tend to look a bit different, mutations are usually more heavy on InputArguments while queries tend to have more nested fields - but in the end, both are just Object Types.

spawnia: I think we can not lump the fields of all args together in one large validation block, but should rather run one validation per field. Each nested sub-field gets its own validation context, and if that fails, only that part of the operation should be nulled.

chrissm79: Can input type fields have args?

spawnia: Nope. From a type system perspective, they are actually the same thing as field args.

chrissm79: >I think we can not lump the fields of all args together in one large validation block, but should rather run one validation per field. Each nested sub-field gets its own validation context, and if that fails, only that part of the operation should be nulled.

I don't feel the same way here. Mutation should fail (as they currently do) if the *input* provided is invalid (as Laravel would). On a query, only fields that have invalid rules/args should be returned as null (since they have a resolver) and validation message(s) attached to the output.

spawnia:
````graphql

Comment (spawnia) [reply src/Schema/Factories/FieldFactory.php]: However, if the root mutation had an input that is a large nested InputObject, and any of that is invalid, then i would fail the whole mutation.

Comment (chrissm79) [reply src/Schema/Factories/FieldFactory.php]: Correct... the mutation args are valid... the *ObjectType's* `bar` field has an incorrect arg and would return `null`.

Comment (spawnia) [reply src/Schema/Factories/FieldFactory.php]: Yup. And we could just go ahead and replace `mutation` by `query` here, and execution-wise, nothing should change!

Comment (chrissm79) [reply src/Schema/Factories/FieldFactory.php]: Just so we're on the same page, that's how it currently works in this PR.

Comment (spawnia) [reply src/Schema/Factories/FieldFactory.php]:
```
-	    if ('Mutation' === $parentOperationType || 'Query' === $parentOperationType)
+       if ('Mutation' === $parentOperationType)

PR #175, opened 2018-06-28
Add tests and fixes to ensure Query and Mutation fields are treated equally

spawnia:
Enhanced the existing tests and added an additional test for nested Inputs, that were previously only considered on Mutations.

Changed the implementation so that no difference is made between queries and mutations.

### DIFF
--- a/src/Schema/Directives/Args/RulesDirective.php
+++ b/src/Schema/Directives/Args/RulesDirective.php
@@ -30,10 +30,7 @@ class RulesDirective extends BaseDirective implements Directive, ArgMiddleware
      */
     public function handleArgument(ArgumentValue $value)
     {
-        // Mutation arguments are handled in the RuleFactory. This check
-        // should be unnecessary when additional interfaces are created for
-        // more fine-grain control.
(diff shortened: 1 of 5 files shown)


chrissm79:
So what should happen if we add the following test to `RuleFactoryTest`?

```php
/**
  * @test
  */
public function itCanGenerateRulesForQueryArguments()
{
    $documentAST = ASTBuilder::generate('
    type User {
    }
    type Query {
        user(foo: String @rules(apply: ["required"])): User
    }');

    $variables = [
        'input' => [
            'email' => 'foo',
        ],
    ];

    $rules = (new RuleFactory())->build(
        $documentAST,
        $documentAST->objectTypeDefinition('Query'),
        $variables,
        'createUser'
    );

    // $rules is empty...
}
```

spawnia:
````php
$this->assertSame([
	'foo' => ['required']
], $rules);
````

spawnia:
Was the rule on a field you put intended?

chrissm79:
No you pushed an update that I hadn't pulled down :-)

PR #179, opened 2018-07-01
Add some more info to the readme to make it a little less empty

spawnia:


hailwood:
@spawnia since you're doing a readme update, want to add the lighthouse logo into the top of the readme?

Always makes the projects look more professional in my humble opinion :)

PR #180, opened 2018-07-01
 Add scalar & return type hints

spawnia:
Because we can depend on PHP 7.0, this enables us to use those features.
I added type hints where they could safely be added - some places require
"mixed" to be set.

In this process, i found & fixed a few minor bugs and also uncovered quite
a lot of cases where the PHPDoc was actually lying. Static type checking
enabled me to find those cases quickly and fix the underlying issues.

### DIFF
--- a/src/Schema/Directives/Nodes/EnumDirective.php
+++ b/src/Schema/Directives/Nodes/EnumDirective.php
@@ -4,6 +4,7 @@ namespace Nuwave\Lighthouse\Schema\Directives\Nodes;
 
 use GraphQL\Language\AST\EnumValueDefinitionNode;
 use GraphQL\Type\Definition\EnumType;
+use GraphQL\Type\Definition\Type;
 use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
 use Nuwave\Lighthouse\Schema\Values\NodeValue;
 use Nuwave\Lighthouse\Support\Contracts\NodeResolver;
@@ -15,7 +16,7 @@ class EnumDirective extends BaseDirective implements NodeResolver
      *
      * @return string
      */
-    public function name()
+    public function name(): string
     {
         return 'enum';
     }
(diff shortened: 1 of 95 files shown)


chrissm79:
@spawnia Not opposed to the internals using return types, however, let's not put them on the directives that users define and would need to update.

spawnia:
How about we do it on the Manipulator types, since they are not part of 2.1, and i remove it from NodeResolver? Have not put it on the other contracts, since i figured the same - we might do it in v3.

chrissm79:
Let's keep it out of all contracts that a user would implement for the time being. Thanks!!

src/Schema/Directives/Nodes/EnumDirective.php
```diff
@@ -54,7 +55,7 @@ public function resolveNode(NodeValue $value)
      *
      * @return string
      */
-    protected function safeDescription($description = '')
+    protected function safeDescription(string $description = ''): string
```
chrissm79: I assume this will effect things elsewhere, but if `null` is passed in this throws an error (since it's forcing a string to be passed). This gets called from line 45 w/ `$field->description` which can return `null`.

spawnia: The default `= ''` actually catches that. See https://3v4l.org/prfFU

chrissm79: Add the following enum to the schema [for this test](https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Schema/Directives/Args/QueryFilterDirectiveTest.php#L30-L38)

```graphql
enum Role {
    ADMIN @enum(value: "ADMIN")
    EMPLOYEE @enum(value: "EMPLOYEE")
}
```

When I do, I get the error:

```
TypeError: Argument 1 passed to Nuwave\Lighthouse\Schema\Directives\Nodes\EnumDirective::safeDescription() must be of the type string, null given, called in /lighthouse-mll/src/Schema/Directives/Nodes/EnumDirective.php on line 45
```

spawnia:
This PR has become kind of a mess, so i am closing it for now, will probably redo it a later point. @chrissm79 when would be a good window to get this in?

We established that:
- safeDescription can not have scalar type hint string
- Interface signatures can not change because of this

It will be really beneficial once we establish this, as it makes our code way more predictable and uncovers some bugs and enables us to fix them trivially. I aim to push the "correctness" of the code, planning to add in `phpstan` and getting rid of all the errors/warnings ...

PR #182, opened 2018-07-04
Parse without location info and serialize effectively

spawnia:
**PR Type**

Performance optimization

**Changes**

The way the DocumentAST was serialized before was way to costly, as the class
structure of the AST was completely serialized. To improve efficiency,
the location information is left out during parsing and the AST is stored
as an Array.

This is the preferred approach for schema caching as described in http://webonyx.github.io/graphql-php/type-system/type-language/#performance-considerations

### DIFF
--- a/src/Schema/AST/DocumentAST.php
+++ b/src/Schema/AST/DocumentAST.php
@@ -7,6 +7,7 @@ namespace Nuwave\Lighthouse\Schema\AST;
 use GraphQL\Language\Parser;
 use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\NodeList;
+use GraphQL\Utils\AST;
 use Illuminate\Support\Collection;
 use GraphQL\Language\AST\DocumentNode;
 use GraphQL\Language\AST\DefinitionNode;
@@ -23,7 +24,7 @@ use GraphQL\Language\AST\TypeExtensionDefinitionNode;
 use GraphQL\Language\AST\InputObjectTypeDefinitionNode;
 use Nuwave\Lighthouse\Support\Exceptions\DocumentASTException;
 
-class DocumentAST
+class DocumentAST implements \Serializable
 {
     /**
      * Check if documentAST is currently locked.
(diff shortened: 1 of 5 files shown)


chrissm79:
@spawnia This makes total sense, but none of my tests are passing for my local project (I have no idea how the tests are passing in the project). I won't have time till this afternoon to dive into it but I get the following error:

```
ErrorException: Trying to get property 'start' of non-object

/site/root/vendor/webonyx/graphql-php/src/Language/AST/Node.php:120
/site/root/vendor/webonyx/graphql-php/src/Language/AST/Node.php:98
/site/lighthouse-mll/src/Schema/AST/ASTHelper.php:63
/site/lighthouse-mll/src/Schema/AST/DocumentAST.php:124
/site/root/vendor/laravel/framework/src/Illuminate/Support/Collection.php:910
/site/lighthouse-mll/src/Schema/AST/DocumentAST.php:127
/site/lighthouse-mll/src/Schema/AST/DocumentAST.php:302
/site/lighthouse-mll/src/Schema/AST/DocumentAST.php:170
/site/lighthouse-mll/src/Schema/AST/ASTBuilder.php:49
/site/lighthouse-mll/src/Schema/AST/ASTBuilder.php:28
/site/lighthouse-mll/src/GraphQL.php:202
/site/lighthouse-mll/src/GraphQL.php:175
/site/root/vendor/laravel/framework/src/Illuminate/Cache/Repository.php:362
/site/root/vendor/laravel/framework/src/Illuminate/Cache/CacheManager.php:304
/site/root/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php:221
# ...
```

It seems that when it clones a definition it tries to look for the start location which is no longer defined. It looks like even if we set the `$recursive` variable to `false` it still throws the same error. I'll try to figure out what's in in my schema that causes this bit of code to run and add a test to the repo so you can see the error.

spawnia:
What version of `webonyx/graphql-php` are you on? My composer.lock says the following:

````json
            "name": "webonyx/graphql-php",
            "version": "v0.11.6",
            "source": {
                "type": "git",
                "url": "https://github.com/webonyx/graphql-php.git",
                "reference": "f438a726cd523bc584e78d866eca270165c42fd5"
            },
````

spawnia:
Maybe we should commit the `composer.lock` as well? Or is that not appropriate for a library such as this?

spawnia:
Just taking a wild guess here - it might have to do something with the object hashing? Since now objects with the same definitions may have the same hash, were they previously could have differed because of the location info.

chrissm79:
That's certainly one possible scenario

spawnia:
Well if you find a fix, feel free to edit the PR! Hope we can make this possible, as this is a massive performance improvement!

In my case, the AST shrinks to ~30% of its previous size. Memory usage was about half when executing.

chrissm79:
@spawnia Absolutely going to try to get this integrated, the benefits are pretty huge so I'll be debugging until I find a solution!

chrissm79:
@spawnia Not sure what was going on but I re-cloned into a new folder and everything worked! Nice job!

PR #186, opened 2018-07-08
Feature: Middleware Pipeline

chrissm79:
### Note: All of this code should be attributed to @olivernybroe and the great work he did on his PR (#118). I [followed up with him](https://github.com/nuwave/lighthouse/commit/788acc04eb77bf7c9dac7704b5bc43b2d54af893#comments) to ensure it was okay to to create this PR on his behalf.

**Changes**

From the original PR notes:

> I have changed Arg directive to use a Laravel Pipeline instead.
This give us the power to actually define if a arg directive should be executed last or if any directive after it should be run also, just like when using Laravel's middlewares.

Originally the value/middleware was intended to be similar to the request/middleware from Laravel. This PR gives us the benefits the middleware pipeline in Laravel provides. There's some additional work to be done on the `FieldResolver` and the generation of Node/Field/Arg values but this provides a great first step.

**Breaking changes**

Any current Node/Field/Arg middleware's handle function(s) will need to pass the value object back into the pipeline for further processing.

For example:

```php
// previous field middleware
public function handleField(FieldValue $value)
{
    return $value->setResolver(function () {
        // ...
    });
}

// new field middleware
public function handleField(FieldValue $value, \Closure $next)
{
    return $next($value->setResolver(function () {
        // ...
    }));
}
```

### DIFF
--- a/src/Schema/Directives/Args/BcryptDirective.php
+++ b/src/Schema/Directives/Args/BcryptDirective.php
@@ -2,8 +2,9 @@
 
 namespace Nuwave\Lighthouse\Schema\Directives\Args;
 
-use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
+use Closure;
 use Nuwave\Lighthouse\Schema\Values\ArgumentValue;
+use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
 use Nuwave\Lighthouse\Support\Contracts\ArgMiddleware;
 
 class BcryptDirective extends BaseDirective implements ArgMiddleware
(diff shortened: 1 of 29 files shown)


spawnia:
How do we want to handle versioning with this?

This affects everyone who uses custom Middleware. The required change is relatively small but still, it is a breaking change. If we were strict about semantic versioning, this would have to be v3

chrissm79:
I've been battling over that for awhile now, I don't want to open the flood gates to breaking changes because we'd be bumping the major version. That being said, I still think semantic versioning is the way to go for this project. That breaks from the way Laravel does things, but I think it's safer.

With the DocumentAST, nested rules and some other features upcoming I think a `v3.0` is fitting. Without these items, I would've left this update out as bumping versions for this alone doesn't seem worth it.

spawnia:
So if i get this right, all commits to master from now on go to v3? If so, i got some major ideas for some restructuring.

chrissm79:
> I don't want to open the flood gates to breaking changes because we'd be bumping the major version

Yeah, that's why I've been considering *not* following semantic versioning... I don't want to introduce a major overhaul due to a breaking change. I would much rather call this `v2.2` if that's the case. Although it has a breaking change I don't want to start rewriting major sections of the internals for the next release.

That being said, if what you're considering can be broken down into phases we can start to incrementally change the internal structure while trying to keep the number of breaking changes to a minimum between versions (while also keeping the PRs small and focused). 

But before doing anything that would be considered a "major" overhaul, you'll want to create an issue that (in as much detail as possible) explains the why/what/how of the proposed changes.

PR #192, opened 2018-07-12
Get the model class for the directives in the same way for all directives

spawnia:
**Related Issue(s)**

Previously, the configuration option for the Model namespace was not considered in
a bunch of cases.

**PR Type**

Bugfix

**Changes**

The common method for getting a model is now reused in all applicable places, thus unifying
the behaviour that a user can expect when referencing models in directives.

**Breaking changes**

Apart from some really far fetched edge cases, no. Should a user reference a model like

	@create(model: "App\\Foo")

this would now reference `App\Models\App\Foo` if it exists.


chrissm79:
@spawnia Nice! Makes things much more consistent

Issue #187, opened 2018-07-08
[Question] Get all, without @paginate  and without custom resolver


robsontenorio:
Cant find any info on docs or issues. 

I can do this, great!

```
type Query{
   users: [User] @paginate(model: "App\\User")
}
```
But, the `count` parameter is required

```
{
  users(count: 10){
    data{
      id
      nome
    }
  }
}
```

I would like simple get all, without paginate. 

```
type Query{
   users: [User] 
}
```

But  ...
```
{
  "data": {
    "users": null
  },
  "errors": [
    {
      "message": "Class 'App\\Http\\GraphQL\\Queries\\User' not found",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

I was wondering if exists some directive to simple get all, like `@find` or `@first`. Maybe `@all`, and a custom resolver wont be required.

And ...

```
type Query{
   users: [User] @all <-----
}
```

it would allow to query simple like this

```
{
  users{
      id
      name    
  }
}
```

robsontenorio:
I have tried, but no lucky. Actually does not make sense.

```
type Query{
  users: [User] @hasMany
}
```

```
Call to a member function getKey() on null
```

spawnia:
Closed in favour of #173 

We plan on implementing this, should be in one of the next versions of Lighthouse.

PR #195, opened 2018-07-12
Introduce @all directive which simply gets all models for a field

### LINKED ISSUE #173: [Feat] Add @all field directive to query all models
At the moment, if you want to the query filter arguments, you have to have a field directive like `@paginate` or `@hasMany`.

There are cases where i want to have a list of models without pagination. Can we just extend the `@model` directive to work with fields, too?

spawnia:
**Related Issue(s)**

Resolves #173 #187 

**PR Type**

Feature

**Breaking changes**

No

### DIFF
--- a/src/Schema/DirectiveRegistry.php
+++ b/src/Schema/DirectiveRegistry.php
@@ -127,7 +127,7 @@ class DirectiveRegistry
             throw new DirectiveException("No directive has been registered for [{$name}]");
         }
 
-        return $handler;
+        return new $handler;
     }
 
     /**
--- a/src/Schema/Directives/Fields/AllDirective.php
+++ b/src/Schema/Directives/Fields/AllDirective.php
@@ -27,9 +27,8 @@ class AllDirective extends BaseDirective implements FieldResolver
      */
     public function resolveField(FieldValue $value)
     {
-        $model = $this->getModelClass();
-        return $value->setResolver(function () use ($model){
-            return $model::all();
+        return $value->setResolver(function (){
+            return $this->getModelClass()::all();
         });
     }
 }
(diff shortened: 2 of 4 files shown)


spawnia:
@chrissm79 Fixed a major bug in the second commit! I guess we did not run into it before, as it required the same directive to be used twice in a single test.

chrissm79:
Is there a need to create a new instance of every directive? I use the `@hasMany` directive 14 in my schema and it doesn't produce any problems. If there's no performance/memory penalty I guess this may be okay, but unless it's necessary I think we should do without this.

spawnia:
Let me start of by saying that i think it is really unintuitive to have the directives share an instance. So i think they should not behave that way.

If we do so, we run into problems whenever we using any instance variable inside of closures.

This works fine, even with a shared instance:

````php
    public function resolveField(FieldValue $value)
    {
		$model = $this->getModelClass();
        return $value->setResolver(function () use ($model){
            return $model::all();
        });
    }
````

This does not, as the result of `$this->getModelClass()` might change at a later point.

````php
    public function resolveField(FieldValue $value)
    {
        return $value->setResolver(function (){
            return $this->getModelClass()::all();
        });
    }
````

While instantiating a new object for each directive has a bit of a performance cost, this should be fairly minimal as the number of directives is pretty small, i guess at most a few hundred. Also, it should be mitigated as the second example also brings performance benefits: All of the work can be done lazily inside of the callback, just when it is really needed.

Skimming through the directives, i found a bunch more cases where `$this` is used inside of the returned callbacks, and bugs might appear. Sure, we can work around that if we are really careful in every single directive. But i would prefer to just skip that whole class of bugs entirely. That will make it way easier for custom directives as well!

chrissm79:
@spawnia I can certainly agree w/ that!!

PR #198, opened 2018-07-13
Simplify schema stitcher and enable partial imports

spawnia:
The main issue this solves for me is partial imports like this:

````graphql
type Query {
    #import queries.graphql
}
````
Previously, that led to a result like this:

````graphql
   foo: Foo
type Query {
    #import queries.graphql
}
````

Now, it cleanly inlines the import and replaces the `#import` marker
````graphql
type Query {
    foo: Foo
}
````

**PR Type**

Feature/Fix


spawnia:
@chrissm79 can you please look at this before adding more features? I would like to avoid resolving conflicts all over again as this PR sits here. Thanks :)

PR #215, opened 2018-07-18
Always resolve scalar types through the ScalarResolver

spawnia:
**PR Type**

Refactor/Tests

**Changes**

This makes handling much easier through the utility already present in BaseDirective.
I created additional tests to fully detail the expected behaviour when no class
or even no directive is given.

**Breaking changes**

Nope


chrissm79:
@spawnia This looks good at first glance! I'll try to dig into it tonight but I might not be able to get it merged until tomorrow morning. Thanks!!

PR #217, opened 2018-07-19
[WIP] [Refactor] Utility Classes

chrissm79:
**PR Type**

Refactor

**Changes**

Currently a WIP. Depreciating several traits in favor of classes.

**Breaking changes**

None (converted traits will be removed in the next release)

### DIFF
--- a/src/Support/Utils/DirectiveUtil.php
+++ b/src/Support/Utils/DirectiveUtil.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Utils;
+
+use GraphQL\Language\AST\DirectiveNode;
+use GraphQL\Language\AST\ListValueNode;
+use GraphQL\Language\AST\ObjectValueNode;
+use GraphQL\Language\AST\FieldDefinitionNode;
+
+class DirectiveUtil
+{
+    /**
+     * Get directive by name.
+     *
+     * @param FieldDefinitionNode $field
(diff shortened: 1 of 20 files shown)


src/Support/Utils/GlobalIdUtil.php
```diff
@@ -0,0 +1,115 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Utils;
```
spawnia: Since this has to do with execution, how about we put it into src/Execution/Utils?

src/Support/Utils/DirectiveUtil.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Utils;
```
spawnia: This could be in src/Schema/Utils? I am not a fan of the Support directory, since it tells us pretty much nothing about what the stuff in there actually does

src/Support/Utils/GlobalIdUtil.php
chrissm79: @spawnia Sounds like a good idea to me, thanks for the feedback!!

spawnia:
Hey @chrissm79 i like the direction of this. Do you mind if i resolve the conflicts on this and finish it up to get it ready to merge?

chrissm79:
Depreciate in favor of #313

PR #220, opened 2018-07-19
Update to webonyx/graphql-php 0.12

spawnia:
### Breaking changes

**Change**

Descriptions are now using the new syntax (# -> "/""")

**Effect**
The old syntax will now get empty descriptions
No error will occurr, and the switch is really easy to make

**Change**

TypeExtensions in the AST use a different class

**Effect**

Unless someone reaches really far into the internals this should not have an effect on users. Since we merge the type extensions before we apply directives that are not schema manipulating, i think the chances of this breaking someones app are really slim. If so, the
required changes are really small.

**Change**

Custom Scalars `parseLiteral` function now has one extra parameter

**Effect**

Concerns users that defined custom scalars.
Really straightforward, just a small change to the method signature.

**Change**

When trying to parse an empty type, the Parser now rejects that.

**Effect**

There is an off chance that someone may have an empty type declared in their schema, which
they then extended to add in the fields. Those users will have to change their schema.

### Summary

I think overall, while this does introduce some breaking changes, they are pretty minor and i am confident that most users will not be affected at all. Because of this, i would advocate towards adding this into v2. If we introduce those changes now, v3 will be less of a big bang and allow us to focus on more important, major changes.


chrissm79:
@spawnia Nice work on this one! 👍

PR #222, opened 2018-07-19
[Feature] .graphqlconfig file generation

chrissm79:
**PR Type**

Feature

**Changes**

Creates a `.graphqlconfig` at specified location.

**Breaking changes**

Current commands have been namespaced.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -1,7 +1,6 @@
 <?php
 
 return [
-    
     /*
     |--------------------------------------------------------------------------
     | GraphQL endpoint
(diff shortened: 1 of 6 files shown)


spawnia:
I like the idea behind namespacing the commands, as we get more and more. However, i would like to hold off on it for the moment, because:
- We currently have only a small number of commands
- Discovery of commands is easier when they have a flat structure
- Calling of commands is simple, as i can just go `php artisan lighthouse:clear` and it automatically executes the `clear-cache` command. With the namespaced command, you always have to type out the full name

So i would advocate towards keeping a flat structure for the moment and make the first word of the command unique of possible, but adding on a more complete name seperated by hyphens.

spawnia:
@chrissm79 is it fine if we close this for now?

PR #232, opened 2018-07-26
Update default-schema.graphql file

morpheus7CS:
Hey,

I think the idea of having a default schema is wonderful. I, however, ran into an issue, which I tried to address with this PR. 

Straight after installation of Lighthouse on a new Laravel install, I was thrilled to test out the default schema Lighthouse generates by default. Laravel immediately threw an error as User model couldn't be found in App\\Models namespace. While I usually keep my models in a Models directory, User is the only one I leave at the default location.

I believe the default installation of a package immediately throwing an error may present a very wrong picture of Lighthouse being non-functional or poorly maintained. I myself have been subscribed to the package updates long enough to know that it definitely isn't the case.

I, therefore, propose the default schema changes the User model location "App\\User" to reflect default Laravel location for User model and avoid the error thrown "on first spin" by a Lighthouse newcomer.

Alternatively, if you choose a different route, I believe it would be beneficial for the interested developers to be made aware of this in the installation/configuration section of the documentation.

Keep up the great work.

### DIFF
--- a/assets/default-schema.graphql
+++ b/assets/default-schema.graphql
@@ -1,21 +1,21 @@
 type Query {
-    users: [User!]! @paginate(type: "paginator" model: "User")
-    user(id: ID @eq): User @find(model: "User")
+    users: [User!]! @paginate(type: "paginator" model: "App\\User")
+    user(id: ID @eq): User @find(model: "App\\User")
 }
 
 type Mutation {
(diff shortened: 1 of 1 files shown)


chrissm79:
@morpheus7CS Thanks for this!!

PR #235, opened 2018-07-30
[Feat] Exception handler

olivernybroe:
**Related Issue(s)**
#81

**PR Type**

Feature & refactoring

**Changes**

- Custom exception handler

This adds support for a kinda Laravel looking exception handler, where you can define how exceptions should be rendered and how they should be reported

- Errorable interface

This trait is related to the new exception handlers. If an exception implements this trait, it will be rendered based on this method.

- Validation messages as separate messages

Each Validation message will now be displayed as a separate message.

**Breaking changes**

- Validation exceptions are now shown as individual messages.
- Validate directive has support for messages.

**WIP**
Some tests are missing.

- [x] Missing tests for implementing Errorable interface


config/config.php
```diff
@@ -116,4 +116,17 @@
...
+    |--------------------------------------------------------------------------
+    | Handlers
+    |--------------------------------------------------------------------------
+    |
+    | This defines the paths for the handlers used in the program. You can
+    | extend them or replace them for customization.
+    |
+    */
+    'handlers' => [
+        'error' => \Nuwave\Lighthouse\Support\Exceptions\Handler::class
```
spawnia: Why not use the container class for that?

src/Support/Contracts/Errorable.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
+
+use Nuwave\Lighthouse\Support\Exceptions\Error;
+
+interface Errorable
+{
+    public function toError() : Error;
+}
```
spawnia: newline

src/Support/Exceptions/Error.php
```diff
@@ -0,0 +1,60 @@
...
+     * @return array
+     */
+    public function toArray() : array
+    {
+        return array_merge($this->custom, [
+            'message' => $this->message,
+            'locations' => $this->locations,
+        ]);
+    }
+}
```
spawnia: newline

tests/Integration/Schema/Directives/Args/ValidateDirectiveTest.php
```diff
@@ -0,0 +1,75 @@
...
+use Tests\Utils\Models\User;
+
+class ValidateDirectiveTest extends DBTestCase
+{
+    use RefreshDatabase;
+
+    /**
+     * @test
+     */
+    public function it_can_add_messages_to_validation()
```
spawnia: camelCase

tests/Integration/Schema/Directives/Args/ValidateDirectiveTest.php
```diff
@@ -0,0 +1,75 @@
...
+                data {
+                    id
+                }
+            }
+        }';
+
+        $result = $this->execute($schema, $query);
+        $this->assertCount(2, $result['errors']);
+    }
+}
```
spawnia: newline

tests/Integration/Schema/Directives/Args/ValidateDirectiveTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $result = $this->execute($schema, $query);
+        $this->assertEquals("custom message", $result['errors'][0]['message']);
+    }
+
+
+    /**
+     * @test
+     */
+    public function it_can_throw_validation_exceptions()
```
spawnia: camelCase

spawnia:
So far, i do not see why we need a config option for the error handler? It could be done via the container class by resolving from an Interface.

Have you considered the @rules directive?

And yeah, test methods are camelCase in the rest of the project, and you should always add a newline at the end of each file.

olivernybroe:
@spawnia You are right, not need for it. Completely forgot that we had the container.

Considered what about the `@rules` directive?

spawnia:
The functionality you added to the ValidateDirective is actually already part of the RulesDirective. They do serve different purposes: @rules allows to add rules to a particular argument, @validate does point a complete field to a validator class.

The `handleArgument` part of the ValidateDirective exists rather for compatibility reasons but will be deprecated in v3.

src/Support/Exceptions/NestedError.php
```diff
@@ -0,0 +1,42 @@
...
+        return new NestedError(
+            collect($exception->errors())->map(function (array $messages) {
+                return collect($messages)->map(function (string $message) {
+                    return new Error($message);
+                })->values()->all();
+            })->values()->flatten()->all()
+        );
+    }
+
+}
```
spawnia: newline. This stuff kinda matters, as it leads to ugly diffs if it is not in there. Github rightfully marks it as missing.

spawnia:
Looking better and better! You can check out my PR in the docs repo to see more about the difference between @rules and @validate https://github.com/nuwave/lighthouse-docs/pull/24

src/Support/Exceptions/NestedError.php
olivernybroe: Haha so sorry :) Keeps forgetting adding it all the places...

olivernybroe:
@spawnia I do know the differences. But sure I can remove it, just thought it made sense to make them work identically on arguments until we remove it.

spawnia:
I can absolutely agree with making them work the same. How about changing it to this, but leaving out the extra test case for it?

```php
    /**
     * Resolve the field directive.
     *
     * @param ArgumentValue $value
     * @param \Closure       $next
     *
     * @return ArgumentValue
     * @deprecated Will be removed in favour of the RulesDirective
     */
    public function handleArgument(ArgumentValue $value, \Closure $next)
    {
        return graphql()->directives()->get('rules')->handleArgument($value, $next);
    }
```

olivernybroe:
@spawnia That doesn't work, as rules directive uses another keyword for the args, which means it would be a breaking change.

src/Schema/Factories/FieldFactory.php
```diff
@@ -193,9 +193,7 @@ protected function wrapResolverWithValidation(\Closure $resolver, Collection $in
                     ]
                 );
 
-                if ($validator->fails()) {
-                    throw with(new ValidationError('validation'))->setValidator($validator);
-                }
+                $validator->validate();
```
chrissm79: Since we're always throwing a `Illuminate\Validation\ValidationException` here is there any way to handle the error output for this? 

For instance, on the frontend of a project we're placing all the validation errors in the `errors.0.validation` array along w/ setting `errors.0.status_code` = 422. Am I able create my own Error Handler class/callback to be used in place of the default Handler to handle custom output (i.e., putting all validation errors inside the first error returned from the server)?

If we don't currently have a way of doing that in this PR it seems that we could do the following:

```php
class Handler
{
    protected $handler = null;

    public function setHandler(\Closure $handler) 
    {
        $this->handler = $handler;
    }

    /**
     * @param array|GraphError[] $errors
     * @return array
     */
    public function handler(array $errors) : array
    {
        if ($this->handler) {
            return call_user_func($this->handler, $errors);
        }
        
        // ...
    }
}
```

and make the Handler a singleton so in a service provider we could do the following:

```php
app(\Nuwave\Lighthouse\Support\Exceptions\Handler::class)
    ->setHandler(function (array $errors) {
        // custom error handler logic...
    });

// or, instead of creating a singleton we could do this:

app('graphql')->exceptionHandler()->setResolver(function (array $errors) {
    // ...
});
```

If I completely spaced out and missed this, my apologizes! Looking at the `itCanThrowValidationExceptions` it just looks like multiple validation errors will put multiple errors in the response and since `ValidationException` are always handled the same way we might not have an option to override the output.

Great work on this PR @olivernybroe!!!

olivernybroe: The handler is resolved with Laravel, so you can just make a binding for it. However we can make an interface and use that if that is more intuitive when making a binding.

To get the old way of showing exceptions, just extend the handler and bind your extended class in the container. Then override render method and call parent render at the end. Here you get the exception, so just make an if for ValidationException and specify how you want it rendered.

chrissm79: Awesome, let's go the interface route 👍

src/Support/Contracts/ExceptionHandler.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
+
+interface ExceptionHandler
+{
+    public function handler(array $errors) : array;
+}
```
spawnia: newline

olivernybroe: @spawnia Haha sorry! :D I always always forget that...

chrissm79:
@olivernybroe Awesome work on this, it looks great! Thanks!!

spawnia:
@olivernybroe I would love to see these new features documented, can you add a guide to the docs? https://github.com/mll-lab/lighthouse-docs/tree/master/docs

If you need assistance in working with the docs, feel free to shoot me a message on Slack.

olivernybroe:
@robsontenorio I was actually working on it. But we refactored it all, so it works in a new way. 
We will be glad to help you with it anyways, you can [come to our Slack](https://lighthouse-php.slack.com/join/shared_invite/enQtMzc1NzQwNTUxMjk3LWI1ZDQ1YWM1NmM2MmQ0NTU0NGNjZWFkMTJhY2VjMDAwZmMyZDFlZTc1Mjc3ZGY0MWM1Y2Q5MWNjYmJmYWJkYmU) and we can push you in the right direction :) 

Also look into this https://github.com/nuwave/lighthouse/blob/master/config/config.php#L127-L139

robsontenorio:
Thanks!!

PR #237, opened 2018-07-30
fix filters when use the same key multiple times. closes #231

### LINKED ISSUE #231: Cannot apply two where directives
**Describe the bug**
I cannot apply two where directives when i use the same key.

**Expected behavior**
Apply two where clauses.

**Schema**

```graphql
type Campaign @model {
  id: ID! @globalId
  posts(
    since: Date @where(clause: "whereDate", key: "publish_at", operator: ">=")
    until: Date @where(clause: "whereDate", key: "publish_at", operator: "<=")
  ): [Post!]!
    @hasMany(type: "relay")
}
```

**Result query**

```sql
SELECT
    *
FROM (
    SELECT
        *
    FROM
        `shr_posts`
    WHERE
        date(`publish_at`) <= '2018-07-31'
    LIMIT 10 offset 0
) AS `posts`
```

**Environment**

Lighthouse Version: v2.2-beta.1
Laravel Version: 5.5
PHP Version: 7.2.7

hosmelq:
**Related Issue(s)**

#231 

**PR Type**

Bugfix

**Changes**

I create a composed key for the where directive based in the argument name and key name.

**Breaking changes**

None

### DIFF
--- a/src/Support/Database/QueryFilter.php
+++ b/src/Support/Database/QueryFilter.php
@@ -61,6 +61,10 @@ class QueryFilter
     public function filter($query, array $args = [])
     {
         foreach ($this->filters as $key => $filter) {
+            if (strpos($key, '.')) {
+                $key = explode('.', $key)[1];
+            }
+
             if (! isset($args[array_get($filter, 'key', $key)])
                 && 0 === count(array_intersect(array_get($filter, 'resolveArgs', []), array_keys($args)))) {
                 continue;
--- a/src/Support/Database/QueryFilter.php
+++ b/src/Support/Database/QueryFilter.php
@@ -61,9 +61,7 @@ class QueryFilter
     public function filter($query, array $args = [])
     {
         foreach ($this->filters as $key => $filter) {
-            if (strpos($key, '.')) {
-                $key = explode('.', $key)[1];
-            }
+            $key = last(explode('.', $key));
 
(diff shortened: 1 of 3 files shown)


src/Support/Database/QueryFilter.php
```diff
@@ -61,6 +61,10 @@ public static function build($query, array $args)
     public function filter($query, array $args = [])
     {
         foreach ($this->filters as $key => $filter) {
+            if (strpos($key, '.')) {
```
spawnia: Would add a comment here:

src/Support/Traits/HandlesQueryFilter.php
```diff
@@ -17,11 +17,16 @@
      */
     protected function injectFilter(ArgumentValue $argument, array $filter)
     {
+        $argName = $argument->getArgName();
         $key = $this->queryFilterKey($argument);
         $query = QueryFilter::getInstance($argument);
 
+        if ($argName != $key) {
```
spawnia: How about we always do this and eliminate this check here and the check in QueryFilter.php?

spawnia:
Looks good overall, i think we might improve on simplicity a bit. The test is great, and once the questions are out of the way, i would be happy to see this merged.

spawnia:
LGTM can be merged!

chrissm79:
@hosmelq Thank you for your work on this, good stuff!!

PR #246, opened 2018-08-06
Add a description for the default DateTime scalar

maarten00:
**Related Issue(s)**

None

**PR Type**

Bugfix

**Changes**

When using the scalar DateTime type in your schema, no description pops up when using introspection. When someone views the docs they will not be able to see which format is returned. 

**Breaking changes**

None


chrissm79:
Thanks @maarten00!!

PR #250, opened 2018-08-07
[WIP] [FIX] Batch Key

### LINKED ISSUE #247: I cannot access to a field twice.
**Describe the bug**
I cannot access to a field twice.

**Schema**

```graphql
type Campaign @model {
  posts: [Post!]! @hasMany(type: "relay")
}

type Post @model {
  campaign: Campaign! @belongsTo
}
```

```graphql
{
  node(id: "Q2FtcGFpZ246MzIy") {
    ... on Campaign {
      _id
      id
      posts(first: 1) {
        edges {
          node {
            _id
            id
            campaign {
              _id
              id
              posts(first: 1) {
                edges {
                  node {
                    _id
                    id
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

**Output/Logs**

```json
{
  "data": {
    "node": {
      "_id": 322,
      "id": "Q2FtcGFpZ246MzIy",
      "posts": {
        "edges": [
          {
            "node": {
              "_id": 69,
              "id": "UG9zdDo2OQ==",
              "campaign": {
                "_id": 322,
                "id": "Q2FtcGFpZ246MzIy",
                "posts": null
              }
            }
          }
        ]
      }
    }
  }
}
```

**Environment**

Lighthouse Version: 2.2-beta.1
Laravel Version: 5.5
PHP Version: 7.2.7

**Additional context**
As you can see in the output, the latest `posts` field is null, i was looking into this and i found that the problem is read the [value](https://github.com/nuwave/lighthouse/blob/master/src/Support/DataLoader/BatchLoader.php#L59).

I added a log in every run before the return.

```php
logger(array_keys($this->keys[$key]));
```

```php
[
    0 => 'relation',
    1 => 'parent',
    2 => 'args',
    3 => 'scopes',
    4 => 'type',
    5 => 'value',
]
```

```php
[
    0 => 'relation',
    1 => 'parent',
    2 => 'args',
    3 => 'scopes',
    4 => 'type',
]
```

in the second run the value is not set.

chrissm79:
**Related Issue(s)**

Fixes #247 

**PR Type**

Bugfix

**Changes**

Creates a batch loader key using the entire patch rather than the field name.

*Need to add a test before merging*

**Breaking changes**

None

### DIFF
--- a/src/Support/DataLoader/BatchLoader.php
+++ b/src/Support/DataLoader/BatchLoader.php
@@ -3,8 +3,8 @@
 namespace Nuwave\Lighthouse\Support\DataLoader;
 
 use GraphQL\Deferred;
-use GraphQL\Type\Definition\ResolveInfo;
 use Illuminate\Database\Eloquent\Model;
+use GraphQL\Type\Definition\ResolveInfo;
 
 abstract class BatchLoader
 {
@@ -33,7 +33,7 @@ abstract class BatchLoader
      */
     public static function key(Model $root, $relation, ResolveInfo $info = null)
     {
-        $path = ! empty(data_get($info, 'path')) ? array_last($info->path) : $relation;
+        $path = ! empty(data_get($info, 'path')) ? implode('_', $info->path) : $relation;
 
         return camel_case($root->getTable().'_'.$path);
     }


Nvservicing:
Hi there, is batchloader enabled by default?

spawnia:
Where it makes sense, yes. Nested relations of lists may be grouped together.

PR #252, opened 2018-08-08
Use Facade\Request instead of request()

WeMustFly:
**Related Issue(s)**

I had the error:
```
Symfony\Component\Debug\Exception\FatalThrowableError: Call to undefined function Nuwave\Lighthouse\request() in .../vendor/nuwave/lighthouse/src/GraphQL.php:159
```

**PR Type**

Bugfix

**Changes**

I use `Illuminate\Support\Facades\Request;` as you use `Illuminate\Support\Facades\Cache;`


chrissm79:
@WeMustFly Thank you for this!

Issue #202, opened 2018-07-16
[Question] 


vantuanprimelabo:
I want to you help me, I can't install lighthouse.
It's Symfony \ Component \ HttpKernel \ Exception \ MethodNotAllowedHttpException
No message when i press url : http://127.0.0.1:8000/public/graphql
 protected function methodNotAllowed(array $others)
    {
        throw new MethodNotAllowedHttpException($others);
    }

vantuanprimelabo:
how to define route url?

hailwood:
It's great to see you getting involved with Lighthouse!
The github issues are for bug reports, feature requests, and issues related to the development **of** Lighthouse.

For support related questions please join the Lighthouse Slack Community as outlined in the ["get involved" section of the readme.](https://github.com/nuwave/lighthouse#get-involved).

You'll receive a faster and more helpful response there :)

PR #255, opened 2018-08-14
Enable using the GraphQL endpoint via GET by default

spawnia:
I do not see any harm in defaulting this to `true`, it will avoid a
very common issue among new users of Lighthouse.

**Related Issue(s)**

#202 

**PR Type**

Config change

**Breaking changes**

Nope, the config can easily be changed back

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -21,7 +21,7 @@ return [
     | This setting controls if GET requests to the GraphQL endpoint are allowed.
     |
     */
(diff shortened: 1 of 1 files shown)


olivernybroe:
Also this doesn't affect us who already have a config file.

chrissm79:
@spawnia Looks good to me, thanks!

PR #266, opened 2018-08-18
Fix some inconsistencies with Extensions

spawnia:
- Add type hints to the extensions classes
- fix up inconsistent config option keys
- require ext-json in composer.json

**Breaking changes**

Minor changes for custom Extensions implementations
ext-json is implicetely required already, better to make it a hard requirement.

### DIFF
--- a/src/Schema/Extensions/GraphQLExtension.php
+++ b/src/Schema/Extensions/GraphQLExtension.php
@@ -14,7 +14,7 @@ abstract class GraphQLExtension implements \JsonSerializable
      *
      * @return DocumentAST
      */
-    public function manipulateSchema(DocumentAST $current, DocumentAST $original): DocumentAST
+    public function manipulateSchema(DocumentAST $current, DocumentAST $original)
     {
         return $current;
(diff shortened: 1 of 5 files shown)


src/Schema/Extensions/GraphQLExtension.php
```diff
@@ -14,7 +14,7 @@
      *
      * @return DocumentAST
      */
-    public function manipulateSchema(DocumentAST $current, DocumentAST $original)
+    public function manipulateSchema(DocumentAST $current, DocumentAST $original): DocumentAST
```
chrissm79: Until this becomes the default way of doing things in Laravel, let's not force the user to add return types on classes they will be extending.

spawnia: Agreed when we are dealing with stuff in Laravel-Land where we are not responsible for dealing with the result (as in the Controller). In this case, we handle the return value and always, positively and strictly need a `DocumentAST` so i would advocate towards keeping it there.

PR #269, opened 2018-08-18
Make the GraphQLController actually return a response

spawnia:


src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -23,13 +24,13 @@ public function __construct(Request $request)
...
-     * Execute graphql query.
+     * Execute GraphQL query.
      *
      * @param Request $request
      *
-     * @return \Illuminate\Http\Response
+     * @return Response
      */
-    public function query(Request $request)
+    public function query(Request $request): Response
```
chrissm79: No problem with it returning a response, but let's not force the user to adopt a return type on classes they may extend.

spawnia: I am all for strict return types. How can we guarantee our code to work with extended classes if they do not have a well-defined return type?

chrissm79: The same way Laravel currently does... we don't :-) 

But seriously, I am all for strict return types as well but I don't want users to be forced to use when extending Lighthouse classes when Laravel does not enforce the same practice. So let's remove for now and if Laravel adopts them later (or just adopts them in Controllers) then we can start adding them into classes that are meant to be extended.

PR #270, opened 2018-08-18
Better error message and style fixes

spawnia:


src/Console/PrintSchemaCommand.php
```diff
@@ -32,11 +32,13 @@ public function handle()
...
         // Clear the cache so this always gets the current schema
         Cache::forget(config('lighthouse.cache.key'));
+
         $schema = SchemaPrinter::doPrint(
             graphql()->prepSchema()
         );
+
         if ($this->option('write')) {
-            $this->info('Wrote schema to the default file storage (usually storage/app)');
+            $this->info('Wrote schema to the default file storage (usually storage/app) as "lighthouse-schema.graphqls".');
```
chrissm79: is `.graphqls` a typo? Seems like it should be `.graphql` for IDEs that can read the file type.

spawnia: Its actually how the command has been before. Some tools expect `.graphqls` file ending. However, i just saw that we use `.graphql` in the config. I opened an issue to unify this. https://github.com/nuwave/lighthouse/issues/275

Let's leave this as is for now and possibly change it later once we have settled it.

PR #277, opened 2018-08-20
Fix default config for `disable_introspection`

hosmelq:
Internally `graphql-php` use strict checks for this so we need `0` instead of `false`.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -1,5 +1,7 @@
 <?php
 
+use GraphQL\Validator\Rules\DisableIntrospection;
+
 return [
     
     /*
@@ -105,7 +107,7 @@ return [
     'security' => [
         'max_query_complexity' => 0,
         'max_query_depth' => 0,
-        'disable_introspection' => 0,
+        'disable_introspection' => DisableIntrospection::DISABLED,
     ],
 
      /*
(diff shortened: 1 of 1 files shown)


spawnia:
https://github.com/webonyx/graphql-php/blob/master/src/Validator/Rules/DisableIntrospection.php

While they do set `ENABLED` to `1`, they annotate `$isEnabled` as a bool. Seems inconsistent. They actually do not have type annotations in place, so it is fine as is.

The tests pass and a boolean shows intent more clearly, so i am closing this for now. If you can come up with a failing test that forces this change, we can reopen.

hosmelq:
By default the introspection does not work, this can be a little tedious for some new people that do not know they have to change this to 0.

You can see [here](https://github.com/webonyx/graphql-php/blob/master/src/Validator/Rules/DisableIntrospection.php#L55) that this is always anable since `false !== 0`.

spawnia:
Oh i see, did not catch that. Thanks!

How about we use the constants, like `DisableIntrospection::DISABLED`?

hosmelq:
@spawnia done.

spawnia:
Thanks!

PR #280, opened 2018-08-29
Introduce the HasOneDirective and the HasOneLoader, improve BatchLoader semantics

### LINKED ISSUE #276: [Feature request] Add `@hasOne` directive
Currently there is only a `@hasMany` directive. 

`hasOne` relationship is very commonly used.  Having a `@hasOne` directive would be helpful😊

spawnia:
**Related Issue(s)**

Resolves #276

**PR Type**

Feature

**Changes**

- Add test case and scenario for a HasOne relationship
- BatchLoader classes are now initialized with a set of immutable params
- Avoid unnecessary steps inbetween for performance & code clarity
- Loaders now indicate clearly what they require as a one-off config and what is required for each field instance

**Breaking changes**

For users that do not reach into internals, nothing changes. Existing directives continue to work as before.

- The API for the BatchLoader abstract class changed so additional Loader implementations have to adapt. `resolve()` is now expected to return a keyed array of results.
- `graphql()->batch()` was changed to `graphql()->batchLoader()` and is now only responsible for returning a unique BatchLoader instance.

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -2,7 +2,6 @@
 
 namespace Nuwave\Lighthouse;
 
-use GraphQL\Deferred;
 use GraphQL\Type\Schema;
 use GraphQL\GraphQL as GraphQLBase;
 use GraphQL\Executor\ExecutionResult;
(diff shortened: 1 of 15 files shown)


chrissm79:
Looks good, thanks @spawnia!

PR #284, opened 2018-08-31
Fix running out of memory when using @inject for the user id

spawnia:
The `@inject` directive fails to inject the user id from the context, as in this example schema:

```graphql
type Query {
  user: User! @inject(context: "user.id", name: "user_id") @field(resolver: "App\\Queries\\User@resolveUser")
}
```

The `RuleFactory` seems to run into an infinite loop, as PHP crashes and runs out of memory.


tonipepperoni:
should be labelled as a bug. I'll have time next to have a go at fixing this. Thx.

tonipepperoni:
```
    protected function getRulesForPath(
        DocumentAST $documentAST,
        FieldDefinitionNode $field,
        string $path,
        bool $traverseOne
    ) {

    	if(strlen($path) === 0) {
    		return null;
	    }

```


I think it's related to this @spawnia . I was working on it today and discovered that the getRulesForPath is having issues because it gets stuck in a recursion when the $path is an empty string. If I added a string length check and escaped the recursion, then things appear to work fine. What do you think of this solution? Perhaps its related to a typing ?

spawnia:
Yeah, as i said in the description of this PR, the RuleFactory is responsible.

Your fix did indeed work - thank you very much! 🎉

chrissm79:
@spawnia Great work on this one!!

PR #286, opened 2018-08-31
Rework Exception handling

spawnia:
- Use the native way of webonyx/graphql-php to render exceptions
- Render validation errors in a spec-compatible manner
- Refactor a whole lot of code

This should really be split into a few commits/PRs, but i wanted to get this out of the way before the weekend.


olivernybroe:
Like the way it is looking so far.

Why did you remove the macro class and added all of them as a function for Collection?

spawnia:
@chrissm79 This is a pretty big one. There at most minor breaking changes to how the schema has to be defined, but Exception handling was certainly changed in a big way.

How about we release `v2.2` with what we have at the moment and merge this in right after?

yaquawa:
Looks so nice to have 💖

PR #291, opened 2018-09-03
Align resolver signature with `@field` directive

yaquawa:
The signature for default resolver is not aligned with `@field` directive's.
For more consistency, I would like to recommend make this little change.

### DIFF
--- a/src/Schema/Factories/FieldFactory.php
+++ b/src/Schema/Factories/FieldFactory.php
@@ -118,8 +118,8 @@ class FieldFactory
     {
         return function ($obj, array $args, $context = null, $info = null) use ($fieldName, $rootOperationType) {
             $class = config("lighthouse.namespaces.{$rootOperationType}").'\\'.studly_case($fieldName);
-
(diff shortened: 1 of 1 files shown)


spawnia:
This little change in code does actually mean quite a large breaking change for everybody that utilizes these resolvers.

In this case the previous approach does have some merits. The way the fallback works is that it does locate a single class per field. I imagine this feature is needed when dealing with more complex, special-case resolvers that are not covered by the existing directives.

However, i do agree that consistency would be a rather important reason for this change. It would allow for the code to be more flexible - moving a resolver that has grown too large in complexity to its own class is just a matter of copying it over. Right now, it involves rewriting it to be a class that accepts the resolvers args in the constructor.

@chrissm79 @olivernybroe what do you think?

olivernybroe:
I am all for consistency. As your say this is a breaking change, however it's not a change which requires users to do that much for making it work again, and in most of the cases it will still work.

I think it would be fine having this in the next minor version and just mention it in upgrade guide, so users know how to handle this.

Issue #290, opened 2018-09-03
[Question] How can I filter model using custom conditions with `Paginator` returned ?


yaquawa:
It's easy enough to get a list of models by just utilizing the `@paginate` directive.

But how can I apply my own filter to a model yet still get the `Paginator` type returned?

For example:
```graphql
type Query {
    # The `query` argument is optional, in that case, no filter applied.
    users(query: QueryCondition): UserPaginator
}

input QueryCondition {
    conditions: [Condition]
    relation: Relation
}

input Condition {
    key: String
    value: String
    operator: Operator
}

enum Operator {
    IN @enum(value: "IN")
    NOT_IN @enum(value: "NOT IN")
    AND @enum(value: "AND")
    EXISTS @enum(value: "EXISTS")
    NOT_EXISTS @enum(value: "NOT EXISTS")
}

enum Relation {
    AND @enum(value: "AND")
    OR @enum(value: "OR")
}
```

How can I get the same return type `Paginator` as `@paginate` gives me? 
I thought I must utilize some internal methods of Lighthouse, but I can't find which to use😂

Any help would be greatly appreciated !

spawnia:
Take a look at https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/Fields/PaginateDirective.php

However, what you are looking for is already possible with the various filter directives such as [`@eq`](https://lighthouse-php.netlify.com/docs/next/directives.html#eq) or [`@in`](https://lighthouse-php.netlify.com/docs/next/directives.html#in). The `next` version of the docs is still missing a guide that gives an overview on how to use filters.

Maybe you can get it started over at https://github.com/nuwave/lighthouse-docs ? Just add a file at `docs/guides-filtering.md` and write a few lines. Your contribution will be much welcome.

yaquawa:
@spawnia Yes, I can use some of those filter directives. However by that way I have to hard code every condition, the patterns of condition are going to explode.

spawnia:
I see what you are getting at. You can take a look at https://github.com/deInternetJongens/Lighthouse-Utils as they implemented quite a lot of filters.

Another big thing would be client-side filters. They actually seem more desirable then server-side filtering because as you said, it gives more power to the API consumer. I am not quite sure what is the best approach to implement them, client-side directives or a fixed set of arguments as you suggested might both be worth exploring.

Issue #289, opened 2018-09-02
[Feature Request] Fallback to default `Union` type resolver.


yaquawa:
Currently, if I don't specify a resolver via `@union` directive on union types, I'll get this error:

```
Union types need to have the @union directive defined to resolve them.
```

Fallback to the `config.lighthouse.namespaces.unions` to find a default resolver just like other types would be nice.😃

spawnia:
Accepting PR's on this one!

spawnia:
This is now a supported feature as of #305

PR #292, opened 2018-09-05
[Refactoring] Add `getResolver` method to `BaseDirective` class

yaquawa:
There are some potential usages of the business logic for `getResolver` from an argument around multiple places. 

For reusability, extracting the business logic into a single method would be better.

The only change is that if a method was not provided, fallback to the `resolve` method instead.

Also, I'm planning to extend the `paginate` directive, and I found I have to get the resolver too. So to preventing repeat my self, I made this PR.

### DIFF
--- a/src/Schema/Factories/FieldFactory.php
+++ b/src/Schema/Factories/FieldFactory.php
@@ -118,8 +118,8 @@ class FieldFactory
     {
         return function ($obj, array $args, $context = null, $info = null) use ($fieldName, $rootOperationType) {
             $class = config("lighthouse.namespaces.{$rootOperationType}").'\\'.studly_case($fieldName);
-
-            return (new $class($obj, $args, $context, $info))->resolve();
+            
+            return app($class)->resolve($obj, $args, $context, $info);
         };
     }
 


yaquawa:
I'm sorry I checked out from my [previous PR](https://github.com/nuwave/lighthouse/pull/291)…
Should I resend this PR again?😿or if the previous PR is OK to merge I'll keep this one.

spawnia:
@yaquawa very nice work, i dug around the code a bit more and was able to find a bunch more places where the generalized `getResolver` function can be used.

Also, i reverted the changes from PR #291 to keep this as focused as possible.

yaquawa:
Thanks @spawnia 
How wonderful…! That's much better now!👍🏼

spawnia:
This PR really pushes for more consistency among directives, as the way that functions are defined on them is now more uniform. Also, we previously did not have proper validation for schema correctness in place, which is now fixed.

Thank you for your great idea and for getting this started!

@chrissm79 I think we can go ahead and merge this, tests are passing without changes so there should be no breaking changes.

yaquawa:
@spawnia Thank YOU!

I just edited my comment a little while ago.

> I prefer `callable` to `Closure` personally, how do you think?

For example, if your resolver is a method of the current class,
instead of `Closure::fromCallable([$this,'resolve'])` I can just use `[$this,'resolve']` which is much simpler I think.

yaquawa:
I just did a research about callable and closure.
Looks like the winner is closure😃

https://why-cant-we-have-nice-things.mwl.be/requests/closure-from-callable-function

spawnia:
So `Closure` it is then?

yaquawa:
@spawnia Yes, seems like `Closure` is better!

yaquawa:
@spawnia Looks like there are some conflicts, would you fix it for me?

yaquawa:
Thank you @spawnia !
I’d appreciate it if you could get this PR merged, because I've made some enhancements related to #289 and #290 base on this PR locally. I have to wait this PR to be accepted to send the new PR.

spawnia:
As this does not entail breaking changes and the tests pass without any change, this is a fine refactor. I like the extra consistency this brings us. @yaquawa go nuts!

yaquawa:
@spawnia Hurray!! Now I can keep going on to the next one!

yaquawa:
@spawnia 
Question!

I see you using `resolve($class)->resolve($obj, $args, $context, $info); ` instead of `app($class)->resolve($obj, $args, $context, $info);` why is that? 

And the function `resolve` is a helper function comes with Laravel, why should we try to define a same function at https://github.com/nuwave/lighthouse/blob/master/src/Support/helpers.php#L84 ?

https://github.com/yaquawa/lighthouse/commit/c84411a11a98812b5e66db17a5608ec51d948288

spawnia:
It's because it makes it clear that we are just resolving a class, instead of the many other things the `app()` function can do. #286 converts all appropriate calls to `resolve()`

We define the helper because it is not available in Lumen.

yaquawa:
@spawnia Now I got it! Thanks!

PR #293, opened 2018-09-05
Add test for recursive relationships

mijgame:
**Related Issue(s)**

Currently, when querying recursive relationships, relationships in the second recursion become null. 
As discussed in the Slack channel, this is a WIP PR that currently adds a failing test that demonstrates this behaviour.

**PR Type**

Probably a bugfix, maybe a feature?

**Changes**

Support recursive relationships.

**Breaking changes**

As far as I am aware, there are no breaking changes unless someone relies on the "second layer parent is null" behaviour.

### DIFF
--- a/tests/Integration/Schema/Directives/Fields/HasManyDirectiveTest.php
+++ b/tests/Integration/Schema/Directives/Fields/HasManyDirectiveTest.php
@@ -182,7 +182,7 @@ class HasManyDirectiveTest extends DBTestCase
         }
         ';
 
-        $result = $this->executeQuery($schema, '
+        $result = $this->queryAndReturnResult($schema, '
         { 
             user { 
                 tasks(first: 2) { 
@@ -243,7 +243,7 @@ class HasManyDirectiveTest extends DBTestCase
         }
         ';
 
-        $result = $this->queryAndReturnResult($schema, '
+        $result = $this->executeQuery($schema, '
         { 
             posts {
                 id
--- a/tests/Integration/Schema/Directives/Fields/HasManyDirectiveTest.php
+++ b/tests/Integration/Schema/Directives/Fields/HasManyDirectiveTest.php
@@ -215,6 +215,52 @@ class HasManyDirectiveTest extends DBTestCase
     /**
      * @test
      */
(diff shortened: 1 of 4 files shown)


spawnia:
> As far as I am aware, there are no breaking changes unless someone relies on the "second layer parent is null" behaviour.

If someone relies on a bug, that is entirely their fault.

mijgame:
@spawnia agreed, but I thought is was worth mentioning.

spawnia:
Thank you for the PR, do you have any ideas yet on how it might be fixed?

mijgame:
@chrissm79 Thanks for your commit, the test was indeed wrong! But I'm even more puzzled by the issue now, as the same situation in my application returns `null` for second level (self-referencing) parent/children...
Currently, in my application the query type looks like this:
```js
type Query {
    categories: [Category!]!
}
```

The `Category`  model in my application is basically the same as the `Post` model in the tests; it has a `parent` relationship and a `children` relationship. The query function looks like this:
```php
  public function resolve()
  {
      return Category::with('parent')->get();
  }
```
So the only difference with the test right now is the use of the `@all` directive. When I try to use the `@all` directive in my application, I get `No directive has been registered for [all]`. I can't find it in the documentation either, is it custom to the test suite? There is a test for the `@all` directive though...

spawnia:
The query as above without any directive will not work. Can you try switching to dev-master so you have the newest version? The `@all` directive is pretty new

mijgame:
@spawnia I have updated to 2.2 and the recursive relationship works now. I don't know what caused the problem/solved it though.. Thank you for your help!

spawnia:
I would actually like this PR in, as it adds another test to our suite.

spawnia:
Fixed the test setup so it does not break our other tests.

@chrissm79 This can be safely merged now.

@mijgame Thanks for your contribution. Looking forward to your next PR. In the future,
try creating a new branch and make your PR from there. Makes it manageable to have multiple PRs open and keep your master clean/in sync with upstream.

chrissm79:
Thannks @spawnia @mijgame!!

PR #296, opened 2018-09-05
remove the unnecessary `user` from context.

yaquawa:
If the incoming request is from a guest, the `user` will be `null`, which will throw an error.

Most of the case, the API should be used by a user, however for public API, such as `createUser` or `loginUser` etc.. there is no user provided, so the `user` of context is not always guaranteed.

I dug around and found nowhere is using the `user` property of `context`, we can get the `user` by ourselves by simply calling the `Auth::user()`, so I don't think the `user` is necessary to be bound to the context object.


spawnia:
We actually need the user in the Context tor some other links, like the InjectDirective. If you want this fixed now, look at what i did with Context and the Controller in #286

yaquawa:
@spawnia WOW! That's a big one! 
You can close this PR now. I hope yours can be merged as soon as possible because I'm using the dev-master version😂

spawnia:
@yaquawa Hijacked your PR 🗡 

This fixes what was broken in #282 so you all can keep riding the bleeding edge!

yaquawa:
@spawnia Thank you very much!

PR #298, opened 2018-09-07
`BaseDirective` class enhancements

yaquawa:
Just made some enhancements to the `BaseDirective` class.

[75fae5e](https://github.com/nuwave/lighthouse/pull/298/commits/75fae5eeba1577e4c01db17f7d83cd0277766885) `str_after($this->directiveArgValue($argumentName), '@')` won't work as expected if the resolver method was not given.

For example 
```graphql
@field(resolver: "App\Models\User")
``` 
won't fallback to `App\Models\User@resolve`.
Instead, simply using the `explode` function, which also gives more simplicity.

------------------

[37bd724](https://github.com/nuwave/lighthouse/pull/298/commits/37bd7240b0ef43632fa528e0c9549062ce312357) Add more flexibility to `getModelClass` method, and uniform the Exception to be thrown.

#### Added `field name` as a fallback
For example:

```graphql
type Query {
    users: [User!]! @paginate(type: "paginator" model: "User")
}
```
You can omit the `model: "User"` part now,
because the field name `users` will be transformed into `Namespace/For/Model/User`,
and if it is a valid class, use it as the model class instead.

The fallback priority would be:
1. `model` argument
2. `model` argument + namespace
3. `field return type` + namespace
4. `field name(converted to singular form)` + namespace

#### Uniformed the Exception
Restrict the Exception to only the `DirectiveException` to clarify where the issue came from.

### DIFF
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Schema\Directives;
 
+use Closure;
 use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\ValueNode;
 use GraphQL\Language\AST\ArgumentNode;
@@ -89,14 +90,14 @@ abstract class BaseDirective implements Directive
     /**
      * Get the resolver that is specified in the current directive.
      *
-     * @param \Closure $defaultResolver Add in a default resolver to return if no resolver class is given.
+     * @param Closure $defaultResolver Add in a default resolver to return if no resolver class is given.
      * @param string $argumentName If the name of the directive argument is not "resolver" you may overwrite it.
      *
      * @throws DirectiveException
      *
-     * @return \Closure
+     * @return Closure
      */
-    protected function getResolver(\Closure $defaultResolver = null, string $argumentName = 'resolver'): \Closure
+    protected function getResolver(Closure $defaultResolver = null, string $argumentName = 'resolver'): Closure
     {
         $baseClassName =
             $this->directiveArgValue('class')
@@ -122,7 +123,7 @@ abstract class BaseDirective implements Directive
             throw new DirectiveException("Method '{$resolverMethod}' does not exist on class '{$resolverClass}'");
         }
 
-        return \Closure::fromCallable([app($resolverClass), $resolverMethod]);
+        return Closure::fromCallable([app($resolverClass), $resolverMethod]);
     }
 
     /**
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -3,6 +3,7 @@
 namespace Nuwave\Lighthouse\Schema\Directives;
 
 use Closure;
+use Exception;
 use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\ValueNode;
 use GraphQL\Language\AST\ArgumentNode;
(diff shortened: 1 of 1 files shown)


spawnia:
Thank you for this PR, unfortunately i do not agree with most changes you made.

> Going from `\Closure` -> `Closure`

The first one immediately makes it clear what class is meant. We will keep it like that, and i actually refactored some of our other code to do the same. #300

> str_after() vs explode()

I added a comment which explains what we are doing clearly. Your suggestion could lead to
an `Undefined offset` error.

> Fallback to singularized field name for the type

Falling back to the fields return type is enough magic, if that is not the correct type then it is better to have the user specify it explicitely, rather then doing even more magic.

spawnia:
You did however uncover a bug with the second fallback, after we attempt to find the model in the default model namespace. I fixed it in #299 

Thanks!

yaquawa:
Hi @spawnia ! Thank you for taking your time to review my PR.

> Use the root namespace explicitly if it is a buit-in type.

Got that!

> Avoid black magic

I see LOL…

> str_after() vs explode()

However, as for this one, firstly I'm using the `??` operator here, so there shouldn't raise any errors.

Secondly, `str_after('Foo@bar', '@')` will return 'bar', 
but `str_after('Foo', '@')` will return 'Foo' however.
If you expect the user to explicitly specify a method after the resolver class, this still is a potential bug, which means that if the class 'Foo' has a `Foo` method, the `Foo` method will be chosen instead of the default `resolve` method.

And, should we really have to restrict the user to specify a method explicitly? If you have a default method for the `class` argument why not give one to `resolver`  too?

Lastly,  I'm curious to know if the returned value is intended:
```php
protected function getModelClass(): string
    {
        $model = $this->directiveArgValue('model');
        // Fallback to using the return type of the field as the class name
        if(! $model && $this->definitionNode instanceof FieldDefinitionNode){
            $model = ASTHelper::getFieldTypeName($this->definitionNode);
        }
        if (! $model) {
            throw new DirectiveException("A `model` argument must be assigned to the {$this->name()} directive on {$this->definitionNode->name->value}");
        }
        if(class_exists($model)){
            return $model;
        }
        $modelWithDefaultNamespace = config('lighthouse.namespaces.models').'\\'.$model;
        if(class_exists($modelWithDefaultNamespace)){
            return $modelWithDefaultNamespace;
        }



        // The result may not be a valid class.
       // Is this is the intended behavior?
        return $this->namespaceClassName($model);
    }
```

spawnia:
> Lastly, I'm curious to know if the returned value is intended: // The result may not be a valid class.

Actually, it is. `namespaceClassName` throws if it is not.

> but str_after('Foo', '@') will return 'Foo' however.

Wow, i did not expect that. I made the switch to explode, fits better here. Thank you!

> And, should we really have to restrict the user to specify a method explicitly? If you have a default method for the class argument why not give one to resolver too?

We do not.

```php
        $resolverMethod =
            $this->directiveArgValue('method')
            ?? $resolverArgumentFragments[1]
            ?? 'resolve';
```

I would love a review on #299 if something is missing.

spawnia:
Yeah, it is set by the `@group` directive. This has to do with the way that Lighthouse handles schema extension.

For better performance, some directives are applied to the schema before it is cached in a step we call schema manipulation. The result is an intermediary schema that is closer to the executable schema (but not quite, since the actual executable schema can not contain server-side directives).

So yeah, `@namespace` is not meant for the end user - it is not in the docs - but only for internal purposes.

yaquawa:
@spawnia Lucid!! Thanks for your explanation! 
Lighthouse is great, I'd like to see more internal architecture topics in the doc however!

It's too magic for newcomers like me. If the doc is as detailed as vue.js's, I believe more people would like to choose this library instead of others.

spawnia:
@yaquawa That is why we need newcomers like you to document their learning process by adding to the docs 😉

PR #302, opened 2018-09-07
[Enhancement] Add `resolver` argument to paginate directive

yaquawa:
## Summary
* Update PHP to 7.1 (to use more strict type hinting)
* Add PHP native return type
* Add `resolver` argument to allow user define their own resolver for the field

## The `resolver` argument
See the discussions at #290 

This PR allows user to define their own resolver instead of the default `model` solution.

The purpose is clear:  **Increasing the flexibility**.

In real world, chances are that user would like to have their own query conditions to filter the result set.

Below is a real example of mine:

```graphql
type Query {
    users(query: QueryCondition): [User!]!  @paginate(type: "paginator" resolver:"App\\GraphQL\\Queries\\Users") @middleware(checks: ["auth:api"])
}

input QueryCondition {
    conditions: [Condition!]!
    relation: Relation = "AND"
    order: Order
    orderby: String
}

input Condition {
    key: String!
    value: String!
    operator: Operator!
}

enum Order {
    ASC @enum(value: "ASC")
    DESC @enum(value: "DESC")
}

enum Relation {
    AND @enum(value: "AND")
    OR @enum(value: "OR")
}

enum Operator {
    IN @enum(value: "IN")
    NOT_IN @enum(value: "NOT IN")
    EQUALS @enum(value: "=")
    NOT_EQUALS @enum(value: "!=")
    LIKE @enum(value: "LIKE")
    NOT_LIKE @enum(value: "NOT LIKE")
    GREATER_THAN @enum(value: ">")
    GREATER_THAN_OR_EQUAL @enum(value: ">=")
    LESS_THAN @enum(value: "<")
    LESS_THAN_OR_EQUAL @enum(value: "<=")
}
```

In `Users::resolve` simply return an instance of `Illuminate\Database\Eloquent\Builder`.
```php

use App\Models\User;
use Illuminate\Database\Eloquent\Builder;

class Users
{
    public function resolve($root, array $args): Builder
    {
        $eloquentBuilder = User::query();

        // query how ever you like ...

        return $eloquentBuilder;
    }
}
```

As you can see, this way, the API consumer can specify the query throw the `query` variable themselves, which can bring massive flexibility to the client side.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "source": "https://github.com/nuwave/lighthouse"
     },
     "require": {
-        "php" : ">= 7.0",
+        "php" : ">= 7.1",
         "ext-json": "*",
         "illuminate/contracts": "^5.4",
         "illuminate/http": "^5.4",
--- a/src/Schema/Directives/Fields/PaginateDirective.php
+++ b/src/Schema/Directives/Fields/PaginateDirective.php
@@ -2,16 +2,21 @@
 
 namespace Nuwave\Lighthouse\Schema\Directives\Fields;
 
+use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Language\AST\FieldDefinitionNode;
 use GraphQL\Language\AST\ObjectTypeDefinitionNode;
 use Illuminate\Pagination\Paginator;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Contracts\Pagination\LengthAwarePaginator;
+use Nuwave\Lighthouse\Schema\Context;
 use Nuwave\Lighthouse\Execution\QueryUtils;
 use Nuwave\Lighthouse\Schema\AST\DocumentAST;
 use Nuwave\Lighthouse\Schema\Values\FieldValue;
-use Nuwave\Lighthouse\Support\Contracts\FieldManipulator;
-use Nuwave\Lighthouse\Support\Contracts\FieldResolver;
-use Nuwave\Lighthouse\Exceptions\DirectiveException;
 use Nuwave\Lighthouse\Support\Traits\HandlesGlobalId;
+use Nuwave\Lighthouse\Support\Contracts\FieldResolver;
+use Nuwave\Lighthouse\Support\Contracts\FieldManipulator;
+use Nuwave\Lighthouse\Exceptions\DirectiveException;
+
 
 class PaginateDirective extends PaginationManipulator implements FieldResolver, FieldManipulator
 {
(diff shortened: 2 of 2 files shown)


yaquawa:
Looks like the new master branch has a lot of failed test ?

spawnia:
How about we rename `resolver` to `builder` in this case, and have it be a method that sits on the Model itself? This should make the intent more clear.

Also, you will need to provide tests that shows how the directive is supposed to work, preferrably also covering the not-so-happy path.

spawnia:
Just had one more thought about this. I really like the idea of giving more power to the client, to enable them to define how the Query is run. We should look into generalizing this concept to more directives that query models, so `@all`, `@paginate`, `@first` and `@hasMany` should be able to use this.

The options for what you can query are pretty well defined, so we may just offer the types you have put in your schema above as a part of the default schema for Lighthouse. Then, instead of putting an additional argument on the directive, you could just add the `QueryConstraints` argument to the field, and the directives that support it may look at it.

yaquawa:
@spawnia 
> How about we rename resolver to builder in this case, and have it be a method that sits on the Model itself? This should make the intent more clear.

Do you mean this ?
```graphql
type Query {
    users(query: QueryCondition): [User!]!  @paginate(type: "paginator" builder: "User")
}
```

and in the `User` calss:
```php
interface Buildable {
    public static function build();
}

class User implements Buildable
{
    public static function build($root, $args, $context, $info): Builder
    {
        // build query or maybe use a `build` trait...
    }
}
```

> We should look into generalizing this concept to more directives that query models, so @all, @paginate, @first and @hasMany should be able to use this.

oh… I just forgot we have many similar directives…
As you said, the other similar directives should have this kind of `queryable` ability too!!

yaquawa:
> The options for what you can query are pretty well defined, so we may just offer the types you have put in your schema above as a part of the default schema for Lighthouse.

**Great idea!**
I think the `Condition` input type is little bit verbose, how about just an array like this?
```
[key, operator(optional), value]
```

> Then, instead of putting an additional argument on the directive, you could just add the `QueryConstraints` argument to the field, and the directives that support it may look at it.

I didn't get this. Do you mean something like this?
```
type Query {
    users: [User!]!  @paginate(type: "paginator" queryConstraints:"true")
}
```

yaquawa:
@spawnia 
I played around and found maybe this is the best solution for me.

```graphql
type Query {
    users: [User!]!  @paginate(type: "paginator" model: "User" queryable: true)
}
```

spawnia:
> `users: [User!]!  @paginate(type: "paginator" model: "User" queryable: true)`

I like that one, as it gives the most flexibility to the end user. It will be more enjoyable
for the back-end programmer too, as they will not have to update their GraphQL server each
time somebody wants a slightly different query.

> `users(query: QueryCondition): [User!]!  @paginate(type: "paginator" querybuilder: "className@build")`

Since pagination is pretty hard to wire up yourself, this might be nice to have. The argument should be `queryBuilder` though, or perhaps just `builder`, and just point to any method that returns a `Builder` instance. That enables use cases where any DB queries can be paginated, even complex joins or whatever.

yaquawa:
@spawnia opened an issue at #312 !

spawnia:
@yaquawa Superseded by #326 

Thanks for the inspiration!

robsontenorio:
@yaquawa can you share your custom paginator builder/resolver that applies all query conditions?

PR #304, opened 2018-09-07
[Refactor] Remove return types on interfaces

chrissm79:
**Related Issue(s)**

None

**PR Type**

Refactor

**Changes**

Removes return types on interfaces that may be implemented by the user. This allows the user to choose to add a return type on their implementation rather than forcing them to do so.

**Breaking changes**

None

### DIFF
--- a/src/Schema/Source/SchemaSourceProvider.php
+++ b/src/Schema/Source/SchemaSourceProvider.php
@@ -3,13 +3,7 @@
 namespace Nuwave\Lighthouse\Schema\Source;
 
 /**
- * Interface SchemaSourceProvider
- * @package Nuwave\Lighthouse\Schema\Source
- *
- * This interface can be implemented to have different methods of providing
- * a schema string. Since the schema is then parsed into an AST and cached,
- * this does not necessarily have to be very performant, e.g. it may be
- * acceptable to download the schema or fetch it from an API.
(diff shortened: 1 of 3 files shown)


src/Schema/Source/SchemaSourceProvider.php
```diff
@@ -3,13 +3,7 @@
 namespace Nuwave\Lighthouse\Schema\Source;
 
 /**
- * Interface SchemaSourceProvider
- * @package Nuwave\Lighthouse\Schema\Source
- *
- * This interface can be implemented to have different methods of providing
```
spawnia: Why delete the comment?

spawnia:
I do not think it is beneficial for the user to omit the return type. In this case, we are passing the result of the interface function right down to `webonyx/graphql-php`.

Let's suppose the user messes up and does not return an array. If we have a strong type hint, they will get a clear error message that tells them exactly where they did something wrong and what. If we do not have it, the error will probably occur somewhere down the line in the underlying library - or even worse - silently fail to output a result.

chrissm79:
@spawnia The purpose is to let the *user* decide if they want to utilize a return type. Laravel does not enforce this anywhere (to my knowledge) in it's codebase, even when it's passing results down to Symfony. If the user decides to leverage return types they can do so but we shouldn't force them to do so.

PR #305, opened 2018-09-08
Rework type resolving for Interfaces, Unions and the global id Node interface

yaquawa:
Related to #289

Now user can omit the @union directive and have their default resolver placed to `lighthouse.namespaces.unions` directory just like other definition types.

Also rearranged some of the namespace importing statements to have them grouped by vendor.


src/Schema/Directives/Nodes/UnionDirective.php
```diff
@@ -4,8 +4,11 @@
 
 use GraphQL\Type\Definition\Type;
 use GraphQL\Type\Definition\UnionType;
+use GraphQL\Type\Definition\ResolveInfo;
```
spawnia: Sort by length

src/Schema/Directives/Nodes/UnionDirective.php
```diff
@@ -46,10 +42,26 @@ function() use ($value){
...
+
+                    if ( ! \method_exists($resolver, 'resolve')) {
+                        throw new DirectiveException("Class {$resolverClass} must have a `resolve` method.");
+                    }
+
+                    return $resolver->resolve($value, $context, $info);
+                }
+
+                return resolve(TypeRegistry::class)->get(
+                    str_after('\\', \get_class($value))
```
spawnia: I just noticed that this is a pretty severe bug (that is still in the previous version, it seems). Apparently, no one actually uses Union types 🙊 

Anyways, can you please provide a test that shows that this works, and fix this one?

src/Schema/Directives/Nodes/UnionDirective.php
```diff
@@ -46,10 +42,26 @@ function() use ($value){
                     return resolve(TypeRegistry::class)->get($type->name->value);
                 })->filter()->toArray();
             },
+            'resolveType' => $this->getResolver(function ($value, Context $context, ResolveInfo $info) {
```
spawnia: Resolvers have 4 arguments

src/Schema/Directives/Nodes/UnionDirective.php
```diff
@@ -46,10 +42,26 @@ function() use ($value){
...
-            'resolveType' => function ($value) use ($resolver) {
-                return $resolver($value);
-            },
+                // Try to locate a fallback resolver corresponds to the config file.
+                $resolverClass = config('lighthouse.namespaces.unions') . '\\' . $unionName;
+
+                if (\class_exists($resolverClass)) {
+                    $resolver = resolve($resolverClass);
+
+                    if ( ! \method_exists($resolver, 'resolve')) {
```
spawnia: Call this with the class name before resolving the class, it will avoid unnecessary work.

src/Schema/Directives/Nodes/UnionDirective.php
```diff
@@ -46,10 +42,26 @@ function() use ($value){
...
+                        throw new DirectiveException("Class {$resolverClass} must have a `resolve` method.");
+                    }
+
+                    return $resolver->resolve($value, $context, $info);
```
spawnia: Again, not the right args for resolvers.

src/Schema/Factories/NodeFactory.php
```diff
@@ -4,22 +4,24 @@
 
 use GraphQL\Type\Definition\Type;
 use GraphQL\Type\Definition\EnumType;
+use GraphQL\Type\Definition\UnionType;
```
spawnia: By length

yaquawa: @spawnia  oh.. I just saw many PHP files in this project actually are sorted by length, and some of them sorted by vendor, if this is the rule I'll follow 👍

src/Schema/Directives/Nodes/UnionDirective.php
yaquawa: @spawnia 

What's wrong with this one…?
Do you mean If the $value is not an object get_class($value) won't get any class?

spawnia: The order of the arguments in `str_after` is reversed

src/Schema/Directives/Nodes/UnionDirective.php
spawnia: Alright, can you add a line comment to our code please?

yaquawa: Sure!

src/Schema/Directives/Nodes/UnionDirective.php
yaquawa: Got that!

yaquawa:
@spawnia Thanks for your review👍🏼 
I just added the test, and fixed some of the issues.

tests/Unit/Schema/Directives/Nodes/UnionTest.php
```diff
@@ -22,9 +20,12 @@ public function itResolvesUnionUser()
             }
         }
         ';
-        $result = $this->executeQuery($this->schema(), $query);
 
-        $this->assertEquals('user.id', array_get($result->data, 'person.id'));
+        foreach ($this->schemas() as $schema) {
```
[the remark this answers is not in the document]
yaquawa: Can you show me some code?
I believe that the singleton never be cleared only if you refresh the app container itself. Or just  rebind the instance itself.
That's why I made this another PR #306 for this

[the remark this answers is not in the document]
spawnia: Solved the issue with the tests so closed #306

spawnia:
@yaquawa I really like the idea behind this whole PR, it makes schema definition even easier and pushes for more consistency.

I expanded your PR to Interfaces and eliminated some code duplication in between them and brought them closer together. Did not get to finish my work today, i will do so tomorrow.

yaquawa:
@spawnia Cool!! Thanks for that!!👍🏼

spawnia:
Hey @yaquawa thanks for your comment.

The rebind is not necessary when you use data providers. PHPUnit rebuilds a fresh context with each set of data. So we do not actually have multiple schemas in one run.

spawnia:
And thank you very much for the great inspiration you provided by starting this PR!

You have come to the project with a fresh set of eyes and immediately found some inconsistencies and points where we can improve. I hope what i made out of your PR is in line with what you originally had in mind.

yaquawa:
@spawnia 
> PHPUnit rebuilds a fresh context with each set of data.

wow, I didn't even notice that, good to know!

> And thank you very much for the great inspiration you provided by starting this PR!

Thank you for your kind words, it's a great pleasure to hear that!💖
Probably this is the best solution I can find in the Laravel land currently, hope that more people can involve to this project.

Still a little busy now, but I'm thinking if I can translate the doc into Japanese and Chinese, so more people can benefits from this project. But the problem is that the current doc is still not that friendly for newcomers😅

spawnia:
@yaquawa @chrissm79 

This is ready for review and can probably be merged. I hope the code is self-evident,
i can provide a short write-up tommorow. Docs will follow after this is final.

chrissm79:
@spawnia @yaquawa Going through this PR now! First question, why are we getting rid of the `@interface` directive completely? I certainly see the value in having a default namespace and creating a nice command line helper to create a class to handle a single interface, but it seems that we lose the ability to define a single class for all interfaces. 

For example:

```graphql
type User @interface(resolver: "App\\InterfaceResolver@employee") {
  # ...
}

type Admin @interface(resolver: "App\\InterfaceResolver@employee") {
  # ...
}

type Post @interface(resolver: "App\\InterfaceResolver@content") {
  # ...
}

type Comment @interface(resolver: "App\\InterfaceResolver@content") {
  # ...
}
```

```php
namespace App;

class InterfaceResolver
{
    public function employee($value)
    {
        if ($value instanceof Admin) {
            return schema()->instance('Admin');
        } 
        
        return schema()->instance('User');
    }

    public function content($value)
    {
        if ($value instanceof Comment) {
            return schema()->instance('Comment');
        } 
        
        return schema()->instance('Post');
    }
}
```

This is a large PR so I may be missing the implementation to define a single interface resolver, and my apologizes if I did! But if this forces users to define a resolver for every individual interface, I say we should keep the `InterfaceDirective` (same goes for the `UnionDirective` if that's the case, just haven't made it that far yet).

spawnia:
You got that correctly, i removed the directive for Union, Scalar and Interfaces. They do now simply look in the default namespace for a class with the same name and for a `resolveType` function on that class. The approach i have been taking with this favours convention over configuration, which in my opinion is quite Laravel-like.

In most cases, the default type resolver will suffice, so you do not actually need to write any PHP.

chrissm79:
@spawnia I always used Interfaces over Unions so I never caught that error either (nice catch!). For now, let's add the Interface and Union directives back and I'll push an update so they both use a single method to create the interface (and a separate method for union) type(s) so we can allow the use of a single class resolver.

spawnia:
```graphql
type User @interface(resolver: "App\\InterfaceResolver@employee") {
  # ...
}
```

Just noticed, why is the @interface directive on the User?

chrissm79:
Because I have a single `users` table that includes a `json` column. If the user is an `Admin` they have additional settings. The vast majority of my interfaces are done as so... they usually have a json column and based on what type they are they may have different json fields.

spawnia:
I think it should be:

```graphql
interface Employee @interface(resolver: "App\\InterfaceResolver@employee")

type User implements Employee {
  # ...
}

type Admin implements Employee {
  # ...
}
```

chrissm79:
Ah, yes, that's my bad... didn't want to copy and paste from my project :-)

src/Providers/LighthouseServiceProvider.php
```diff
@@ -81,9 +81,12 @@ function () {
 
         if ($this->app->runningInConsole()) {
             $this->commands([
-                \Nuwave\Lighthouse\Console\ClearCacheCommand::class,
```
chrissm79: Was this meant to be deleted?

spawnia: Good catch, that was an accident!

spawnia:
I think that having a 1:1 relationship between Interfaces and classes is the most simple and clear way. That way, there is no room for ambiguity and it should be easy to manage. I think that the best approach for Lighthouse is to just implement one canonical way of doing things but allow backdoors for the user to plug into if they really need it.

However, i see how in your case it might be nice to have a common Interface class that holds all the typeResolvers. How about we add the directives back in but mark them as deprecated for v3? They can always be added back in by the user, but i would stop supporting them officially.

chrissm79:
Well if we're using the same code (which I will add) to create the Interfaces and Unions either by a directive or the NodeFactory why do we need to get rid of it? We find it quite handy to have the definitions in a single file as most of the logic to determine the type is just a few lines of code and it seems wasteful to create a separate class for each of them.

It's actually just like the NodeFactory... we wouldn't extract each function to resolve a type into it's own class.

spawnia:
@chrissm79 Added them back in.

chrissm79:
Nice work on this one @spawnia and @yaquawa!

yaquawa:
@spawnia Thanks for the rework!!

and I agree with @chrissm79 recently I realized that it's very handy to have resolver for the developer who wants to develop a package for Lighthouse, if there is no directive for it, they have to inject the resolver into Lighthouse, which is pretty tricky.

PR #306, opened 2018-09-08
[Enhancement] Add a `clearCache` method to enable clear document AST cache

yaquawa:
Currently we can only clear the cache from artisan command.
This PR adds a `clearCache` method to enable clear document AST cache programmatically.

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -197,6 +197,19 @@ class GraphQL
     }
 
     /**
+     * Clear the documentAST cache.
+     *
+     * @return GraphQL
+     */
+    public function clearCache(): self
+    {
+        $this->documentAST = null;
+        Cache::forget(config('lighthouse.cache.key'));
+
+        return $this;
+    }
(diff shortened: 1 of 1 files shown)


spawnia:
Can you give me an example of why you would want to use this instead of `Artisan::call('lighthouse:clear-cache')`?

I would like to avoid adding additional public methods that users may depend on if we do not really need them.

yaquawa:
@spawnia For test purpose. 
If you want to test multiple schemas in a single test session, you have to rebind the `graphql` singleton to clear the documentAST cache.
`Artisan::call('lighthouse:clear-cache')` won't set the `$this->documentAST` to `null`, so there is no way to refresh the documentAST cache from outside.

spawnia:
This is not really needed right now, as PHPUnit data providers solve the issue with tests. Closing this, unless an actual need for it comes up.

Until then, i prefer to keep the API as small and clean as possible.

PR #313, opened 2018-09-10
Deprecate HandlesGlobalId trait

spawnia:
- Global ID related methods were moved into Nuwave\Lighthouse\Execution\Utils\GlobalId
- Move cursor related methods to Nuwave\Lighthouse\Execution\Utils\Cursor
- Added tests and replaced calls to new util classes

**Related Issue(s)**

Supersedes #217 

Thank you so much @chrissm79, as this is based on the great work you did already.

**PR Type**

Refactor

**Breaking changes**

Removed option to overwrite the GlobalID Utils.

@chrissm79 I am open to adding them back in, but since the config options were not documented, i figured we could keep it simple and remove them.


chrissm79:
Great work @spawnia, looks good!

spawnia:
Thanks @chrissm79. I think moving away from Traits makes our code more clear, testable and DRY.

Looking forward to seeing that test coverage once we delete all the deprecated classes/methods.

PR #317, opened 2018-09-14
Make the `load` and `tryRegisterClassName` to public method in `DirectiveRegistry` class

yaquawa:
I'm creating a package for Lighthouse, so I have to load my directives into the `DirectiveRegistry`, but I don't want to repeat myself…  so does other developers I think.

### DIFF
--- a/src/Schema/DirectiveRegistry.php
+++ b/src/Schema/DirectiveRegistry.php
@@ -32,16 +32,17 @@ class DirectiveRegistry
 
     /**
      * Create new instance of the directive container.
+     * @throws \ReflectionException
      */
     public function __construct()
     {
         $this->directives = collect();
 
         // Load built-in directives from the default directory
-        $this->load(realpath(__DIR__.'/Directives/'), 'Nuwave\\Lighthouse\\');
+        $this->load(realpath(__DIR__ . '/Directives/'), 'Nuwave\\Lighthouse\\', \dirname(__DIR__));
 
         // Load custom directives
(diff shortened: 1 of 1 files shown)


src/Schema/DirectiveRegistry.php
```diff
@@ -130,157 +156,119 @@ public function get($name)
...
+     * Get all directives of a certain type that are associated with an AST node.
      *
      * @param Node $node
+     * @param string $directiveClass
      *
-     * @return \Illuminate\Support\Collection
+     * @return Collection
      */
-    protected function directives(Node $node)
+    protected function associatedDirectivesOfType(Node $node, string $directiveClass): Collection
```
yaquawa: Question, what does the "associated" in the method name mean?
was it meant for this line `$this->hydrate($directive, $node);`?

yaquawa:
@spawnia Nice clean up!

src/Schema/DirectiveRegistry.php
spawnia: I caught myself adding a comment to the PHPDoc that explains how this gets just the directives that are associated with the specific node that is passed into this function. As a general rule of thumb, i try to avoid such comments so i just made the method name more explicit.

yaquawa: I think `directivesOfType` is clear enough. It's confusing with the "associated" added, because user have to think "associate to what ?" for that name. 
At least, I got confused💦.

PR #318, opened 2018-09-15
Delete unnecessary assignment in DocumentAST.php

yaquawa:
I don't think we need that assignment.


spawnia:
@yaquawa You were right, the code also was a bit hard to follow in that place, so i polished it a bit. It is a tiny bit more efficient now, and easier to understand.

Thank you!s

PR #321, opened 2018-09-15
[WIP] [Refactor] Type Factory

chrissm79:
**Related Issue(s)**

#305 

**PR Type**

Refactor

**Changes**

Reintroduces resolvers for Interface and Union directives as well as adding the ScalarDirective back into Lighthouse. These are needed to help support 3rd party defined types as well as introducing a TypeFactory to serve as a single source to create these supported types.

**Breaking changes**

None

### DIFF
--- a/src/Execution/GraphQLValidator.php
+++ b/src/Execution/GraphQLValidator.php
@@ -12,9 +12,9 @@ class GraphQLValidator extends Validator
     /**
      * Run the validator's rules against its data.
      *
-     * @return array
-     *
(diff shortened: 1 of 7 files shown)


spawnia:
I added the ScalarDirective back in, as well as some tests.

The tests in NodeFactoryTest, InterfaceTest and UnionTest show how those types can be defined. They also demonstrate that the original functionality of overriding the class location or the type resolver function was still kept.

PR #322, opened 2018-09-15
Add ScalarDirective back in and tests that show how Scalars and Unions can still use custom classes and resolvers

spawnia:
Also, the InterfaceTest shows how Interfaces can use custom type resolvers as well.

**Related Issue(s)**

#321 

**PR Type**

Regression Fix, Tests


chrissm79:
Looks good, thank @spawnia!

Issue #319, opened 2018-09-15
hasMany directive is ignoring Anonymous Global Scopes on model


trasteroandy:
Hello,

I have this line in a graphql type:

```
cuentas: [Cuenta]! @hasMany
```

and the Laravel Model "Cuenta" has declared an [Anonymous Global Scope](https://laravel.com/docs/5.7/eloquent#query-scopes):

```
class Cuenta extends Model
{
    protected static function boot()
    {
        parent::boot();
        static::addGlobalScope('habilitado', function (Builder $builder) {
            $builder->where('habilitado', '=', true);
        });
    }
```

It works fine in normal code in laravel, but when the client (Apollo) calls the graphql server, the response is using some query that **ignores the global scope** "habilitado".

**Expected behavior**

Don't know if LightHouse use the default relations of the model, but the query that LightHouse uses to retrieve "cuentas" should include:
```
where habilitado = true
```
and it doesn't.

**Environment**

Lighthouse Version: 2.1
Laravel Version: 5.6
PHP Version: 7.2

Thanks a lot, best regards

spawnia:
Can you try my fixed branch https://github.com/mll-lab/lighthouse/tree/native-load-with and see if that does it for you?

trasteroandy:
Hi, I have tried that fixed branch, but got the error:

No directive has been registered for [scalar]

becouse the file "lighthouse/src/Schema/Directives/Nodes/*ScalarDirective.*
php" has been removed. I have this code in my schema.graphql definition:

scalar DateTime @scalar(class:
"Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\DateTime")


How should I proceed?
thanks in advance

spawnia:
Hey @trasteroandy, thanks for hanging in there!

I just pushed & merged a hotfix, since #320 was just a WIP and merged by accident. I added a test case to the https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Schema/Directives/Fields/HasManyDirectiveTest.php which does show that the global scope is indeed called in this case.

Sorry about that thing with the `@scalar` directive, that regression was fixed in #322. You should be able to try `dev-master` now and see if your Scope is called.

If not, please report back with some more details - or even better: Add a failing test to https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Schema/Directives/Fields/HasManyDirectiveTest.php

Thanks for getting involved.

trasteroandy:
[...]

PR #323, opened 2018-09-16
 Revert #320 and use fetch() instead of load() again

spawnia:
**Related Issue(s)**

Adds test for #319 
Reverts misguided WIP #320 

https://github.com/laravel/framework/issues/16217

**PR Type**

Test/Fix

**Changes**

- Add test to ensure the global scope is called when using the @hasmany directive
- Add TODO for removing custom fetch logic once Laravel resolves their issue with eager loading

**Breaking changes**

Nope

### DIFF
--- a/src/Providers/LighthouseServiceProvider.php
+++ b/src/Providers/LighthouseServiceProvider.php
@@ -96,6 +96,8 @@ class LighthouseServiceProvider extends ServiceProvider
      */
     protected function registerCollectionMacros()
     {
+        // TODO remove and just use load() as soon as Laravel fixes https://github.com/laravel/framework/issues/16217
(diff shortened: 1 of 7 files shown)


spawnia:
@chrissm79 #320 was a misguided WIP, reverted it.

Your solution for fixing eager loading via `fetch` is actually brilliant, i think it should be part of Laravel core so i added TODO's that remind us of removing it once that happens 😉

chrissm79:
@spawnia Thanks for reverting, I didn't even catch that! And yeah, I dropped that note in the Laravel repo quite a while ago but hopefully your update will get things going, I would love to see it make its way into Laravel core as well!!

PR #326, opened 2018-09-16
Allow passing a custom builder to @paginate instead of a model

spawnia:
Supersedes #302

This enables the user to use the complex logic that is within the @paginate directive
with a completely custom query.

- Add 'builder' argument to @paginate
- Add test that verifies a custom builder can be called

**Deprecation**

Deprecate alternate way of defining a field resolver through class/method

**Breaking changes''

Fail hard if no valid resolver argument was found but an argument for specifying such
was given. This might cause some users to get Errors, but in a good way - since it means
their previous definition did not work.

They can then either fix the directive argument or simply omit it to use the default, if possible.

**Refactor**

- Refactor BaseDirective->getResolver() to a more general getMethodArgument that contains
less magic and is more composable
- Add directiveHasArgument method to both BaseDirective and ASTHelper
- Refactor existing classes to use the new methods
- More types

### DIFF
--- a/src/Schema/AST/ASTHelper.php
+++ b/src/Schema/AST/ASTHelper.php
@@ -115,6 +115,21 @@ class ASTHelper
     }
 
     /**
+     * Does the given directive have an argument of the given name?
+     *
+     * @param DirectiveNode $directiveDefinition
+     * @param string $name
+     *
+     * @return bool
+     */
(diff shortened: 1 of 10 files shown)


chrissm79:
Love it! Nice work @spawnia!

yaquawa:
@spawnia Thanks! I really need this one!

Issue #236, opened 2018-07-30
@globalId doesn't work for relations in mutations


jordyvandomselaar:
**Describe the bug**
We have the following GraphQL code:

```
type Mutation{
	createFoo(relation_id: ID! @globalId, other_field: String) @create(model: "Foo")
}
```

When we run this `mutation` in `graphIQL` we get a `QueryException`:

> invalid input syntax for integer: "UmVsYXRpb246MQ=="

It appears the `@globalId` doesn't work on relation `ID` fields in mutations.

**Expected behavior**
The mutation to run

**Schema**

```graphql
type Mutation{
	createFoo(relation_id: ID! @globalId, other_field: String) @create(model: "Foo")
}
```

**Output/Logs**

```
invalid input syntax for integer: "UmVsYXRpb246MQ=="
```

**Environment**

Lighthouse Version: dev-master
Laravel Version: 5.6
PHP Version: 7.2

**Additional context**
I can't show you the actual data, so the above data are examples.

PR #328, opened 2018-09-17
Allow using @globalID on args for decoding

spawnia:
**Motivation**

Resolves #236

The docs displayed that usage, but it did not actually work.

**Changes**

The `@globalId` can now be used on Arguments to decode a global id. The value that is then passed along
is an array that consists of [$type, $id].

This is used in the definition of the `node` endpoint, that was simplified as a result:
It is now defined in the ASTBuilder again (going back and forth on this one :D ), which in the end
is a cleaner solution that also allows to leverage caching.

Did some formatting cleanups on the go, as i usually do.

**Breaking changes**

Decoding the global ID on fields is now no longer a configurable option. This is because
global IDs are internal, and fields only ever output values, so all they need is encoding.

### DIFF
--- a/src/Execution/Utils/GlobalId.php
+++ b/src/Execution/Utils/GlobalId.php
@@ -67,22 +67,4 @@ class GlobalId
         list($type, $id) = self::decode($globalID);
         return $type;
     }
-    
-    /**
-     * This function is the resolver for global id fields.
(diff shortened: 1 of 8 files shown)


chrissm79:
Nice work @spawnia!!

Issue #333, opened 2018-09-19
QueryBuilder is not using the connection specified on the model


jeroenpf:
**Describe the bug**

The QueryBuilder always uses the default connection, regardless whether or not an Eloquent model has the $connection property set to a different connection than the default. 

**Expected behavior**

The QueryBuilder should respect the $connection property in the Eloquent model (if set) and use it to run the query.

PR #334, opened 2018-09-19
Fix for #333 - Use connection specified on the model in QueryBuilder

jeroenpf:
**Related Issue(s)**

Resolves [#333](https://github.com/nuwave/lighthouse/issues/333)

**PR Type**

Bugfix

**Changes**

Use the connection set on the model when building the query in the QueryBuilder


spawnia:
Thanks, good work!

SiamKreative:
Good stuff, thanks 👍

PR #341, opened 2018-09-20
Allow schema import via glob in SchemaStitcher

### LINKED ISSUE #314: [Feat] Schema import multiple files via glob
I`m using import file syntax as described

https://lighthouse-php.netlify.com/docs/schema-extending.html

I`m organizing files in multiples folder, and this works as expected

schema.graphql
```
#import ./post/post.graphql
#import ./post/category.graphql

.... and a dozen of files will be imported ....
```

Is it possible a folder import syntax like this?

```
#import ./post/*
```

enzonotario:
**Related Issue(s)**

Resolves #314 

**PR Type**

Feature

**Changes**

If the file name to import has a wildcard, it loads via `glob`.

**Breaking changes**

No breaking changes.

### DIFF
--- a/src/Schema/Source/SchemaStitcher.php
+++ b/src/Schema/Source/SchemaStitcher.php
@@ -60,9 +60,18 @@ class SchemaStitcher implements SchemaSourceProvider
                 }
 
                 $importFileName = trim(str_after($line, '#import '));
-                $importFilePath = realpath(dirname($path).'/'.$importFileName);
 
-                return self::gatherSchemaImportsRecursively($importFilePath);
+                if (! str_contains($importFileName, '*')) {
+                    $importFilePath = realpath(dirname($path).'/'.$importFileName);
+
+                    return self::gatherSchemaImportsRecursively($importFilePath);
+                }
+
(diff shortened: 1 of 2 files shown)


chrissm79:
Great work on this @enzonotario!!!

robsontenorio:
Nice @enzonotario ! So, what is the final import syntax?

enzonotario:
Thanks!

In your case, that you have a folder `routes/graphql/post` with 2 files, it could be:
```
#import post/*.graphql
```

Now I will try to add this to docs.

robsontenorio:
@enzonotario  A noob question: why not auto import all files placed on `routes/graphql/*` ?

Is there any scenario where i could not require all schema files?

enzonotario:
Hey, we are 2 noobs then!

I don't know. But maybe it is better to let to the developer to decide.

Maybe anyother here can answer you.

spawnia:
@robsontenorio imports have to be explicit, as it can sometimes matter where inputs are placed. This also enables you to reuse parts of your schema definition, for example a common list of fields or arguments in multiple types.

Auto-importing all would only work if all files only had complete type definitions.

PR #342, opened 2018-09-20
Description print

### LINKED ISSUE #297: Description displays incorrectly
For some reasons, the description looks like this😂
(I'm using the `dev-master` ver.)

image

### LINKED ISSUE #336: Paginator default resolver fallback

/* A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] */

I think that define paginate builder in schema file is not beautiful code.

/* A clear and concise description of what you want to happen. */

Auto call custom paginator it is defined.

/* A clear and concise description of any alternative solutions or features you've considered. */

Like query resolver same logic.For example auto call function config('lighthouse.namespaces.queries')\Users@paginator.

/* Add any other context or screenshots about the feature request here. */

spawnia:
**Related Issue(s)**

Resolves #297
Resolves #336 

**Changes**

This makes it so that we now convert descriptions correctly everywhere they occur. Previously, they showed up bugged or not at all in a bunch of places. #297

- Fix description display globally
- Enable descriptions for args
- Enable setting a default value for args

To enable that, i had to rework how Arguments are converted as well as how Query filters are injected.

- `@whereBetween` and `@whereNotBetween` can now be used on the same column in the same field.
- All query directives may now specify a custom key if needed (i considered renaming it to column?)

Also, i found some general bugs inbetween:

- Fixed a bug that prevented `@paginate` to be used without explictly passing a model class. #336 
- Only add default value to client side directive if given, previously it was set to null if not given

- Added more types
- Some formatting

**Breaking changes**

Only internals are affected, schema's work same as before. The changes should be pretty minor.

- The API to add in query filters changed slightly but is way easier now. Users that added their own Query filters should be able to convert them quickly.
- Aligned the signature for `ASTHelper::getNamespaceForDirective` with the other functions in that class: $haystack, $needle
- Moved QueryFilter into `Nuwave\Lighthouse\Execution` where the other query stuff sits. This does not affect custom directives, as they only used it indirectly.
- The ArgumentValue now has different properties and methods.


spawnia:
@chrissm79 This required bigger changes then i originally thought. I could have done a 80 % solution for descriptions only, but decided to go all in.

Thanks in advance for digging in this quite large PR 😜

chrissm79:
@spawnia Thanks for all the work on this, I'm going to try and get this merged in this afternoon after reviewing!

assets/default-schema.graphql
```diff
@@ -1,3 +1,4 @@
+"A date string with format 'Y-m-d H:i:s', e.g. '2018-01-01 13:00:00'"
```
robsontenorio: According graphql spec it would be """A date string with format ...""" (triple double quote). Am i wrong?

spawnia: Both are allowed.

chrissm79:
Great work on this one @spawnia!

PR #343, opened 2018-09-21
[Bug fix]PartialParser can not parse `argument`

yaquawa:
A quick bug fix for PartialParser.


spawnia:
Good catch, tahnk you!

PR #344, opened 2018-09-21
[FIX] Interfaces not providing possible types

chrissm79:
**PR Type**

Bugfix

**Changes**

Currently just a test to show that interfaces do not provide a `possibleType`

**Breaking changes**

None (at the moment)


spawnia:
This seems like it could be a bug in webonyx

chrissm79:
@spawnia I'm starting to come to that conclusion as well.

chrissm79:
Created an issue upstream (will try to add a PR w/ a failing test there as well)

https://github.com/webonyx/graphql-php/issues/355

spawnia:
Let's integrate the Schema into our TypeRegistry. Having the types in both is wasteful.

chrissm79:
yep, agreed

chrissm79:
@spawnia actually, i may have gotten ahead of myself here... won't we run into a circular reference issue if we don't have a [registry](http://webonyx.github.io/graphql-php/type-system/object-types/#recurring-and-circular-types) that holds references to the type instances?

spawnia:
I have an idea for that in mind. Am on mobile right now so i can not type it out fully. I will get back to this in a few hours, hang tight in the meantime

chrissm79:
@spawnia Since this is a quick hotfix I'm going to merge this one in but create a new refactor PR and make any updates to the TypeRegistry there and we'll get it merged in. Thanks!!

Issue #346, opened 2018-09-24
Error when using $with as a string instead of an array


robjbrain:
I've been getting this error:

> GraphQL Error: {"code":0,"message":"array_filter() expects parameter 1 to be array, string given","trace":"#0 [internal function]: Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'array_filter() ...', '/home/vagrant/c...', 225, Array)
> #1 /home/vagrant/code/main/vendor/nuwave/lighthouse/src/Support/DataLoader/QueryBuilder.php(225): array_filter('position', Object(Closure))

It relates to this file and line: https://github.com/nuwave/lighthouse/blob/4db038a53a713769e341533e91ce6f684d2be383/src/Support/DataLoader/QueryBuilder.php#L250

The cause is if you have a with like this:

```php
class Stadium extends Model
{
    // This will cause the error
    protected $with = 'stands';

    // This is fine
    protected $with = ['stands'];
```

But `$with` can be a string and a string does work (aside from lighthouse).

I would suggest checking that it's an array first? Or possibly just forcing it to be an array?

`array_filter((array)$withProperty->getValue($model)`

PR #347, opened 2018-09-24
Forced the withProperty to an array type.

robjbrain:
**Related Issue(s)**

Fixing this issue: https://github.com/nuwave/lighthouse/issues/346
**PR Type**

Bugfix

**Changes**

Forced the $with to be an array


spawnia:
Can you please add a test that shows your fix is working to ensure we have no regression in the future?

spawnia:
Worked your fix into #348

spawnia:
Merging this as a hotfix, #351 showed this is a pressing issue.

PR #348, opened 2018-09-24
A whole bunch of fixes i accumulated over the last few days

### LINKED ISSUE #346: Error when using $with as a string instead of an array

spawnia:
**Changes**

- Fix default model fallback for `@paginate`
- Performance optimizations
- Throw clearer error messages more early on schema definition errors
- Add missing query validation back in
- Work in the fix from #347, Resolves #346
- The QueryBuilder now works with SQLite too
- Fix the travis tests, as they were having time-outs when generating coverage
- Run an actual test matrix with PHP 7.0, 7.1 and 7.2 as well as Laravel 5.5, 5.6 and 5.7
- Ensure PHP 7.0 compatibility that was previously broken
- (De-)serialize the DocumentAST complete with the $locked property
- Refactorings, reformatting, comments

**Breaking changes**

The generated types for `@paginate` and `@hasMany` do now have different names. This does not affect queries, but could affect users that somehow rely on it via introspection? Do not think it is really an issue.

The TypeRegistry now throws when a type is not found instead of failing silently. This will only help uncover bugs.

Queries are now validated properly. Again, this will only uncover bugs.

Laravel 5.4 was kicked out of `composer.json`. This is not an actual breaking change, as the tests were not really using 5.4 before. As i made the switch, i realized we are incompatible in many places - too much to fix in my opinion.

### DIFF
--- a/src/Support/DataLoader/QueryBuilder.php
+++ b/src/Support/DataLoader/QueryBuilder.php
@@ -214,12 +214,9 @@ class QueryBuilder
             $withProperty = $reflection->getProperty('with');
             $withProperty->setAccessible(true);
 
-            $with = array_filter(
-                $withProperty->getValue($model),
-                function ($relation) use ($model) {
-                    return ! $model->relationLoaded($relation);
-                }
-            );
+            $with = array_filter((array) $withProperty->getValue($model), function ($relation) use ($model) {
+                return ! $model->relationLoaded($relation);
+            });
 
             if (! empty($with)) {
                 $collection->load($with);


mfn:
This PR is getting out of hand! ✋ 

j/k :D

spawnia:
@mfn You are absolutely right. I keep slipping from one issue to the next though, and i tend to dislike solving them with some kind of workaround 😃

chrissm79:
@spawnia Just an update on this while I'm reviewing, the update is really slowing down the performance on my internal project. I wouldn't be too concerned about non-cached performance, but the cached schema is still 2x slower than a before the update so the performance issue seems to be outside of schema generation. I'm working on getting subscriptions updated but if I can I'll dig around and see if I can narrow down what's causing the performance issues. 

| Measurement | Before Update | After Update |
| --- | --- | --- |
| Query (non-cached) | 410ms | 1.4sec |
| Query (cached) | 60ms | 132ms |
| Test Suite | 32sec | 1.6min |

spawnia:
The issue was that i switched calling `lock()` on the DocumentAST before storing it to cache. However, the serialization did not actually store that attribute, so as the DocumentAST was restored from cache, it was marked as unlocked. This led to operations being way slower.

I changed the DocumentAST so that it also serializes whether it is locked. I feel that is the cleanest solution, as it is reasonable to expect it to be the same after deserializing it.

chrissm79:
@spawnia Great progress on this! I'm still looking into something... seems that a M-to-M polymorphic relationship (`morphToMany`) has stopped working in my queries. I only have one of those relationships but I have a test w/ a query for it and it fails w/ an error message stemming from the `QueryBuilder ` that I'm trying to debug.

Error:
`Trying to get property 'my_foreign_key' of non-object`

Trace:
```json
[{
    "file": "/root/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php",
    "line": 271,
    "call": "Illuminate\\Foundation\\Bootstrap\\HandleExceptions::handleError(8, 'Trying to get property \\'my_foreign_key\\' of non-object', '/root/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php', 271, array(3))"
  },
  {
    "file": "/root/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php",
    "line": 241,
    "call": "Illuminate\\Database\\Eloquent\\Relations\\BelongsToMany::buildDictionary(instance of Illuminate\\Database\\Eloquent\\Collection(1))"
  },
  {
    "file": "/root/lighthouse-mll/src/Support/DataLoader/QueryBuilder.php",
    "line": 121,
    "call": "Illuminate\\Database\\Eloquent\\Relations\\BelongsToMany::match(array(1), instance of Illuminate\\Database\\Eloquent\\Collection(1), 'myRelationship')"
  },
  {
    "file": "/root/lighthouse-mll/src/Support/DataLoader/QueryBuilder.php",
    "line": 77,
    "call": "Nuwave\\Lighthouse\\Support\\DataLoader\\QueryBuilder::loadRelation(instance of Illuminate\\Database\\Eloquent\\Builder, instance of Closure, array(1), array(4))"
  },
  {
    "file": "/root/lighthouse-mll/src/Providers/LighthouseServiceProvider.php",
    "line": 107,
    "call": "Nuwave\\Lighthouse\\Support\\DataLoader\\QueryBuilder::eagerLoadRelations(instance of Illuminate\\Database\\Eloquent\\Builder, array(1))"
  }
  # ...
]
```

spawnia:
@chrissm79 Replicated the failure you mentioned through a test, found the root cause and fixed it!

I threw out a bit too much from the QueryBuilder, but did not catch the regression because there were no tests in place. It is nice to have you and your big project as a backup, although i think that it is most healthy for Lighthouse to have tests of its own. I did work one in, so we are fine in the future.

Thanks for your thorough review. I hope we are good now.

PR #349, opened 2018-09-27
Allow adding to the base schema via `BuildingAST` event

### LINKED ISSUE #338: What is the best way to insert my schema to the base schema?
Hi, I'm building a package for Lighthouse currently(schema auto generation based on data model).
And I have a use case for inserting my own schema into the base schema. 

Currently, I think I have only two options to achieve that.

1. Rebinding the `SchemaSourceProvider`.
2. Create a directive that implements `NodeManipulator`.

Both have its limitation and weak points.

## 1. Rebinding the `SchemaSourceProvider`.
This is good for who want to just attach their schema without having to worry about the base schema.

but sometimes, there may come to a case that developers want to have some knowledge about the base schema. 

For example, if a developer wants to insert their own scalar type definitions to the base schema, but they should make sure it won't conflict with the base schema. So it is required to search all the scalar types in base schema, and check if it has existed if it has then skipped that one. 
This is just a simple example; there should have more and more complex cases. So when you want to conditionally insert your schema base on the base schema, you want to do it with solution 2(below).

And even a simple case that does not require a conditional inserting, the problem occurs also when you use multiple packages developed for Lighthouse. Which is.. only the very last rebinding works... rebinding before the last one will just be ignored.

So for simple schema inserting the solution can be this; dispatch an event for just that purpose.

in `GraphQL.php:193`
```php
protected function buildAST(): DocumentAST
{
    // Listeners of `BuildingAST` event are expected to return their schema string.
    $additionalSchemas = implode("\n", (array)event(new BuildingAST($this)));
    $schemaString = $this->schemaSourceProvider->getSchemaString();

    return ASTBuilder::generate($schemaString . $additionalSchemas);
}
```

## 2. Create a directive that implements `NodeManipulator`.
This is needed if I want to conditionally insert my schema as referred above.

But, the problem is…
1. The usage of `NodeManipulator` is intended for just directive only, so you do this means you created an internal directive which has no meaning by itself, bad for semantics.
2. The current API only allows you to insert the node one by one by using the `setDefinition` in `Schema/AST/DocumentAST.php:330`, which is obviously inefficiency for massive nodes inserting. There should have an API for massive insertion too.

I want to know if there are better ways for this, and if not what you think about this?

yaquawa:
Resolves #338


yaquawa:
@spawnia Thanks for your review! 
Yeah, it should be described in docs too. "developer's guide book" something like that.

chrissm79:
@yaquawa @spawnia Great work on this addition!

PR #353, opened 2018-10-01
Removed dependency on Facades

### LINKED ISSUE #352: Forced use of Facades is back..
Hi all,

The following commits re-introduced the dependency on Facades, resulting in RunTimeExceptions if Facades are not enabled: "A facade root has not been set."

commit 1050eb6c9be6978a9268ac43304fc9b9763e57c4
commit c7e448859b7130363eb51d5c1b4f07bb73584d39
commit a0a31176d528bd4206cac6cb0ab88ca2f00ba537

I can't stress enough that Facades are userland helpers and don't belong in libraries since Facades are optional.  So please do not accept any PR's / commits that re-introduce Facades..  I can imagine that "old habits die hard", but working without Facades is really not that hard at all.

The offending lines are (currently) here: https://github.com/nuwave/lighthouse/blob/4db038a53a713769e341533e91ce6f684d2be383/src/Providers/LighthouseServiceProvider.php#L164

All it takes is to use the following syntax to get rid of Facade dependencies:
```php
$this->app['validator']->extendImplicit
```

The same applies to the other 2 uses of Facades in https://github.com/nuwave/lighthouse/blob/4db038a53a713769e341533e91ce6f684d2be383/src/GraphQL.php#L84 and https://github.com/nuwave/lighthouse/blob/4db038a53a713769e341533e91ce6f684d2be383/src/GraphQL.php#L178

@chrissm79 May I kindly suggest to turn of Facades in your Laravel setup so these kind of issues are caught early on? (Or even better, use Lumen instead :wink: )

I'll make a PR to avoid Lumen / Laravel users without Facades being forced to work with forked versions..

**Environment**

Lighthouse Version: master
Laravel Version: n/a
Lumen Version: 5.6.38
PHP Version: 7.2.7

4levels:
**Related Issue(s)**

Resolves #352

**PR Type**
Dependency reduction (-Facades)

**Changes**

Removed unneeded use of Facades

**Breaking changes**

No breaking changes


spawnia:
@4levels Thanks for your contribution, you are correct in that we should not use facades.

Added a hint to our contribution guidelines #355

4levels:
Hi @spawnia,

Thanks for the quick actions and the added info on the contribution guidelines!  Still very happy to use Lighthouse and glad to be able to help (a little)

:+1:

PR #358, opened 2018-10-02
Multiple paginators on same type

### LINKED ISSUE #357: Can not have multiple paginators of the same type
This is a regression caused by #348 

> Copying this message from @enzonotario from Slack

In `PaginateDirectiveTest@itCanCreateQueryPaginators`, for example, if I put:

```graphql
type User {
            id: ID!
            name: String!
        }
        
        type Query {
            users: [User!]! @paginate
            users2: [User!]! @paginate
        }
```

it throws: `Schema must contain unique named types but contains multiple types named "UserPaginator"`. (edited)
I have some `@hasMany` and `@paginate` with same type but in different locations. For example:

```graphql
type Car {
    properties: [Property!] @paginate
}

type Phone {
   properties: [Property!] @paginate
}
```

before, it works, since (I think) the `paginatorTypeName` was concatenated with the `parentTypeName`:

```php
protected function paginatorTypeName(FieldDefinitionNode $fieldDefinition, ObjectTypeDefinitionNode $parent)
    {
        return studly_case(
            $this->parentTypeName($parent)
            . $this->singularFieldName($fieldDefinition)
            . '_Paginator'
        );
    }
```

enzonotario:
**Related Issue(s)**

Resolves #357 

**PR Type**

Bugfix

**Changes**

When use `@paginate` and `@hasMany` directives (and thus `PaginatorManipulator`), if there is another type defined with the same name, it throws an exception.
This PR handles it and, if the type is already defined, does what we have been doing before (previously to the PR #348) 

**Breaking changes**

None unless a developer has extended to `PaginatorManipulator` (since some function's definitions has changed)

### DIFF
--- a/src/Schema/Directives/Fields/PaginationManipulator.php
+++ b/src/Schema/Directives/Fields/PaginationManipulator.php
@@ -56,8 +56,8 @@ abstract class PaginationManipulator extends BaseDirective
     protected function registerConnection(FieldDefinitionNode $fieldDefinition, ObjectTypeDefinitionNode $parentType, DocumentAST $documentAST): DocumentAST
     {
         $fieldTypeName = ASTHelper::getFieldTypeName($fieldDefinition);
-        $connectionTypeName = "{$fieldTypeName}Connection";
-        $connectionEdgeName = "{$fieldTypeName}Edge";
+        $connectionTypeName = $this->connectionTypeName($fieldDefinition, $parentType, $documentAST);
+        $connectionEdgeName = $this->connectionEdgeName($fieldDefinition, $parentType, $documentAST);
         $connectionFieldName = addslashes(ConnectionField::class);
 
         $connectionType = PartialParser::objectTypeDefinition("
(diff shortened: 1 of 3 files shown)


hosmelq:
I tested this and i got this error.

```
Type error: Argument 1 passed to Nuwave\Lighthouse\Schema\Directives\Fields\PaginationManipulator::Nuwave\Lighthouse\Schema\Directives\Fields\{closure}() must be an instance of GraphQL\Language\AST\ObjectTypeDefinitionNode, instance of GraphQL\Language\AST\EnumTypeDefinitionNode given
```

I think you can use `GraphQL\Language\AST\Node` interface instead.

I also got this.

```
Schema must contain unique named types but contains multiple types named "CampaignConnection" (see http://webonyx.github.io/graphql-php/type-system/#type-registry).
```

I have defined the `CampaignConnection` in some types.

spawnia:
I do not think this is a clean solution. This will result in the first type being called something like `UserPaginator` and the second type `QueryUsers2Paginator`, even though they are actually the same. Also, i am pretty sure that resolution for the second type will fail, as it will not find the default model - since the type is named differently.

enzonotario:
So instead of do this, would be better to just avoid de register of a new type (if it already exists)?

spawnia:
That should be fine. I am on it.

PR #359, opened 2018-10-02
resolves #284 injection

tonipepperoni:
fixes out of memory issue with injection directive, resolved with temporary array.

resolves #284


spawnia:
What you did is just not passing along the Context and the ResolveInfo argument to the resolver. That is not a solution to the problem unfortunately, and would end up causing problems further down the chain of resolvers.

PR #361, opened 2018-10-03
[Bugfix & Refactor]`eagerLoadCount` cleanup & potential bug fix

yaquawa:
This piece of code may cause problems because some properties may changes when it was converted to the array by using the `toArray` method. 
In my case, because I'm using the binary UUID for primary key, so the `id` property of model will be converted into a string representation of the binary UUID, so if it `toArray` the model, the `id` will never be the binary UUID anymore.

```php
foreach ($models as $model) {
    if (isset($dictionary[$model->{$key}])) {
        $model->forceFill($dictionary[$model->{$key}]
            ->toArray());
    }
}
```

Also, cleanup the code a bit.

### DIFF
--- a/src/Providers/LighthouseServiceProvider.php
+++ b/src/Providers/LighthouseServiceProvider.php
@@ -5,6 +5,7 @@ namespace Nuwave\Lighthouse\Providers;
 use Illuminate\Support\Str;
 use Nuwave\Lighthouse\GraphQL;
 use Illuminate\Support\Collection;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\ServiceProvider;
 use GraphQL\Type\Definition\ResolveInfo;
 use Nuwave\Lighthouse\Schema\NodeRegistry;
@@ -115,8 +116,9 @@ class LighthouseServiceProvider extends ServiceProvider
                     $eagerLoadRelations = [$eagerLoadRelations];
                 }
 
-                $query = $this->first()::withCount($eagerLoadRelations);
-                $this->items = resolve(QueryBuilder::class)->eagerLoadCount($query, $this->items);
+                $queryWithCount = $this->first()::withCount($eagerLoadRelations);
+
+                $this->items = resolve(QueryBuilder::class)->reloadWithBuilder($queryWithCount, $this->items);
             }
 
             return $this;
@@ -127,7 +129,7 @@ class LighthouseServiceProvider extends ServiceProvider
                 if (is_string($eagerLoadRelations)) {
                     $eagerLoadRelations = [$eagerLoadRelations];
                 }
-
+//
                 $this->items = $this->fetchCount($eagerLoadRelations)->items;
                 $query = $this->first()::with($eagerLoadRelations);
                 $this->items = resolve(QueryBuilder::class)
--- a/src/Support/DataLoader/QueryBuilder.php
+++ b/src/Support/DataLoader/QueryBuilder.php
@@ -13,7 +13,10 @@ use Illuminate\Database\Eloquent\Relations\Relation;
 class QueryBuilder
 {
     /**
-     * Eager load count on collection of models.
+     * Reload a collection of models, using a given builder instance.
+     *
+     * This can be used to eagerly load the relations with an additional
+     * clause, such as a count subselect.
      *
      * Thanks to marcus13371337
      * https://github.com/laravel/framework/issues/17845#issuecomment-313701089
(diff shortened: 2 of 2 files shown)


yaquawa:
@spawnia And the method name `eagerLoadCount` actually doesn't make sense, because it has nothing to do with the `Count`, consider change the method name as well.

spawnia:
@yaquawa very nice find, i also found that previous code hideous - now it looks like proper Laravel code.

spawnia:
How about `reloadWithBuilder`?

yaquawa:
@spawnia That makes sense for me!
Actually, maybe this `eagerLoadCount` method shouldn't be called in the first place. 
It causes some extra database querying, what it tries to get is the count of the relational objects, it should check if there is already a count property there, if there is, then you don't need to reload the models. 

What do you think?

spawnia:
I see that the only place where `fetchCount` is actually used is in the `fetchForPage` macro below. Maybe it is possible to combine the two queries together.

yaquawa:
@spawnia Yes, that could be an option as well!

spawnia:
Hm so this is the basic eager loading query. They can be combined with a UNION ALL, so there might be multiples:

`select * from "posts" where "posts"."user_id" in (?) limit 3 offset 0`

You would have to combine it with this count subselect:

`select "users".*, (select count(*) from "posts" where "users"."id" = "posts"."user_id") as "posts_count" from "users" where "users"."id" in (?)`

Trying to figure out if that is even possible in SQL.

yaquawa:
@spawnia Yes, the number of `UNION ALL`s is equal to the number of parent models, if the parent model is `user` and the related model is `post`, and there are 10 parent models, here is the SQL what it has actually generated.

I'm not a DB export, I don't know if this will or will not lead performance issues, 
but this is how it will be done by handwriting the SQL I think.

```sql
(select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0) 
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
union all (select * from `posts` where `posts`.`user_id` in (?) limit 1 offset 0)
```

```sql
select "users".*, (select count(*) from "posts" where "users"."id" = "posts"."user_id") as "posts_count" from "users" where "users"."id" in (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
```

spawnia:
@yaquawa that is exactly what happens right now. Do you have an idea how those queries could be optimized further?

yaquawa:
@spawnia 

> that is exactly what happens right now. Do you have an idea how those queries could be optimized further?

Must be some solutions, but I just found the bugs, not yet dig deeper right now.

One more thought about the `Collection::macro`.
I think this responsibility should not be bound to the `Collection` class. 
Because the macro we added is very specific for our own use case, it assumes the items of the collection to be model, shouldn't be added to the generic `Collection` for the semantic reason, actually, I'm confused with this when I saw this code the first time.

spawnia:
I somewhat agree on the collection macros. On the one hand, it is convenient to have the methods on there, since they do solve a problem that is related to optimizing eager loads on Collections of models.

On the other hand, it might not be obvious to users or to integrators where those functions come from - a general problem with Macros - and we might have name conflicts with users who wish to add Macros themselves.

Would love to hear some more opinions on that, although i think this is a different issue.

yaquawa:
> since they do solve a problem that is related to optimizing eager loads on Collections of models.

Didn't notice that😓

yaquawa:
@spawnia 
Here is what I thought for the correct way to deal with this.

1. check if the model already has a `count` attribute, if yes, return the original model.
2. retrieve only the `count` from DB, no need to retrieve the model entirely again.

yaquawa:
@spawnia Currently working on solving the `union all` and unnecessary model reloading issues, hold this PR for a moment please🙏🏼.

spawnia:
@yaquawa In this case, `UNION ALL` is necessary. Otherwise, we can not do the `LIMIT`/`OFFSET` per relationship.

chrissm79:
Thanks @yaquawa and @spawnia!

spawnia:
@chrissm79 I think this PR is fine as is, if there are further improvements there can be a seperate PR.

yaquawa:
@chrissm79  Thanks you check this! As @spawnia said no need to revert it, I'll send another one later on.

PR #362, opened 2018-10-03
Ensure that the pagination/connection/relay type is only registered once

### LINKED ISSUE #357: Can not have multiple paginators of the same type

spawnia:
Resolves #357

A clean fix for this required a rework of the DocumentAST.

I changed the mechanism that it uses to store the definitions internally. Now, it is safe to just call setDefinition, without worrying about locking, immutability or anything else of that matter.

- Store the definitions in a map, no need to lock it anymore
- Unify the Exception type to ParseException
- Create QueryAST for handling parsing of incoming query documents
- Remove unneeded object hashing logic


hosmelq:
Hello @spawnia i am getting the following errors.

```
Unknown argument \"first\" on field \"attachments\" of type \"ShSubscription\".
Cannot query field \"edges\" on type \"Attachment\".
```

This is my schema.

```graphql
type ShSubscription @model(class: "App\\Models\\Subscription") {
  attachments: [Attachment!]! @hasMany(relation: "files", type: "relay")
  id: ID! @globalId
}
```

enzonotario:
I'm also getting errors:
`Cannot query field \"paginatorInfo\" on type \"Cars\"`
`Cannot query field \"data\" on type \"Cars\"`

spawnia:
Thanks for checking this out. I can reproduce that error, working on the root cause right now. You can expect a fix soon.

yaquawa:
@spawnia Same issue here, I suggest let the user define the correct return type, which should be this.


```graphql
attachments_with_pagination: AttachmentConnection! @hasMany(relation: "files", type: "relay")
attachments: [Attachment!]! @hasMany(relation: "files")
```
This is better for me.

The problem of this approach is that the `AttachmentPaginator` can cause syntax error for IDE syntax auto-completion like PHPStorm. This can be solved by generating the pagination type definition to a user specified file, so IDE won't complain us.

spawnia:
@yaquawa The schema definition will stay as-is, the user does not have to care about the inbetween type of the Paginator/Connection.

To get proper autocompletion use `php artisan lighthouse:print-schema`, which gives you the compiled, final schema.

hosmelq:
I tested the @yaquawa's solution and works well.

spawnia:
The problem was rooted within how the DocumentAST works.

I changed the mechanism that it uses to store the definitions internally. Now, it is safe to just call `setDefinition`, without worrying about locking, immutability or anything else of that matter.

@chrissm79 This should be good now.

yaquawa:
@spawnia does this means `setDefinition` always overwrite the old definition (if any)?

yaquawa:
@spawnia I thought that should it'll be nicer to have a `force` option? Say, default to `true`, if set it to `false` it'll throw. 

This may be helpful for third-party developers to prevent overwriting the existing definition.

spawnia:
I am open to adding an `addDefinition` function that throws if the definition already exists.

yaquawa:
@spawnia Oh, another method, that could be the best solution!

chrissm79:
thanks @spawnia!

PR #364, opened 2018-10-05
Do not try to encode a cursor when the result of a Connection is empty

hosmelq:
**PR Type**

Bugfix

I am getting the following error when the paginator is empty.

```
Argument 1 passed to Nuwave\Lighthouse\Execution\Utils\Cursor::encode() must be of the type integer, null given
```


spawnia:
@hosmelq Good catch. Remember to add a test next time, it ensures we do not have a regression in the future. Thank you!

hosmelq:
Thank you @spawnia, i forgot the test.

PR #367, opened 2018-10-06
Fix paginator total attribute being null with visibility set on models

stayallive:
**PR Type**

Bugfix

**Changes**

Consider the following schema:

```graphql
type Query {
    me: User @auth
}

type User {
    name: String!
    tasks: [Task!]! @hasMany(type: "paginator")
}

type Task {
    task: String!
}
```

With the following `User` mode:

```php
class User extends Model {
    protected $visible = ['id'];
}
```

And the following query:

```graphql
query {
  me {
    name
    tasks(count: 3) {
      paginatorInfo {
        total
      }
      data {
        task
      }
    }
  }
}
```

This fails with a `InvariantViolation`: `Cannot return null for non-nullable field PaginatorInfo.total.`.

Because `->toArray()` is being used to get all the attributes from the model and re-setting them on the eager loaded model only the `$visible` attributes are taken, while visibility should be ignored here otherwise the `*_count` fields will not be properly copied over.

**Breaking changes**

None, bugfix.

### DIFF
--- a/src/Support/DataLoader/QueryBuilder.php
+++ b/src/Support/DataLoader/QueryBuilder.php
@@ -40,7 +40,7 @@ class QueryBuilder
 
         foreach ($models as $model) {
             if (isset($dictionary[$model->{ $key }])) {
-                $model->forceFill($dictionary[$model->{ $key }]->toArray());
+                $model->forceFill($dictionary[$model->{ $key }]->getAttributes());
             }
         }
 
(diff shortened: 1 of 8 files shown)


stayallive:
This looks already fixed on the master branch, but currently broken on 2.2.

spawnia:
@stayallive So this is backporting the fix for `2.1.x` right?

stayallive:
@spawnia 2.2 actually but there is no branch for that 👍 (possibly 2.1 too but did not check)

chrissm79:
Thanks @stayallive!!

PR #368, opened 2018-10-07
[Enhancement] Rework the business logic of relation directives

yaquawa:
resolver #361

Reworked the business logic of relation directives.

There are many "copy and paste" works in the current code, we can encapsulate the common business logic into specific classes, so the code is easier to follow and more robust for making changes.

## List of changes.
* Replaced the `QueryBuilder` with `ModelRelationLoader`
* Removed Collection macros due to the `ModelRelationLoader`.
* Avoid unnecessary `union all` when the return type is not pagination type.
* Use `PaginationManipulator` as a helper class instead of base class.
* Abstracted the business logic of relation directives.
* Removed the meaningless `scope` argument of `@HasOne` directive.
* Added the `@belongsToMany` directive.

## Why removed the `QueryBuilder` and Collection macros

The `QueryBuilder` is only used for relation loader purpose. 
Actually it is more than just for "Querying", so the name of it is not quite self-descriptive. 
Changing its name to `ModelRelationLoader` fits better I think.

By converting the `QueryBuilder`  to `ModelRelationLoader`, now as the name implies, `ModelRelationLoader` has the responsibility of "loading relation for models", so it's natural to also integrate the `Collection` macros into the `ModelRelationLoader` as well.
This way the code itself is much easier to follow and it will avoid naming conflict of `Collection` macros.


yaquawa:
@spawnia Thanks for your review!! Way more better now!❤️

spawnia:
@chrissm79 

How would the schema definition look like in the use-case you are describing?

chrissm79:
@spawnia something like this:
```graphql
type User {
  latestPost(active: Boolean): Post @hasOne(scopes: ["byActiveState"])
}
```

spawnia:
@chrissm79 I changed it so the scopes are applied everytime, no matter the kind of relation. Should be the most consistent.

yaquawa:
@chrissm79 Good to know that tricky tips!
@spawnia Thanks for the refactor, even more simpler now 😂!

chrissm79:
@spawnia @yaquawa Thanks for adding that back!

Still looks like the resolver arguments aren't being passed into the scope, but as soon as that's cleared up it looks like we can merge this in!

spawnia:
@chrissm79 Wrote a failing test and then fixed it.

chrissm79:
@spawnia @yaquawa Great work on this one!

PR #380, opened 2018-10-12
Add `@with` directive for eager-loading any relation

stayallive:
**Changes**

This adds the `@loadRelation` directive. 

I use this is my personal projects because some fields that are just simple types depend on a relation being present / loaded (for eager loading).

This allows me to have that and still return the simple value from the attribute preventing doubled or 1+N query problems.

The test is pretty lame, but it paints the clearest picture I think. This might need to still test when you retrieve multiple records it really only loads once (eager).

----

This would probably still need an exception if you try to use pagination or scopes or something because that could become awkward, or not... did not put much thought in that, wanted to see if this going the right way or I'm just thinking the wrong way around.


src/Schema/Directives/Fields/LoadRelationDirective.php
```diff
@@ -0,0 +1,60 @@
...
+        return $next(
+            $value->setResolver(
+                function (Model $parent, array $resolveArgs, $context = null, $resolveInfo = null) use ($resolver) {
+                    $loader = BatchLoader::instance(
+                        RelationBatchLoader::class,
+                        \array_slice($resolveInfo->path, 0, \count($resolveInfo->path) - 2),
+                        $this->getLoaderConstructorArguments($parent, $resolveArgs, $context, $resolveInfo)
+                    );
+
+                    return new Deferred(function () use ($loader, $resolver, $parent, $resolveArgs, $context, $resolveInfo) {
```
spawnia: This might be simplified by having the loader just return a `Deferred` everytime and just calling it directly in fields such as `HasMany`

spawnia:
Looks pretty good so far. I wonder if this way of loading the fields should just be the default and used in the current relationship directives. Because of the way Eloquent allows the relationship to be accessed, the default resolver works fine as a final resolver.

This would eliminate the need for having more directives. The only thing that `@hasMany` add is an option to override the relation name, so i guess they would provide a resolver for that particular case.

stayallive:
That would be the preferred way indeed, I think that would even be an easier change possibly... good idea @spawnia, will work that out, should be an even simpeler change.

stayallive:
@spawnia great idea, however that won't work considering pagination for example. This really only loads the complete data, it doesn't paginate (and so is expected as far as I am concerned).

stayallive:
@spawnia this works great in production :)

The voodoo I had with the path was because I thought it had something to do with the cache path for the loader... but also works with the normal path with the same amount of queries... so it's fine I guess 👍 

As far as I'm concerned it LGTM.

spawnia:
How about we rename this `@load` or maybe `@with`? Having *relation* both in the directive name and the argument name feels clunky.

I am not too sure if this is conceptually closer to eager loading or lazy eager loading.

stayallive:
@spawnia I'm not even sure why it's `@loadRelation` here since I use it as `@load` in my own projects (might be because some Slack discussion about it), but `@with` would also be fine and maybe make more sense since Laravel uses the same wording.

spawnia:
Thanks @stayallive for your contribution to Lighthouse.

chrissm79:
@stayallive Great job, thank you!!

PR #395, opened 2018-10-16
Run middleware on a per-field basis

### LINKED ISSUE #363: Run middleware inside of GraphQL execution

When querying a field that does use Laravel Middleware via `@middleware`, and an Exception is thrown in the Middleware, the Request fails before we are in the GraphQL execution context. The resulting Response is not a proper GraphQL response, it does not conform to the specification.

The problem with this becomes even more apparent when you consider that one might query multiple fields in a single request. A failure in one field would cause the whole query to abort. However, the correct behaviour would be to return an error for one field and resolve the other field as usual.

We might use the Context, which is the third resolver argument, to call the Middleware on a per-field basis, while we are already during GraphQL execution.

This would also solve another issue: We are currently forced to load all fields of the root Query and Mutation types to evaluate if they have middleware. Deferring the Middleware execution would allow us to load those fields lazily, just like any other field.

**Possible issues**

- We have to ensure that we always have the request available as part of the Context object
- This would change the behaviour of Middleware execution in a possibly breaking way

**TODOs**

- [x] Get inner middlewares from directives working
- [ ] Consider #385 
- [ ] Global middleware from the config vs. directive config
- [ ] Write comprehensive tests
- [ ] Document the changes

### LINKED ISSUE #385: Disable global middleware for a subset of fields

Currently i can set a global middleware for AUTH on config.
```
 'route' => [      
        'middleware' => ['auth:api']
    ],
```
Because 99% of my requests are protected. But in some cases 1% can be public available for unauthenticated users.

So, right now, if i use the global middleware i cant set another requests to public.
When most of queries are protected, it is easier set the global middleware than  manually on each query. 

If i choose to make it public, something like this would be useful. So it would override the global middleware.

```php
extend type Query @group(middleware: []){  
   // many fields
}
```

spawnia:
**Related Issue(s)**

Resolves #363 
Resolves #385

**PR Type**

Feature

**Changes**

The major benefit of this PR can be seen in this test:

```php
    /**
     * @test
     */
    public function itPassesOneFieldButThrowsInAnother()
    {
        $this->schema = '
        type Query {
            fail: Int @middleware(checks: ["Tests\\\Utils\\\Middleware\\\Authenticate"])
            pass: Int
                @middleware(checks: ["Tests\\\Utils\\\Middleware\\\AddFooProperty"])
                @field(resolver: "'. addslashes(self::class).'@resolveFooMiddleware")
        }
        ';

        $result = $this->queryViaHttp('
        {
            fail
            pass
        }
        ');

        $this->assertSame(42, array_get($result, 'data.pass'));
        $this->assertSame(Authenticate::MESSAGE, array_get($result, 'errors.0.message'));
        $this->assertSame('fail', array_get($result, 'errors.0.path.0'));
        $this->assertNull(array_get($result, 'data.fail'));
    }
```

Previous to this PR, the whole query would have failed and returned a response that is not in any way spec-compliant.

Now, it returns a spec-compliant response. A failure in one field does not cause the entire query to abort. Middleware are no longer limited to root fields, they can be used on any field.

Thanks to @jbbr i added a few more improvements:

The `@middleware` directive can now be used on object types, too. This makes its behaviour consistent with that of `@group`. Both of them apply a transformation that happens during schema manipulation, that adds the `@middleware` directive to their fields. Middleware is only ever run on a per-field basis.

Additional info is in #363.

**Caveats**

The Request object is shared between fields. If the middleware of one field modifies the Request, this does influence other fields. This is however a behaviour i would expect, the `webonyx/graphql-php` specifically mentions that the context is shared between fields. If seperate Request objects are needed, we can add a configuration option in a later PR. I think it is not needed for now.

Middlewares that are used during field execution do not receive the complete `Response` object when calling `$next($request)` but rather the slice of data that the particular field returned. If more specific capabilities are needed here, it is advised to use `FieldMiddleware` directives, which are aware of the GraphQL schema. `@middleware` is rather meant as a quick and easy way to reuse existing middleware for inbound checks.
In a similar vein, the `terminate` method of middleware is also **not** called.

**Breaking changes**

Yes, middleware's now behave differently in that they are only specific to fields. However, this ensures spec-compliant responses and gives more granular control. Global middlewares that fail the whole request may still be registered through the config.


src/Schema/Directives/Fields/MiddlewareDirective.php
```diff
@@ -27,52 +51,49 @@ public function name()
...
-        if (is_string($checks)) {
-            $checks = [$checks];
-        }
+        $resolver = $value->getResolver();
 
-        return $checks;
+        return $next(
+            $value->setResolver(
+                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver, $middleware) {
+                    dump($resolveInfo->fieldName, $middleware);
```
jbbr: Note: These dumps break json output (workaround: use dump-server)

spawnia: Thanks for the review, this is still a WIP, i will remove the dumps once it is finished. Any opinions on the rest?

jbbr: I checked this out to benchmark and test it, unfortunately my relatively unperformant schema does not work with latest master or this branch, so I'm still figuring out whats causing the issues.
I will give some more feedback later

jbbr:
This does work really fine in a simple test project.

Observations:

- Throwing `\Exception` or GraphQL `Error` now both result in perfect JSON/GraphQL errors without exposing exception details in production mode. I previously modified my global Exception handler to get that behavior. But with this PR it's even possible to continue processing not-failed queries. Neat!

- When using `@group(middleware` on `extend type Query`, a throw in the middleware will still cause all requests to fail. So switching to middleware per field is required here.
Would it be possible to get the per-field behavior also for group middleware?

- Using `@middleware` on `type Query` does not throw an error but also middlewares are not used. This could cause unexpected security issues if someone not aware of this tries to use `@middleware` like this
```graphql
extend type Query @middleware(checks: ["auth:api", "can:list-users"]) {
    users: [User!]! @all
}
# Everyone can access `query {users {}}`; no error is thrown
```
I think this is by design as directives can, but mustn't provide mutators for object/field/...  definitions.
Still this could cause confusion

- Performance: I still have to get this working in a bigger project with some performance issues to see if this improves performance.

src/Schema/Directives/Nodes/GroupDirective.php
```diff
@@ -77,14 +69,26 @@ protected function setMiddlewareDirectiveOnFields($objectType)
             return $objectType;
         }
 
-        $middlewareValues = '["'.implode('", "', $middlewareValues).'"]';
-        $middlewareDirective = PartialParser::directive("@middleware(checks: $middlewareValues)");
+        $middlewareValues = addslashes(
```
jbbr: `addslashes` is causing issues here:

```graphql
extend type Query @group(middleware: ["some", "other"] {
   someField: Type @all
}
```

incorrectly expands to

```graphql
@middleware(checks: ["some\", \"other"])
```

Having only 1 middleware in `@group`-Directive works but if there are 2 or more, it fails with:
```
{"errors":[{"debugMessage":"Function name must be a string","message":"Internal server ...
```

spawnia: Nice find, was an easy fix. I just `addslashes` before imploding the array now. Thanks!

spawnia:
Thank you @jbbr for your review, was really helpful. I was able to clean up some faulty behaviours and have addressed your questions in the description above.

chrissm79:
Great job on this one @spawnia!!

src/Schema/Directives/Fields/MiddlewareDirective.php
```diff
@@ -27,52 +63,114 @@ public function name()
...
+            $value->setResolver(
+                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver, $middleware) {
+                    return $this->pipeline
+                        ->send($context->request())
+                        ->through($middleware)
+                        ->then(function (Request $request) use ($resolver, $root, $args, $resolveInfo){
+                            return $resolver(
+                                $root,
+                                $args,
+                                $this->createsContext->generate($request),
```
mfn: Why is the context regenerated here and not passed through from [line 75](https://github.com/nuwave/lighthouse/pull/395/files#diff-1871bf72c1cbac916932b2c17b024df8R75)?

spawnia: I think this might not be necessary anymore. It might be there because middleware is often used for authentication, so we have to make sure the `Context` is updated with the current user.

If the regeneration is not necessary, i would like to see that covered with a test, then we can remove it. Open to PRs.

PR #398, opened 2018-10-18
Add getRoothPath to SchemaStitcher

maarten00:
**Related Issue(s)**

None

**PR Type**

Feature

**Changes**

Our Lighthouse Utils plugin overwrites the schema root path to generate a new schema. I have upgraded the plugin to work with 2.3 and was happy to see a setter was introduced to do this, so I don't have to do it through the config. I still need to retrieve the original schema path before overwriting it, so it would be even better if I could also retrieve it without having to go through the config. 
Also see: https://github.com/deInternetJongens/Lighthouse-Utils/blob/develop/src/Generators/SchemaGenerator.php#L165

**Breaking changes**

Not that I know of

### DIFF
--- a/src/Schema/Source/SchemaSourceProvider.php
+++ b/src/Schema/Source/SchemaSourceProvider.php
@@ -17,6 +17,13 @@ interface SchemaSourceProvider
     public function setRootPath(string $path);
 
     /**
+     * Get schema root path.
+     *
+     * @return string
+     */
+    public function getRootPath();
+
+    /**
      * Provide the schema definition.
      *
      * @return string
--- a/src/Schema/Source/SchemaStitcher.php
+++ b/src/Schema/Source/SchemaStitcher.php
@@ -34,6 +34,14 @@ class SchemaStitcher implements SchemaSourceProvider
     }
 
     /**
+     * @return string
+     */
+    public function getRootPath(): string
+    {
+        return $this->rootSchemaPath;
+    }
+
+    /**
      * Stitch together schema documents and return the result as a string.
      *
      * @return string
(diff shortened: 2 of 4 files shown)


spawnia:
Did you consider https://lighthouse-php.netlify.com/docs/next/plugin-development.html#add-schema-definitions yet?

I think this eliminates to mess with the root path, i actually want to kick this method from the `SchemaSourceProvider` interface. It is too specific, since there might be other ways of obtaining a schema string that do not agree with the concept of a root path.

maarten00:
Hmmm I'm not sure if that would solve my problem. I actually need the whole object as provided by `graphql()->prepSchema()`, not the string.

spawnia:
I looked at how you add to the schema. You should be able to just do something like this in your ServiceProvider (this is not working at all, just too show the concept):

```php
    resolve('events')->listen(
        BuildingAST::class,
        function (BuildingAST $buildingAST) use ($definitionFileDirectories) {
            $typeDefinitionPaths = $this->definitionsParser->getGraphqlDefinitionFilePaths(
                $definitionFileDirectories['types']
            );
            $relativeTypeImports = $this->generateGraphqlRelativeImports($typeDefinitionPaths);
            $placeholderQuery = 'type Query{placeholder:String}';
            return "$relativeTypeImports\r\n$placeholderQuery";
        }
    );
```

And then just call `graphql()->prepSchema()` when you need access to the whole schema.

maarten00:
Adding to the schema is actually only half of it. To decide what I'm going to add to the schema, a temporary schema is generated with all the user defined types. It preps that temporary schema into an object and I look at that object to generate new queries and mutations. 
 
You make a fair poin tthough. If I change the way the package works I could eliminate that need alltogether.

On the other hand, if the prepSchema function accepts an optional argument with the full path to the schema root, that would also solve my problem without having to rewrite a bunch of code. That would be my best case scenario actually and would also eliminate the need for the getter **and** the setter.

PR #399, opened 2018-10-18
Prevent duplicate type extension nodes and duplicate fields

### LINKED ISSUE #387: Using @group in "extend type" causes incorrectly nested types when using @paginate
**Describe the bug**

When using `@paginate` directive within an `extend type Query @group` the emitted schema is incorrect.

Introspection/Playground Schema:
image

Query examples:

```graphql
{
  users(count: 5) {
    data { id email }
  }
}

# Error: Cannot query field \"id\" on type \"UserPaginator\".
```

```graphql
{
  users(count: 5) {
    data { data { id email } }
  }
}

# Error: "Argument 1 passed to Nuwave\\Lighthouse\\Schema\\Types\\PaginatorField::dataResolver() must implement interface Illuminate\\Contracts\\Pagination\\LengthAwarePaginator, instance of App\\User given, called in test/vendor/nuwave/lighthouse/src/Schema/Directives/Fields/FieldDirective.php on line 52",
```

**Expected behavior**

Query example 1 should work fine and emitted schema should only contain `UserPaginator` and not `UserPaginatorPaginator`

**Schema**

```graphql
type User {
    id: ID!
    email: String!
}
type Query {
    me: User @auth
}
extend type Query @group {
    users: [User!]! @paginate
}
```

**Environment**

Lighthouse Version: 2.3
Laravel Version: 5.7
PHP Version: 7.1

**Additional context**

Issue was introduced with Lighthouse 2.3. Downgrading to 2.2 fixes the issue.

jeroenpf:
**Related Issue(s)**

Resolves #387 

**PR Type**

Bugfix

**Changes**

Prevent type extension nodes to be added more than once.

I added the unit test that was added by @jbbr and it passes with the provided fix

# Edit by @spawnia

**Breaking Changes**

The ASTBuilder now throws when merging type extensions that define duplicate fields. Whereas this might cause failures for some users, it uncovers behaviour that was previously bugged. It now fails early instead of adding a hard to find bug, so i think it is a strict improvement over the previous behaviour.

### DIFF
--- a/src/Exceptions/DefinitionException.php
+++ b/src/Exceptions/DefinitionException.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Nuwave\Lighthouse\Exceptions;
+
+use Exception;
+use GraphQL\Error\ClientAware;
+
+class DefinitionException extends Exception implements ClientAware
+{
+    /**
+     * Returns true when exception message is safe to be displayed to a client.
+     *
+     * @api
+     * @return bool
+     */
+    public function isClientSafe()
+    {
+        return false;
+    }
+
+    /**
+     * Returns string describing a category of the error.
+     *
+     * Value "graphql" is reserved for errors produced by query parsing or validation, do not use it.
+     *
+     * @api
+     * @return string
+     */
+    public function getCategory()
+    {
+        return 'schema';
+    }
+}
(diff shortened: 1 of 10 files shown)


chrissm79:
Nice work on this one!!

robsontenorio:
@jeroenpf @spawnia  

This does not work anymore. Maybe is this PR related to this issue?

```graphql
extend type Query {
     users: [User] @paginate
}
```

spawnia:
@robsontenorio Can you be more specific please?

jbbr:
Thank @jeroenpf !
I also looked at this issue and tried to fix it but I wasn't able to figure it out in an acceptable period of time.

PR #412, opened 2018-10-29
[Fix] Lumen request array access

### LINKED ISSUE #411: Controller handling of batched queries breaks Lumen
**Describe the bug**

The batched feature breaks Lumen, in `GraphQLController::query`
Using the array access on Request breaks Lumen.  It seems related to the simplified routing that Lumen uses:
```
Call to a member function parameters() on array
in Illuminate\Http\Request.php line 624

624:             $this->all() + $this->route()->parameters(),
```

If I force set `$batched = false`, everything works as expected.  If however I set it to true, the schema cannot be built anymore either, so it seems the whole batched feature is not Lumen compatible due to this.  Can you explain how a batched request is supposed to look variable wise?

I've been trying to understand why the call to `isset($request[0])` was added.  Is it possible to rework this statement to a more readable $request->input('xxx') or any of the other documented methods on the Request class?
One could also argue that this is a Lumen / Laravel bug and not a Lighthouse bug, but since accessing the request by index is nowhere documented in the Laravel docs, I think Lighthouse shouldn't rely on this either. I'll report this as well in the Lumen repo, but I doubt anyone will bother since my previous experiences with the Laravel dev's is far from optimal. - Seems that this has been an issue in Lumen / Laravel since 2015 - https://github.com/laravel/lumen-framework/issues/119
According to that issue, this was resolved in a PR that only got merged last month in the 5.7 branch but this clearly doesn't seem to fix it.

I'm also surprised to see that in the default Lighthouse config, the batched queries feature is enabled by default. -> `'batched_queries' => true,`

Also mentioned here:  https://github.com/nuwave/lighthouse/commit/e57494fcb0b154d6ce9a2e58b7bfa9bc85f4984c#r31071978

Lighthouse Version: master
Laravel Version: 5.7
PHP Version: 7.1

**Additional context**

By var_dumping the request in my local Lumen and Laravel Lighthouse projects, I see a difference in how the $request variable is layed out:
In Lumen:
```
object(Laravel\Lumen\Http\Request)#105 (27) {
  ["json":protected]=>
  object(Symfony\Component\HttpFoundation\ParameterBag)#119 (1) {
    ["parameters":protected]=>
    array(1) {
      ["query"]=>
      string(1414) "
  query IntrospectionQuery {
    __schema {
```

in Laravel:
```
object(Illuminate\Http\Request)#42 (27) {
  ["json":protected]=> NULL 
  ["convertedFiles":protected]=> NULL
  + thousands of lines more
```
Note that I can only get the var_dump in Laravel to work if I request the schema in a new browser tab and abort the request as soon as something appears.  If I don't abort, Google chrome crashes since this var_dump is literally +100Mb of text and my VM peaks out indefinitely trying to send all this data back to the browser.

chrissm79:
**Related Issue(s)**

Resolves #411 

**PR Type**

Bugfix (Lumen)

**Changes**

Converts request object to an array before performing an `isset` on a key.

**Breaking changes**

None

### DIFF
--- a/src/Support/Http/Controllers/GraphQLController.php
+++ b/src/Support/Http/Controllers/GraphQLController.php
@@ -50,7 +50,7 @@ class GraphQLController extends Controller
      */
     public function query(Request $request)
     {
-        $batched = isset($request[0]) && config('lighthouse.batched_queries', true);
+        $batched = isset($request->toArray()[0]) && config('lighthouse.batched_queries', true);
 
         $this->extensionRegistry->requestDidStart(
             new ExtensionRequest($request, $batched)
--- a/src/Support/Http/Controllers/GraphQLController.php
+++ b/src/Support/Http/Controllers/GraphQLController.php
@@ -50,6 +50,7 @@ class GraphQLController extends Controller
      */
     public function query(Request $request)
     {
+		// If the request is a 0-indexed array, we know we are dealing with a batched query
         $batched = isset($request->toArray()[0]) && config('lighthouse.batched_queries', true);
 
         $this->extensionRegistry->requestDidStart(


src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
      */
     public function query(Request $request)
     {
-        $batched = isset($request[0]) && config('lighthouse.batched_queries', true);
+        $batched = isset($request->toArray()[0]) && config('lighthouse.batched_queries', true);
```
spawnia:
```suggestion
		// If the request is a 0-indexed array, we know we are dealing with a batched query
        $batched = isset($request->toArray()[0]) && config('lighthouse.batched_queries', true);
```

PR #414, opened 2018-11-01
Enable usage of the default field namespace for @complexity

spawnia:
**PR Type**

Feature

**Changes**

The default namespace resolution for `@complexity` now works just like `@field`.
This makes them more consistent.

**Breaking changes**

Nope


chrissm79:
thanks @spawnia!

PR #425, opened 2018-11-10
DO-NOT-MERGE Rework how AST nodes are converted into the executable schema

### LINKED ISSUE #184: Treat InputObjectValues as InputValueDefinitions, not as FieldDefinitions

The way input types are currently handled is mixed up between other types and fields. The fields of InputObjects are actually not FieldDefinitionNode but rather InputValueDefinitionNode.

This leads to cases where directives might be applied at a place where they do not actually make sense. Also, it prevents us adding strict types to the methods that deal with fields/arguments, because they might recieve different DefinitionNodes.

Change the NodeFactory/FieldFactory and related Directives so that they cleanly seperate Input objects and treat them seperately.

By distinguishing between them, our code becomes more clear in intent and more logical. We can add strong type hints to ensure we always know the exact type of data we are dealing with.

spawnia:
Resolves #184

There is now a clean seperation between FieldDefinitions and InputObjectValues.
They were previously muddled together, which caused our type annotations to be
unnecessarily loose and required some workarounds.

**Breaking changes**

The way the Field|Node|Arg-Value objects are
constructed were changed and compatibility with the previous ValueFactory can not be guaranteed.
I cleaned up some unneeded methods in the Value objects to streamline their behaviour.


spawnia:
@chrissm79 This PR is meant just for ease of review. It contains a single failing test, which will pass once we remove the `@validate` directive.

#427 contains the entire set of changes.

PR #427, opened 2018-11-13
Clean validation

### LINKED ISSUE #184: Treat InputObjectValues as InputValueDefinitions, not as FieldDefinitions

### LINKED ISSUE #430: Bad argument default value in mutation
**Describe the bug**
I define this mutation:

```graphql
createUser(
  name: String!
  type: String = "admin" @rules(apply: ["in:admin,user"])
): User
```

In playground I see this default value in the introspection:

```graphql
createUser(
  name: String!
  type: String = "{\\\\\\\"kind\\\\\\\":\\\\\\\"StringValue\\\\\\\",\\\\\\\"value\\\\\\\":\\\\\\\"user\\\\\\\",\\\\\\\"block\\\\\\\":false}"
): User
```

And I make this mutation in Playground:

```graphql
mutation {
  createUser(name: "Abelardo Addams") {
    id
  }
}
```

Then I log the request and I see:

```php
array (
  'name' => 'Abelardo Addams',
  'type' => 
  GraphQL\Language\AST\EnumValueNode::__set_state(array(
     'kind' => 'EnumValue',
     'value' => 'user',
     'loc' => NULL,
  )),
)
```

**Expected behavior**

**Schema**

```graphql
createUser(
  name: String!
  type: String = "user"
): User
```

**Environment**

Lighthouse Version:2.6.0
Laravel Version:5.7.11
PHP Version:7.1.12

**Additional context**

graphql-php version:.12.6

spawnia:
**Related Issue(s)**

Closes #425
Resolves #184
Resolves #430

**Changes**

This contains all the changes from #425.

Diverging input paths can now have distinct set of rules.
Because of the nested nature that input arguments may have, they can be structurally
different. If such a nested input was given as a list, all of them were treated the
same. Now they have rules generated that is exactly appropriate for the given input.

The new `@rulesForArray` directive allows defining validation rules for arrays themselves.
This can be useful if you want to validate you get a certain number of inputs.

Parse the default definitions back into PHP values when building from schema.
This eliminates the need for a bunch of safety checks in pagination and makes default
definitions actually work for regular queries.

**Breaking changes**

The `@validate` directive was removed. Its custom validator functionality has been broken for quite a few releases, so i expect no breakages there. Users that defined rules through it will just have to switch over to using `@rules` syntax. No major conceptual changes/rebuilds needed.


chrissm79:
Hey @spawnia, I'm going to try and get you some more insight on this one when I can but I'm running into a few errors when running this against my project. The first one is an undefined key in the `BatchLoader` [here](https://github.com/nuwave/lighthouse/blob/master/src/Execution/DataLoader/BatchLoader.php#L99) but that *may* be do to something in my code so I'm still verifying.

The next error has to do w/ a rule on an array field. The schema looks something like this:

```graphql
type Mutation {
  createField(
    form_id: ID!
    label: String @rules(apply: ["..."])
    required: Boolean @rules(apply: ["..."])
    type: String @rules(apply: ["..."])
    order: Int
    options: [String] @rules(apply: ["required_if:type,CHECKBOX_LIST,..."])
  ): Field
}
```

In my test I am sending up an array of strings for the `options` argument (as per the schema) but I'm getting the following validation errors:

screen shot 2018-11-15 at 11 34 47 am

I'll get back to you on the BatchLoader error when I can dig into it a bit more. Thanks, the rest of the PR looks nice and clean!!

spawnia:
@chrissm79 took care of the issue with the array validation. I generated the rules in a way that was conceptually wrong. Added an actual integration test to ensure this does not happen again.

No clue about the BatchLoader yet.

chrissm79:
@spawnia Thanks for updating!! I should get this re-tested and merged this afternoon!

chrissm79:
Great job on this on @spawnia!!

PR #431, opened 2018-11-20
Allow user to be guest when using the @can directive

yaquawa:
**PR Type**

refactor + behavior change

**Changes**

currently, if we use the `@can` directive, the current user must be authenticated.

while Laravel itself allow the `$user` to be either a guest or an authenticated user by specifying the first argument (`$user`) as a **nullable** argument or an argument with the default value with `null` in the policy class's methods. This PR make it aligned to the Laravel's `can` method.

The authentication check logic should be separate from this directive, for example the `@middleware` is the best place to do that.

Additionally, I found use the keyword `ability` is more understandable in most of cases(aligned to Laravel's naming convention) than using `if`, so I added an option to allow user to use `ability` instead of `if`.

Last thing that might worth to mention is that I separated the validation logic into the `validate` method to get more extensibility. I found validation logic can varies quite differently in some cases. Making the validation logic stay in a method so one can make a new directive by extending the CanDirective class and change it's behavior very easily.


yaquawa:
Hi @spawnia 😄
just added some tests, but I don't know if this is good enough for you.

plus the test will fail less than Laravel 5.7 because this is the new behavior as of ver 5.7.

https://github.com/laravel/framework/blob/5.7/src/Illuminate/Auth/Access/Gate.php#L403-L408

src/Schema/Directives/Fields/CanDirective.php
```diff
@@ -36,25 +37,64 @@ public function handleField(FieldValue $value, \Closure $next)
...
+     * @throws \Exception
+     *
+     * @return array
+     */
+    protected function getArguments(): array
+    {
+        $modelClass = $this->getModelClass();
+        $args = (array) $this->directiveArgValue('args');
+
+        array_unshift($args, $modelClass);
```
spawnia: What is going on here? Please elaborate with a comment

tests/Unit/Schema/Directives/Fields/CanDirectiveTest.php
```diff
@@ -5,26 +5,31 @@
...
+     *
+     * @param string $argumentName
      */
-    public function itThrowsWhenNotAuthenticated()
+    public function itThrowsIfNotAuthorized(string $argumentName)
     {
-        $schema = '
+        $this->be(new User());
+
+        $schema = sprintf('
```
spawnia: String concatenation is preferred for consistency

src/Schema/Directives/Fields/CanDirective.php
```diff
@@ -36,25 +37,64 @@ public function handleField(FieldValue $value, \Closure $next)
...
+
+    /**
+     * @param Authenticatable|null $user
+     * @param Gate                 $gate
+     * @param string               $ability
+     * @param array                $args
+     *
+     * @throws AuthorizationException
+     */
+    protected function validate($user, Gate $gate, string $ability, array $args)
```
spawnia:
```suggestion
    protected function authorize($user, Gate $gate, string $ability, array $args)
```

tests/Unit/Schema/Directives/Fields/CanDirectiveTest.php
yaquawa: I thought this is much more cleaner than concatenation…
Also, PHPStorm shows us where the variables are used.

phpstorm

What do you think?

If you really want me to rollback it to the concatenation style, I'll do it later on 🖐🏼

src/Schema/Directives/Fields/CanDirective.php
yaquawa: (renamed)

src/Schema/Directives/Fields/CanDirective.php
yaquawa: (added a comment)

tests/Unit/Schema/Directives/Fields/CanDirectiveTest.php
spawnia: I prefer inlining and most of all, i prefer consistency.

yaquawa: OK! I just changed them to string concatenation 😃

spawnia:
This is looking pretty good, can you add a few lines to the `master` docs on how to use this?

Possibly with a hint that guest user functionality requires Laravel 5.7.

yaquawa:
@spawnia Done!

chrissm79:
Excellent work @yaquawa, thank you!

PR #434, opened 2018-11-21
Update BaseDirective.php

sadnub:
My project requires multiple namespaces for models. This PR will allow using an array in the models config and will also preserve the old functionality.

**PR Type**

Feature

**Changes**

PR will allow this in the config:

```
namespaces => [
  models => [
    "App\\Containers\\Core\\Models",
    "App\\Containers\\ServiceDesk\\Models",
    ...
  ]
]


spawnia:
This seems like a slippery slope to take. What if two models have the same basename in different directories?

PR #435, opened 2018-11-25
 Remove deprecated traits, classes, methods, directives, various options and cruft

spawnia:
This is a large cleanup PR that removes behaviour that was kept in Lighthouse for compatibility reasons.

Now that we are nearing a v3, we can clean up our code and lay a more solid foundation for the continued development of Lighthouse. The changes made have a number of benefits:
- Code Coverage goes up
- There 

I will outline the changes in the description below. Some of the changes do only affect internals, others are cosmetic in nature - only a few will actually affect users.

**Remove the user mutations from the default schema**

They were not working properly and thus are more likely to cause confusion then to help. 

Non-breaking, as the default schema is just a starting point.

**Remove introspection queries from assets**

They were not used in any way and are very easy to get outdated. Having them in this repo is not the right place.

Non-breaking as they are just random text-files that were not actively used for anything.

**Remove deprecated methods**

Removed methods that were either renamed, moved somewhere else or made obsolete by utilizing proper Laravel DI.

All of those have been marked as deprecated for quite a few releases. Alternatives have been suggested in the `@deprecated` PHPDoc. Mostly internals are affected.

**Limit the `@field` directive to using the resolver argument**

The old style of using seperate `class` and `method` arguments is now no longer supported.
The docs are using the `resolver` argument exclusively, so few users should be affected.

It is simple to migrate:

```graphql
type Query {
  foo: Foo @field(class: "MyFoo" method: "bar")
}
```

turns into this:

```graphql
type Query {
  foo: Foo @field(resolver: "MyFoo@bar")
}
```

You just have to search-and-replace in most cases:
- `@field(class:` -> `@field(resolver:`
- `" method: "` -> `@`

**Remove the `@security` directive**

Was removed from the docs a while ago in favour of defining docs via the config.

**Remove the `resolver` argument from `@interface` and `@union`**

It has been renamed to `resolveType` which makes it more clear at what the function is supposed to do and leaves less room for confusion. Again, a simple search-and-replace should suffice for any users that still have the old syntax:

- `@interface(resolver:` -> `@interface(resolveType:`
- `@union(resolver:` -> `@union(resolveType:`

**Remove the majority of Traits**

They have all been marked deprecated for a while and a lot of them have even been in a non-functional state. Their usage has been for internal purposes, alternatives were mentioned in the `@deprecated` PHPDoc.


chrissm79:
Great job on this one @spawnia 👍

PR #436, opened 2018-11-26
Support querying for models with composite primary key

robsontenorio:
**Related Issue(s)**

None. Just sent PR before opening a issue.

**PR Type**

Feature

**Changes**

Some legacy systems does not have a default 'ID' as primary key. Even worse, there are built with composite primary keys, with no surrogate key. 

By default Laravel does not support composite keys. But, there are some traits that enable this behavior, allowing declare composite primary keys:

```php
class Product extends Model
{
    protected $primaryKey = ['barcode', 'uuid'];
      
    // Just override default `getKey()` method
    // This has nothing to do with Lighthouse 
    // This is not part of this PR implementation

    public function getKey()
    {
        $attributes = [];
        foreach ($this->getKeyName() as $key) {
            $attributes[$key] = $this->getAttribute($key);
        }
        return $attributes;
    }
}
```

This PR introduces support for QUERYING models when it has composite primary keys. By simply building a new `key` for models that will be loaded via @belongsTo directive ( by checking if the primary key is composite or not).

```graphql
        type Color {
            id: ID!
            name: String
        }
                
        type Product {
            barcode: String!
            uuid: String!
            name: String!
            color: Color @belongsTo

        }
            
        type Query {
            products: [Product] @paginate
        }
```

**Breaking changes**

None.


src/Execution/DataLoader/ModelRelationFetcher.php
```diff
@@ -419,4 +419,19 @@ protected function getRelationInstance(string $relationName): Relation
...
+
+    /**
+     * Get model key. Support composite primary keys.
+     * Ex: $primaryKey = ['key1', 'key2'];.
+     *
+     * @param Model $model
+     *
+     * @return string
+     */
+    protected function getModelKey(Model $model)
```
spawnia: This function is kind of duplicated from BatchLoader, can we keep this DRY?

robsontenorio: What do you suggest? Don’t where to place it.

robsontenorio: @spawnia done!

chrissm79:
@robsontenorio Great job on this on 👍

PR #437, opened 2018-11-26
Added @search directive to docs

erikroznbeker:
**Related Issue(s)**

-

**PR Type**

Docs

**Changes**

Added directive @search to Docs

**Breaking changes**
-

### DIFF
--- a/docs/2.6/api-reference/directives.md
+++ b/docs/2.6/api-reference/directives.md
@@ -750,6 +750,32 @@ If your class is not in the default namespace, pass a fully qualified class name
 scalar DateTime @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\DateTime")
 ```
 
+## @search
+
+Define custom search function in Eloquent model.
+```graphql
+type Query {
+  users(find:String): [User!] @search
+}
+```
+
+```php
+<?php
+
+namespace App\Models;
+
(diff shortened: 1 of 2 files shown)


spawnia:
I just had a look at the [tests for the `@search` directive](https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Schema/Directives/Args/SearchDirectiveTest.php), seems like the directive is supposed to be used with Laravel Scout.

The example in your test surely can not work, as the SearchDirective only goes on arguments. I get what you are trying to do though, you can just use a custom resolver to achieve the same thing.

	php artisan lighthouse:query Users

```php
<?php

use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;

class Users
{
    public function resolve($rootValue, array $args, GraphQLContext $context = null, ResolveInfo $resolveInfo)
    {
        return User::where(  'name' , 'LIKE', '%'.$args['search'].'%')
                   ->orWhere('email', 'LIKE', '%'.$$args['search'].'%')
				   ->get();
    }
}
```

@olivernybroe Would you care to write up a section on how `@search` is supposed to be used in your opinion?

olivernybroe:
@spawnia  Yeah, the search directive is meant for being used with Laravel Scout only.  In theory using it like he wrote is possible, but is not how it should be used.

It is meant to be used as a full text search with Laravel Scout. 
I'll write up some text about it this week :)

Issue #400, opened 2018-10-18
Lazy-load types from the serialized AST


spawnia:
**Is your feature request related to a problem? Please describe.**

While the caching of the AST does provide pretty good performance, there are issues when
dealing with really large schemas.

**Describe the solution you'd like**

The basic idea is summed up pretty nicely [in this comment](https://github.com/webonyx/graphql-php/issues/229#issuecomment-359704382).

We have to parse the whole schema initially to do schema manipulation, such as adding additional types for `@paginate`, and merging type extensions. Once we have our final schema AST, we can split it up.

**Challenges**

We have to solve the issue of orphaned types for Introspection, see [the discussion here](https://github.com/webonyx/graphql-php/issues/355#issuecomment-423576537)

maarten00:
I think this would be a great idea. Especially when queries/mutations have a lot of arguments result in a significant slowdown when building a schema. This happens just after the cached AST has been retrieved, Lighthouse loops over all the queries and their arguments. If I'm only performing one query, it seems pretty ineffecient to parse all other queries even though we already know they will not be called anyway. 

You could also check if the user is trying to perform a query or a mutation before unserializing and parsing the whole schema. Since it's not possible to perform a mutation and a query in the same call, this would mean you could completely ignore the other part. That could potentially also save a lot of parsing time.

spawnia:
I would aim to do the schema building lazily as well. `webonyx/graphql-php` allows to build the schema with a type loader. We basically need a function with the following signature:

```php
function(string $typeName): Type
```

In there, we just do roughly the following:

```php
$typeDefinition = $this->cache->get($typeName);
$resolvedType = $this->nodeFactory->handle($typeDefinition);
```

No need to differentiate between Queries/Mutations, `webonyx` has got us covered and will only load the types that are needed for the exact query that was given.

maarten00:
@spawnia I'd like to have a go at this myself, not sure if I'll get anywhere but the conceptually it shouldn't be too hard to get a working version. I'd like to know if I'm working on the right section or not. 
I'm currently thinking that for a proof of concept, I could rework the `build` function of `SchemaBuilder`. I was thinking I should pass the request and try to decide which typeDefinitions to load/register. Would that give me a significant improvement or do I need to do this earlier in the Schema parsing process?

If I can't find a way to speed up schema parser, we might have to drop Lighthouse for our project.

spawnia:
[...]

Issue #372, opened 2018-10-09
Use `@bcrypt` in input type


winter-ice:
**Is your feature request related to a problem? Please describe.**

I want to use  `@bcrypt`  in `Input` type.

**Describe the solution you'd like**

Currently,We can only use `@bcrypt`  in `Mutation`,Like this

```graphql
type Mutation {
    createUser(name:String!,email:String!,password:String! @bcrypt): User @create
}
```

This is the way i want:

```graphql
input CreateUser{
    name: String!
    email:String!
    password: String! @bcrypt
}

type Mutation {
    createUser(input:CreateUser!): User @create(flatten:true)
}
```

spawnia:
Yes, this is currently not possible. Unfortunately, it is not trivial to implement - accepting PR's!

yaquawa:
@spawnia How about something like this?

```php
namespace Nuwave\Lighthouse\Schema\Directives\Args;

use Nuwave\Lighthouse\Schema\Values\ArgumentValue;
use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
use Nuwave\Lighthouse\Schema\Values\FieldValue;
use Nuwave\Lighthouse\Support\Contracts\ArgMiddleware;

class BcryptDirective extends BaseDirective implements ArgMiddleware, InputObjectFieldTransformer
{
    /**
     * Directive name.
     *
     * @return string
     */
    public function name(): string
    {
        return 'bcrypt';
    }

    /**
     * Apply transformations on the ArgumentValue.
     *
     * @param ArgumentValue $argumentValue
     * @param \Closure       $next
     *
     * @return ArgumentValue
     */
    public function handleArgument(ArgumentValue $argumentValue, \Closure $next): ArgumentValue
    {
        $argumentValue->addTransformer(
            function ($password) {
                return bcrypt($password);
            }
        );

        return $next($argumentValue);
    }

    public function handleTransformation(FieldValue $fieldValue, \Closure $next): FieldValue
    {
        $fieldValue->addTransformer(
            function (FieldValue $fieldValue) {
                // modify the value of the files
                return $fieldValue;
            }
        );

        return $next($fieldValue);
    }

}
```

yaquawa:
@spawnia Thinking of how to fix this…

spawnia:
No need to have a seperate interface. InputObjectFields and FieldArguments are both of the same underyling AST type, `InputValueDefinitionNode`.

[...]

PR #441, opened 2018-11-26
Simplify argument directive structure and optimize execution

yaquawa:
In a nutshell, this PR is sent for **performance optimization**.

## Architecture changes 

Make it handle the argument directives at the **resolving time** but not the parsing time.

## Changes

1. Removed `ArgMiddleware`,  using newly added interface `ArgFilterDirective` and `ArgTransformerDirective` and `ArgValidationDirective` instead.
2. A lot easier to write an argument directive because we can deal with **actual argument value** within the handler method now.
3. Handle underlying argument directives when the argument value is an **input object**.
4. The `@bcrypt` directive also works on input object now (actually all "transformers" do so now).
5. Simplified the `@rule` directive 
6. Removed the `RuleFactory` class and related classes.
7. Evaluate the argument directives in the written order.

## Introduced a new class `ErrorBuffer`
Currently, it is used for gathering errors generated from `@rules` directive.

Also this error buffering mechanism may be very helpful when dealing with nested query error handling.

Actually, I'm building a schema auto-generation framework based on Lighthouse, the crud validation system is pretty complex and I found I can't done without this error buffering mechanism, so for me, kills two birds with one stone 😊

## Known issues
This is not specific for this PR, but worth to mention I think.
Because of #400 , even if we evaluate the `ArgMiddleware::handleArgument` at the resolving time, simplified the `@rule` directive, still can not get much performance gains.
still there are a lot of unnecessary processing if we don't solve #400, hopefully one day it can be fixed.

---------------
Resolver: #372


yaquawa:
Hi @spawnia  @chrissm79 
I have a question here,
The `FieldMiddleware` and `FieldResolver` can be handled as at the resolving time too I think, are there any edge cases that require them to be evaluated at the parsing time ? If not I think we should evaluate them at the resolving time too.

spawnia:
Big one, @yaquawa, I hope i will be able to dive into this in the next few days.

I like the idea of the `DirectiveFactory`, it looks like a more flexible but also performant approach. The naming convention for directives makes sense!

The `@rulesForArray` should be kept as-is. The definition is intentionally close to `@rules` and the name makes the difference clear.

yaquawa:
Hi @spawnia , thanks for reviewing!

I think what really matters in this PR is just the above 2 classes. The rest of them are just effected by the 2 classes accordingly.

* [ArgumentFactory.php](https://github.com/yaquawa/lighthouse/blob/reworkArgMiddleware/src/Schema/Factories/ArgumentFactory.php)
* [FieldFactory.php](https://github.com/yaquawa/lighthouse/blob/reworkArgMiddleware/src/Schema/Factories/FieldFactory.php)

As for the `@rulesForArray`, what about user provide something won't be able to applied to array? say the `email` rule. How about keep the name but change the signature only?

yaquawa:
@spawnia I just implemented the new @rulesForArray and fixed the test `tests/Integration/ValidationTest.php`.

spawnia:
Good. I pulled your branch and am just reviewing it, fixing up a bunch of stuff

tests/Unit/Schema/Directives/Args/RulesDirectiveTest.php
```diff
@@ -26,17 +26,15 @@ public function itCanValidateQueryRootFieldArguments()
...
+                    'message' => 'Validation failed for the field [foo].',
                     'category' => 'validation',
-                    'extensions' => [
-                        'validation' => [
-                            'bar' => [
-                                'The bar field is required.'
-                            ]
-                        ]
-                    ]
+                    'validation' => [
```
spawnia: Let's keep the extra messages under extensions always, according to the spec recommendation.

tests/Integration/ValidationTest.php
```diff
@@ -163,4 +168,4 @@ protected function assertValidationKeysSame(array $keys, array $result)
 
         $this->assertSame($keys, array_keys($actual));
     }
-}
+}
```
spawnia: newline

tests/Unit/Schema/SecurityTest.php
```diff
@@ -105,4 +105,4 @@ protected function assertIntrospectionIsDisabled(string $schema)
 
         $this->assertSame(DisableIntrospection::introspectionDisabledMessage(), $result['errors'][0]['message']);
     }
-}
+}
```
spawnia: Why?

spawnia:
Also removed two Contracts `HasRootArgumentValue` and `HasResolverArguments` as they were not actually used for anything.

I noticed you nested the arrays under `items`, which i removed as i could not see any reason
why that would be necessary?

chrissm79:
Hi @yaquawa, making my way through the PR, really like the direction of the simplified `handleArgument`! 

Is there a reason the `GraphQLValidator` was removed along w/ the `required_with_mutation` rule? I didn't see it mentioned in the PR but I see that the code was taken out. The rule I might be able to work around but the `GraphQLValidator` is pretty essential.

src/Support/Contracts/Directive.php
```diff
@@ -9,5 +9,5 @@
      *
      * @return string
      */
-    public function name();
+    public function name(): string;
```
chrissm79: This is a public facing interface, let's remove the required return type here.

spawnia: Did it.

chrissm79:
If a validator is created per field is it possible to have a `required_with` rule since it would reference a different field (or any validation that requires another field to validate)?

spawnia:
@chrissm79 as of right now, not really. I figured we might improve upon the flexibility of
ArgumentMiddleware by allowing them to wrap the field resolver they belong to, just like FieldMiddleware does.

What do you think?

spawnia:
I found we can save ourselves quite a bit of hassle going back and forth between ArgFactory and FieldFactory if we just do the bulk of the work in FieldFactory. This makes more sense to me now anyway, as the majority of the ArgumentFactory actually dealt with processes within the field resolution process.

That together with a more polished handling of the QueryFilter saves quite a few extra lines, bringing the line count of this PR into the negative :tada:

chrissm79:
>@chrissm79 as of right now, not really. I figured we might improve upon the flexibility of
ArgumentMiddleware by allowing them to wrap the field resolver they belong to, just like FieldMiddleware does.

That sounds like a possible solution.. could you provide a short code example of what that would look like? If we have access to the field resolver we could also keep the `GraphQLValidator` since we could grab the resolver arguments.

yaquawa:
@chrissm79 @spawnia

Thanks for taking time to check this!

I’m afraid that I don’t have enough time to check all the changes today!

But here is what I can answer right now.

> I noticed you nested the arrays under items, which i removed as i could not see any reason
why that would be necessary?

this is for the `rulesForArray`, to separate the errors for array it self and it’s items’s errors.

> However, i do not see how one could easily access the whole resolver arguments inside an ArgMiddleware, maybe @yaquawa knows something about that.

You can get it by implementing the `HasResolverArguments` (which you have deleted)

> If a validator is created per field is it possible to have a required_with rule since it would reference a different field (or any validation that requires another field to validate)?

Yeah, just I said before, I realized that a while ago, we can solve this by collecting the rules and messages in the directives and create a validator inside the FieldFactory.

spawnia:
@yaquawa done for now.

About the nested key `items`: I believe that we can already differentiate between validation for an array itself and its contents.

```
foo <- validation for the array itself
foo.0 <- validation for the nth item inside of the array
```

`HasResolverArguments` is a bit limited, in that the ArgumentMiddleware can not modify
the resolver arguments and they might get outdated. I would prefer to pass through the actual resolver.

Doing all validation in one full swoop is pretty much what we did before, so it is definitely possible. I think the ErrorBuffer allows the process to be simplified a bit.

How about we do one layer of arguments at a time? So first round of validation is all the direct LeafTypes and ListTypes, next round goes into one of the nested InputObjects. Always passing along the correct slice of arguments and gathering the errors in the buffer.
That would allow for `required_with` rules to consider direct neighbors.

If we go back to the all-in-one approach, is it good to allow rules to access nested inputs? Or possibly even parent inputs? `zipCode: Int @rules(apply: ["required_with:address.country"])`

yaquawa:
> If we go back to the all-in-one approach, is it good to allow rules to access nested inputs? Or possibly even parent inputs?

Not tried this kind of validation before, but I think this is possible
https://laravel.com/docs/5.7/validation#validating-arrays

yaquawa:
> HasResolverArguments is a bit limited, in that the ArgumentMiddleware can not modify
the resolver arguments and they might get outdated. I would prefer to pass through the actual resolver.

If I’m right, allow the ArgumentMiddleware access the resolver means we must handle it at the parsing time.. and add more complexity to the directive class.
In my  opinion,to allow user define server site directive is the key  feature of Lighthouse, we should keep the directive class as simple as we can, so that users of Lighthouse even don’t have any knowledge of the internal mechanisms can create their own custom directive very easily.

yaquawa:
@spawnia yes, that is real for me too. But we should have a way to avoid the “full scan” issue, for example how about passing the resolver arguments it self by reference ? But honestly I still haven’t got why you want to access the resolver in the argumentMiddleware in the first place..?

chrissm79:
> But honestly I still haven’t got why you want to access the resolver in the argumentMiddleware in the first place..?

There could be reasons such as having access to the `ResolveInfo` class to determine if the client placed a client side directive on an argument that we may want access to inside the `ArgumentMiddleware` to determine how/if to manipulate it.

spawnia:
I see the performance issue you are hinting at. While direct field arguments are not that hard, InputObjects make this way more complicated, because you can not know ahead of time how deep you need to go. Maybe passing the args by reference is the way to go then.

The most pressing issue i see right now are complex, multi-arg validation rules.

spawnia:
Oh and another thing about the `ArgMiddleware`, how about we get rid of the second argument and have the function simply take a value, do something with it, then return it?

Since the flow of the values is one-directional, other then HTTP requests or field queries, there is no actual need to do processing of the "response". There is none from the perspective of an argument.

The only case where it is currently used is `@rules`, and i actually consider the usage in there harmful. What if the user wants to validate first, then transform the input? We can let the user decide on that simply by the order in which they define the directives. (This should probably go in the docs somewhere, as it is not immediately obvious.)

yaquawa:
if we treat the validation "as a whole", actually we have only 2 options

1. validate then transform 
2. transform then validate

```php
    public function handleArgument($argumentValue, \Closure $next)
    {
        // validate...

        return $next($value);
    }

```

yaquawa:
@spawnia I can understand that you want the `@rules` to be applied rely on the order of where it was placed. I'm all for that, but what do you mean for this?
> Let's just do validation outside of a arg transformer.

Do you have any idea how we can finally achieve such a result? 
Because I'm thinking that evaluate directives by their definition order while we can still "validate the arguments as a whole", can really make it be a difficult problem to solve.

spawnia:
```php
foreach($argumentDirectives as $directive) {
  if($this->isRulesDirective($directive){
    break;
  }  

  if($directive instanceof ArgTransformer){
    $value = $directive->transform($value);
    // pop argument directive
  }
}

// do validation for all arguments  

foreach($remainingArgumentDirectives as $directive) {
  if($directive instanceof ArgTransformer){
    $value = $directive->transform($value);
  }
}
```

spawnia:
```php
foreach($argumentDirectives as $directive) {
  if($this->isRulesDirective($directive){
    break;
  }  

  if($directive instanceof ArgTransformer){
    $value = $directive->transform($value);
    // pop argument directive
  }
}

// do validation for all arguments  

foreach($remainingArgumentDirectives as $directive) {
  if($directive instanceof ArgTransformer){
    $value = $directive->transform($value);
  }
}
```

yaquawa:
@spawnia 
just 2 questions:

1. what is the `$remainingArgumentDirectives`?
2. when will the `@rules` directive to be evaluate? after `ArgTransformer`? or before?

spawnia:
@yaquawa updated the code snippet, i think it should be more clear now.

spawnia:
That is not what the code does.

It `break`s the loop of evaluating transformers after it hits the first rules directive. Notice how the check takes a single directive.

yaquawa:
@spawnia still can't understand what the code trying to to exactly.. Can you write it as pseudocode?

yaquawa:
did you mean :

1. evaluate directives before `@rules` or `@rulesForArray` 
2. validation all field
3. (if no errors thrown) evaluate directives after `@rules` or `@rulesForArray`

spawnia:
@yaquawa you got it, that is what i am trying to do.

The implementation could work in a number of different ways, but the most important point is
that **the order in which directives are placed determines the order in which they are applied**.

If you decide on the order by extendending one of two interfaces, you would need two directives to do the same thing, e.g. `@trimBeforeValidation` and `@trimAfterValidation`.

yaquawa:
@spawnia Thanks!!
I've started working on this now, and I found you placed the error messages back to the `extensions` key, can you tell me why should we put the errors in the `extensions` key..?

spawnia:
@yaquawa because the spec recommends to do so.

yaquawa:
@spawnia Done with the current tests and some additional tests for now. Maybe should add more tests..?

**Changes**
* Evaluate directives "in order"
* Added a new transformer interface
* Removed unused `ArgMiddleware` interface
* Added a `@Trim` directive
* Require Laravel 5.7 now

***Notice:***
Validation won't work properly until I updated Laravel to the latest 5.7 😂
So I edited the `composer.json` to update some packages to Laravel 5.7.

spawnia:
Nice work @yaquawa

I think we should mention how the ordering is now relevant in the docs, any other doc-relevant updates in this?

yaquawa:
@spawnia Not yet added any docs. 
But I think we can some topics to "how to create a custom directive" after this PR have been merged.

By the way, I see tests passed through Laravel 5.5~5.7 via Travis CI, so I think this must be my local environment's bad. But.. can we leave it Laravel 5.5 and update the PHP requirement to 7.1 ?(so we can utilize the `void` and nullable return types.)

spawnia:
@yaquawa I think this PR is ready to go.

Can you update the description and maybe add a little something to the docs?

spawnia:
@yaquawa Great writeup! It clearly shows how simple those new arg directives are, the `ArgMiddleware` would have been a pain to explain.

spawnia:
@chrissm79 This one should be ready to go!

chrissm79:
@yaquawa @spawnia Noticed an issue when testing. It looks like `@rules` aren't being applied when added to an input argument. Still wrapping my head around the code to see if I can find the culprit but I added a test to showcase the issue.

I'm also going to add the `GraphQLValidator` back so we can get access to the Root field, Context and ResolveInfo when creating custom validation rules.

yaquawa:
@chrissm79 I didn't expect a rule can be applied to an input object. But I can fix it.

> I'm also going to add the GraphQLValidator back so we can get access to the Root field, Context and ResolveInfo when creating custom validation rules.

As I mentioned before, you can get the the Context and ResolveInfo when creating custom validation rules by have the directive implemented `HasResolverArguments` (currently deleted because no where use it back I can add it back).

chrissm79:
> @chrissm79 I didn't expect a rule can be applied to an input object. But I can fix it.

Awesome, thanks!!!

> Currently, we are not using the GraphQLValidator and I think we don't need to use it.

Let's keep it in the codebase.

yaquawa:
@chrissm79 Got it! Great use case!

> I'll add docs for it in a separate PR.
Thanks!!

yaquawa:
@chrissm79 I just fixed the issue, would this be fine now?

yaquawa:
@chrissm79 Thanks for the test! Fixed that too.

src/Schema/Factories/FieldFactory.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Schema\Factories;
 
+use GraphQL\Type\Definition\NonNull;
```
spawnia: import order

chrissm79:
@yaquawa Thanks for taking care of those!! Great work!

PR #442, opened 2018-11-27
Optional parameter before required

kilya11:
**Related Issue(s)**

This will work unexpectedly if less then 4 arguments will be provided. See also the [example #5](http://php.net/manual/en/functions.arguments.php#functions.arguments.default).

**PR Type**

Reformat

**Changes**

Change the function signature to a valid one.

**Breaking changes**

no


spawnia:
Will be solved as part of rehauling the type definitions in #518.

PR #443, opened 2018-11-27
Change now()-helper to Carbon::now()

juusopiikkila:
**Related Issue(s)**

Lumen doesn't have a now()-helper as Laravel does so this enables Lumen to use cache and tracing.

**PR Type**

Bugfix (Lumen)

**Changes**

Changed now() to Carbon::now()

**Breaking changes**

None


spawnia:
Good find, thanks @juusopiikkila

PR #449, opened 2018-11-29
Docs (2.6): Add `key` argument for @eq operator

robsontenorio:

### DIFF
--- a/docs/2.6/api-reference/directives.md
+++ b/docs/2.6/api-reference/directives.md
@@ -327,7 +327,6 @@ pass the actual column name as the `key`.
 
 ```graphql
 type User {
-  # this will place a equal operator for `cat` database column
   postsByCategory(category: String @eq(key: "cat")): [Post] @hasMany
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/2.6/api-reference/directives.md
```diff
@@ -322,6 +322,14 @@ type User {
   postsByCategory(category: String @eq): [Post] @hasMany
 }
 ```
+If `category` field does not match with the database column, you can use the `key` argument.
```
spawnia:
```suggestion
If the name of the argument does not match the database column,
pass the actual column name as the `key`.
```

docs/2.6/api-reference/directives.md
```diff
@@ -322,6 +322,14 @@ type User {
...
+
+```graphql
+type User {
+  # this will place a equal operator for `cat` database column
```
spawnia:
```suggestion
```

spawnia: Let's keep the code GraphQL sections clear of redundant comments

chrissm79:
Thanks @robsontenorio!

PR #454, opened 2018-11-30
[DOCS] Style Adjustments

chrissm79:
**Related Issue/Intent**

N/A

**Changes**

Style adjustments to docs

**Breaking changes**

None


robsontenorio:
@chrissm79 both Read docs buttons are broken. You need to use Vue router to navigate from home page. Just check previous version.

chrissm79:
Thanks @robsontenorio! I also had some old content in the homepage... just updated it 👍 Still wrapping my head around some of Vue/Vuepress but as far as docs go it's a pretty good experience

PR #463, opened 2018-12-03
Retrieve missing docs

robsontenorio:
- [x] Added Docs for master version

When Lighthouse VuePress docs was released, some new docs from old repository were lost.

I will try to do it for all another docs files.


robsontenorio:
Checked all files. Done! All synced.

spawnia:
Thanks @robsontenorio

PR #467, opened 2018-12-08
Add missing imports.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

Stay consistent with Laravel code style 

**Changes**

Add missing imports.

**Breaking changes**

No BC

### DIFF
--- a/src/Exceptions/AuthenticationException.php
+++ b/src/Exceptions/AuthenticationException.php
@@ -2,7 +2,9 @@
 
 namespace Nuwave\Lighthouse\Exceptions;
 
-class AuthenticationException extends \Illuminate\Auth\AuthenticationException implements RendersErrorsExtensions
+use Illuminate\Auth\AuthenticationException as BaseAuthenticationException;
+
+class AuthenticationException extends BaseAuthenticationException implements RendersErrorsExtensions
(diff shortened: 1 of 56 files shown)


lucasmichot:
@spawnia - I believe most people contributing to this project are already contributing to one of the laravel projects.

The intent of this PR was to make these imports match the way they are handled in `laravel/framework` - making it simpler for anyone to contribute.

mfn:
> The intent of this PR was to make these imports match the way they are handled in

That's two edged thing. FWIW, I do like the `Base*` approach :-)

However, if we're following Laravel, there's also this: Laravel project insists on ordering imports "by length of characters".

IDK which human finds value in this, but unless it's sorted alphabetically, this is useless for humans IMO.

Also style guides should be as clear as possible without too much exception => **that** is the real overhead.

Either import ALL classes or none. Having to know/keeping track of exceptions to that, that is real mental overhead.

Just my 2c ;-)

spawnia:
I see how there is value in sticking to what Laravel does when it comes to code style.

Does the StyleCI Laravel preset have any rules concerning the direct use of root namespaces preceded by a `\`?

PR #468, opened 2018-12-08
[WIP] Enable Travis for PHP 7.3.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

**Changes**

Enable Travis for PHP 7.3.

**Breaking changes**

No beaking change


spawnia:
Still have to wait on `php-cs-fixer` to update before we switch. Thanks for the PR though, i will leave it open in the meantime.

lucasmichot:
@spawnia does it even make sense to keep `php-cs-fixer`?

- Running `composer style`shows that the current code base does not comply these rules
- We already use https://styleci.io but with different rules

Why not just use StyleCI - it's free for public repositories and simply works fine

PR #469, opened 2018-12-09
Fix custom primary keyed models

okaybroda:
- [✓] Added or updated tests
- [✓ ] Added Docs for all relevant versions

**Related Issue/Intent**
Use primary key defined in model to execute update

**Changes**

Use primary key defined in model to execute update and a new test case


src/Execution/MutationExecutor.php
```diff
@@ -89,7 +89,7 @@ protected static function handleHasManyCreate(Collection $multiValues, HasMany $
     public static function executeUpdate(Model $model, Collection $args, HasMany $parentRelation = null): Model
     {
         $model = $model->newQuery()->findOrFail(
-            $args->pull('id')
+            $args->pull($model->getKeyName())
```
spawnia: I would be fine with having this as a fallback, but the first attempt should still be to get the primary key as `id`, no matter the actual name in the model/database.

Having the primary key defined as `id` makes it way easier to handle client side caching, like Apollo, that relies on this convention.

To me, it also makes sense to not pollute the API with implementation details from the database. Why prefix the `id` field with the name of the entity and not all the fields?

spawnia:
As i said, we need to keep `id` as the primary option and can fallback to the underlying key name.

Also, can you please update the docs for `master` in this PR?

src/Execution/MutationExecutor.php
okaybroda: I agree that the right way to query/update is to use id. I made this PR to make it work with existing client implementations that use custom id for their graphql types.

I've already modified the code to use id and fallback if it's unavailable.

spawnia:
Thanks @okaybroda, good work!

PR #479, opened 2018-12-09
Feat/subscriptions - Lumen compatibility

4levels:
- [ ] Added or updated tests

**Changes**

Relaxed type checks on Application to keep Lumen in the running - replaced ->get with ->make in the SubscriptionProvider.
Not 100% sure if this is a viable approach

**Breaking changes**

### DIFF
--- a/src/Subscriptions/BroadcastManager.php
+++ b/src/Subscriptions/BroadcastManager.php
@@ -2,7 +2,7 @@
 
 namespace Nuwave\Lighthouse\Subscriptions;
 
-use Illuminate\Foundation\Application;
+use Illuminate\Container\Container as Application;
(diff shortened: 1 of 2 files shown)


4levels:
Hi all,
the reason I'm using Container as Application is that a Lumen Application is not extended from `Illuminate\Foundation\Application`, it extends `Laravel\Lumen\Application` instead.  Since both the Laravel Application and the Lumen Application are extending `Illuminate\Container\Container`, this makes it work with both Laravel and Lumen.

chrissm79:
Thanks @4levels! #479 is still a WIP so for now I'm going to close this PR as the code is still being worked on. However, feel free to post items such as this directly in the PR comments and I'll get it worked into the code and it will keep me clear of conflicts. Thanks again @4levels, excited to get subscriptions merged in!!

PR #484, opened 2018-12-09
Update the master docs to the latest changes

spawnia:
- optimise -> optimize
- Change namespaces to the new defaults
- Always prefer `[TypeName!]!` list definitions
- Some formatting


chrissm79:
thanks @spawnia!

PR #485, opened 2018-12-09
Promote StyleCI.

lucasmichot:
- [ ] Added or updated tests
- [ ] Added Docs for all relevant versions

**Related Issue/Intent**

Compy contributing instruction and promote StyleCI.

**Changes**

According to the docs, StyleCI is used (which is a smart choice).
HAving Laravel settings make it very familiar for any Laravel contributor, to also contribute to this project. 
This PR also removes PHP-CS-Fixer, which does not need to be used; also it rules are different from Laravel styleguide.

@chrissm79 / @spawnia  - I believe that StyleCI needs to be enabled for this project. I am not sure it is, as I don't see any notification on open PRs.

**Breaking changes**

PR #488, opened 2018-12-09
Avoid array helpers functions.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

Avoid array helpers functions.

**Changes**

array_has, array_has and array_pull are just helpers which are handy inside a laravel application.
Within a package, there is not point using them; it's faster to directly call the real methods of `Arr`class. 

**Breaking changes**

No breaking change.

### DIFF
--- a/src/Schema/Directives/Fields/PaginateDirective.php
+++ b/src/Schema/Directives/Fields/PaginateDirective.php
@@ -2,9 +2,9 @@
 
 namespace Nuwave\Lighthouse\Schema\Directives\Fields;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Support\Arr;
 use Nuwave\Lighthouse\Execution\QueryUtils;
 use Nuwave\Lighthouse\Schema\AST\ASTHelper;
 use GraphQL\Language\AST\FieldDefinitionNode;
--- a/src/Schema/Extensions/ExtensionRequest.php
+++ b/src/Schema/Extensions/ExtensionRequest.php
@@ -2,8 +2,8 @@
 
 namespace Nuwave\Lighthouse\Schema\Extensions;
 
-use Illuminate\Http\Request;
 use Illuminate\Support\Arr;
+use Illuminate\Http\Request;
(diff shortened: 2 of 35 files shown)


lucasmichot:
See https://github.com/nuwave/lighthouse/pull/488/commits/9da7ef5ba1ec1ea1dc2f350942270092aaaaba27 @spawnia

lucasmichot:
Can this not be managed with https://github.com/nuwave/lighthouse/pull/485 ?

spawnia:
Maybe, maybe not. This PR is wrong as is.

lucasmichot:
See https://github.com/nuwave/lighthouse/pull/488/commits/fadb11641b00d9908a8a5c480361c9fa738dc75c @spawnia

mfn:
Because they forced upon users via `composer.json`:
```json
    "autoload": {
        "psr-4": {
            "Nuwave\\Lighthouse\\": "src/"
        },
        "files": [
            "src/Support/helpers.php"
        ]
}
```
This means everyone installing lighthouse gets the `helpers.php` loaded, whether they want or not.

Worse: some of the global helpers are used within the project itself (so they're not purely for convenience of users of the lighthouse library).

This means you a) force load global stuff to users which b) may or may not clash with other projects you're not ware of 💥

This is never good. Laravel does the same and it bites _some_ users (spoiler: I'm one of them).

And the funny side of this: it's simply not necessary.

Either provide namespaces functions (which can be a autoload PITA I've heard) or just slap a _namespaced_ class somewhere with static methods.

That it's possible to override their behaviour (all the `function_exists` checks) is even worse:
- it's actually "not so" easy to override specifically on your own
- but if any other project also defining a same-named function is by accident loaded before, you've your clash already

Just a perspective of someone who had the pleasure integrating Laravel itself (not lighthouse) with other (legacy) software and just ended up in a world of pain, a situation which doesn't necessarily have to exist with a bit tactfulness.

Just my 2c :-)

spawnia:
@mfn i really dig your reasoning and feel inclined to get rid of the helpers. Would you mind posting an issue with what you just wrote so we can gather some more input before acting?

Otherwise i will have to steal what you just wrote and post it myself 😉

mfn:
Thanks, created https://github.com/nuwave/lighthouse/issues/499

PR #494, opened 2018-12-09
Remove prestissimo from Travis build.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

None

**Changes**

Disable prestissimo. This is totally useless as we already cache composer dependencies in Travis.
This can sometimes slow down the build.

This build might take more time just because it was removed, but next one will be as fast as usual

**Breaking changes**

None


spawnia:
Did not realize that travis caches deps between builds, nice catch.

PR #495, opened 2018-12-09
Let composer decide which packages it should install.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

None

**Changes**

Let composer decide which packages it should install in Travis

**Breaking changes**

No


spawnia:
This seems way simpler than the current way of doing things.

PR #508, opened 2018-12-14
[FIX] Adjust key in BatchLoader for batched queries

chrissm79:
- [X] Added or updated tests

**Related Issue/Intent**

Currently, if a batched query is sent up to Lighthouse that references the same field the same key will be produced in the `BatchLoader` causing an error (the `key` of any subsequent query will be missing). 

**Changes**

To resolve the issue `Nuwave\Lighthouse\GraphQL` now holds a reference to the `currentBatchIndex`. When set, the `BatchLoader` will prepend the index to the key making the field name unique across multiple queries.

**Breaking changes**

None

### DIFF
--- a/src/Execution/DataLoader/BatchLoader.php
+++ b/src/Execution/DataLoader/BatchLoader.php
@@ -47,6 +47,11 @@ abstract class BatchLoader
     {
         // The path to the field serves as the unique key for the instance
         $instanceName = static::instanceKey($pathToField);
+        $currentBatchIndex = app('graphql')->currentBatchIndex();
+
+        if (null !== ! $currentBatchIndex) {
+            $instanceName = "batch_{$currentBatchIndex}_{$instanceName}";
+        }
 
         // Only register a new instance if it is not already bound
         $instance = app()->bound($instanceName)
--- a/src/Execution/DataLoader/BatchLoader.php
+++ b/src/Execution/DataLoader/BatchLoader.php
@@ -47,6 +47,9 @@ abstract class BatchLoader
     {
         // The path to the field serves as the unique key for the instance
         $instanceName = static::instanceKey($pathToField);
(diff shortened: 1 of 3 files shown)


src/GraphQL.php
```diff
@@ -55,6 +58,16 @@ public function __construct(ExtensionRegistry $extensionRegistry, SchemaBuilder
         $this->pipeline = $pipeline;
     }
 
+    /**
+     * Get current batch index.
```
spawnia:
```suggestion
     * Returns the index of the current batch if we are resolving
     * a batched query or `null` if we are resolving a single query.
```

tests/Integration/Execution/DataLoader/BatchLoaderTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+        $query = '
+        query User($id: ID!) {
+            user(id: $id) {
+                email
+                tasks {
+                    name
+                }
+            }
+        }';
```
spawnia:
```suggestion
        }
        ';
```

src/Execution/DataLoader/BatchLoader.php
```diff
@@ -47,6 +47,11 @@ public static function instance(string $loaderClass, array $pathToField, array $
     {
         // The path to the field serves as the unique key for the instance
         $instanceName = static::instanceKey($pathToField);
+        $currentBatchIndex = app('graphql')->currentBatchIndex();
```
spawnia:
```suggestion
		// If we are resolving a batched query, we need to assign each
		// query a uniquely indexed instance
        $currentBatchIndex = app('graphql')->currentBatchIndex();
```

chrissm79:
Thanks for the feedback @spawnia!

PR #509, opened 2018-12-14
 Refresh newly created models before returning them

### LINKED ISSUE #501: Default values set through the database fail to be returned during @create
**Describe the bug**

When creating a new model via `@create` that gets a default value assigned on the database level, this value can not immediately be returned.

**Expected behavior**

I am able to create a new model instance and retreive all fields immediately.

**Solution**

Probably the most easy solution would be to call `refresh()` on the model instance after returning and before saving.

spawnia:
- [x] Added or updated tests

**Related Issue/Intent**

Resolves #501

**Changes**

Make sure that values that are set through the database insert
are immediately available for return, e.g. default values.

Add a test for creating nested has many relations on the side.

**Breaking changes**

No

### DIFF
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -34,6 +34,10 @@ class MutationExecutor
             });
         });
 
+        // Make sure that values that are set through the database insert
+        // are immediately available for return, e.g. default values
+        $model->refresh();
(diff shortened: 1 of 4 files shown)


chrissm79:
Nice work @spawnia!

PR #511, opened 2018-12-16
Being able to set injected value with dot notation deeper

liepaja:
- [x] Added or updated tests

**Related Issue/Intent**

Allows inject with dot notation. For example if using input object. This allows to do as follows:
`type Mutation {
  createPost($input: CreateObjectInput!): Post
    @create
    @inject(context: "user.id", name: "input.user_id")
}`

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -476,15 +476,23 @@ type Mutation {
   createPost(title: String!, content: String!): Post
     @create
     @inject(context: "user.id", name: "user_id")
-
-  createTask(input: CreateTaskInput!): Task
-    @create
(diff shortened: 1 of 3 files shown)


spawnia:
Can you please add a test and update the docs on that?

liepaja:
I don't see mutations in the tests, any guidelines for that?

spawnia:
I suggest you add a test case to InjectDirectiveTest, you can look at something like the CreateDirectiveTest to see how mutations are set up there.

liepaja:
Hmm, am i missing something, i don't see CreateDirectiveTest

spawnia:
@liepaja good work, thanks!

PR #512, opened 2018-12-16
[Feat] Make mutating directives transactional (@create and @update)

liepaja:
- [x] Added or updated tests
- [x] Added Docs for all relevant versions / is applicable when explaining nested mutation

Wrap `@create` and `@update` directives in a transaction. This ensures that if any of the nested mutations within fail, no changes are made to the database.

**Breaking changes**

The default is now to use transactions for mutations, but this can be disabled in the config.


src/Schema/Directives/Fields/CreateDirective.php
```diff
@@ -7,9 +7,21 @@
 use Nuwave\Lighthouse\Execution\MutationExecutor;
 use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
 use Nuwave\Lighthouse\Support\Contracts\FieldResolver;
-
+use Illuminate\Database\DatabaseManager;
```
spawnia: sort imports by length

src/Schema/Directives/Fields/CreateDirective.php
```diff
@@ -40,7 +52,10 @@ function ($root, array $args) {
                     ? reset($args)
                     : $args;
 
-                return MutationExecutor::executeCreate($model, collect($args));
+                $this->db->connection()->beginTransaction();
+                $mutEx = MutationExecutor::executeCreate($model, collect($args));
```
spawnia: no abbr, mk it hard to underst. ;)

src/Schema/Directives/Fields/CreateDirective.php
```diff
@@ -40,7 +52,10 @@ function ($root, array $args) {
                     ? reset($args)
                     : $args;
 
-                return MutationExecutor::executeCreate($model, collect($args));
+                $this->db->connection()->beginTransaction();
```
spawnia: Let's use the transaction closure instead

tests/Integration/Schema/Directives/Fields/CreateDirectiveTest.php
```diff
@@ -212,4 +212,64 @@ public function itCreatesAnEntryWithDatabaseDefaultsAndReturnsItImmediately()
...
+                id
+                name
+                tasks {
+                    id
+                    name
+                }
+            }
+        }
+        ';
+        $result = $this->execute($schema, $query);
```
spawnia: adding some newlines here can not hurt

spawnia:
`@update` should function the same, so please wrap that in a transaction closure as well.

tests/Integration/Schema/Directives/Fields/CreateDirectiveTest.php
```diff
@@ -212,4 +212,64 @@ public function itCreatesAnEntryWithDatabaseDefaultsAndReturnsItImmediately()
...
+        $data = ['id' => 1, 'name' => 'foo'];
+        $this->assertDatabaseMissing('users', $data);
```
spawnia: Let's just workaround this test failure by doing this check manual style.

src/Schema/Directives/Fields/CreateDirective.php
liepaja: transaction with Closure is a void function, How exactly do you imagine this?

tests/Integration/Schema/Directives/Fields/CreateDirectiveTest.php
liepaja: PHP style fixer is complaining if it is added

src/Schema/Directives/Fields/CreateDirective.php
liepaja: ` $this->db->connection()->beginTransaction(function () use ($model, $args) {
                MutationExecutor::executeCreate($model, collect($args));
         }
  );`

how would you return the populated model from closure?

tests/Integration/Schema/Directives/Fields/CreateDirectiveTest.php
liepaja: PHP style fixer (php-cs-fixer) is complaining if it is added

mfn:
This change does not seem to be optional for create/update, or am I mistaken?

liepaja:
No, is that ever an issue @mfn ?

tests/Integration/Schema/Directives/Fields/CreateDirectiveTest.php
spawnia: Single newlines are fine

mfn:
If you using events from models within transactions, the events get dispatched immediately and (non-sync background) jobs may pick up this change before the transaction is commited and may see a different state of the database.

(Very) Recent discussion: https://github.com/laravel/ideas/issues/1441

This can be a hard to track source of problems, I therefore suggest to make this optional and not default and add a documentation about why.

liepaja:
It looks like events indeed bring inconsistency. However its hard to decide whether having half of mutations being executed and an event after is better then no mutations and an event. Would you disagree @mfn?

liepaja:
I believe there is no such case where 
> Return an unordered map containing data and errors.
is the case. Its ether error or data in lighthouse.
I agree on having the option within config which defaults to transactions.

spawnia:
If we allow partially successful mutations, we **must** ensure that we return the data of the successful operations until then and the error.

Otherwise, it just becomes impossibly hard for the client to know at what point it went wrong.

Should that prove to be too difficult, i suggest we only offer transactional mutations.

yaquawa:
Transaction is good to have, but the performance is really bad.. I believe that not every mutation must to be a transaction. The directive should give user an option to select if it's going to use the transaction or not.(Considering using it with the `@can` directive, it should handle partial mutations.)

yaquawa:
@liepaja 

If I'm not missing something, at this time we can't select if perform as transaction or not at the directive level like `@update(transactional_mutations: false)`.

Another thing I noticed is that the connection which transaction applied on should be the model's connection but not the default connection.

liepaja:
I will add this functionality shortly.
What do you mean by

> Another thing I noticed is that the connection which transaction applied on should be the model's connection but not the default connection.

yaquawa:
@liepaja Each model can have different connection, the transaction should be applied to each connection specific to each model.

liepaja:
> @liepaja Each model can have different connection, the transaction should be applied to each connection specific to each model.

Can you provide documentation confirming that? I really don't expect eloquent opening different connections within 1 synchronous closure.

liepaja:
> @liepaja Each model can have different connection, the transaction should be applied to each connection specific to each model.
> 
> see "Database Connection" in https://laravel.com/docs/5.7/eloquent

Hmm, good point.

yaquawa:
@liepaja if the database grows really large, actually you want to separate the each model to connect to different databases, and I think it's not uncommon use case.

liepaja:
> want to separate the each model to connect to different databases, and I think it's not uncommon use case.

I am not sure that this is common, but indeed in can happen. This is not really solvable as within nested mutation, another model can then have another connection.

liepaja:
I believe there is no better solution than just using default connection.

spawnia:
> if the database grows really large, actually you want to separate the each model to connect to different databases, and I think it's not uncommon use case.

I really doubt that. Performance for joins would by abysmal. Something like sharding would be the more obvious choice, and the database engine should still support transactions.

liepaja:
> Can you copy over the config to the master docs?

What do you mean?

spawnia:
I do not like the per-directive overwrite of transactional handling. Think about the perspective of the client.

Unless it is somehow documented in each and every field, there is no way for them to know how each field behaves. And even if that is the case, it leads to a crazy inconsistent API.

It is better and easier to communicate to clients to have consistent behaviour for all fields. Let's remove that option.

spawnia:
The config is duplicated in the docs: https://github.com/nuwave/lighthouse/blob/master/docs/master/getting-started/configuration.md

Probably not the best approach, but reasonably easy to keep up to date and it saves the user from leaving the docs to view the config options.

yaquawa:
> It is better and easier to communicate to clients to have consistent behaviour for all fields. Let's remove that option.

I think it's nothing bad to keep it as an option.. At the end of the day, the behavior of mutation is determined by developers not the clients.

spawnia:
@yaquawa additional options add mental overhead and complexity. Lighthouse should be kept as simple as possible, with reasonable defaults.

If somebody really wants to change the behaviour, which i doubt anyone would want and which would be a really bad idea, they can still implement their own version of `@create`/`@update`.

We can't always prevent developers from shooting themselves in the foot, but at least we can make it hard for them.

src/Schema/Directives/Fields/CreateDirective.php
```diff
@@ -33,14 +46,21 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
+                $model = new $modelClassName();
 
                 $flatten = $this->directiveArgValue('flatten', false);
                 $args = $flatten
                     ? reset($args)
                     : $args;
 
-                return MutationExecutor::executeCreate($model, collect($args));
+                if (! $this->directiveArgValue('transaction',
+                    config('lighthouse.transactional_mutations', true))) {
```
spawnia: Let's remove this option.

yaquawa:
@spawnia I understand what are you pointing, me too agree most of your opinions.
But here is what I thought
> additional options add mental overhead and complexity. Lighthouse should be kept as simple as possible, with reasonable defaults

In this specific case, actually it won't add much complexity to the code.

>  doubt anyone would want and which would be a really bad idea

If the mutation is not a nested mutation, we are forced to perform with transaction with can be a huge performance issue.

src/Schema/Directives/Fields/CreateDirective.php
liepaja: I will wait for an hour or so for consensus from everyone

mfn:
> The directive should give user an option to select if it's going to use the transaction or not

I do like this.

This is also how I've seen "others" solve this problem, e.g. REST frameworks where you can specify per endpoint if it should be transactional, whether it's a read-only or write transaction, etc.

I don't know how the compatibility is with nested transactions. Last time I used them with MySQL (<8) they were a problem (not supported or didn't work with Laravel, whatever). They do work in Postgres "now" (first `->begin()` start a transaction, other `->begin()` set savepoints).

liepaja:
I suggest for now merging transactions with global config, as it is possible to set it to false and have previous functionality. New PR should be created for optional features. Otherwise this is going back and forth.

liepaja:
I started using this package from my fork and stumbled across 1 issue - model->refresh() doesn't work properly in a transaction.
As far as i know, it was used for default values, however in my case all default values are present.

yaquawa:
@liepaja I didn’t get why the `$model->refresh()` should be called, if you want the default value to be set, you can set it in the `$attributes` property or in `created` event.

liepaja:
> @liepaja I didn’t get why the `$model->refresh()` should be called, if you want the default value to be set, you can set it in the `$attributes` property or in `created` event.

This was not added in this PR, but in #509. I just moved its position.

yaquawa:
@liepaja got that! Thanks!
It seems like an extra perfermence issue to me.. maybe we can have an option to turn off the refresh invocation?

liepaja:
> @liepaja got that! Thanks!
> It seems like an extra perfermence issue to me.. maybe we can have an option to turn off the refresh invocation?

That is true as refresh rehydrates. I believe this is something for a new PR.

liepaja:
@chrissm79 , can this be merged?

chrissm79:
@liepaja I should be able to get this merged in today

chrissm79:
fantastic job on this one @liepaja 👍

Issue #396, opened 2018-10-17
Support BelongsToMany relationships in nested mutations


spawnia:
**Is your feature request related to a problem? Please describe.**

Currently only `BelongsTo` and `HasMany` are supported. While it is generally
possible to use two `HasMany` relationships in place of a `BelongsToMany`
relationship, this is neither practical nor intuitive.

**Describe the solution you'd like**

Provide a way to use nested mutations for [updating Many-to-Many relationships](https://laravel.com/docs/5.7/eloquent-relationships#updating-many-to-many-relationships).

Issue #332, opened 2018-09-18
Support other Eloquent relationships


spawnia:
**Is your feature request related to a problem? Please describe.**

Many users have come up with use cases where they require some of the other relationship types of Eloquent, such as Polymorphic relationships or Pivot Relations. In many cases, they work fine without any directives or do also work when using `@belongsTo` or `@hasMany`, although edge cases do come up from time to time.

**Describe the solution you'd like**

Implement additional directives for Eloquent's other relationship types. I am sure we can
do some nice optimizations and make Lighthouse even more simple to use.

https://laravel.com/docs/5.7/eloquent-relationships#defining-relationships

spawnia:
@enzonotario have you started work on `@morphOne` and `@morphMany`? I would like to get those in before the next release.

enzonotario:
not yet, but I can work on that this weekend!

sascha1337:
I would like to get the Roles of my UserModel, im using spatie/laravel-permissions.
It's adding a roles() method with return MorphToMany.
How would you guys put the reverse into the Schema - morphedByMany as inverse ?
Documentation does not show this directive, but there is a Model in the Tests using morphedByMany here:
https://github.com/nuwave/lighthouse/blob/36552f2886296d0e324c3482b5cef155a53b9501/tests/Utils/Models/Tag.php

Any clue ? When @morphedByMany implementation ?

lorado:
Could you try `@belongsToMany`? It works somehow, however I didn't look into the code, so don't know how it exactly works.

You can currently fetch the data also without directive, but you will not have performance benefits from data loaders, as every field will be resolved separately, means every time laravel goes to DB.

sascha1337:
Sounds like a workaround, would be amazing to get this fixed / implemented :-)

mwangaben:
waiting for  morphedByMany

PR #207, opened 2018-07-17
[Feat] Nested mutations

spawnia:
**Related Issue(s)**

Resolves #233 #208 #131

**PR Type**

Feature

**Changes**

This introduces the option for crud mutations to contain data for multiple models in one single mutation.

I used some ideas from https://www.prisma.io/docs/reference/prisma-api/mutations-ol0yuoz6go/#nested-mutations and they way they are doing nested mutations, and tried to combine that with the way Eloquent models work.

**Example Schema**

````graphql
type User {
  id: ID!
  email: String!
  posts: [Post!] @hasMany
}

type Post {
  id: ID!
  title: String!
  author: User! @belongsTo
}

type Mutation {
  createUser(input: CreateUserInput!): User
  createPost(input: CreatePostInput!): Post
  updateUser(input: UpdateUserInput!): User
  updatePost(input: UpdatePostInput): Post
  deleteUser(id: ID!): User
  deletePost(id: ID!): Post
}

input CreateUserInput {
  email: String!
  posts: CreatePostRelation
}

input CreatePostRelation {
  create: [CreatePostInput!]
}

input CreatePostInput {
  title: String!
  author: ID
}

input UpdateUserInput {
  id: ID
  email: String
  posts: UpdatePostRelation
}

input UpdatePostRelation{
  create: [CreatePostInput!]
  update: [UpdatePostInput!]
  delete: [ID!]
}

input UpdatePostInput {
  id: ID
  title: String!
  author: ID
}
````

**Example Create**

````graphql
mutation {
  createUser(
    input: {
      email: "adf@asdf.de"
      posts: {
        create: [
          { title: "post a" }
          { title: "post b" }
        ]
      }
    }
  ) {
    id
  }
}
````

**Example Update**

````graphql
mutation {
  updateUser(
    input: {
	  id: 2
      email: "adf@asdf.de"
      posts: {
        create: [
          { title: "post c" }
        ]
        update: [
          { id: 2, title: "post cysadf" }
        ]
        delete: [ 23, 4 ]
      }
    }
  ) {
    id
  }
}
````

PR #514, opened 2018-12-17
[Feat] Populating nested mutations with other relationships for @create

liepaja:
- [x] Added or updated tests

**Related Issue**

#396 #332

**Changes**

Adding other relationship capabilities to nested mutation.

| **T** indicates presence of a test

Existing nested mutation relationships from #207:
`@create`:
- **BelongsTo**
- [x] no arg | **T** moved to connect
- **HasMany**
- [x] create | **T**

Nested mutation relationships from current #514:
`@create`:
------------------------------- One to one/many
- **BelongsTo**
- [x] create | **T**
- [x] connect | **T**
- **HasOne**
- [x] create | **T**
------------------------------- Many to many
- **BelongsToMany**
- [x] create | **T**
- [x] connect | **T**
------------------------------- Morph one to one/many
- **MorphOne**
- [x] create | **T**
- **MorphMany**
- [x] create | **T**
- **MorphTo**
- [x] no args | **T**

**Breaking changes**

Instead of passing existing belongsTo ID as variable, it should be wrapped in a `connect` argument.

```graphql
type User {
  ....
}

type Post {
  id: ID!
  title: String!
  author: User! @belongsTo
}

input CreatePostRelation {
  create: [CreatePostInput!]
}
```
From:
```graphql
input CreatePostInput {
  title: String!
  author: ID
}
```
To:
```graphql
input CreatePostInput {
  title: String!
  author: {
   connect: ID || connect: [ID] / based on many to many or one to many
  }
}
```

### DIFF
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -26,7 +26,7 @@ class MutationExecutor
     public static function executeCreate(Model $model, Collection $args, Relation $parentRelation = null): Model
     {
         $reflection = new \ReflectionClass($model);
-        list($hasMany, $remaining) = self::partitionArgsByRelationType($reflection, $args, HasMany::class);
+        [$hasMany, $remaining] = self::partitionArgsByRelationType($reflection, $args, HasMany::class);
 
         list($morphMany, $remaining) = self::partitionArgsByRelationType($reflection, $remaining, MorphMany::class);
 
@@ -45,6 +45,7 @@ class MutationExecutor
             $relation = $model->{$relationName}();
 
             collect($nestedOperations)->each(function ($values, string $operationKey) use ($relation) {
+
                 if ('create' === $operationKey) {
                     self::handleMultiRelationCreate(collect($values), $relation);
                 }
@@ -104,7 +105,7 @@ class MutationExecutor
             $relation = $model->{$relationName}();
 
             collect($nestedOperations)->each(function ($values, string $operationKey) use ($relation) {
-                if ('create' === $operationKey) {
+                if ($operationKey === 'create') {
                     self::handleMultiRelationCreate(collect($values), $relation);
                 }
             });
@@ -123,7 +124,7 @@ class MutationExecutor
     protected static function saveModelWithBelongsTo(Model $model, Collection $remaining, Relation $parentRelation = null): Model
     {
         $reflection = new \ReflectionClass($model);
-        list($belongsTo, $remaining) = self::partitionArgsByRelationType($reflection, $remaining, BelongsTo::class);
+        [$belongsTo, $remaining] = self::partitionArgsByRelationType($reflection, $remaining, BelongsTo::class);
 
         // Use all the remaining attributes and fill the model
         $model->fill(
@@ -188,8 +189,6 @@ class MutationExecutor
      * @param Collection   $args           the corresponding slice of the input arguments for updating this model
      * @param HasMany|null $parentRelation if we are in a nested update, we can use this to associate the new model to its parent
      *
-     * @throws ModelNotFoundException
-     *
      * @return Model
      */
     public static function executeUpdate(Model $model, Collection $args, ?HasMany $parentRelation = null): Model
@@ -202,7 +201,7 @@ class MutationExecutor
         $model = $model->newQuery()->findOrFail($id);
 
         $reflection = new \ReflectionClass($model);
-        list($hasMany, $remaining) = self::partitionArgsByRelationType($reflection, $args, HasMany::class);
+        [$hasMany, $remaining] = self::partitionArgsByRelationType($reflection, $args, HasMany::class);
 
         $model = self::saveModelWithBelongsTo($model, $remaining, $parentRelation);
 
@@ -211,17 +210,18 @@ class MutationExecutor
             $relation = $model->{$relationName}();
 
             collect($nestedOperations)->each(function ($values, string $operationKey) use ($relation): void {
-                if ('create' === $operationKey) {
+
+                if ($operationKey === 'create') {
                     self::handleMultiRelationCreate(collect($values), $relation);
                 }
 
-                if ('update' === $operationKey) {
+                if ($operationKey === 'update') {
                     collect($values)->each(function ($singleValues) use ($relation) {
                         self::executeUpdate($relation->getModel()->newInstance(), collect($singleValues), $relation);
                     });
                 }
 
-                if ('delete' === $operationKey) {
+                if ($operationKey === 'delete') {
                     $relation->getModel()::destroy($values);
                 }
             });
@@ -276,7 +276,7 @@ class MutationExecutor
                     return false;
                 }
 
-                return $returnType->getName() === $relationClass;
+                return $relationClass === $returnType->getName();
             }
         );
     }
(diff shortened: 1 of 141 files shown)


spawnia:
@liepaja Very impressed with your methodical approach to this.

Would love for nested mutations to be more powerful!

I do disagree with the breaking change about the `BelongsTo` relationship. The name of the underlying column should not matter to the end user of the schema.

In the example you made, it seems reasonable as the foreign key has a great speaking name. It falls apart as soon as the relation was renamed or the application was possibly built on a legacy database schema.

If the column name was `user_id` but the relation is named `author`, it is much clearer to name the input after the relation. The column might be named even worse, something like `post_author_uuid_foreign`. Would you still want to use that in the schema?

Eloquent does a good job of hiding such gory implementation details, so we should leverage that!

src/Execution/MutationExecutor.php
```diff
@@ -2,25 +2,39 @@
 
 namespace Nuwave\Lighthouse\Execution;
 
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
```
spawnia: Please pay attention to the import order. `php-cs-fixer` can do that automatically.

src/Execution/MutationExecutor.php
```diff
@@ -2,25 +2,39 @@
...
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 
+
 class MutationExecutor
 {
+
+
```
spawnia: No extranous blank lines

src/Execution/MutationExecutor.php
```diff
@@ -48,7 +80,7 @@ public static function executeCreate(Model $model, Collection $args, HasMany $pa
      *
      * @return Model
      */
-    protected static function saveModelWithBelongsTo(Model $model, Collection $remaining, HasMany $parentRelation = null): Model
+    protected static function saveModelWithBelongsTo(Model $model, Collection $remaining, Relation $parentRelation = null): Model
```
spawnia: Make sure to update the docblocks as well

src/Execution/MutationExecutor.php
```diff
@@ -77,14 +117,23 @@ protected static function saveModelWithBelongsTo(Model $model, Collection $remai
     protected static function handleHasManyCreate(Collection $multiValues, HasMany $relation)
     {
         $multiValues->each(function ($singleValues) use ($relation) {
-            self::executeCreate(
-                $relation->getModel()->newInstance(),
-                collect($singleValues),
-                $relation
-            );
+            self::executeCreate($relation->getModel()->newInstance(), collect($singleValues), $relation);
```
spawnia: Such complicated method calls should be split into multiple lines.

This helps readability and the ability to debug issues. For example, if `$relation->getModel()->newInstance()` fails, you would get an exact line number pointing to that.

src/Execution/MutationExecutor.php
```diff
@@ -107,22 +156,18 @@ public static function executeUpdate(Model $model, Collection $args, HasMany $pa
 
         $model = self::saveModelWithBelongsTo($model, $remaining, $parentRelation);
 
-        $hasMany->each(function ($nestedOperations, string $relationName) use ($model) {
+        $hasMany->each(function ($nestedOperations, $relationName) use ($model) {
```
spawnia: Can this ever be anything else then a string?

src/Execution/MutationExecutor.php
```diff
@@ -156,9 +201,35 @@ public static function executeUpdate(Model $model, Collection $args, HasMany $pa
...
+
+    /**
+     * @param Model $model
+     * @param Collection $args
+     *
+     * @return Collection
+     */
+    protected static function extractMorphToArgs(Model $model, Collection $args): Collection
+    {
+        return $args->partition(function ($value, $key) use ($model) {
```
spawnia: Consider extracting the partitioning into its own method and pass the type of relation as an argument.

tests/Utils/Models/Hour.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Tests\Utils\Models;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\MorphTo;
+
+class Hour extends Model
+{
+    protected $fillable = ['from', 'to', 'weekday'];
+
```
spawnia: no extranous blank lines

src/Execution/MutationExecutor.php
spawnia: Why was this type hint removed?

src/Execution/MutationExecutor.php
```diff
@@ -107,26 +187,22 @@ public static function executeUpdate(Model $model, Collection $args, HasMany $pa
...
         $model = self::saveModelWithBelongsTo($model, $remaining, $parentRelation);
 
-        $hasMany->each(function ($nestedOperations, string $relationName) use ($model) {
+        $hasMany->each(function ($nestedOperations, $relationName) use ($model) {
             /** @var HasMany $relation */
             $relation = $model->{$relationName}();
 
-            collect($nestedOperations)->each(function ($values, string $operationKey) use ($relation) {
-                if ($operationKey === 'create') {
+            collect($nestedOperations)->each(function ($values, $operationKey) use ($relation) {
```
spawnia: Why did you remove the `string` type hint?

liepaja: Messed up branches, this is old commit

liepaja:
Any suggestions about MorphTo?

liepaja:
Currently when having an update mutation, apollo attaches __typename. What do you think @spawnia , should it be handled, or users must put this manually into input object?

spawnia:
@liepaja i am not quite sure what you are saying. Can you eloborate with a concrete example?

spawnia:
Apollo adds the `__typename` field to queries automatically, not the Inputs. If you directly use the results of a query as the arguments for a mutation, you might run into the problem you are experiencing.

While the GraphQL spec does reserve `__typename` for query fields, it is a perfectly valid name for an actual input argument, so we should not do any special treatment on the server side.

We ran into this issue in our work project, our solution was to omit the `__typename` fields from a query result before passing them into a mutation as inputs. https://github.com/jonschlinkert/omit-deep works quite well for us.

liepaja:
Yeah using json from query was my case. Didn't track it prior. So yeah you are right

liepaja:
@spawnia StyleCI is nice, but how do you automate it?

spawnia:
@liepaja you can click on the StyleCI check, download the diff file and `git apply` it. If you use PHPStorm, just drag-and-drop it into the editor.

spawnia:
Nice job @liepaja

How about `@update`, have something planned for that as well?

liepaja:
@spawnia i decided to split pull requests, as it would get to unstructured. After this will be merged will open one for @update

chrissm79:
@liepaja Superb work on this!!! Did you want to merge in master and fix the conflict? I could do it if you prefer but it's your PR so it's up to you. Thanks!!

liepaja:
@chrissm79 resolved

PR #517, opened 2018-12-19
Bump webonyx/graphql-php to ^0.13 and php to >= 7.1

### LINKED ISSUE #218: [Discussion] Require PHP >= 7.1
`webonyx/graphql-php` requires PHP 7.1 in the newer version, so if we want to upgrade the library, we have to require it as well. This also means that we would cut cords with Laravel 5.5, since it only requires PHP 7.0

In the process of updating, we could work even more on the correctness and type safety of the code, by adding in nullable/optional types.

Should we wait on this until v3 or do you think it is reasonable to do it as part of a minor release?

spawnia:
Breaking changes for our users are minimal, as we wrap webonyx most of the time.

Read https://github.com/webonyx/graphql-php/blob/v0.13.0/UPGRADE.md for more information.

Resolves #218

- [x] Added or updated tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ cache:
 language: php
 
 php:
-- 7.0
 - 7.1
 - 7.2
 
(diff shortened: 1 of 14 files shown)


chrissm79:
nice work @spawnia 👍

PR #518, opened 2018-12-19
Laravel & PHP 7.1 style type definitions

spawnia:
This is a refactoring PR with the purpose of cleaning up type definitions
throughout the codebase.

We now fully embrace the Laravel code style, all PHPDoc blocks were fixed.

Missing type hints are added and existing type hints were refined with the PHP 7.1 nullable
types.

The tests were restructured to make all requests solely through the usual HTTP test methods, leveraging Laravel's built in TestResponse helper class and unifying the test setup. This allows the resolvers to always expect a `GraphQLContext`, so we do not force the user to break PSR-2 anymore and can guarantee it is passed along.

Many small fixes to definitions, spelling, type hints, comments and more were made through PHPStorm inspections and static analysis.


src/Schema/Extensions/DeferExtension.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Schema\Extensions;
 
+use Illuminate\Http\Response;
```
spawnia: Order imports by length

src/Schema/Extensions/DeferExtension.php
```diff
@@ -61,8 +62,9 @@ public static function name()
      * Handle request start.
      *
      * @param ExtensionRequest $request
+     * @return void
```
spawnia: Add a new line between param and return definitions

src/Schema/Extensions/ExtensionRequest.php
```diff
@@ -56,7 +56,7 @@ public function context(): GraphQLContext
      *
      * @return string
      */
-    public function queryString($index = null): string
+    public function queryString(int $index = null): string
```
spawnia: Change to `?int $index = null` to be explicit

spawnia:
@chrissm79 Just finished this beast. This should make it so we have a nice base to put on the final touches for `v3`.

This got larger than i anticipated, but i feel it will make the coming PR's way easier, since we no longer have a mix of formatting/refactoring with functional changes.

chrissm79:
Great work on this on @spawnia, reviewing now!

src/Schema/Extensions/GraphQLExtension.php
```diff
@@ -10,19 +10,18 @@
...
      * Manipulate the schema.
      *
-     * @param DocumentAST $documentAST
-     *
-     * @return DocumentAST
+     * @param  \Nuwave\Lighthouse\Schema\AST\DocumentAST  $documentAST
+     * @return \Nuwave\Lighthouse\Schema\AST\DocumentAST
      */
-    public function manipulateSchema(DocumentAST $documentAST)
+    public function manipulateSchema(DocumentAST $documentAST): DocumentAST
```
chrissm79: This is an abstract class that is meant to be extended by the user so let's remove the return types on this class.

src/Subscriptions/SubscriptionBroadcaster.php
```diff
@@ -100,11 +106,11 @@ function (Subscriber $subscriber) use ($root) {
...
      * Authorize the subscription.
      *
-     * @param Request $request
+     * @param  \Illuminate\Http\Request  $request
      *
-     * @return array
+     * @return \Illuminate\Http\Response
      */
-    public function authorize(Request $request)
+    public function authorize(Request $request): Response
```
chrissm79: This can also be a `Illuminate\Http\JsonResponse`

src/Schema/Extensions/GraphQLExtension.php
spawnia: Done

src/Subscriptions/SubscriptionBroadcaster.php
spawnia: How about responses in general? What should the return type in the controller and in `GraphQLResponse` be?

chrissm79: `GraphQLResponse` is an interface which can be extended by the user so no return type goes there. As for the Controller, although users can technically return an array if they prefer, Laravel uses `\Illuminate\Http\Response` in its docblocks w/out an enforced return type so let's keep it the same.

spawnia: Since we do use the `StreamedResponse`, how about we annotate it as `\Symfony\Component\HttpFoundation\StreamedResponse`. All response types i have identified are actually subclasses of that.

chrissm79: 👍 That works for me. Pipe this through some local testing and get it merged in if everything is good (should be this afternoon). Thanks @spawnia!!

spawnia: Actually i meant to say to change all hints to `\Symfony\Component\HttpFoundation\Response`, the last commit does exactly that. I think you got that, though :)

chrissm79:
Great job on this one @spawnia 👍

PR #519, opened 2018-12-20
Prevent name conflict between argument names and non-relation methods when executing nested mutations

### LINKED ISSUE #507: Argument names in nested mutations conflict with non-relation methods on the model
**Describe the bug**
I am using the Spatie Permissions package and there is a 'guard' field on the Role class. When I try to return that field with a query, it will run the Model::guard() method.

**Expected behavior**
Return guard field from the database

**Schema**

```graphql
#types
type Role{
    id: ID!
    name: String!
    guard: String
}

#Inputs
input AddRoleInput {
    name: String! @rules(apply: ["required"])
    guard: String
}

input EditRoleInput {
    id: ID! @rules(apply: ["required"])
    name: String
    guard: String
}

#Add Queries to Base Query
extend type Query {
    roles: [Role!]! @all
    role(id: ID! @eq): Role! @find
}

#Add Mutations to Base Mutation
extend type Mutation {
    addRole(input: AddRoleInput!): Role! @create(flatten: true)
    editRole(input: EditRoleInput!): Role! @update(flatten: true)
    #editRoles(
    #    ids: [ID!]! 
    #    input: EditUsersInput!): [User!]! @bulk(flatten: true)
    deleteRole(id: ID!): Role! @delete
    deleteRoles(id: [ID!]!): [Role!]! @delete
}
```

**Output/Logs**
```
TypeError: Too few arguments to function Illuminate/Database/Eloquent/Model::guard(), 0 passed in /var/www/simple-rmm-appeval()'d code on line 1 and exactly 1 expected
```

**Environment**

Lighthouse Version: 2.6.3
Laravel Version:5.7.16
PHP Version: 7.2.12

**Additional context**

This seems to be the method it is getting caught up on. I'm not sure if there is a way to check if the method exists only on the model child class.
```
    protected static function extractHasManyArgs(Model $model, Collection $args): Collection
    {
        return $args->partition(function ($value, $key) use ($model) {
            return method_exists($model, $key) && ($model->{$key}() instanceof HasMany);
        });
    }
```

liepaja:
- [x] Added or updated tests
- [x] Added or updated docs

**Related Issue/Intent**

Resolves #507

**Changes**

Detect relationships when executing nested mutations by using reflection instead of calling the methods and examining the result.

This prevents non-relationship methods from being called, which may lead to errors and/or side effects.

**Breaking Changes**

Users are required to define the return type of a relationship method in their Eloquent Model for the detection to work.

### DIFF
--- a/docs/master/guides/relationships.md
+++ b/docs/master/guides/relationships.md
@@ -144,8 +144,31 @@ class Post extends Model
 Lighthouse allows you to create, update or delete your relationships in
 a single mutation.
 
-By default all mutations are wrapped in a database transaction, so if any of the nested
-operations fail, the whole mutation is aborted and no changes are written to the database.
+You have to define return types on your relationship methods so that Lighthouse
+can detect them.
+
+```php
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class Post extends Model 
+{
+    // WORKS
+    public function user(): BelongsTo
+    {
+        return $this->belongsTo(User::class);
+    }
+
+    // DOES NOT WORK
+    public function comments()
+    {
+        return $this->hasMany(Comment::class);        
+    }
+}
+```
+
+By default, all mutations are wrapped in a database transaction.
+If any of the nested operations fail, the whole mutation is aborted
+and no changes are written to the database.
 You can change this setting [in the configuration](../getting-started/configuration.md).
 
 ### Belongs To
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Execution;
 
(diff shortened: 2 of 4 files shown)


spawnia:
I switched the method of detection to what i proposed in the comment.

The approach you chose will fail if a user themselves has defined a method with the same name on the Model.

Another problem it had is that it calls the method, which may lead to side effects.

spawnia:
This seems to fail in PHP 7.0, so we can merge it after we made the switch to require 7.1. #517

spawnia:
@liepaja Please stick to our formatting standard for the docblocks, use `php-cs-fixer`.

liepaja:
Didn't know it was not edited by php-cs-fixer. Reseted

spawnia:
Thanks @liepaja for your contribution.

chrissm79:
Great work on this on @liepaja, thank you!

PR #520, opened 2018-12-21
Replace `DirectiveRegistry` with `DirectiveFactory` to lazy load directives

yaquawa:
The implementation of `DirectiveRegistry` load **ALL** directives for every single request even if user didn't use any directives at all.

The newly added `DirectiveFactory` lazy loads directives on demand which can save unnecessary computing resource.

Also this implementation allow users and plugin developers overwrite the default directives that come with Lighthouse.

**Resolving flow**

1. `DirectiveFactory::create($directiveName)`
2. Assume the class name is `$directiveBaseNamespaces . studly_case($directiveName) . 'Directive'`
3. Create one or throw if no such a class found or not implements the `Directive` interface

***Notice***
This require the class name follow the naming convention shown in `2`. however I think this is not a bad idea.

## Breaking changes
Before this PR, to add custom directives user specifies the directive path in the config file like:

```php
'directives' => [__DIR__.'/../app/GraphQL/Directives'],
```

Now instead of path you specify the base namespace of your directives like:

```php
    'namespaces' => [
        'models' => 'App',
        'queries' => 'App\\GraphQL\\Queries',
        'mutations' => 'App\\GraphQL\\Mutations',
        'interfaces' => 'App\\GraphQL\\Interfaces',
        'unions' => 'App\\GraphQL\\Unions',
        'scalars' => 'App\\GraphQL\\Scalars',

        // just like before, this can be multiple namespaces 
        'directives' => ['App\\GraphQL\\Directives'],
    ],
```

## Migration
If you've been adding directives to Lighthouse as a plugin package, checkout [this test](https://github.com/nuwave/lighthouse/blob/8e3127cabadcfd4f2b14e166fffd083c9733e22f/tests/Integration/Events/RegisteringDirectiveBaseNamespacesTest.php#L20-L25) to see the new directive registration style.


spawnia:
@yaquawa Really like this, gives better flexiblity.

Can you please update description with the breaking changes and how to migrate with small code examples?

A few words in the package development docs about the event based registration mechanism would be great too.

Thank you!

yaquawa:
@spawnia Thanks for the review!! 

> Can you please update description with the breaking changes and how to migrate with small code examples?

A few words in the package development docs about the event based registration mechanism would be great too.

done!

chrissm79:
@yaquawa excellent work on this one!! 👍

PR #523, opened 2018-12-27
Promote StyleCI.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

Compy contributing instruction and promote StyleCI.

**Changes**

According to the docs, StyleCI is used (which is a smart choice).
HAving Laravel settings make it very familiar for any Laravel contributor, to also contribute to this project. 
This PR also removes PHP-CS-Fixer, which does not need to be used; also it rules are different from Laravel styleguide.

@chrissm79 / @spawnia  - I believe that StyleCI needs to be enabled for this project. I am not sure it is, as I don't see any notification on open PRs.

**Breaking changes**

**Notes**
Replace https://github.com/nuwave/lighthouse/pull/485


chrissm79:
@lucasmichot Thanks for this! I did get a error stating that the config was misconfigured w/ the deprecated option `linting` (mentioned in the changelog [here](https://docs.styleci.io/change-log)). Is there another option that we should be using in its place or just simply remove it?

lucasmichot:
You're welcome @chrissm79 !
The linting option can safely be removed 👍

chrissm79:
Awesome, went ahead and took care of it in #522 👍

PR #527, opened 2018-12-27
Console comments and return types.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

None

**Changes**

Add FQCN docblocks and return types + typehint functions

**Breaking changes**

None

### DIFF
--- a/src/Console/PrintSchemaCommand.php
+++ b/src/Console/PrintSchemaCommand.php
@@ -6,8 +6,6 @@ use Illuminate\Console\Command;
 use GraphQL\Utils\SchemaPrinter;
 use Illuminate\Cache\Repository;
 use Illuminate\Contracts\Filesystem\Filesystem;
-use Nuwave\Lighthouse\Exceptions\ParseException;
-use Nuwave\Lighthouse\Exceptions\DirectiveException;
 
 class PrintSchemaCommand extends Command
 {
--- a/src/Console/PrintSchemaCommand.php
+++ b/src/Console/PrintSchemaCommand.php
@@ -14,8 +14,10 @@ class PrintSchemaCommand extends Command
      *
      * @var string
      */
-    protected $signature = ' lighthouse:print-schema
-        {--W|write : Write the output to a file}';
+    protected $signature = '
+        lighthouse:print-schema
+        {--W|write : Write the output to a file}
+    ';
 
     /**
      * The console command description.
--- a/src/Console/PrintSchemaCommand.php
+++ b/src/Console/PrintSchemaCommand.php
@@ -14,7 +14,8 @@ class PrintSchemaCommand extends Command
      *
      * @var string
      */
-    protected $signature = ' lighthouse:print-schema {--W|write : Write the output to a file}';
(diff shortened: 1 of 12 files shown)


spawnia:
Hey, thanks for the PR.

I have #518 going atm which is going to represent a pretty big overhaul of type definitions. Planning to finish it soon.

I see you made some changes here which are a bit different, can you elaborate on them?

https://github.com/nuwave/lighthouse/pull/527/files#diff-095a7fefa5268aca0fb669eae3df78f7R29 has got me particulary confused.

src/Console/PrintSchemaCommand.php
```diff
@@ -16,10 +14,8 @@ class PrintSchemaCommand extends Command
      *
      * @var string
      */
-    protected $signature = '
-        lighthouse:print-schema
-        {--W|write : Write the output to a file}
-    ';
+    protected $signature = ' lighthouse:print-schema {--W|write : Write the output to a file}';
```
spawnia: Let's keep this as is, the multi-line way is more readable as it looks like the output from running the help.

lucasmichot: See https://github.com/nuwave/lighthouse/pull/527/commits/391b756e19a83325c9ccc2525bda89d849eb7fdd @spawnia

src/Console/PrintSchemaCommand.php
```diff
@@ -30,13 +29,11 @@ class PrintSchemaCommand extends Command
     /**
      * Execute the console command.
      *
-     * @param Repository $cache
-     * @param Filesystem $storage
-     *
-     * @throws DirectiveException
-     * @throws ParseException
+     * @param  \Illuminate\Cache\Repository  $cache
```
spawnia: I am a bit confused by this change in signature.

Is the Laravel code style really to not annotate Exceptions?

How about the FQCN, are they always done like that in Doc-Blocks?

lucasmichot: Laravel code styles include them.
Neverthless there is no `throw ...` in `\Nuwave\Lighthouse\Console\PrintSchemaCommand::handle`.

Other functions are responsible for throwing  DirectiveException or ParseException

lucasmichot: If F1 calls F2, that calls F3, that calls F4
and If F4 can throw an exception
F1, F2 & F3 are not annotated.

spawnia: PHPStorm hints at adding all Exceptions that occur somewhere down the function chain and are uncaught. I think that behaviour makes sense, as often the public methods of a class might not be the ones that actually throw the Exceptions, but it would be really inconvenient to have to dig in to the whole callstack to see what can happen.

lucasmichot: Up to you..
In this case there are really tons of functions that are missing these annotations

spawnia: I think we should be consistent first and foremost. Feel free to discuss in #531

chrissm79:
Thanks @lucasmichot!

PR #548, opened 2019-01-03
Treat subscriptions just like other fields

spawnia:
- [x] Added or updated tests

**Related Issue/Intent**

Subscriptions now can be defined just like other fields, default namespaces
will work consistently.

**Changes**

- Default namespaces for Subscriptions now work without directives, just like
those for Queries and Mutations do.
- Split subscription docs into defining fields and triggering subscriptions, each with a few examples.

**Breaking changes**

Changed type hints for the methods `resolve` and `encodeTopic` on `Nuwave\Lighthouse\Schema\Fields\SubscriptionField`.

Both methods do have default implementations, so there are probably just a few users who did overwrite them.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,4 @@ services:
   php:
     image: spawnia/php7.1-fpm:dev-1.3.1
     volumes:
-      - ./:/var/www
-    environment:
-      - APP_ENV=local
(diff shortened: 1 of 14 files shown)


chrissm79:
@spawnia Great work on this one! Should have it tested and merged in soon 👍

PR #562, opened 2019-01-14
Rename the "queue" argument for @broadcast to "shouldQueue"

spawnia:
This naming is more in line with the well-known Laravel interface `ShouldQueue` and also clearly expresses the boolean nature of the argument.


chrissm79:
Agreed, thanks @spawnia!

PR #578, opened 2019-01-21
Update directives.md

iliyaZelenko:
Looks like you have a mistake here, or not?


spawnia:
Thanks @iliyaZelenko, great find!

PR #581, opened 2019-01-21
Prevent crash when invalid JSON variables are given

claudiodekker:
- [x] Added or updated tests

**Related Issue/Intent**

When an invalid or empty (JSON) `variables` parameter is given (default behaviour of  https://github.com/graphql/graphiql), `json_decode` can return `null`, causing the following exception:
screenshot 2019-01-21 at 21 50 23

**Changes**

- Adds an empty array as fallback for when json_decode fails / returns null.

**Breaking changes**

 None.

### DIFF
--- a/src/Support/Http/Controllers/GraphQLController.php
+++ b/src/Support/Http/Controllers/GraphQLController.php
@@ -127,11 +127,11 @@ class GraphQLController extends Controller
      */
     protected function ensureVariablesAreArray($variables): array
     {
-        return is_string($variables)
-            ? json_decode($variables, true)
-            : $variables === null
-                ? []
-                : $variables;
+        if (is_string($variables)) {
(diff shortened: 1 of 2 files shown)


spawnia:
I think this is a clear improvement over the current behaviour, so i am inclined to merge this as is.

Generally, i think we might do an even better job of validating inputs by throwing and notifiying the user. A hard failure is better then a silent one, which an invalid JSON might cause here.

claudiodekker:
Perhaps that would indeed be a better solution.

Unfortunately, though, I currently don't have much time available to look into this and create a proper and clean solution. If you want, I can revert the changes I made to the test so that behaviourally nothing news is introduced for now. How does that sound?

spawnia:
This is fine as it is, i will merge it for now as it is an improvement over the current behaviour. Thanks @claudiodekker

PR #584, opened 2019-01-22
Add guard argument to auth directive

AMondayOnFebruary:
- [x] Added or updated tests

**Changes**

Adds a `guard` argument to the auth directive so one can specify a different guard to resolve the authenticated user

**Breaking changes**

No breaking changes, if the `guard` argument is not specified it uses the default guard resolver

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -10,7 +10,8 @@ type Query {
 }
 ```
 
-If you need to use a guard besides the default to resolve the authenticated user, you can pass the guard name as the `guard` argument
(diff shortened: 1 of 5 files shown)


spawnia:
Thanks @AMondayOnFebruary, great work.

I added an additional adjustment and got rid of our redefined `auth()` helper. While the test passed in the Laravel-based test environment, the implementation would have failed with Lumen apps which do not have the `auth()` helper and would have fallen back to our own - which did not allow a `guard` argument.

Using proper DI now.

chrissm79:
Great job @AMondayOnFebruary, thank you!!

PR #585, opened 2019-01-22
Add lighthouse-graphql-passport-auth to resources

joselfonseca:
GraphQL mutations for Laravel Passport using Lighthouse PHP. Ability to generate tokens using the password grant in mutations specially for iOS and Android clients.

- [ ] Added or updated tests
- [X] Added Docs for all relevant versions

**Related Issue/Intent**

Add a plugin to the docs that allows us to have 3 mutations to get, refresh and revoke an access token from Laravel passport.

**Changes**

None
**Breaking changes**

None


joselfonseca:
Thanks @chrissm79 BTW great GraphQL server. Will be using it a lot.

PR #587, opened 2019-01-24
Document @rulesForArray directive and fix bugs

spawnia:
- [x] Updated tests

Docs were completely missing before.

**Bugfix**

`@rulesForArray` was being applied twice if the content of the array was a scalar, once to the array itself and once to the contents.

**Enhancement**

Array based rules only work correctly if the rule `"array"` is used. As this is redundant when using `@rulesForArray` anyway, it is added automatically if not present yet.


chrissm79:
Thanks @spawnia

PR #597, opened 2019-02-02
Use Laravel broadcast connection parameters for Pusher broadcaster

stayallive:
This allow you to configure which broadcast connection parameters are used to configure the Pusher broadcaster, this gives you a bit more flexibility IMO.


chrissm79:
Thanks @stayallive!

PR #598, opened 2019-02-02
Prefix the Pusher channels to easily identify them

stayallive:
This way the channels can be easily identified and allows me to hook up some custom authentication so I can use a single Pusher client for "normal" websockets and (Apollo) subscriptions.

Also gives me some indication what the channel is supposed to be doing instead of a completely random string 😄

### DIFF
--- a/src/Subscriptions/Subscriber.php
+++ b/src/Subscriptions/Subscriber.php
@@ -130,7 +130,7 @@ class Subscriber
      */
     public static function uniqueChannelName(): string
     {
-        return 'private-'.Str::random(32).'-'.time();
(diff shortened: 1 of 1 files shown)


stayallive:
Since there is no good way I saw to access the driver specific config (from the `lighthouse.broadcasters` config where I would place it since it's Pusher specific in this case) in the Subscription class where this channel name is generated. And it didn't seem needed IMHO, also prevents user configuration errors like the 164 channel limit Pusher imposes.

But if this is accepted I'm sure a follow up could possibly add a configuration key for it or a PR with configuration option could supersede this PR.

PR #619, opened 2019-02-26
Update composer.json with Laravel 5.8 dependencies.

francisco-ltech:
- [ ] Updated composer.json with Laravel 5.8 dependencies

**Related Issue/Intent**

None

**Changes**

The latest Laravel dependencies are for 5.7. Updated composer.json to also include 5.8

**Breaking changes**

None

PR #621, opened 2019-02-27
Provide Laravel 5.8 support.

lucasmichot:
- [x] Added or updated tests

**Related Issue/Intent**
Upgrade to Laravel 5.8

**Changes**
This PR adds support to Laravel 5.8

**Breaking changes**
The master branch cannot support Laravel <5.8 anymore.
The reason for that is the change of signature on setUp and tearDown (`: void`) introduced in orchestral 3.8, so that it can support Phpunit 8.

---

TODO:

- [x] Check caching move to duration in seconds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,23 +13,11 @@ env:
 matrix:
   include:
     - php: 7.1
-      env: LARAVEL='5.5.*'
-    - php: 7.1
-      env: LARAVEL='5.6.*'
-    - php: 7.1
-      env: LARAVEL='5.7.*'
+      env: LARAVEL='5.8.*'
     - php: 7.2
-      env: LARAVEL='5.5.*'
-    - php: 7.2
-      env: LARAVEL='5.6.*'
-    - php: 7.2
-      env: LARAVEL='5.7.*'
+      env: LARAVEL='5.8.*'
     - php: 7.3
-      env: LARAVEL='5.5.*'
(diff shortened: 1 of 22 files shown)


lucasmichot:
This PR replaces https://github.com/nuwave/lighthouse/pull/619

olivernybroe:
@lucasmichot Don't upgrade to orchestral 3.8 then. I don't think it would be a good idea to move support for older Laravel versions, or at least keep support for 5.7.

olivernybroe:
Also as I commented in the other PR, this does not take into account the changes made to cache.

lucasmichot:
> Don't upgrade to orchestral 3.8 then
Orchestral 3.7 does not support Illuminate 5.8

lucasmichot:
> Also as I commented in the other PR, this does not take into account the changes made to cache.

This is a WIP PR and noted in the todos list

lucasmichot:
@olivernybroe on the Cache seconds topics, there is nothing to do has we already pass the duration in seconds (which was a smart decision).
A Carbon (also DateTimeInterface) instance is created and handled accordingly in `\Illuminate\Support\InteractsWithTime::secondsUntil`

---

No BC on this side

lucasmichot:
I let someone else search for an alternative testing package and provide a PR.
I will use my fork for 5.8 support

PR #625, opened 2019-02-28
Actually use Y-m-d H:i:s for DateTime scalar

### LINKED ISSUE #599: A input of type: DateTime should be accepted in the same format as as output
**Describe the bug**
DateTime currently serializes via `Carbon::toAtomString()` which is in this format e.g. `"2019-07-07T20:46:28+00:00"`, but when it is called by @update it only parses the input string via `Carbon::DEFAULT_TO_STRING_FORMAT` which is `"Y-m-d H:i:s"` 

**Expected behavior**
DateTime should accept inputs in the same format as it's output. From the example above, I should be able to pass the exact same string back in an @update or @create as I got it.

**Proposed changes**

Change Line 71 in lighthouse/src/Schema/Types/Scalars/DateTime.php from
`            return Carbon::createFromFormat(Carbon::DEFAULT_TO_STRING_FORMAT, $value);`

to
`            return Carbon::parse($value);`      

Now both input formats work

spawnia:
- [x] Added or updated tests

**Related Issue/Intent**

Resolves #599 	

**Changes**

The description for DateTime scalars was lying and the behaviour was inconsistent.
Now they are always formatted/parsed as Y-m-d H:i:s

### DIFF
--- a/src/Schema/Types/Scalars/DateTime.php
+++ b/src/Schema/Types/Scalars/DateTime.php
@@ -20,11 +20,12 @@ class DateTime extends ScalarType
     public function serialize($value): string
     {
         if ($value instanceof Carbon) {
-            return $value->toAtomString();
+            return $value->toDateTimeString();
         }
 
-        return $this->tryParsingDateTime($value, InvariantViolation::class)
-            ->toAtomString();
+        return $this
+            ->tryParsingDateTime($value, InvariantViolation::class)
+            ->toDateTimeString();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


chrissm79:
Thanks @spawnia!

datashaman:
Why though?

This change makes the datetime-rendered value go from an unambiguous timezone-aware output format to an ambiguous timezone-naive output format.

Before we could tell from the string what timezone it is in, now we have no idea.

Am I missing something?

spawnia:
@datashaman We can add another scalar `DateTimeTz` for that purpose.

PR #628, opened 2019-02-28
Add support for file uploads

### LINKED ISSUE #77: Handling file uploads
Hey @chrissm79,

I'm curious about your thoughts on file uploads?

Right now I'm quite happy handling the uploads through a separate controller that simply accepts the file, marks it as transient and returns an ID.

Then for a mutation we simply send across a list of these transient file ids, and the mutation connects the file to the resource and marks it as a completed file, any files that are not marked as completed get deleted after 30 minutes.

For me this has the advantage of being able to use pretty much any standard file uploader on the frontend.

But I know there are some who like to keep everything within graphql, there's been some discussion going on at https://github.com/Folkloreatelier/laravel-graphql/issues/125 regarding this.

What are your thoughts here?

alberthaff:
- [x] Added or updated tests
- [x] Conform to the changes in #645 

**Related Issue/Intent**

Resolves #77 

**Changes**

This PR adds support for multipart/form-requests. This allows the user to upload files through  mutations. Implemented [graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec) as suggested by multiple others. 

However, this implementation does not require any third party libraries.

**Breaking changes**

There should be none.


alberthaff:
@spawnia @chrissm79 Where do you think it would be appropriate to place the docs for this?

There is a new scalar called `Upload` that has to be included the the users schema in order for this to work.

olivernybroe:
Regarding docs, I think adding it under directives should suffice. 
Remember to mention that it depends on both the Scalar and using the default controller.

tests/Unit/Schema/Types/Scalars/UploadTest.php
```diff
@@ -0,0 +1,52 @@
...
+class UploadTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function itThrowsIfSerializing(): void
+    {
+        $this->expectException(Error::class);
+
+        (new Upload())->serialize('');
```
spawnia: remove braces after new

tests/Utils/Mutations/Upload.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Tests\Utils\Mutations;
+
+use Exception;
+use Illuminate\Http\UploadedFile;
+
+/**
+ * This is used solely as a placeholder resolver, as schemas without a valid
```
spawnia: remove this comment

tests/Integration/GraphQLTest.php
```diff
@@ -214,4 +265,61 @@ public function itIgnoresInvalidJSONVariables(): void
...
+            uploadAvatar(user: $user, file: $file) {
+                id
+                url
+            }
+        }
+        ';
+        $res = $this->postGraphQLMultipart(
+            [
+                'operations' => [
+                    'query' => $query,
```
spawnia: inline temp variables

tests/Integration/GraphQLTest.php
```diff
@@ -214,4 +265,61 @@ public function itIgnoresInvalidJSONVariables(): void
...
+                    ],
+                ],
+                'map' => [
+                    '0' => ['variables.file'],
+                ],
+                0 => UploadedFile::fake()->create('image.jpg', 500),
+            ]
+        );
+
+        $res->assertJson([
```
spawnia: chain such calls

tests/Integration/GraphQLTest.php
```diff
@@ -86,6 +99,44 @@ function (Task $task): array {
...
+                tasks {
+                    name
+                }
+            }
+        }
+        ';
+        $this->postGraphQLMultipart(
+            [
+                'operations' => [
+                    'query' => $query,
```
spawnia: inline

src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -88,12 +90,21 @@ public function query(Request $request): Response
      */
     protected function execute(Request $request, GraphQLContext $context): array
     {
+        $operations = json_decode($request->input('operations'));
+
+        $query = optional($operations)->query ?: $request->input('query', '');
```
spawnia: Is optional even necessary here?

src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -64,8 +64,10 @@ public function __construct(
      */
     public function query(Request $request): Response
     {
-        // If the request is a 0-indexed array, we know we are dealing with a batched query
-        $batched = isset($request->toArray()[0]) && config('lighthouse.batched_queries', true);
+        // If the request is a 0-indexed array and the request isn't a multipart-request,
+        // we know we are dealing with a batched query
+        $batched = isset($request->toArray()[0]) && $this->isMultipartRequest($request) == false &&
```
spawnia: split across multiple lines cleanly

src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -88,12 +90,21 @@ public function query(Request $request): Response
      */
     protected function execute(Request $request, GraphQLContext $context): array
     {
+        $operations = json_decode($request->input('operations'));
+
+        $query = optional($operations)->query ?: $request->input('query', '');
+        $variables = (array) optional($operations)->variables ?: $request->input('variables', []);
```
spawnia: Bundle these together maybe so you just have to check for operations once?

src/Schema/Types/Scalars/Upload.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Serialize an internal value, ensuring it is a valid date string.
+     *
+     * @param $value
+     * @throws Error
+     */
+    public function serialize($value)
+    {
+        throw new Error('"Upload" cannot be serialized');
```
spawnia: Do not throw a user facing `Error` here.

spawnia:
I think this feature warrants a seperate page in `guides` - maybe called `File Uploads`?

alberthaff:
Thanks for the changes.

I noticed that there aren't any docs for the two other built-in scalars. Wouldn't it be an idea to make a page called "Scalars" under "API Reference" and then add all of them there?

spawnia:
> Wouldn't it be an idea to make a page called "Scalars" under "API Reference" and then add all of them there?

That would be a great idea!

Parts of this feature definitely have their place in the API Reference, but those should mostly just give technical details and link to the comprehensive guide that explains the usage.

alberthaff:
Great. I'll add both pages to this PR.

src/Schema/Types/Scalars/Upload.php
alberthaff: Should I just throw a `\Exception` for the development related errors, or do you have a special exception for cases like this?

spawnia: We have been using \GraphQL\Error\InvariantViolation from webonyx so far.

docs/master/api-reference/scalars.md
```diff
@@ -0,0 +1,55 @@
...
+    appointments: [Appointment] @paginate(type: "connection")
+}
+
+type Mutation {
+    addAppointment(text: String!, date: DateTime!): Appointment @create
+}
+
+```
+
+## Date
```
olivernybroe: Add a one liner to explain what this scalar does.

docs/master/api-reference/scalars.md
```diff
@@ -0,0 +1,55 @@
...
+**Parse:**
+
+Expects an ISO date string, eg.: `2019-01-15`. Parses to a Carbon-object.
+
+**Serialize:**
+
+Convert a date or Carbon-object to an ISO 8601 date string, eg. `2019-01-15`.
+
+## DateTime
+
```
olivernybroe: Add a one liner to explain what this scalar does.

docs/master/api-reference/scalars.md
```diff
@@ -0,0 +1,55 @@
...
+type Mutation {
+    addAppointment(text: String!, date: DateTime!): Appointment @create
+}
+
+```
+
+## Date
+**Parse:**
+
+Expects an ISO date string, eg.: `2019-01-15`. Parses to a Carbon-object.
```
olivernybroe:
```suggestion
Expects an ISO 8601 date string (`Y-m-d`), eg.: `2019-01-15`. Parses to a Carbon-object.
```

docs/master/api-reference/scalars.md
```diff
@@ -0,0 +1,55 @@
...
+
+**Serialize:**
+
+Convert a date or Carbon-object to an ISO 8601 date string, eg. `2019-01-15`.
+
+## DateTime
+
+**Parse:**
+
+Expects an ISO 8601 datetime string, eg.: `2019-01-15T23:15:33`. Parses to a `Carbon`-object.
```
olivernybroe: This is actually not correct, the formatter it's trying to parse is `Y-m-d H:i:s`, and ISO 8601 is defined as `Y-m-d\TH:i:sO`.

docs/master/guides/file-uploads.md
```diff
@@ -0,0 +1,59 @@
+# File Uploads
```
olivernybroe: Add text explaining this requires the default Controller.

docs/master/guides/file-uploads.md
```diff
@@ -0,0 +1,59 @@
...
+
+Once the scalar is added, you can add it to a mutation.
+
+````graphql
+type Mutation {
+    addProfilePicture(file: Upload!): ProfilePicture @field(resolver: "App\\GraphQL\\Mutations\\Models\\ProfilePictureMutator@add")
+}
+````
+
+## Uploading a file
```
olivernybroe: Maybe add a `CURL` example or something similar, as that might be easier for people to understand.

alberthaff:
> Remember to mention that it depends on both the Scalar and using the default controller.

Thanks. I'll add that to the guide.

src/Support/Http/Controllers/GraphQLController.php
alberthaff: Yes. Otherwise an ErrorException will be thrown when using a traditional POST-request:
`Trying to get property 'query' of non-object`

However, I have wrapped `json_decode ` in `optional()` instead of using it for each parameter.

src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -88,12 +91,21 @@ public function query(Request $request): Response
      */
     protected function execute(Request $request, GraphQLContext $context): array
     {
+        $operations = optional(json_decode($request->input('operations')));
```
olivernybroe: Do not use `optional()`. the helper is not defined in lumen.

olivernybroe: Also it is not supported in Laravel 5.5 according to Travis CI ;)

spawnia:
I think the tests for file uploads are wrong - the request that is sent from the test methods is not the same as an actual upload via `curl` as described in the spec.

We might have to use Guzzle for making a proper multipart form request: http://docs.guzzlephp.org/en/latest/request-options.html#multipart

#645 changes how the controller handles requests quite a bit, so let's hold off on merging this and rebase it on top of that once we got it through.

docs/master/guides/file-uploads.md
```diff
@@ -1,71 +1,79 @@
...
 
-Lighthouse conforms to the [graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec)
-which use multipart/form-requests to transfer files in the same request as the query.
+Lighthouse allows you to upload files using a multipart form request
+as defined in [graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec).
 
-This allows for file uploads through mutations.
+::: warning
+The upload functionality is implemented in Lighthouse's default controller `\Nuwave\Lighthouse\Support\Http\Controllers\GraphQLController`.
+When using a custom controller, you will have to add support for it yourself.
```
alberthaff: 👍

alberthaff:
@spawnia Thanks! I'll take a look at it tomorrow. :-)

alberthaff:
@spawnia Okay. I will rewrite `postGraphQLMultipart` inside `TestCase` so that it relies on Guzzle for requests.

Yes, let's hold on until #645 is merged. I will continue work on my client meanwhile.
While we may have to change some of the code in `GraphQLController`, we will most likely stick to the `graphql-multipart-request-spec`, right? If so, i will proceed to implement it in my client.

src/Schema/Types/Scalars/Upload.php
```diff
@@ -10,47 +10,53 @@
 
 class Upload extends ScalarType
 {
-    public $name = 'Upload';
-
-    public $description = ''; // TODO Add description / refer to docs.
```
alberthaff: Are name and description not used by Lighthouse?

tests/Utils/Mutations/Upload.php
```diff
@@ -10,18 +10,14 @@ class Upload
...
     {
-        if (! $args['file'] instanceof UploadedFile) {
-            throw new Exception('Argument "file" is not of type "UploadedFile"');
-        }
-
-        return [
-            'id' => 123,
-            'url' => 'http://localhost.dev/image_123.jpg',
-        ];
+        return $args['file'] instanceof UploadedFile;
```
alberthaff: Much cleaner!

spawnia:
Great. It does not necessarily have to be Guzzle, but we must ensure that the request is an actual multipart form request. Try to call a running server with the curl example, you will see how it is different from what was being tested now.

Yeah, let's stick to the spec for sure.

You could check out https://github.com/mll-lab/lighthouse/tree/event-based-request-lifecycle
Would be great to have you try it since it changes up some things quite a bit and allows you to implement file uploads on top.

src/Schema/Types/Scalars/Upload.php
spawnia: They are put in from the schema definition. This allows users to change them if needed.

alberthaff:
The test currently posts to posthere.io, but i may need some help getting it to post directly to the HttpKernel. Are we going to start a local PHP server for the tests, or is it possible to simulate an HTTP request like Laravel's built-in methods?

composer.json
```diff
@@ -20,6 +20,7 @@
     "require": {
         "php": ">= 7.1",
         "ext-json": "*",
+        "GuzzleHttp/guzzle": "^6.3",
```
olivernybroe: As guzzle are only used for a test, this should be in `require-dev`

alberthaff:
An other solution could be to use Laravel's functionality for simulating post requests, and generating the body with:

```php
$payload; // Payload generated in postGraphQLMultipart
$multipartBody = (new GuzzleHttp\Psr7\MultipartStream($payload))->getContents();
```

`$multipartBody` is a valid multipart/form request body. We just have to figure out how to post it to the HttpKernel.

composer.json
alberthaff: Thanks

olivernybroe:
@spawnia We cannot use Guzzle or something similar like that without starting a web server.
I think we should use the `call` method and then pass the files in through the `files` arg. 

This will create a Request object with files, which is what we want. We do not need to test if Laravel is able to create a Request object correctly, only to test that we are able to parse a correct Request object.

composer.json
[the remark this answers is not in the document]
spawnia: nevermind

tests/TestCase.php
```diff
@@ -199,8 +199,52 @@ protected function postGraphQL(array $data, array $headers = []): TestResponse
...
+            $parameters['map'] = json_encode($data['map']);
+        }
+
+        // Add files
+        $files = [];
+
+        // If array is zero-index, there are files attached to the query
+        if (isset($data[0])) {
+            // Loop through all files, and add them to the payload
+            for ($i = 0; isset($data[$i]); $i++) {
```
spawnia: Why is this so complicated? Can't this be `$files = $data;`

alberthaff: `$data` also contains `operations` and `map`. Would you prefer following?
```php
unset($data['operations']);
unset($data['map']);

$files = $data; // unnecessary - can just be added directly as an argument in $this->call()
```

spawnia: The keys for the files are actually arbitrary, so let's just pass them as strings.

```php
[
	'operations' => //
	'map' => //
	'0' => //
]
```

alberthaff: So we are reverting to the original test?

spawnia: Just checking if this simulates the request properly in my local environment, will report back soon.

spawnia:
@alberthaff nice job with figuring out the correct usage of `$this->call()`.

Since we do have so few tests that use `postGraphQLMultipart`, i just stripped it way down and reduced its complexity.

This kind of API works great for tests, as we can simulate all kinds of calls through it.

tests/TestCase.php
```diff
@@ -222,41 +222,20 @@ protected function postGraphQL(array $data, array $headers = []): TestResponse
      * This is used for file uploads conforming to the specification:
      * https://github.com/jaydenseric/graphql-multipart-request-spec
      *
-     * @param  mixed[]  $data
+     * @param  mixed[]  $parameters
+     * @param  mixed[]  $files
      * @return \Illuminate\Foundation\Testing\TestResponse
      */
-    protected function postGraphQLMultipart(array $data): TestResponse
+    protected function postGraphQLMultipart(array $parameters, array $files): TestResponse
```
alberthaff: 👍

alberthaff:
All thanks to @olivernybroe :-)

Great. That is a much better solution.

Should I start changing the code in `GraphQLController` to conform to the changes in #645, or are there more changes on the way?

spawnia:
Can't think of any more changes that will clash with the file uploads feature.

The `GraphQLController` contains very little logic now. The logic for unpacking the multipart structure into a proper structure lives mostly within `\Nuwave\Lighthouse\Execution\GraphQLRequest` now.

alberthaff:
@spawnia I moved the logic into `LighthouseRequest`. Should we still display a warning in the docs about overwriting `LighthouseRequest`?

spawnia:
> @spawnia I moved the logic into `LighthouseRequest`. Should we still display a warning in the docs about overwriting `LighthouseRequest`?

Don't think we need that anymore. It should be pretty clear to the user that they can screw things up if they replace interfaces without addding in all necessary functionality.

spawnia:
We should add a test case and proper implementation for batched multipart requests.

alberthaff:
Great. I'll update the docs and add support for batched multipart requests tonight.

alberthaff:
@spawnia @chrissm79 Should be ready for review now.

chrissm79:
Amazing work on this one @alberthaff! I'll give this a review and get it merged in no later than tomorrow if there aren't any issues 👍 Thanks for your work on this, great feature to add to the arsenal!!!

alberthaff:
@chrissm79 Thanks!

spawnia:
Will go through it tomorrow.

docs/master/guides/file-uploads.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Client-side usage
+
+In order to upload a file, you must send a `multipart/form-data` request.
+Use any of the [available client implementations](https://github.com/jaydenseric/graphql-multipart-request-spec#client)
+or look at the [specification examples](https://github.com/jaydenseric/graphql-multipart-request-spec#multipart-form-field-structure) to roll your own.
+
+To test the example above, prepare a file you can upload.
+
+    echo "test content" > my_file.txt
```
olivernybroe: Should have the keyword `bash` on for highlight

docs/master/guides/file-uploads.md
```diff
@@ -0,0 +1,74 @@
...
+
+Then, send a request to upload the file to your server:
+
+    curl localhost/graphql \
```
olivernybroe: Should have the keyword `bash` on for highlight

src/Execution/LighthouseRequest.php
```diff
@@ -4,7 +4,7 @@
 
 use Illuminate\Http\Request;
 
-class LighthouseRequest implements GraphQLRequest
+class LighthouseRequest extends BaseRequest
```
olivernybroe: @spawnia would we prefer this class to implement `GraphQLRequest` instead as that make it clearer to understand that this is actually an implementation of the interface.

spawnia: We don't do that with the directives and the `BaseDirectives`, so it would stay consistent and do the same here.

chrissm79:
Excellent work on this @alberthaff, thank you!!

PR #635, opened 2019-02-28
Redirect to custom domain

alberthaff:
**Changes**

Added `_redirects`, that instructs netlify to redirect all visitors to lighthouse-php.com.

Netfliy docs: https://www.netlify.com/docs/redirects


chrissm79:
Awesome, thanks @alberthaff!

PR #645, opened 2019-03-04
Event based request lifecycle

spawnia:
- [x] Added or updated tests

**Related Issue/Intent**

Unify how to hook into parts of the request lifecycle via events.

This gives more control to end users as well as plugin developers, without having
to learn a Lighthouse specific way of hooking into the lifecycle - it's all based on events.

**Changes**

What were previously extensions - Subscriptions, Tracing and Defer - was rebuilt using
the event system.

A new `GraphQLRequest` singleton was added to wrap the raw HTTP request and add GraphQL
semantics and make it accessible throughout the codebase.

Multiple events were added and documented in the API reference.

**Breaking changes**

This replaces the extension registration with a more flexible event-based approach.

From a user perspective, extensions will have to be registered through a more standard
ServiceProvider, rather then the Lighthouse config.


spawnia:
@olivernybroe @chrissm79 @lucasmichot would love some feedback before i type up the docs for this :)

src/Defer/Defer.php
```diff
@@ -59,72 +61,40 @@ class DeferExtension extends GraphQLExtension
 
     /**
      * @param  \Nuwave\Lighthouse\Support\Contracts\CanStreamResponse  $stream
+     * @param  \Nuwave\Lighthouse\GraphQL  $graphQL
      * @return void
      */
-    public function __construct(CanStreamResponse $stream)
+    public function __construct(CanStreamResponse $stream, GraphQL $graphQL)
```
lucasmichot: ❤️

src/Tracing/Tracing.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    protected $resolverTraces = [];
+
+    /**
+     * Set the tracing directive on all fields of the query to enable tracing them.
+     *
+     * @param  \Nuwave\Lighthouse\Events\ManipulatingAST  $manipulatingAST
+     * @return void
+     */
+    public function handleManipulatingAST(ManipulatingAST $manipulatingAST): void
```
lucasmichot: I would just call it `handle`

spawnia: As you scroll further down this class, you will see it handles many different events :)

tests/Integration/Subscriptions/BroadcastManagerTest.php
```diff
@@ -52,39 +58,42 @@ public function itCanExtendBroadcastManager(): void
...
-                {
-                    return $data;
-                }
-            };
+            return $broadcaster;
         });
 
-        $data = ['foo' => 'bar'];
-        $broadcaster = $this->broadcastManager->driver('foo');
+        /** @var Broadcaster $broadcaster */
```
lucasmichot: I would use the FQN here too

tests/Integration/Subscriptions/StorageManagerTest.php
```diff
@@ -20,13 +27,115 @@ class StorageManagerTest extends TestCase implements GraphQLContext
...
+
+                /**
+                 * Unserialize the context.
+                 *
+                 * @param  string $context
+                 * @return \Nuwave\Lighthouse\Support\Contracts\GraphQLContext
+                 */
+                public function unserialize(string $context)
+                {
+                    return new class() implements GraphQLContext {
```
lucasmichot: no need of braces `()`

tests/Integration/Subscriptions/StorageManagerTest.php
```diff
@@ -20,13 +27,115 @@ class StorageManagerTest extends TestCase implements GraphQLContext
...
+    }
+
+    /**
+     * @see \Nuwave\Lighthouse\Support\Contracts\GraphQLContext::user()
+     *
+     * @return \Tests\Utils\Models\User
+     */
+    public function user(): User
+    {
+        return new User([
```
lucasmichot: Would use factory make

lucasmichot:
@spawnia I need a bit more time to make a full review, but for now it looks very promising 🎉

tests/Integration/Subscriptions/StorageManagerTest.php
spawnia: Can you elaborate on that?

src/Defer/Defer.php
spawnia: ;)

tests/Integration/Subscriptions/BroadcastManagerTest.php
spawnia: Done

src/Defer/Defer.php
```diff
@@ -310,18 +282,12 @@ protected function maxNestedFieldsResolved(int $nested): bool
...
-        // TODO: Properly parse variables array
-        // TODO: Get debug setting
-        $this->data = graphql()
-            ->executeQuery(
-                $this->request->request()->input('query', ''),
-                $this->request->context(),
-                $this->request->request()->input('variables', [])
-            )
-            ->toArray(config('lighthouse.debug'));
+        $this->result = app()->call(
```
lucasmichot: 👍

src/Defer/DeferrableDirective.php
```diff
@@ -11,25 +11,25 @@
...
 use Nuwave\Lighthouse\Support\Contracts\FieldMiddleware;
-use Nuwave\Lighthouse\Schema\Extensions\ExtensionRegistry;
 
 class DeferrableDirective extends BaseDirective implements Directive, FieldMiddleware
 {
+    const NAME = 'deferrable';
+
     /**
-     * @var \Nuwave\Lighthouse\Schema\Extensions\ExtensionRegistry
+     * @var Defer
```
lucasmichot: FQN

lucasmichot: Search this case-sensitive regexp `@var\s*[A-Z]`

src/Events/ManipulatingAST.php
```diff
@@ -0,0 +1,26 @@
...
+use Nuwave\Lighthouse\Schema\AST\DocumentAST;
+
+class ManipulatingAST
+{
+    /**
+     * The AST that can be manipulated.
+     *
+     * @var \Nuwave\Lighthouse\Schema\AST\DocumentAST
+     */
+    public $ast;
```
lucasmichot: I would have `$documentAST` here, just  a personal preference though

src/Execution/GraphQLRequest.php
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @return bool
+     */
+    public function advanceBatchIndex(): bool
+    {
+        if ($this->hasMoreBatches()) {
+            $this->batchIndex++;
+
+            return true;
+        } else {
```
lucasmichot: Useless `else... return`

lucasmichot: I guess you can do:
```php
if ($result = $this->hasMoreBatches()) {
    $this->batchIndex++;
}

return $result;
```

src/GraphQL.php
```diff
@@ -65,72 +62,79 @@ class GraphQL
...
+    protected $eventDispatcher;
+
+    /**
+     * The AST builder.
      *
-     * @var int|null
+     * @var \Nuwave\Lighthouse\Schema\AST\ASTBuilder
      */
-    protected $currentBatchIndex = null;
+    private $astBuilder;
```
lucasmichot: Then nobody can extend this? Maybe better to have it `protected`?

src/GraphQL.php
```diff
@@ -139,21 +143,23 @@ public function executeBatchedQueries(array $requests, GraphQLContext $context,
      * To render the ExecutionResult, you will probably want to call `->toArray($debug)` on it,
      * with $debug being a combination of flags in \GraphQL\Error\Debug
      *
-     * @param  string  $query
+     * @param  string|\GraphQL\Language\AST\DocumentNode  $query
```
lucasmichot: Or the other way around: `\GraphQL\Language\AST\DocumentNode|string`

src/GraphQL.php
```diff
@@ -139,21 +143,23 @@ public function executeBatchedQueries(array $requests, GraphQLContext $context,
...
-        string $query,
+        $query,
         GraphQLContext $context,
         ?array $variables = [],
         $rootValue = null,
         ?string $operationName = null
     ): ExecutionResult {
-        $operationName = $operationName ?: app('request')->input('operationName');
+        $this->eventDispatcher->dispatch(
+            new StartExecution()
```
lucasmichot: Useless new braces

src/GraphQL.php
```diff
@@ -166,7 +172,19 @@ public function executeQuery(
             $this->getValidationRules() + DocumentValidator::defaultRules()
         );
 
-        $result->extensions = $this->extensionRegistry->jsonSerialize();
+        // Listeners of this event should return an array comprised of
+        // a single key and the extension content as the value.
+        $extensionResults = $this->eventDispatcher->dispatch(
+            new GatheringExtensions()
```
lucasmichot: same
Search insensitive regexp: `new [a-z\\]*\(\)`

src/Schema/AST/ASTBuilder.php
```diff
@@ -12,39 +12,47 @@
 use Nuwave\Lighthouse\Schema\Factories\DirectiveFactory;
 use Nuwave\Lighthouse\Support\Contracts\NodeManipulator;
 use Nuwave\Lighthouse\Support\Contracts\FieldManipulator;
-use Nuwave\Lighthouse\Schema\Extensions\ExtensionRegistry;
 
 class ASTBuilder
 {
+    /**
+     * @var \Nuwave\Lighthouse\Schema\Factories\DirectiveFactory
+     */
```
lucasmichot: `@return void`

src/Schema/AST/ASTBuilder.php
```diff
@@ -229,9 +237,8 @@ protected static function addPaginationInfoTypes(DocumentAST $document): Documen
           lastPage: Int
         }
         ');
-        $document->setDefinition($pageInfo);
 
-        return $document;
+        return $document->setDefinition($pageInfo);
```
lucasmichot: yes 👍

src/Subscriptions/Subscriber.php
```diff
@@ -19,86 +23,105 @@ class Subscriber
...
+        $this->channel = $data['channel'];
+        $this->args = $data['args'];
+        $this->context = $this->contextSerializer()->unserialize(
+            $data['context']
         );
-        $instance->args = Arr::get($data, 'args', []);
-        $instance->operationName = Arr::get($data, 'operation_name');
-        $instance->queryString = Arr::get($data, 'query_string');
+        $this->query = AST::fromArray(
+            \unserialize($data['query'])
```
lucasmichot: Let's promote root function without backslah: `unserialize`

src/Subscriptions/Subscriber.php
```diff
@@ -19,86 +23,105 @@ class Subscriber
...
+     * @return false|string
+     */
+    public function serialize()
+    {
+        return json_encode([
+            'operation_name' => $this->operationName,
+            'channel' => $this->channel,
+            'args' => $this->args,
+            'context' => $this->contextSerializer()->serialize($this->context),
+            'query' => \serialize(
```
lucasmichot: same here

src/Subscriptions/Subscriber.php
```diff
@@ -19,86 +23,105 @@ class Subscriber
...
+            ),
+        ]);
     }
 
     /**
      * Set root data.
      *
      * @param  mixed  $root
-     * @return $this
+     * @return \Nuwave\Lighthouse\Subscriptions\Subscriber
```
lucasmichot: so `$this` then

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -156,22 +158,27 @@ public function subscriptions(Subscriber $subscriber): Collection
     }
 
     /**
-     * Get all current subscribers.
+     * Reset the collection of subscribers when a new execution starts.
      *
-     * @return string[]
+     * @param \Nuwave\Lighthouse\Events\StartExecution $startExecution
      */
```
lucasmichot: `@return void`

src/Subscriptions/SubscriptionServiceProvider.php
```diff
@@ -0,0 +1,70 @@
...
+
+        // Register the routes for the configured broadcaster. The specific
+        // method that is used can be changed, so we retrieve its name
+        // dynamically and then call it with an instance of 'router'.
+        $broadcaster = config('lighthouse.subscriptions.broadcaster');
+        if ($routesMethod = config("lighthouse.subscriptions.broadcasters.{$broadcaster}.routes")) {
+            [$router, $method] = Str::parseCallback($routesMethod, 'pusher');
+
+            call_user_func(
+                [app($router), $method],
```
lucasmichot: in most Laravel provider, the syntax is `$app[$router]`

src/Tracing/TracingDirective.php
```diff
@@ -10,19 +10,35 @@
...
+     *
+     * @param  \Nuwave\Lighthouse\Tracing\Tracing  $tracing
+     * @return void
+     */
+    public function __construct(Tracing $tracing)
+    {
+        $this->tracing = $tracing;
+    }
+
+    const NAME = 'tracing';
```
lucasmichot: Let's put  constants before variables and methods

src/Tracing/TracingDirective.php
```diff
@@ -38,20 +54,23 @@ public function handleField(FieldValue $value, Closure $next): FieldValue
...
-
+        return $value->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver) {
             $start = Carbon::now();
-            $result = $resolver($root, $args, $context, $info);
+            $result = $resolver($root, $args, $context, $resolveInfo);
 
             ($result instanceof Deferred)
-                ? $result->then(function (&$items) use ($info, $start, $tracingExtension) {
-                    $tracingExtension->record($info, $start, Carbon::now());
+                ? $result->then(function () use ($resolveInfo, $start): void {
```
lucasmichot: Or maybe just:

```php
$end = Carbon::now();

if (($result instanceof Deferred)) {
    $result->then(function () use ($resolveInfo, $start, $end): void {
        $this->tracing->record($resolveInfo, $start, $end);
    });
} else {
    $this->tracing->record($resolveInfo, $start, $end);
}
```

src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -26,126 +24,77 @@ class GraphQLController extends Controller
...
+     * @param  \Illuminate\Contracts\Events\Dispatcher  $eventsDispatcher
+     * @param  \Nuwave\Lighthouse\Support\Contracts\GraphQLResponse  $createsResponse
      * @return void
      */
     public function __construct(
-        ExtensionRegistry $extensionRegistry,
         GraphQL $graphQL,
         CreatesContext $createsContext,
-        GraphQLResponse $graphQLResponse
+        Defer $defer,
```
chrissm79: Looks like this might have been accidentally left in?

tests/Integration/Defer/SetUpDefer.php
```diff
@@ -1,27 +1,27 @@
...
 {
     /**
      * @var \Nuwave\Lighthouse\Support\Http\Responses\MemoryStream
      */
     protected $stream;
 
-    protected function setUpInMemoryStream($app)
+    /**
+     * @param \Illuminate\Foundation\Application $app
+     */
```
lucasmichot: `@return void`

tests/Integration/Subscriptions/StorageManagerTest.php
lucasmichot:
```php
return factory(User::class)->make([
    'first_name' => 'John',
    'last_name' => 'Doe',
    'email' => 'john@doe.com',
]);
```

lucasmichot:
2nd pass @spawnia  - looks very good so far @spawnia 👍

chrissm79:
@spawnia Starting to dig into this one today. First off, I think the event based approach is a great idea as it keeps things nice a separated allowing listeners to just hook into what concerns them 👍 

In a project, I have a need to manipulate the request object before it makes it's way to the GraphQL execution. For instance, I'm implementing [APQ](https://blog.apollographql.com/automatic-persisted-queries-and-cdn-caching-with-apollo-server-2-0-bf42b3a313de) and I need to swap out the hashed key w/ the full query string out of cache. Right now, I see that I can hook into `startRequest` but it doesn't have a reference to the `GraphQLRequest` object so I can manipulate the input.

On the response side, it would also be nice if we could manipulate the response (if needed) before it's sent out to the client, so it would be great if we could add an additional hook `WillSendResponse` to w/ a `$response` property (from [here](https://github.com/nuwave/lighthouse/blob/70cc8d88ebb0dca1e277037ecd63970e1958f7ea/src/Support/Http/Controllers/GraphQLController.php#L77-L79)) that would allow listeners to manipulate the response prior to it being sent down to the client.

Great work on this one!!

chrissm79:
If we have `StartRequest`, `StartExecution`, `WillSendResponse`, maybe we could rename the others to `BuildExtensionMeta` and `ManipulateAST` rather than `GatheringExtensions` and `ManipulatingAST`?

tests/Integration/Subscriptions/StorageManagerTest.php
spawnia: Just realized we can remove the method entirely

src/Schema/AST/ASTBuilder.php
```diff
@@ -12,39 +12,47 @@
...
-use Nuwave\Lighthouse\Schema\Extensions\ExtensionRegistry;
 
 class ASTBuilder
 {
+    /**
+     * @var \Nuwave\Lighthouse\Schema\Factories\DirectiveFactory
+     */
+    private $directiveFactory;
+
+    public function __construct(DirectiveFactory $directiveFactory)
```
lucasmichot: missing docblocks

src/GraphQL.php
spawnia: It's a string most of the time, and the base library we pass this to annotates this like so:

`     * @param string|DocumentNode $source`

so i am going to leave it like this.

src/Events/ManipulatingAST.php
spawnia: Done

spawnia:
> If we have `StartRequest`, `StartExecution`, `WillSendResponse`, maybe we could rename the others to `BuildExtensionMeta` and `ManipulateAST` rather than `GatheringExtensions` and `ManipulatingAST`?

Then we might do `RegisterDirectiveBaseNamespaces` and `BuildAST` to get rid of all the `-ing`.

Maybe we can find a naming convention that makes it clear which events expect a return/can be used to modify things?

spawnia:
@chrissm79 is it enough for those use cases to simply use Middleware?

src/Defer/Defer.php
```diff
@@ -40,12 +42,12 @@ class DeferExtension extends GraphQLExtension
     /**
      * @var bool
      */
-    protected $defer = true;
+    protected $deferMore = true;
```
olivernybroe: What does this actually do? 
It feels weird when reading it.

src/Defer/DeferrableDirective.php
```diff
@@ -39,7 +39,7 @@ public function __construct(ExtensionRegistry $extensions)
      */
     public function name(): string
     {
-        return 'deferrable';
+        return self::NAME;
```
olivernybroe: If you are adding a constant for name, then let's move this to `BaseDirective`, so you won't have to define both the constant and the method. 

Users can still override the method in the rare cases where they need more advanced logic.

src/Defer/DeferrableDirective.php
```diff
@@ -55,42 +55,41 @@ public function handleField(FieldValue $value, Closure $next): FieldValue
...
+     * @param  \GraphQL\Type\Definition\ResolveInfo  $resolveInfo
      * @return bool
      *
      * @throws \Nuwave\Lighthouse\Exceptions\ParseClientException
      */
-    protected function shouldDefer(TypeNode $fieldType, ResolveInfo $info): bool
+    protected function shouldDefer(TypeNode $fieldType, ResolveInfo $resolveInfo): bool
     {
-        if (strtolower($info->operation->operation) === 'mutation') {
+        if (strtolower($resolveInfo->operation->operation) === 'mutation') {
```
olivernybroe: Maybe change this to also use `Str::lower()` instead as we use `Str` lots of other places.
The only reason for not to do it, is that this is a base php method.

src/Execution/GraphQLRequest.php
```diff
@@ -0,0 +1,146 @@
+<?php
+
+namespace Nuwave\Lighthouse\Execution;
+
+use Illuminate\Http\Request;
+use Nuwave\Lighthouse\Support\Contracts\CreatesContext;
+use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;
+
+class GraphQLRequest
```
olivernybroe: This class basically implemented the `Iterator` interfaces for handling batches. What if we make a class for batches and make that implement `iterator` and then this class simply has an instance of the batch object?
Or make this class implement `Iterator`.

src/Execution/Utils/Subscription.php
```diff
@@ -25,7 +25,9 @@ public static function broadcast(string $subscriptionField, $root, ?bool $should
     {
         // Ensure we have a schema and registered subscription fields
         // in the event we are calling this method in code.
-        app(GraphQL::class)->prepSchema();
+        /** @var \Nuwave\Lighthouse\GraphQL $graphQL */
+        $graphQL = app(GraphQL::class);
```
olivernybroe: Why was this split into a variable instead of inline it?

src/LighthouseServiceProvider.php
```diff
@@ -38,25 +36,33 @@ class LighthouseServiceProvider extends ServiceProvider
...
      */
-    public function boot(): void
+    public function boot(ValidationFactory $validationFactory): void
     {
-        $this->mergeConfigFrom(__DIR__.'/../../config/config.php', 'lighthouse');
+        $this->mergeConfigFrom(__DIR__.'/../config/config.php', 'lighthouse');
 
         $this->publishes([
-            __DIR__.'/../../config/config.php' => config_path('lighthouse.php'),
+            __DIR__.'/../config/config.php' => config_path('lighthouse.php'),
```
olivernybroe: `config_path` helper is only used here. 
Let's remove the global helper and just do the stuff in the class itself.

src/LighthouseServiceProvider.php
```diff
@@ -38,25 +36,33 @@ class LighthouseServiceProvider extends ServiceProvider
...
         ], 'config');
 
         $this->publishes([
-            __DIR__.'/../../assets/default-schema.graphql' => config('lighthouse.schema.register'),
+            __DIR__.'/../assets/default-schema.graphql' => config('lighthouse.schema.register'),
```
olivernybroe: Remove `config` helper.

src/LighthouseServiceProvider.php
```diff
@@ -87,15 +93,36 @@ public function register(): void
...
         $this->app->singleton(NodeRegistry::class);
         $this->app->singleton(TypeRegistry::class);
         $this->app->singleton(CreatesContext::class, ContextFactory::class);
         $this->app->singleton(CanStreamResponse::class, ResponseStream::class);
-        $this->app->singleton(GraphQLResponse::class, Response::class);
 
-        $this->app->singleton(SchemaSourceProvider::class, function () {
-            return new SchemaStitcher(config('lighthouse.schema.register', ''));
+        $this->app->bind(GraphQLResponse::class, function (): GraphQLResponse {
+            return new class implements GraphQLResponse {
```
olivernybroe: Why are we creating an anonymous class here? I think we should create a class for this.

src/LighthouseServiceProvider.php
```diff
@@ -87,15 +93,36 @@ public function register(): void
...
+                /**
+                 * Create GraphQL response.
+                 *
+                 * @param  array $data
+                 * @return \Symfony\Component\HttpFoundation\Response
+                 */
+                public function create(array $data)
+                {
+                    return response($data);
```
olivernybroe: Don't use helper method `response`.

src/Schema/AST/ASTHelper.php
```diff
@@ -248,13 +248,15 @@ function (DocumentAST $document, ObjectTypeDefinitionNode $objectType) use ($dir
...
 
-                    $objectType->fields = new NodeList(collect($objectType->fields)
-                        ->map(function (FieldDefinitionNode $field) use ($directive): FieldDefinitionNode {
-                            $field->directives = $field->directives->merge([$directive]);
-
-                            return $field;
-                        })
-                        ->all());
+                    $objectType->fields = new NodeList(
+                        collect($objectType->fields)
```
olivernybroe: Do not use `collect` helper.

src/Schema/Directives/Fields/DeprecatedDirective.php
```diff
@@ -29,15 +29,14 @@ public function name()
...
+    public function handleField(FieldValue $value, Closure $next): FieldValue
     {
         $reason = $this->directiveArgValue('reason');
 
         if (! $reason) {
-            $parentName = $value->getParentName();
-            $message = "The @{$this->name()} directive requires a `reason` argument [defined on {$parentName}]";
-
-            throw new DirectiveException($message);
+            throw new DirectiveException(
```
olivernybroe: We should prob. define in our `contributing.md` that when we throw exceptions if the message should have a `.` at the end or not.

src/Schema/Factories/DirectiveFactory.php
```diff
@@ -52,16 +53,18 @@ class DirectiveFactory
...
     /**
      * DirectiveFactory constructor.
+     *
+     * @param  \Illuminate\Contracts\Events\Dispatcher  $dispatcher
      * @return void
      */
-    public function __construct()
+    public function __construct(Dispatcher $dispatcher)
     {
         $this->directiveBaseNamespaces = collect([
```
olivernybroe: Helper method...

src/Schema/Factories/DirectiveFactory.php
```diff
@@ -52,16 +53,18 @@ class DirectiveFactory
...
             // User defined directives (top priority)
             config('lighthouse.namespaces.directives'),
```
olivernybroe: Helper method

src/Schema/Types/GraphQLSubscription.php
```diff
@@ -29,7 +30,9 @@ public function can(Subscriber $subscriber)
      */
     public function encodeTopic(Subscriber $subscriber, string $fieldName)
     {
-        return strtoupper(snake_case($fieldName));
+        return strtoupper(
```
olivernybroe: Change to `Str::upper`?  It used `mb_strtoupper` which adds multi byte support

src/Schema/Types/GraphQLSubscription.php
```diff
@@ -41,7 +44,9 @@ public function encodeTopic(Subscriber $subscriber, string $fieldName)
      */
     public function decodeTopic(string $fieldName, $root)
     {
-        return strtoupper(snake_case($fieldName));
+        return strtoupper(
```
olivernybroe: Change to `Str::upper`?  It used `mb_strtoupper` which adds multi byte support

src/Subscriptions/Subscriber.php
```diff
@@ -19,79 +23,98 @@ class Subscriber
...
-     * Unserialize subscription.
+     * Unserialize subscription from a JSON string.
      *
      * @param  string  $subscription
-     *
-     * @return static
+     * @return $this
      */
-    public static function unserialize(string $subscription): self
+    public function unserialize($subscription): self
```
olivernybroe: What can this be other than string? Either add that in `@param` or typehint it to `string` again.

src/Subscriptions/Subscriber.php
```diff
@@ -136,7 +143,7 @@ public static function uniqueChannelName(): string
     /**
      * @return \Nuwave\Lighthouse\Subscriptions\Contracts\ContextSerializer
      */
-    protected function serializer(): ContextSerializer
+    protected function contextSerializer(): ContextSerializer
     {
         return app(ContextSerializer::class);
```
olivernybroe: laravel helper used.

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -119,20 +126,14 @@ public function subscriber(Subscriber $subscriber, string $channel): self
...
         // sure the schema has been generated.
-        app(GraphQL::class)->prepSchema();
+        $this->graphQL->prepSchema();
 
-        $documentNode = Parser::parse($subscriber->queryString, [
-            'noLocation' => true,
-        ]);
-
-        return collect($documentNode->definitions)
+        return collect($subscriber->query->definitions)
```
olivernybroe: helper used

src/Subscriptions/SubscriptionServiceProvider.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $eventDispatcher->listen(
+            GatheringExtensions::class,
+            SubscriptionRegistry::class.'@handleGatheringExtensions'
+        );
+
+        // Register the routes for the configured broadcaster. The specific
+        // method that is used can be changed, so we retrieve its name
+        // dynamically and then call it with an instance of 'router'.
+        $broadcaster = config('lighthouse.subscriptions.broadcaster');
```
olivernybroe: helper used

src/Subscriptions/SubscriptionServiceProvider.php
```diff
@@ -0,0 +1,70 @@
...
+        if ($routesMethod = config("lighthouse.subscriptions.broadcasters.{$broadcaster}.routes")) {
```
olivernybroe: helper used

tests/Integration/Defer/DeferTest.php
```diff
@@ -310,6 +303,21 @@ public function itCancelsDefermentAfterMaxExecutionTime(): void
         }
         ";
 
+        /** @var \Nuwave\Lighthouse\Defer\Defer $defer */
+        $defer = app(Defer::class);
```
olivernybroe: helper used

tests/Integration/Defer/DeferTest.php
```diff
@@ -368,6 +362,20 @@ public function itCancelsDefermentAfterMaxNestedFields(): void
...
```
olivernybroe: helper used

src/Defer/DeferrableDirective.php
spawnia: Unfortunately we cannot guarantee child directives to define a constant.

I think we could change the function to `public static name(): string`?

src/Defer/Defer.php
spawnia: Renamed to `$acceptFurtherDeferring`

src/Defer/DeferrableDirective.php
olivernybroe: We could define a default for it or just let it fail if they did not define it. This is something we set in `BaseDirective`, so it is just a helper for naming.
The contract still requires an implementation of `name()` which returns a `string`.

src/Defer/DeferrableDirective.php
spawnia: Not sure if this is better, maybe you can open up an issue about this?

src/Execution/Utils/Subscription.php
spawnia: To enable autocompletion via the annotation.

src/Execution/GraphQLRequest.php
spawnia: I tried that in the first place, but found it complicates usage quite a bit. The way this is used now enables the GraphQLController to iterate through with a very simple do-while loop.

Using standardized interfaces seems good generally, maybe you can give it a shot and see if it improves things?

src/LighthouseServiceProvider.php
spawnia: https://github.com/nuwave/lighthouse/pull/647

src/LighthouseServiceProvider.php
spawnia: I figured in this case the implementation is so trivial. Any downsides to the approach i chose? Open to changing it to a class

src/LighthouseServiceProvider.php
spawnia: In favour of?

olivernybroe: Take in the config repository in the constructor.

src/LighthouseServiceProvider.php
spawnia: I think it is needed for Lumen compat: https://github.com/nuwave/lighthouse/blob/master/.github/CONTRIBUTING.md#laravel-feature-usage

src/Schema/AST/ASTHelper.php
spawnia: We can do that in a seperate PR throughout the codebase.

src/LighthouseServiceProvider.php
olivernybroe: If I want to check which implementation of `GraphQLResponse` it is, then this makes it a lot harder :)

src/Schema/AST/ASTHelper.php
spawnia: https://github.com/nuwave/lighthouse/issues/652

src/Execution/Utils/Subscription.php
mfn: Interesting fact: tools like https://github.com/barryvdh/laravel-ide-helper generate a meta file which tells (at least PhpStorm) that `app()` (and its over variants) are factory function and that they return what is passed in.

That said: I've no idea if/how this would work within a library project and I'm certainly not challenging this change!

src/LighthouseServiceProvider.php
olivernybroe: Oh did not see that in `contributing.md`.
Could you remove that from `contributing.md`? 
The issue is because Lumen and Laravel both have they own response factory, but to solve this you should just do `new Response` directly. As both fo the factories just creates the `Response` class. 
As our use case of the factory is so simple, there really is no need for the factory here.

src/Schema/AST/ASTHelper.php
lucasmichot: Good idea @spawnia

src/Execution/Utils/Subscription.php
olivernybroe: @mfn Yes if you just want autocompletion for phpstorm you can add in a `.phpstorm.meta.php`, and in that do a simple
```
namespace PHPSTORM_META {
   override(\app(0), map([
      ...
   ])
}
```
and that also works in packages.

However we are actually using a helper method here, which could be nice if we could remove that all together instead.
Could we just supply the classes it needs in the constructor or supply the container in the constructor of the class?

src/Defer/DeferrableDirective.php
olivernybroe: https://github.com/nuwave/lighthouse/issues/653

src/Defer/DeferrableDirective.php
spawnia: I think requiring this constant to be set is more magic then just enforcing it via the interface.

src/Execution/Utils/Subscription.php
spawnia: https://github.com/nuwave/lighthouse/issues/654

src/LighthouseServiceProvider.php
spawnia: Done

src/LighthouseServiceProvider.php
spawnia: That would break use cases where someone has subbed out the `ResponseFactory` with their own implementation.

src/Schema/Directives/Fields/DeprecatedDirective.php
spawnia: https://github.com/nuwave/lighthouse/issues/655

src/Subscriptions/Subscriber.php
spawnia: The interface does not allow that.

src/LighthouseServiceProvider.php
olivernybroe: hmm yeah, you are right. 

Either we need a third approach or we have to live with using the helper or live with users not able to use a custom response factory.

Should we make a discussion about this?

src/Execution/Utils/Subscription.php
mfn: > Could we just supply the classes it needs in the constructor or supply the container in the constructor of the class?

Well in this case this is a static method so I guess constructor injection isn't possible 😅

src/LighthouseServiceProvider.php
spawnia: I am fine with using a bunch of helpers where it is convenient and does not cause trouble. We should avoid defining our own or unnecessarily using them when they are just aliasing something we can call directly.

chrissm79:
> @chrissm79 is it enough for those use cases to simply use Middleware?

Getting access to `GraphQLRequest` gives you access to all the helpers it has available and `WillSendResponse` gets fired for each level of deferred fields which I'm unsure how it would be handled via middleware since it's streamed.

spawnia:
@chrissm79 Good point, agree with everything you say.

Can you add those events to the PR? I don't know the exact use case, so it might be easier to implement a solution that covers your needs and verify it is working for you.

src/Execution/Utils/Subscription.php
olivernybroe: Oh did not realise that. 

Yeah I guess it would be weird that the method took them in as arguments.

alberthaff:
Starting to prepare #628 for the changes in this PR...

However, I'm getting following error when i try to run `composer install`(vender & composer.lock has been removed):

```
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover

In ProviderRepository.php line 208:
                                                                  
  Class 'Nuwave\Lighthouse\Providers\LighthouseServiceProvider' not found  
                                                                  

Script @php artisan package:discover handling the post-autoload-dump event returned with error code 1
```

When i change the provider namespace in lighthouse/composer.json to `Nuwave\\Lighthouse\\LighthouseServiceProvider`, i instead get following error:

```
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover

In LighthouseServiceProvider.php line 94:
                                                                               
  Too few arguments to function Nuwave\Lighthouse\LighthouseServiceProvider::  
  register(), 0 passed in /Users/alberthaff/git/utecca/utecca-api/vendor/lara  
  vel/framework/src/Illuminate/Foundation/Application.php on line 598 and exa  
  ctly 1 expected                                                              
                                                                               

Script @php artisan package:discover handling the post-autoload-dump event returned with error code 1
```

I have also tried to change the namespace for the provider inside `app.php` to the new one.
Is there something i'm doing wrong?

I'm running Laravel 5.7.

spawnia:
Good find @alberthaff, solved both just now.

alberthaff:
Thank you @spawnia. Everything seems to work now.

alberthaff:
@spawnia You are probably already aware of this, but there are some issues with Laravel 5.5 and 5.6 in #628 after the merge:
https://travis-ci.org/nuwave/lighthouse/builds/502977910

Is it possible to enable travis build for draft PR's? That would make it a lot easier to find issues before marking a PR "ready for review".

spawnia:
@alberthaff thanks for the heads up. Well this makes this feature pretty useless.

alberthaff:
@spawnia Yeah, I hope they add support for draft PR's in Travis. I got a similar surprise with the older Laravel versions, when I marked my PR as "ready for review".

chrissm79:
@spawnia I'm going to try and get through the review later today/tomorrow but just a quick observation.. It looks like the previous extensions now rely on a `ServiceProvider` to register listeners. Can we get that added to the docs?

spawnia:
> It looks like the previous extensions now rely on a ServiceProvider to register listeners. Can we get that added to the docs?

Will add that for sure. I think i will keep the structure similar to what it is now, at least for the v3 release and just add info on how to properly register the extensions.

spawnia:
@chrissm79 added and edited the docs, waiting on feedback towards the missing pieces.

I would actually go ahead and rename
- `BuildingAST` to `BuildSchemaString`
- `GatheringExtensions` to `BuildExtensionsResponse`
- `ManipulatingAST` to `ManipulateAST`
- `RegisteringDirectiveBaseNamespaces` to `RegisterDirectiveNamespaces`

I would prefer `ManipulateResponse` in favour of `WillSendResponse`.

And yeah, we could include a reference to the GraphQLRequest object in StartRequest, makes total sense!

spawnia:
Does someone have an idea what the root cause of the test failures could be?

alberthaff:
@spawnia I noticed that many of the failed assertions for Laravel 5.5 receives, what looks like a batched response.
However, i haven't identified what the cause is yet. Could it be somewhere in `GraphQLRequest`or it's dependants, that a function / helper which is not supported by Laravel 5.5 is used?
```
53) Tests\Integration\GraphQLTest::itResolvesQueryViaPostRequest
Unable to find JSON: 

// This is a normal response (to be expected)
[{
    "data": {
        "user": {
            "email": "christelle.durgan@example.net",
            "tasks": [
                {
                    "name": "Vitae nihil quia et."
                },
                {
                    "name": "Natus omnis ipsam laudantium."
                },
                {
                    "name": "Et et ut a praesentium at distinctio eaque."
                },
                {
                    "name": "Voluptatem debitis illo a autem."
                },
                {
                    "name": "Debitis maxime nulla perferendis ut."
                }
            ]
        }
    }
}]

// Server returns a batched response
within response JSON:
[[
    {
        "data": {
            "user": {
                "email": "christelle.durgan@example.net",
                "tasks": [
                    {
                        "name": "Vitae nihil quia et."
                    },
                    {
                        "name": "Natus omnis ipsam laudantium."
                    },
                    {
                        "name": "Et et ut a praesentium at distinctio eaque."
                    },
                    {
                        "name": "Voluptatem debitis illo a autem."
                    },
                    {
                        "name": "Debitis maxime nulla perferendis ut."
                    }
                ]
            }
        }
    }
]].
```

spawnia:
> Maybe this should be named httpResponse?

I want to differentiate it from Defer, which is also a HTTP response - but a streamed one

spawnia:
@chrissm79 this should be good to go.

Maybe squash and merge this in, the commit history is kind of messy.

chrissm79:
@spawnia Awesome 👍 I'll be digging into this soon to get this one merged in!!

chrissm79:
Hey @spawnia, made some small tweaks just to get some needed items in extensions from own project, but overall it's makes a lot of sense to go this direction. I am getting an error though and I'm trying to hunt it down... 

```
Illuminate\Contracts\Container\BindingResolutionException: Target [Nuwave\Lighthouse\Subscriptions\Contracts\ContextSerializer] is not instantiable while building [Nuwave\Lighthouse\Subscriptions\SubscriptionRegistry].
```

Which seems to originate [here](https://github.com/nuwave/lighthouse/blob/b9050146f4e67b751962582b70d3cded9512618d/src/Schema/Values/FieldValue.php#L251-L252). Going to step away for a bit but I'll dig into it in the morning so we can get this one merged in. Great work!!

chrissm79:
> I don't think this is necessary, as the HTTP request is readily available from the container anyway.

I need the original instance because I have a listener that I'm going to put into a queue so it's going to be outside the regular request... that being said, since I can swap out the implementation I can simple extend the `LighthouseRequest` and just expose the method myself 👍

spawnia:
@chrissm79 that would be great. Seems weird to have this method on the interface and not even use it in Lighthouse core. If we remove it that will make it easier on implementors :)

chrissm79:
Nice work on this one @spawnia 👍

jdousdebes:
> Hey @spawnia, made some small tweaks just to get some needed items in extensions from own project, but overall it's makes a lot of sense to go this direction. I am getting an error though and I'm trying to hunt it down...
> 
> ```
> Illuminate\Contracts\Container\BindingResolutionException: Target [Nuwave\Lighthouse\Subscriptions\Contracts\ContextSerializer] is not instantiable while building [Nuwave\Lighthouse\Subscriptions\SubscriptionRegistry].
> ```
> Which seems to originate [here](https://github.com/nuwave/lighthouse/blob/b9050146f4e67b751962582b70d3cded9512618d/src/Schema/Values/FieldValue.php#L251-L252). Going to step away for a bit but I'll dig into it in the morning so we can get this one merged in. Great work!!

Hello!.
I'm getting the same error, it fails in that line of code on FieldValue.php. Could you please explain how did you solve this problem?.

If I comment the subscription definition on the schema, executing 
`php artisan lighthouse:validate-schema` throws a valid response.

Thanks!

chrissm79:
Hey @jdousdebes, I just forgot to include the service provider for the extension. Now, instead of having the extensions defined in the `lighthouse.php`, you just add their service provider in your `app.php` file. Here's the docs:

https://lighthouse-php.com/master/extensions/subscriptions.html#setup

jdousdebes:
I had defined the subscription extension in config/app.php, but that didn't work, I cleared the cache, reinstalled packages, and every possible trick, but that didn't help solve the issue.

But strangely, a brand new clone of my repo did work, it's all good now; so if anyone else is facing this issue and everything seems to fail, do a git clone of your repo and it might solve your issues.

Thanks for the response, and also thanks to Spawnia that helped me on slack!.

Issue #499, opened 2018-12-10
Remove global function helpers entirely


mfn:
**Is your feature request related to a problem? Please describe.**

Interoperability with any kind of other software.

**Describe the solution you'd like**

Spin-off from https://github.com/nuwave/lighthouse/pull/488#issuecomment-445569623 : 

Forced global helper functions are always a PITA when integrating with other software.

Unfortunately for Laravel itself, they started this and for some poor souls it provides a hard barrier when having to integrate (especially) legacy software with modern ones like Laravel. Or lighthouse.

The biggest problem is that the global functions in `helpers.php` are forced upon everyone installing lighthouse due to the current `composer.json`:
```json
    "autoload": {
        "psr-4": {
            "Nuwave\\Lighthouse\\": "src/"
        },
        "files": [
            "src/Support/helpers.php"
        ]
}
```
This means everyone installing lighthouse gets the `helpers.php` loaded, whether they want or not.

Worse: some of the global helpers are used within the project itself (so they're not purely for convenience of users of the lighthouse library).

This means you a) force load global stuff to users which b) may or may not clash with other projects you're not ware of 💥

This is never good. Laravel does the same and it bites _some_ users (spoiler: I'm one of them).

And the funny side of this: it's simply not necessary.

Either provide namespaces functions (which can be a autoload PITA I've heard) or just slap a _namespaced_ class somewhere with static methods.

That it's possible to override their behavior (all the `function_exists` checks) is even worse:
- it's actually "not so" easy to override specifically on your own
- but if any other project also defining a same-named function is by accident loaded before, you've your clash already

Just a perspective of someone who had the pleasure integrating Laravel itself (not lighthouse) with other (legacy) software and just ended up in a world of pain, a situation which doesn't necessarily have to exist with a bit tactfulness.

**Describe alternatives you've considered**

n/a

**Additional context**

This issue serves as a discussion to come to a conclusion if and when to get rid of them, what the alternative if any will be, etc.

spawnia:
I am all for removing those helpers, let's go for it!

mfn:
Thanks `\o/`

PR #647, opened 2019-03-04
[WIP] Get rid of helpers.

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**
Get rid of helpers.
Fix : https://github.com/nuwave/lighthouse/issues/499

**Changes**

**Breaking changes**

They will be some if developers still use the helpers functions.

--

@spawnia can you have a deeper look at `config_path` helper removal, I am not sure that I undersand the original intend of it (Lumen?) - It might have been added a long time ago too


src/Support/Helper.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support;
+
+use Closure;
+use Nuwave\Lighthouse\GraphQL;
+use Nuwave\Lighthouse\Exceptions\DefinitionException;
+
+class Helper
```
lucasmichot: I am happy to rename this class. `Util` maybe?
Or also happy to create 2 invokable classes (might be the way to go...)
I don't have any strong opinion on this @spawnia

spawnia: Not a huge fan of invokable classes. I would name it `Helpers`

src/Console/PrintSchemaCommand.php
```diff
@@ -39,7 +40,7 @@ public function handle(Repository $cache, Filesystem $storage): void
         $cache->forget(config('lighthouse.cache.key'));
 
         $schema = SchemaPrinter::doPrint(
-            graphql()->prepSchema()
+            Helper::graphql()->prepSchema()
```
spawnia: Let's use proper DI instead of this one, really don't like the way it masks how it actually get's a singleton class under the hood.

src/Providers/LighthouseServiceProvider.php
```diff
@@ -45,7 +45,7 @@ public function boot(): void
         $this->mergeConfigFrom(__DIR__.'/../../config/config.php', 'lighthouse');
 
         $this->publishes([
-            __DIR__.'/../../config/config.php' => config_path('lighthouse.php'),
+            __DIR__.'/../../config/config.php' => base_path('config/lighthouse.php'),
```
olivernybroe: This is not the same. 
I can move my `config` folder another place if I want to.
You should do something like this https://github.com/nuwave/lighthouse/pull/645/files#diff-59794b479ab0add30d46c8336d474322R54 instead

src/Console/PrintSchemaCommand.php
olivernybroe: Just take `GraphQL` as a parameter in the args

src/Console/PrintSchemaCommand.php
```diff
@@ -39,7 +40,7 @@ public function handle(Repository $cache, Filesystem $storage): void
         $cache->forget(config('lighthouse.cache.key'));
```
olivernybroe: Take Config repository as a arg instead

src/Schema/Directives/Fields/FieldDirective.php
```diff
@@ -35,7 +36,7 @@ public function resolveField(FieldValue $fieldValue): FieldValue
             $fieldValue->defaultNamespacesForParent()
         );
 
-        $resolver = construct_resolver($namespacedClassName, $methodName);
+        $resolver = Helper::constructResolver($namespacedClassName, $methodName);
```
olivernybroe: Move the `constructResolver` to `BaseDirective` or a trait instead.

src/Schema/Directives/Fields/ComplexityDirective.php
```diff
@@ -37,7 +38,7 @@ public function handleField(FieldValue $fieldValue, Closure $next): FieldValue
                 $fieldValue->defaultNamespacesForParent()
             );
 
-            $resolver = construct_resolver($namespacedClassName, $methodName);
+            $resolver = Helper::constructResolver($namespacedClassName, $methodName);
```
olivernybroe: Move the `constructResolver` to `BaseDirective` or a trait instead.

src/Schema/Extensions/DeferExtension.php
```diff
@@ -312,7 +313,7 @@ protected function executeDeferred(): void
     {
         // TODO: Properly parse variables array
         // TODO: Get debug setting
-        $this->data = graphql()
+        $this->data = Helper::graphql()
```
olivernybroe: can it take GraphQL as a constructor parameter instead?

src/Subscriptions/SubscriptionBroadcaster.php
```diff
@@ -85,7 +86,7 @@ public function broadcast(GraphQLSubscription $subscription, string $fieldName,
         $this->iterator->process(
             $subscribers,
             function (Subscriber $subscriber) use ($root): void {
-                $data = graphql()->executeQuery(
+                $data = Helper::graphql()->executeQuery(
```
olivernybroe: Take as a constructor arg instead?

src/Support/Helper.php
olivernybroe: I think we should remove this class all together. 
It feels like a code smell

tests/TestCase.php
```diff
@@ -241,7 +242,7 @@ protected function buildSchema(string $schema): Schema
     {
         $this->schema = $schema;
 
-        return graphql()->prepSchema();
+        return Helper::graphql()->prepSchema();
```
olivernybroe: you have access to app directly here.

PR #650, opened 2019-03-06
Add docker config for Xdebug

spawnia:
This adds a Dockerfile as well as the appropriate docker-compose config to run PHP with Xdebug enabled.


spawnia:
Left the previous dev container as is, this is purely additive.

PR #661, opened 2019-03-12
Handle Scout pagination arguments

### LINKED ISSUE #592: How to use @search directive with scopes
what is the proper way to use scopes with the @search directive. 

if i try to use @search directive with scopes i get the following error:

"debugMessage": "Method Laravel\\Scout\\Builder::NAME_OF_THE_SCOPE does not exist."

it is trying to call the scope method on Scout

xahs:
- [x] Added or updated tests

**Related Issue/Intent**

Resolves #592

**Changes**

```php
// Nuwave\Lighthouse\Schema\Directives\Fields\PaginateDirective

protected function getPaginatedResults(array $resolveArgs, int $page, int $first): LengthAwarePaginator
{
    ...

    if ($query instanceof \Laravel\Scout\Builder) {
        return $query->paginate($first, 'page', $page);
    }

    return $query->paginate($first, ['*'], 'page', $page);
}
```

Please review my test case, it could be simpler I think.

### DIFF
--- a/src/Schema/Directives/Fields/PaginateDirective.php
+++ b/src/Schema/Directives/Fields/PaginateDirective.php
@@ -138,6 +138,10 @@ class PaginateDirective extends BaseDirective implements FieldResolver, FieldMan
             $resolveArgs[3]
         );
 
+        if ($query instanceof \Laravel\Scout\Builder) {
+            return $query->paginate($first, 'page', $page);
+        }
+
         return $query->paginate($first, ['*'], 'page', $page);
     }
 
--- a/src/Schema/Directives/Fields/PaginateDirective.php
+++ b/src/Schema/Directives/Fields/PaginateDirective.php
@@ -3,6 +3,7 @@
 namespace Nuwave\Lighthouse\Schema\Directives\Fields;
 
 use Illuminate\Support\Str;
+use Laravel\Scout\Builder as ScoutBuilder;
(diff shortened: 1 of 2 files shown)


src/Schema/Directives/Fields/PaginateDirective.php
```diff
@@ -138,6 +138,10 @@ protected function getPaginatedResults(array $resolveArgs, int $page, int $first
             $resolveArgs[3]
         );
 
+        if ($query instanceof \Laravel\Scout\Builder) {
```
lucasmichot: Maybe import `\Laravel\Scout\Builder`?

xahs: I think that in some scenarios, when we check for an ```instanceof``` common class names like ```Builder```, it seems more readable to put the full namespace. 

```php
if ($query instanceof \Laravel\Scout\Builder) {
    return $query->paginate($first, 'page', $page);
}

if ($query instanceof Builder) {
    return $query->paginate($first, 'page', $page);
}
```

Open for discussion...

spawnia: We always import classes in code. In this case aliasing to `ScoutBuilder` should make things more readable.

spawnia:
Great job on this @xahs

Thanks!

PR #663, opened 2019-03-13
Get rid of collect() helper.

### LINKED ISSUE #652: Usage of the collect helper
Should we replace `collect()` with `new Collection()` throughout the codebase?

lucasmichot:
- [ ] Added or updated tests

**Related Issue/Intent**

We tend to avoid using helpers, and prefer `Arr` and `Str` classes.
This PR gets rid of `collect()` helper and use directly `\Illuminate\Support\Collection`

Resolves #652 

**Changes**
Use of `->toBase()` method in 2 tests.

**Breaking changes**
None

### DIFF
--- a/src/Execution/DataLoader/BatchLoader.php
+++ b/src/Execution/DataLoader/BatchLoader.php
@@ -4,7 +4,7 @@ namespace Nuwave\Lighthouse\Execution\DataLoader;
 
 use Exception;
 use GraphQL\Deferred;
-use Nuwave\Lighthouse\GraphQL;
+use Illuminate\Support\Collection;
 use Nuwave\Lighthouse\Execution\GraphQLRequest;
 use Nuwave\Lighthouse\Support\Traits\HandlesCompositeKey;
 
@@ -84,7 +84,7 @@ abstract class BatchLoader
      */
     public static function instanceKey(array $path): string
     {
-        return collect($path)
+        return (new Collection($path))
             ->filter(function ($path) {
                 // Ignore numeric path entries, as those signify an array of fields
                 // Those are the very purpose for this batch loader, so they must not be included.
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -83,9 +83,9 @@ class MutationExecutor
             /** @var \Illuminate\Database\Eloquent\Relations\BelongsTo $relation */
             $relation = $model->{$relationName}();
 
-            collect($nestedOperations)->each(function ($values, string $operationKey) use ($relation, $model, $relationName): void {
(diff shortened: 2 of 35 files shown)


src/Execution/DataLoader/BatchLoader.php
```diff
@@ -4,7 +4,7 @@
 
 use Exception;
 use GraphQL\Deferred;
-use Nuwave\Lighthouse\GraphQL;
```
lucasmichot: Also remove this unused import.

tests/Integration/Defer/DeferDBTest.php
```diff
@@ -252,7 +252,7 @@ public function itCanDeferNestedListFields(): void
 
         $deferredCompanies = $chunks[2];
         $this->assertCount(6, $deferredCompanies);
-        collect($deferredCompanies)->each(function (array $item) use ($companies): void {
+        Collection::make($deferredCompanies)->toBase()->each(function (array $item) use ($companies): void {
```
lucasmichot: It's faster to used `->toBase()` and we avoid confusion by importing `\Illuminate\Support\Collection` and `\Illuminate\Database\Eloquent\Collection`

src/Execution/MutationExecutor.php
```diff
@@ -83,9 +83,9 @@ protected static function saveModelWithBelongsTo(Model $model, Collection $args,
             /** @var \Illuminate\Database\Eloquent\Relations\BelongsTo $relation */
             $relation = $model->{$relationName}();
 
-            collect($nestedOperations)->each(function ($values, string $operationKey) use ($relation, $model, $relationName): void {
+            Collection::make($nestedOperations)->each(function ($values, string $operationKey) use ($relation, $model, $relationName): void {
```
spawnia: Just noticed, can't we just use foreach here?

lucasmichot: We could, but I would prefer to do that in another PR

lucasmichot:
> I would prefer (new Collection()) instead of Collection::make(), another layer of indirection stripped away

Make sense 👍

tests/Integration/Defer/DeferDBTest.php
spawnia: Or we just use `foreach`

src/Execution/MutationExecutor.php
spawnia: Sure. I am finishing up nested updates tomorrow, will do it then.

lucasmichot:
See https://github.com/nuwave/lighthouse/pull/663/commits/30d89e37ccab6ab78185141ddc73016c7e6fae78 @spawnia

spawnia:
Can you add that phpunit cache file to gitignore?

lucasmichot:
Done

Issue #657, opened 2019-03-07
New layout for docs


olivernybroe:
I often see questions on Slack about users asking how to do X and X, which are things that are mentions in our docs, however almost everything the user want's to do is mentioned in `API References`.

I think we need a structure to easily browse the functionality of the framework, in a more newcomer friendly way.
The `API References` are nice, but it is a reference, it is not something you go into when you wanna learn how to paginate.

An example structure could be something like this

- Getting Started
- The Basics ( This will cover something that is also covered in other topics, but just shorter and then referencing for details)
  * Schemas
  * Directives
- Types
  * Getting Started
  * Scalars
  * Enums
- Queries
  * Getting Started
  * Pagination
  * Resolvers
  * Relay
- Mutations
  * Getting Started
  * Resolvers
  * Relay
- Security
  * Getting Started
  * Authentication
  * Authorization
  * Validation 
- Extensions
  * Getting Started
  * Monitoring
  * Deferred Fields
  * File Upload
  * Subscriptions
- Testing
  * Getting Started
  * GraphQL Playground
  * Integration Testing
  * Unit Testing
  * Mocking
- Package Development
  * Getting Started
  * Extending the schema
  * Extending the request
- API References

Didn't want to write it all out, but I think it shows what I mean.

spawnia:
I like the structure you are suggesting, should make things easier to find.

We should keep the API reference though, it is essential to have it for looking up the name of arguments or similar things.

I am open to switching to this new structure for a `3.1` release. For the initial v3, we should keep the structure the same, to make the transition easier.

olivernybroe:
> We should keep the API reference though, it is essential to have it for looking up the name of arguments or similar things.
Totally agree. Just forgot to add it to my suggestions :)

eriktisme:
Is it also possible to add an option to search the docs?

olivernybroe:
@eriktisme that is more of a new feature for the docs. I totally agree with we should have that feature, but this discussion is more about what should actually be in the docs.

You are more than welcome to make a pull request which adds search in the current docs.

eriktisme:
@olivernybroe Alright I'll try and create a pull request which adds search to the docs

spawnia:
@olivernybroe we can continue on this, v3 docs are released.

PR #668, opened 2019-03-14
New structure for docs

olivernybroe:
This pull request is for implementing the new architecture for the docs.
Before this is ready we need to write a lot of docs, and for that any help is appreciated.

Resolves: #657


spawnia:
I think this is too early, it should be easier to do once the v3 docs are tagged.

olivernybroe:
@spawnia You are absolutely right. My point was to start making it ready as a lot of text has to be written. So when we release v3, we can have the new docs layout a couple of weeks/months after.

olivernybroe:
We could also instead of doing it all at once just do some parts of it here and there?
So for example adding a section about package development and write that out, and then another pull request for another section? 
This way it will have a more natural flow, instead of everything being re-arranged at once.

@spawnia What would you prefer, all the changes at once or over time?

spawnia:
Let's get the `v3.0` docs finished up, tag them and then work towards a new structure for `v3.1`

robsontenorio:
@spawnia should i proceed with Search Docs feature before this one? We will switch to Vuepress alpha to get Search Docs feature.

olivernybroe:
@robsontenorio Yeah, just go ahead with that. This will take a long time before it is ready.

spawnia:
Hey @olivernybroe we should be able to get to this pretty soon! I hope to find the time to make the v3 docs ready soon, while keeping the structure.

Next step will be a restructuring :)

docs/master/directives/custom-directives.md
```diff
@@ -1,4 +1,8 @@
-# Custom Directives
+# Creating Your Own Directive
+
+[[toc]]
```
spawnia: What's that?

olivernybroe: table of content :)

docs/master/sidebar.js
```diff
@@ -1,42 +1,93 @@
...
+        title: 'Directives',
+        children: [
+            ['directives/getting-started', 'Getting Started'],
+            ['directives/custom-directives', 'Creating Your Own'],
         ]
     },
     {
         title: 'Extensions',
         children: [
+            ['extensions/getting-started', 'Getting Started'],
```
spawnia: I think that each of those probably deserves it's own top-level entry.

We might have information in Extensions on how one could write their own, coupled with the Event system.

docs/master/directives/custom-directives.md
spawnia: Isn't the sidebar sufficient for that?

spawnia:
Let's remove the `[[toc]]` markers, they are redundant and don't add anything.

image

spawnia:
@olivernybroe worked on it a bit more. I am already liking this new structure way more.

olivernybroe:
Finishing this up completely will require a lot of writing for us. I think we will have to live with some of the pages being written pretty sloppily just to get it out. Once it is out we can easily add better writings to the pages as we can focus on one page at a time.

spawnia:
I think we should not let this PR sit for too long, new features and docs changes are happening all the time.

Let's make sure we properly ported the existing docs over and merge it right then and there.
Going from there, we have a solid base and can iterate in small steps.

spawnia:
@olivernybroe i gave it a bunch more love :)

I am wondering what we can write about testing, any suggestions?

olivernybroe:
@spawnia I looks awesome. Amazing work on the write ups.
Not that I think it's a bad thing necessarily, but why did you remove the queries and mutations section?

Regarding the testing section, I think it's something we really need. I for example wrote an article about it and it has around 20 reads a day. So it's something people are looking for. 
My plan for it is basically like the pages I created in it. 
- Playground is just what it is and how to set it up and then a link to your github repo for more info.
- Integration Testing is the basic post request to your url
- Unit Testing is for not doing a post request to a url, but to supply a query string directly to lighthouse
- Mocking might be something we need a few code changes for. But basically do so I can make a post request to my graphql endpoint and mock the result and also mock what it should receive.

olivernybroe:
@spawnia I was going to write the part about mocking, but for doing it the most "Laravel" way, we will need to create a facade. Do you have anything against that we add a facade for GraphQL?

spawnia:
I think that while queries and mutations differ in what they are allowed to do on the server, they are very similar otherwise. I found that we could the individual parts rather nicely within `The Basics/Fields` or in `Eloquent`.

Great starting point for the testing section. I am inclined to wait for #802 to properly support test helpers. The assertions can be expanded a bit, too. I found that `assertJson` is pretty nice.

Playground does not really fall within testing in my opinion. I guess we could add a section about dev-environment niceties?

Integration vs. Unit: How would a direct call be better then utilizing the `postJson` helper?

Mocking: Not sure what we could do there, any use-cases?

olivernybroe:
Playground: My thought about having playground there is because you can use it to test your api, just not automated tests.

Integration vs. Unit: I have removed unit testing section as I only saw problems doing it that way. However use cases could be to test if your endpoint works, without actually testing your middlewares and the whole controller flow at the same time.

Mocking: The point here is for example if I write a browser test where I want to test that the save button sends a query to graphql, but I do not care if the query actually works.

Regading helper functions for testing, let's leave that discussion to #802.

olivernybroe:
I have an idea for easier browsing the directives. What if we made a list like Laravel has for validation rules? https://laravel.com/docs/5.8/validation#available-validation-rules
Often When I go to the directive I scout through the names to find the one I am looking for, but that's kinda hard when you have one long list. 
What do you think about adding that in @spawnia?

spawnia:
> What do you think about adding that in @spawnia?

Honestly, i never liked that navigation structure. You have to keep jumping back and forth if you are looking around for a rule that fits what you need.

The sidebar is quite nice IMO. If the list were larger then a hundred directives or so, it would start to become too long, but as of now, i think it is quite convenient.

spawnia:
Hey @alberthaff and @olivernybroe i appreciate the effort on the Frontend docs section, but i feel like it is out of scope for this project.

If a client is specifically geared to integrate with Lighthouse, such as Buoy seems to be, i am fine with adding it to the resource section. Let's keep the docs limited to building a server with Lighthouse.

spawnia:
@chrissm79 we can merge this in after #802

chrissm79:
Incredible effort on this one @olivernybroe @spawnia, thank you!! 👍

PR #673, opened 2019-03-16
Remove global helper functions

### LINKED ISSUE #499: Remove global function helpers entirely

spawnia:
**Related Issue/Intent**

Supersedes #647, thanks @lucasmichot 
Fixes #499 

**Changes**

Remove helper functions and replace their usage.

**Breaking changes**

Yup, anyone depending on those functions. We did not advertise their usage in the docs, so it should be limited.


chrissm79:
Nice work on this one @spawnia, thanks!!

PR #674, opened 2019-03-16
Fix tracing precision

### LINKED ISSUE #672: [3.0] Tracing does not follow the spec correctly
Quoting: https://github.com/apollographql/apollo-tracing#response-format

- The duration of a request is in nanoseconds, relative to the request start, as an integer.
- The startOffset of parsing, validation, or a resolver call is in nanoseconds, relative to the request start, as an integer.
- The duration of parsing, validation, or a resolver call is in nanoseconds, relative to the resolver call start, as an integer.

The spec states durations and offsets are in **nanoseconds**. Now I know nanoseconds are not supported by PHP, so this might be a little tricky. But microseconds are supported fine and happen to be a 1000 smaller than a nanosecond (yay maths ;)).

So something like this should solve it (with the greatest precision we can achieve in PHP):

```php
    public function record(ResolveInfo $resolveInfo, Carbon $start, Carbon $end): void
    {
        $this->resolverTraces [] = [
            // ...
            'startOffset' => $start->diffInMicroseconds($this->requestStart) * 1000,
            'duration' => $start->diffInMicroseconds($end) * 1000,
        ];
    }
```

However... Carbon only supports the microsecond helpers since 2.6 and Laravel can use Carbon `^1.26.3 || ^2.0` so I think there are 2 ways forward.

- Implement our own `diffInMicroseconds` calls so we don't depend on Carbon
- Require Carbon `^2.6`

WDYT?

---

I should mention what is wrong with doing it in seconds I suppose:

<details>
<summary>Example trace</summary>

```json
{
  "data": {
    "viewer": {
      "name": "Alex Bouma",
      "portfolio": {
        "username": "alex"
      }
    }
  },
  "extensions": {
    "tracing": {
      "version": 1,
      "startTime": "2019-03-16T16:27:03.598Z",
      "endTime": "2019-03-16T16:27:03.660Z",
      "duration": 0,
      "execution": {
        "resolvers": [
          {
            "path": [
              "viewer"
            ],
            "parentType": "Query",
            "returnType": "User",
            "fieldName": "viewer",
            "startOffset": 0,
            "duration": 0
          },
          {
            "path": [
              "viewer",
              "name"
            ],
            "parentType": "User",
            "returnType": "String!",
            "fieldName": "name",
            "startOffset": 0,
            "duration": 0
          },
          {
            "path": [
              "viewer",
              "portfolio"
            ],
            "parentType": "User",
            "returnType": "Portfolio!",
            "fieldName": "portfolio",
            "startOffset": 0,
            "duration": 0
          },
          {
            "path": [
              "viewer",
              "portfolio",
              "username"
            ],
            "parentType": "Portfolio",
            "returnType": "String!",
            "fieldName": "username",
            "startOffset": 0,
            "duration": 0
          }
        ]
      }
    }
  }
}
```

</details>

All durations and offsets are 0, instant API! Which is bad for metrics (although good for management).

stayallive:
Fixes #672.

### DIFF
--- a/src/Tracing/Tracing.php
+++ b/src/Tracing/Tracing.php
@@ -15,11 +15,20 @@ use Nuwave\Lighthouse\Events\BuildExtensionsResponse;
 class Tracing
 {
     /**
-     * The point in time where the request was initially started.
+     * The timestamp the request was initially started.
      *
      * @var \Carbon\Carbon
      */
     protected $requestStart;
+
+    /**
+     * The precise point in time where the request was initially started.
+     *
+     * This is either in seconds with microsecond precision (float) or nanoseconds (int).
+     *
+     * @var float|int
+     */
+    protected $requestStartPrecise;
 
     /**
      * Trace entries for a single query execution.
@@ -34,6 +43,7 @@ class Tracing
      * Set the tracing directive on all fields of the query to enable tracing them.
      *
      * @param  \Nuwave\Lighthouse\Events\ManipulateAST  $ManipulateAST
+     *
      * @return void
      */
     public function handleManipulateAST(ManipulateAST $ManipulateAST): void
@@ -52,7 +62,8 @@ class Tracing
      */
     public function handleStartRequest(StartRequest $startRequest): void
     {
-        $this->requestStart = $startRequest->moment;
+        $this->requestStart = Carbon::now();
+        $this->requestStartPrecise = $this->getTime();
     }
 
     /**
@@ -74,15 +85,16 @@ class Tracing
      */
     public function handleBuildExtensionsResponse(BuildExtensionsResponse $buildExtensionsResponse): ExtensionsResponse
     {
-        $end = Carbon::now();
+        $requestEnd = Carbon::now();
+        $requestEndPrecise = $this->getTime();
 
         return new ExtensionsResponse(
             'tracing',
             [
                 'version' => 1,
-                'startTime' => $this->requestStart->format("Y-m-d\TH:i:s.v\Z"),
-                'endTime' => $end->format("Y-m-d\TH:i:s.v\Z"),
-                'duration' => $end->diffInSeconds($this->requestStart),
+                'startTime' => $this->requestStart->format(Carbon::RFC3339_EXTENDED),
+                'endTime' => $requestEnd->format(Carbon::RFC3339_EXTENDED),
+                'duration' => $this->diffTimeInNanoseconds($this->requestStartPrecise, $requestEndPrecise),
                 'execution' => [
                     'resolvers' => $this->resolverTraces,
                 ],
@@ -94,19 +106,63 @@ class Tracing
      * Record resolver execution time.
      *
      * @param  \GraphQL\Type\Definition\ResolveInfo  $resolveInfo
-     * @param  \Carbon\Carbon  $start
-     * @param  \Carbon\Carbon  $end
+     * @param  float|int  $start
+     * @param  float|int  $end
      * @return void
      */
-    public function record(ResolveInfo $resolveInfo, Carbon $start, Carbon $end): void
+    public function record(ResolveInfo $resolveInfo, $start, $end): void
     {
         $this->resolverTraces [] = [
             'path' => $resolveInfo->path,
             'parentType' => $resolveInfo->parentType->name,
             'returnType' => $resolveInfo->returnType->__toString(),
             'fieldName' => $resolveInfo->fieldName,
-            'startOffset' => $start->diffInSeconds($this->requestStart),
-            'duration' => $start->diffInSeconds($end),
+            'startOffset' => $this->diffTimeInNanoseconds($this->requestStartPrecise, $start),
+            'duration' => $this->diffTimeInNanoseconds($start, $end),
         ];
+    }
+
+    /**
+     * Get the system's highest resolution of time possible.
+     *
+     * This is either in seconds with microsecond precision (float) or nanoseconds (int).
+     *
+     * @return float|int
+     */
+    public function getTime()
+    {
+        return $this->platformSupportsNanoseconds()
+            ? hrtime(true)
+            : microtime(true);
+    }
+
+    /**
+     * Diff the time results to each other and convert to nanoseconds if needed.
+     *
+     * @param  float|int  $start
+     * @param  float|int  $end
+     * @return int
+     */
+    private function diffTimeInNanoseconds($start, $end): int
+    {
+        if ($this->platformSupportsNanoseconds()) {
+            return $end - $start;
+        }
+
+        // Difference is in seconds (with microsecond precision)
+        // * 1000 to get to milliseconds
+        // * 1000 to get to microseconds
+        // * 1000 to get to nanoseconds
+        return (int) (($end - $start) * 1000 * 1000 * 1000);
+    }
+
+    /**
+     * Test if the current PHP version has the `hrtime` function available to get a nanosecond precision point in time.
+     *
+     * @return bool
+     */
+    private function platformSupportsNanoseconds(): bool
+    {
+        return function_exists('hrtime');
     }
 }
--- a/src/Tracing/TracingDirective.php
+++ b/src/Tracing/TracingDirective.php
@@ -3,7 +3,6 @@
 namespace Nuwave\Lighthouse\Tracing;
 
 use Closure;
-use Carbon\Carbon;
 use GraphQL\Deferred;
 use GraphQL\Type\Definition\ResolveInfo;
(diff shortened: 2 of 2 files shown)


src/Tracing/Tracing.php
```diff
@@ -17,10 +17,17 @@ class Tracing
     /**
      * The point in time where the request was initially started.
      *
-     * @var \Carbon\Carbon
+     * @var float
```
spawnia: This tells me nothing about the unit

src/Tracing/Tracing.php
```diff
@@ -17,10 +17,17 @@ class Tracing
...
      */
     protected $requestStart;
 
+    /**
+     * The point in time where the request was initially started.
+     *
+     * @var float|int
+     */
+    protected $requestStartPrecise;
```
spawnia: Definitely don't want to have that twice.

Does Carbon 1 not store microsecond time?

src/Tracing/Tracing.php
```diff
@@ -33,7 +40,8 @@ class Tracing
     /**
      * Set the tracing directive on all fields of the query to enable tracing them.
      *
-     * @param  \Nuwave\Lighthouse\Events\ManipulateAST  $ManipulateAST
+     * @param  \Nuwave\Lighthouse\Events\ManipulateAST $ManipulateAST
```
spawnia: It was correct before, there have to be two spaces.

src/Tracing/TracingDirective.php
```diff
@@ -44,8 +43,8 @@ public function name(): string
     /**
      * Resolve the field directive.
      *
-     * @param  \Nuwave\Lighthouse\Schema\Values\FieldValue  $value
-     * @param  \Closure  $next
+     * @param  \Nuwave\Lighthouse\Schema\Values\FieldValue $value
+     * @param  \Closure                                    $next
```
spawnia: Wrong formatting

stayallive:
@spawnia there is a bit of complexity involved because of the following.

I discovered PHP 7.3 does actually support getting a nanosecond precise point in time, so I changed the code to either use the `microtime(true)` value which is only milliseconds precise and if the `hrtime` function exists use the nanosecond time.

The nanosecond time returned by `hrtime`, unlike `microtime`, is not a unix timestamp with nanosecond precision but from an [arbitrary point in time](http://php.net/manual/en/function.hrtime.php) so we cannot use that to set the RFC 3339 string the format wants. I changed to "just" use Carbon for that purpose and do all the offset tracking and total time calculations with the (if possible) nanosecond precise timestamps.

src/Tracing/Tracing.php
stayallive: See comment explaining what is happening and why.

src/Tracing/Tracing.php
```diff
@@ -15,12 +15,19 @@
...
     /**
-     * The point in time where the request was initially started.
+     * The timestamp the request was initially started.
      *
      * @var \Carbon\Carbon
      */
     protected $requestStart;
 
+    /**
+     * The precise point in time where the request was initially started.
```
spawnia: So this is nanoseconds? Let's add that in the comment

spawnia:
@stayallive that makes total sense, nice work then.

chrissm79:
@stayallive Excellent work! Do you think it would be a good idea to get rid of Carbon entirely here? Haven't done a comparison but I have seen a enormous amount of Carbon instances being created on larger datasets. Thanks!!

stayallive:
@chrissm79 this will bring down the Carbon instances to 2 already. Since we cannot use the nanosecond precise time from PHP 7.3 to get back to a normal date I am still using a Carbon instance for the beginning and end of the request timestamps which are by the spec way less precise and Carbon there is fine for the easy formatting. All the recorded “frames” are using floats or ints (depending on the php version) now. 

Before I want to call this ready I would like to have Apollo Engine to ingest this and make sure both the hrtime and the microtime fallback produce sensible defaults. But no luck yet so hoping to find out if it’s something with the tracing response or something else (https://spectrum.chat/apollo/apollo-platform/trying-to-get-the-engine-proxy-to-consume-my-tracing~8faa9cf4-bdff-46da-99b1-2eaeb2505b70). I know of no other aggregater that stores and parses the format so I think this is my only option to see if the data is correct (apart from doing some napkin maths).

chrissm79:
> @chrissm79 this will bring down the Carbon instances to 2 already. 

Perfect 👍 

> Before I want to call this ready I would like to have Apollo Engine to ingest this and make sure both the hrtime and the microtime fallback produce sensible defaults. 

Not not positive but it looks like the precision may be slightly off in your response. When I test using GraphQL Playground I get the trace stack to appear with a response that looks like the following:

Screen Shot 2019-03-17 at 6 57 04 AM

```json
{
  "data": {
    "createAccount": {
      "user": {
        "id": 12
      }
    }
  },
  "extensions": {
    "lighthouse_subscriptions": {
      "version": 1,
      "channels": []
    },
    "tracing": {
      "version": 1,
      "startTime": "2019-03-17T13:54:59.404Z",
      "endTime": "2019-03-17T13:54:59.890Z",
      "duration": 486094000,
      "execution": {
        "resolvers": [{
          "path": ["createAccount"],
          "parentType": "Mutation",
          "returnType": "LoginPayload",
          "fieldName": "createAccount",
          "startOffset": 110699000,
          "duration": 375105000
        }, {
          "path": ["createAccount", "user"],
          "parentType": "LoginPayload",
          "returnType": "User",
          "fieldName": "user",
          "startOffset": 485886000,
          "duration": 6000
        }, {
          "path": ["createAccount", "user", "id"],
          "parentType": "User",
          "returnType": "Int!",
          "fieldName": "id",
          "startOffset": 485922000,
          "duration": 21000
        }]
      }
    },
    "persistedQuery": {
      "version": 1
    }
  }
}
```

My `duration` and `startOffset`s are represented in nanoseconds (10 ^-9 s). Looking at your response it seems that they are 10 ^-8 s (or 1 digit short of a nanosecond). Maybe that's all it is?

stayallive:
Most likely. Didn’t realise that playground supports it. I’ll test and polish with that in mind :) 👌

stayallive:
I was doing something funky with rounding and converting `microtime(true)` to actual microseconds, all fixed now.

image

The trace looks a bit off because the missing parts we don't trace currently:

```json
      "parsing": {
        "startOffset": <>,
        "duration": <>,
      },
      "validation": {
        "startOffset": <>,
        "duration": <>,
      },
```

Causing the resolvers to show just behind the request I believe. If there are some ideas / hooks to also instrument those steps that would be cool (we can work towards that in a next iteration too ofcourse).

Also the dateformat caused some issues at my end because of timezone formatting so I used the RFC3339 extended constant in Carbon (actually DateTime but to prevent another import) to fix that too.

Seeing as it seems to work in GraphQL playground I'm calling it better than it was for now and will see if I can get Apollo Engine to work later.

stayallive:
@olivernybroe I did not even think to look that far because of the docs said `>=7.3` good point. 

Not a bad idea to implement a StopWatch that can either use microtime or hrtime StopWatch under the hood so we can separate that logic either...

I think it's not too to bad having it this way but having a seperated stopwatch could be nicer... although it also adds some bloat and more class instances being created. But not against it persé.

Will hack on that to see where we can get with it.

stayallive:
@olivernybroe after further investigation.

> http://php.net/manual/en/hrtime.installation.php
> http://php.net/manual/en/function.hrtime.php

Only the `hrtime` function is bundled with PHP 7.3, not the full extension as far as I can tell. Furthermore, the pecl library does not even include the `hrtime` function at all. 

So that would mean a 3x implementation of the StopWatch, once for PECL, `hrtime` and `microtime`. I think that is going a bit overboard IMHO especially considering having the hrtime pecl library is not really needed after 7.3 unless you need the StopWatch classes or some other niceties that that pecl library provides.

So not 100% sure it's worth the effort, since the way I see it 7.3 supersedes the pecl library for what we are doing here. But gladly be convince we should support all 3 options if we all think it's better.

spawnia:
@stayallive seems like we should not depend on `StopWatch` then. It's nice to support a precision upgrade when `hrtime` is present, let's make it easy to switch everything to just that once we require `php >= 7.3`.

olivernybroe:
@stayalive ah sorry, I remembered it as they added the whole library and not only `hrtime`. But nice research!

My thought is that the nicest approach would be to have our own stopwatch interface and have two implementations, one using `hrtime` and one using `microtime`. Then in either the class we will use stopwatch or in the service provider, define which implementation of it to use, depending on what is possible.
This might be too overkill, but the code would be clean.

However this could also just be a pr in the future to improve it.

chrissm79:
Excellent work on this one @stayallive, thank you!!

Issue #244, opened 2018-08-03
Policies that require a model instance


acro5piano:
Thank you for the great work!

I use v2.2-beta.1 and I found `@can` directive throws an error.

**Describe the bug**

When I use policy, lighthouse always throws error.

**Expected behavior**
can Delete a post if the user granted in policy

**Schema**

schema:

```gql
type Mutation {
    deletePost(id: ID!): User 
        @can(if: "update", model: "App\\Post") 
        @delete(model: "App\\Post")
}
```

query:

```gql
mutation deletePost {
  deletePost(id: 1) {
    id
  }
}
```

**Output/Logs**

throws this error:

```
Too few arguments to function App\\Policies\\PostPolicy::update(), 
1 passed in /var/www/html/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 481 
and exactly 2 expected
```

**Environment**

Lighthouse Version: v2.2-beta.1
Laravel Version: 5.5.41
PHP Version: 7.2.0

**Additional context**

I researched what is happening, and found the following code may cause the problem.

```php
// src/Schema/Directives/Fields/CanDirective.php

// ...
                $model = $this->getModelClass() ?: get_class($args[0]);
                $can = collect($policies)->reduce(function ($allowed, $policy) use ($model) {
                    if (! app('auth')->user()->can($policy, $model)) {
                        return false;
                    }
// ...
```

https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/Fields/CanDirective.php#L38-L43

Actually the above code pass model "class" string instead of model "instance" to the policy.

So if we change this behavior to use `find` and pass the instance to the policy just like other directive, it will be fixed. 
How do you think?

```
            $model = $this->getModelClass()::find($id);
```

https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/Fields/DeleteDirective.php#L49

spawnia:
Right now, the `@can` directive only supports authorization that are not specific to a single model instance https://laravel.com/docs/5.6/authorization#methods-without-models

I am working on a PR that refines the current behaviour a bit further and am adding test cases as well as documentation for it. Will update here once i got it done.

Not quite sure on how we can solve authorization on specific models. My main problem is that we do have multiple ways of resolving model instances, so just querying by ID might not work for a lot of use cases. On the other hand, we can not really utilize the Model instance after it has been resolved, because in the case of mutations, it may already have changed.

Tough one.

[...]

PR #684, opened 2019-03-26
Enable passing in model instance to @can directive

eriktisme:
- [x] Added or updated tests

**Related Issue/Intent**
Enable passing model instance to determine if a user is authorized to perform a given action on this model.

Resolves https://github.com/nuwave/lighthouse/issues/244

**Changes**
Check if an ID is passed in the query when `@can` directive is given.

**Breaking changes**
No breaking changes to my knowledge.

### DIFF
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -134,14 +134,10 @@ abstract class BaseDirective implements Directive
      */
     protected function findModelById($id = null): ?Model
     {
-        if (is_null($id)) {
-            return null;
-        }
-
         /** @var Model $modelClass */
         $modelClass = $this->getModelClass();
 
-        return $modelClass::where(app($modelClass)->getKeyName(), $id)->first();
+        return $modelClass::find($id);
     }
 
     /**
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -134,10 +134,7 @@ abstract class BaseDirective implements Directive
      */
     protected function findModelById($id = null): ?Model
     {
-        /** @var Model $modelClass */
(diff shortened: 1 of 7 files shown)


src/Schema/Directives/BaseDirective.php
```diff
@@ -125,6 +125,25 @@ protected function getModelClass(string $argumentName = 'model'): string
...
     }
 
+    /**
+     * Find the model instance by the given key.
+     *
+     * @param string|int|null $id
+     * @return \Illuminate\Database\Eloquent\Model|null
+     * @throws \Nuwave\Lighthouse\Exceptions\DirectiveException
+     */
+    protected function findModelById($id = null): ?Model
```
spawnia: Why not just use `$modelClass::find()` here?

eriktisme: You are right! I first passed in the whole array of arguments. Now I'm directly passing in the ID `find` is indeed better

spawnia:
Hm, what about `@can` in scenarios with multiple models? How does Laravel handle that?

eriktisme:
> Hm, what about `@can` in scenarios with multiple models? How does Laravel handle that?

I myself have never used one with multiple models in one check. I will check, one moment!

spawnia:
@eriktisme seems like it is not supported by Laravel anyway. I think this PR should be good to go.

@chrissm79 what do you think?

eriktisme:
@spawnia Thank you, you too!

chrissm79:
Great job @eriktisme 👍Thanks for this!

acro5piano:
Thank you @eriktisme !

BasMulders:
This is great. Thanks a lot! 🙌

plunkettscott:
👏 A much-needed addition. Thanks @eriktisme!

PR #688, opened 2019-03-26
WIP: Throw if missing relation return type

### LINKED ISSUE #687: Relationship mutation does not throw exception when relationship return type is not declared in Laravel
**Describe the bug**

If the relationship mutation feature introduced in lighthouse beta.3 is used, it is a documented requirement that the return type be declared on the function definition using PHP 7.0+ syntax. The reality is that many Laravel projects were started prior to PHP 7.0 release and thus have not implemented return type declarations as a coding practice on their projects.

As of beta.3, if a relationship mutation is attempted on a relationship without the return type defined, Lighthouse will basically skip the action on the relationship and attempt to proceed otherwise. This carries various risks - on a well-designed data model, this should hopefully throw an internal error, but in some cases it may go undetected and commit the other changes and attempt to the return the response to the API without reporting the issue.

**Expected behavior**

An exception should be thrown when a relationship mutation cannot be performed since the return type is not defined on the relationship definition in Laravel.

**Environment**

Lighthouse Version: beta.3
Laravel Version: 5.7.x
PHP Version: 7.3

**Additional context**

Like an idiot I spent a long time troubleshooting why relationship mutations were not working until I remembered the note in documentation that return types must be defined.

spawnia:
Resolves #687

### DIFF
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -12,6 +12,7 @@ use Illuminate\Database\Eloquent\Relations\MorphOne;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
+use Nuwave\Lighthouse\Exceptions\DefinitionException;
 use Illuminate\Database\Eloquent\Relations\MorphToMany;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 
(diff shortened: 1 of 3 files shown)


spawnia:
@SirLamer looks like this check is running into issues with other-non type hinted methods that might be on the model.

This might be methods on the base model class, such as `guard()` but also on the user defined model.

I don't think this check is feasible to do, as we can not magically tell if a user intends a method to be a relationship function or not, without a clear indication in the type hint.

Unless someone comes up with a way around that, i guess we will have to live with the issue outlined you mentioned in #687

SirLamer:
I'm sorry but I am just picking up use of this package so I am not especially familiar with how it works behind the scene, otherwise I might have been more inclined to submit a PR myself.

That's true, though, that a method on the model could just be returning a computed result or something.

spawnia:
I will close this PR, as i can not think of a simpler solution. Nested Mutations are an advanced feature and as such requires some extra setup, namely adding return type hints to the relationship methods.

PR #689, opened 2019-03-28
Convert schema-defined enum default values to internal values

spawnia:
- [x] Added or updated tests

This resolves a bug where a schema definition like this:

```graphql
        type Query {
            foo(
                bar: Baz = FOOBAR
            ): Int
        }
        
        enum Baz {
            FOOBAR @enum(value: "internal")
        }
```

would pass `string("FOOBAR")` to the field resolver instead of the internal value.


chrissm79:
Awesome, thanks @spawnia!

PR #692, opened 2019-03-28
Allow pivot columns on a belongsToMany @update nested mutation

### LINKED ISSUE #691: Dealing with attributes on the pivot table of a BelongsToMany relation
**Situation**

I'm trying to implement @update on a belongsToMany relationship. The pivot table also has columns I want to update, like this:

```php
class Patient extends Model
{
    // many to many RELATIONSHIPS
    public function medications(): BelongsToMany
    {
        return $this->belongsToMany(Medication::class, 'patients__medications', 'patient_id', 'medication_id')->withPivot('dose_mg', 'is_rescue_medication'); //->as("details");
    }
}
```

my example schema: 

```graphql
#schema.graphql excerpts
# Mutations {}
updateMedication(input: MedicationInput): Medication! @update(flatten: true)

# Types
# -- Medication
input MedicationInput {
    id: ID
    name: String
    pivot: MedicationDetailsInput
}

input MedicationDetailsInput {
    patient_id: ID
    medication_id: ID
    dose_mg: Float
    is_rescue_medication: Boolean
}
input MedicationRelation {
  update: [MedicationInput]
  connect: [MedicationDetailsInput]
  disconnect: [ID]
  create: [MedicationInput]
  delete: [MedicationInput]
  sync: [MedicationDetailsInput]
}
```

**Problem**

I get an SQL error that laravel can't find the columns (of course, they don't exist because they are on the pivot table) on the related table to update.

I debugged it down to the `executeUpdate` function in `MutationExecutor.php` , where the pivot property is still inside `$remaining` and gets treated as a normal property on the related Model instead of the special case of being pivot columns.

**Expected**

pivot columns should be updated,

**Workarounds**

there was a workaround suggested:

> You can treat your BelongsTo relation simply as two BelongsTo relationships to work around this.

by @spawnia in this issue thread #549

JoeSchr:
- [X] Added or updated tests

**Related Issue/Intent**

Resolves #691 

**Changes**

pivot is recognized as a special case while @updating nested mutation is processed.

**RFC**
I needed to access the protected property `accessor` on `$parentRelation` to make this work, I didn't see any other way? [See laravel API docs](https://laravel.com/api/5.8/Illuminate/Database/Eloquent/Relations/BelongsToMany.html)

so I had to make a kind of hacky helper function. I don't really know where it could be better put? Or if there would be a less hacky way of doing it?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,6 +53,7 @@
     },
     "scripts": {
         "test": "phpunit --colors=always",
+        "test:failing": "phpunit --colors=always --group=failing",
         "test:unit": "phpunit --colors=always --testsuite Unit",
         "test:integration": "phpunit --colors=always --testsuite Integration"
     },
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -301,6 +301,15 @@ input CreatePostInput {
 }
 ```
 
+If the name of the Eloquent model does not match the return type of the field,
+or is located in a non-default namespace, set it with the `model` argument.
+
+```graphql
+type Mutation {
+    createPost(title: String!): Post @create(model: "Foo\\Bar\\MyPost")
+}
+```
+
 ## @delete
 
 Delete a model with a given id field. The field must be an `ID` type.
@@ -327,6 +336,15 @@ deleted models.
 ```graphql
 type Mutation {
     deletePosts(id: [ID!]!): [Post!]! @delete
+}
+```
+
+If the name of the Eloquent model does not match the return type of the field,
+or is located in a non-default namespace, set it with the `model` argument.
+
+```graphql
+type Mutation {
+    deletePost(id: ID!): Post @delete(model: "Bar\\Baz\\MyPost")
 }
 ```
 
@@ -1174,7 +1192,8 @@ type Mutation {
 }
 ```
 
-If the name of the Eloquent model does not match the return type of the field, set it with the `model` argument.
+If the name of the Eloquent model does not match the return type of the field,
+or is located in a non-default namespace, set it with the `model` argument.
 
 ```graphql
 type Mutation {
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -73,22 +73,22 @@ class MutationExecutor
     {
         $pivotAttributes = null;
 
+        /** @var Collection $remaining */
         [$belongsTo, $remaining] = self::partitionArgsByRelationType(
             new ReflectionClass($model),
             $args,
             BelongsTo::class
         );
 
-        // fetches pivot property from $remaining so
-        // $model->save() is not trying to work on nonexistant columns
-        if ($parentRelation instanceof BelongsToMany) { // only BelongsToMany have pivots
-            // check if there is a pivot
-            $accessor = $parentRelation ? self::accessProtected($parentRelation, 'accessor') : false;
-            if ($accessor && $remaining->has($accessor)) { // there is still a pivot to resolve in $remaining data
-                $pivotAttributes = $remaining[$accessor]; // save for later
-                unset($remaining[$accessor]); // remove pivot from $remaining
-            }
+        // If the parent relation is of the type BelongsToMany, we extract
+        // properties that are meant to be saved onto a pivot table from
+        // the remaining args, as they do not belong in the model itself.
+        if ($parentRelation instanceof BelongsToMany) {
+            $pivotAttributes = $remaining->pull(
+                $parentRelation->getPivotAccessor()
+            );
         }
+
         // Use all the remaining attributes and fill the model
         $model->fill(
             $remaining->all()
@@ -153,6 +153,7 @@ class MutationExecutor
             if ($pivotAttributes) {
                 $parentRelation->updateExistingPivot($model, $pivotAttributes);
             }
+
             $parentRelation->syncWithoutDetaching($model);
         }
 
@@ -558,26 +559,5 @@ class MutationExecutor
                 }
             });
         });
-    }
-
-    /**
-     * Access a private or protected property and return it's value.
-     *
-     * @param  object  $obj Object to be accessed
-     * @param  string  $prop Property of object
-     * @return Value   value of property
-     */
-    protected static function accessProtected($obj, $prop)
-    {
-        $reflection = new ReflectionClass($obj);
-        $property = $reflection->getProperty($prop);
-        $notAccessible = $property->isPrivate() || $property->isProtected();
-        $property->setAccessible(true);
-        $value = $property->getValue($obj);
-        if ($notAccessible) {
-            $property->setAccessible(false);
-        }
-
-        return $value;
     }
 }
--- a/src/Execution/MutationExecutor.php
+++ b/src/Execution/MutationExecutor.php
@@ -71,12 +71,24 @@ class MutationExecutor
      */
     protected static function saveModelWithPotentialParent(Model $model, Collection $args, ?Relation $parentRelation = null): Model
     {
+        $pivotAttributes = null;
+
         [$belongsTo, $remaining] = self::partitionArgsByRelationType(
             new ReflectionClass($model),
             $args,
             BelongsTo::class
         );
 
+        // fetches pivot property from $remaining so
+        // $model->save() is not trying to work on nonexistant columns
+        if ($parentRelation instanceof BelongsToMany) { // only BelongsToMany have pivots
+            // check if there is a pivot
+            $accessor = $parentRelation ? MutationExecutor::accessProtected($parentRelation, "accessor") : false;
+            if ($accessor && $remaining->has($accessor)) { // there is still a pivot to resolve in $remaining data
+                $pivotAttributes = $remaining[$accessor];// save for later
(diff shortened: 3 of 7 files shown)


JoeSchr:
I think there is nothing to add in the documentation, I just works as expected now?

spawnia:
> I think there is nothing to add in the documentation, I just works as expected now?

Expectations may vary. I wouldn't know how to use this - in fact i still don't. A comprehensive example in the docs is required.

JoeSchr:
copy, will get to it next week

spawnia:
@JoeSchr i spotted a pretty significant issue with this implementation.

```
    input UpdateUserInput {
        id: ID!
        name: String
        organizational: UpdateOrganizationalPivotInput
    }
```

You use this input object as a child of the role mutation input in `updateRole`. Things work fine then.

The problems arise when you try to use the `UpdateUserInput` as part of a mutation that is not descendent from Roles. It will not remove the `organizational` input from the attributes to fill as the parent relationship to roles is not known then.

On a similar note, the type definition for `User` has a similar issue:

```graphql
type User {
    id: ID!
    name: String
    organizational: OrganizationalPivot
}
```

The `organizational` attribute will only be available if we are querying the user as a descendant of a role.

I think the easiest way to solve these issues is to just use the underlying BelongsTo relationships of a BelongsToMany if you need to access the pivot attributes or mutate them.

As it stands, the design proposed in this PR is too narrow - it falls apart really quickly once you consider the need to reuse the types.

I guess you might work around the issue by adding a seperate type `UserWithRole` and `UserWithRoleInput` and vice versa that can be used when the user is used as a descendent of a specific role, but at this point i think the additional complexity is not worth it.

JoeSchr:
Hi
@spawnia congratulations to V3 and thanks for looking into my PR.

Also thanks for fixing my hack with `getPivotAccessor`, somehow I just missed that in the official Laravel API documentation. 😓

But now to get right into it: I'm very confused about the points/reasons you made to say, this design doesn't work.

I will try to explain:

> The problems arise when you try to use the UpdateUserInput as part of a mutation that is not descendent from Roles. It will not remove the organizational input from the attributes to fill as the parent relationship to roles is not known then.

> The organizational attribute will only be available if we are querying the user as a descendant of a role.
You use this input object as a child of the role mutation input in updateRole. Things work fine then.

I don't understand how this is relevant, because I can only imagine a use case where you want to work on a many2many relation and then the attribute will be present. If I'm not working on data where there is an existing descendent/parent relationship already present, I fail to see why I would even try to update `organizational` (or better `pivot`) as there is no logical way to even try to update this kind of data then and if tried should just fail (the same way as just entering wrong data)? 

Maybe I didn't choose the right examples in my tests, because I only wanted to adapt an existing test (because I'm not particular familliar with the testing suite) instead of shoehorning a "more real" example into it.  This would be very important to me, because I beg to differ about this being too narrow a PR? 

This is how I currently use it:
```graphql

#mutation   
updatePatient(input: PatientInput): Patient! @update(flatten: true)

#inputs 
input PatientInput {
    id: ID
    ...
    medications:MedicationRelation
}

input MedicationRelation {
  update: [MedicationInput]
  disconnect: [ID]
  create: [MedicationInput]
  delete: [MedicationInput]
...
}

input MedicationInput {
    id: ID
    name: String
    #details is the custom named pivot
    details: MedicationDetailsInput
}

input MedicationDetailsInput {
    patient_id: ID
    medication_id: ID
    dose_mg: Float
    is_rescue_medication: Boolean
}
```

I see not other way for me to connect 2 tables `patients` and `medications` then via a many2many table and there absolutely needs to be access to update the pivot data on the intermediate table, because it doesn't fit anywhere else.

If now coming back to 

> The problems arise when you try to use the UpdateUserInput as part of a mutation that is not descendent from Roles. It will not remove the organizational input from the attributes to fill as the parent relationship to roles is not known then.

and I would try to update `details` by just using `MedicationInput`
e.g. 
```graphql
mutation {
   createMedication(input: MedicationInput): Medication! @create(flatten: true)
   updateMedication(input: MedicationInput): Medication! @update(flatten: true)
}
```
with
```php
updateMedication(input: { id:1 name:"Renamed medication" details: { dose_mg: 11 } })
```

That would just be not smart to do and doesn't make any sense to even try. So I would just get a rightfully error and then rethink what I'm trying to do?

Being able to work with pivot data on a many 2 many relationship directly is pretty essential? At least I couldn't imagine how else you could achieve this while working with a relational database? 

Also they way to give access to pivot data via an pivot attribute (which can be renamed) was the decision of Laravel, so being just conform with their way of handling it, was the reason why I called this "working just as expected", if a User would try to do this by coming from reading the Laravel nested relationship documentation: [Retrieving Intermediate Table Columns](https://laravel.com/docs/5.8/eloquent-relationships#many-to-many). 
```php
$user = App\User::find(1);

foreach ($user->roles as $role) {
    echo $role->pivot->created_at;
}
```

Maybe I should have made that reference more clear? I just don't see a reason to "overengineer" how to update intermediate table data now, only because somebody could use it wrongly (but they then also wouldn't understand, how their relationships are designed anyway), if it's already used in this way by Laravel?


> I think the easiest way to solve these issues is to just use the underlying BelongsTo relationships of a BelongsToMany if you need to access the pivot attributes or mutate

Would you care to give me also an example how this could be done? Maybe this will unblock my understanding of how other people tend to use Lighthouse/Laravel nested relationships?

spawnia:
I think about this from the perspective of a client.

```graphql
type User {
    id: ID!
    name: String
    organizational: OrganizationalPivot
}
```

If i find this type in a schema, i would expect to be able to query the `organizational` data whenever i have a user type available in my query, for example when querying a user by ID. The same goes for an input type. The pivot attributes both within a type and an input object only work within the context of the associated relationship.

If you are writing backend code and know about the underlying structure, this makes sense
for the developer. I know that this concept exists in Laravel, i agree it can be useful, but i also think it's a leaky abstraction.
From a client perspective, this behaviour is quite opaque without an extra description or deeper knowledge about the underlying data model.

The abstraction offered by GraphQL should free clients to think about those issues. When i think about my data as a graph, i expect a User type to always be the same thing, no matter where in the graph it sits or where i got it.

When you think about what the `BelongsToMany` relationship actually is, you will find that it is in fact comprised of a `HasMany` and a `BelongsTo` relationship.

In terms of your concrete example, you should be able to update a medication like this:

```graphql
mutation {
	updateMedication(input: {
		id: 1
		name: "Lysergicream"
		# That would be a HasMany relationship from Mediation to the pivot model
		medicationUser: {
			update: [{
				dose_mg: 11
				# This is a belongsTo relationship from your pivot model to User
				user: {
                    connect: 1
                }
           }]
        }
    }) {
        id
    }
}

```

JoeSchr:
Thanks @spawnia for taking the time to write an example for the `HasMany/BelongsTo` approach. 

And thanks for clarifying, I now get the perspective you are coming from! 

Honestly for me this is now more of a philosophical point, about how opinionated an extension for a framework like Lighthouse should be. In my opinion, which should be no great surprise, I would think it best to offer more options and let the schema designer decide. If they don't care as much about having a "clean" public API,  because it's not their usecase, they could roll with a pivot attribute. 

If the designers are more the mindset you described above:
> The abstraction offered by GraphQL should free clients from thinking about those issues. When i think about my data as a graph, i expect a User type to always be the same thing, no matter where in the graph it sits or where i got it.

they could simple omit they pivot attribute from the schema with no harm done and use a `HasMany/BelongsTo` approach like you demonstrated it above. (Btw. I think this would be a great addition to the documentation in any way)
  
In my opinion this PR would just offer more options, how to build things, instead of making it needlessly more complicated/verbose, without taking something away for people of an other mindset. Also there seems to be some interest in it.

Regardless I will accept your decision, if you don't want to add it and close the PR then.

spawnia:
@JoeSchr i am glad that my reasoning is now clear, thanks for your understanding.

While this PR adds more options to design the schema, it doesn't really add something that is not possible right now. The added flexibility is rather "just" syntactic sugar.

I am afraid there are too many pitfalls in implementing this correctly. A proper explanation of how to design the schema safely and navigating around the problems i outlined, would have to be pretty lengthy.

When we started doing nested mutations in our personal project, i was pretty apprehensive about taking it into the Lighthouse core. From the start, it was predestined to be pretty opinionated. As of now, the complexity is already quite high, and i don't really want to add to it unless it is necessary.

There is a balance to strike from a package development perspective between flexibility, robustness and keeping the complexity low. All things considered, i am not willing to take this feature into the core.

However, i think there is a lot of value in this discussion. As you said, decomposing a `BelongsToMany` relation should be mentioned in the docs, it is not quite obvious at first.

I also opened up #725 to discuss how we can ease the extensibility of nested mutations. While it is possible for you to implement your own version of `@create` and `@update` with your own `MutationExecutor`, i am aware it requires you to fork a significant chunk of functionality - more than i would like it to be necessary. A better abstraction would make this easier.

JoeSchr:
> I also opened up #725 to discuss how we can ease the extensibility of nested mutations. While it is possible for you to implement your own version of @create and @update with your own MutationExecutor, i am aware it requires you to fork a significant chunk of functionality - more than i would like it to be necessary. A better abstraction would make this easier.

Yeah, that's what I will probably do for some time, as long as it doesn't get to be a major PIA to keep it in sync with upstream.

PR #699, opened 2019-04-01
Replace incidental usage of private with protected

spawnia:
Change all `private` accessors to `protected`.

There is value in having a smaller public API (which `protected` class members are a part of by nature of inheritence), but the current usage of `private` was rather incidental.

Let's at least have it consistent for now.


chrissm79:
Thanks @spawnia!

PR #700, opened 2019-04-01
Flatten directive namespace structure

spawnia:
Since directives may be used in multiple places, it is not really practical
to organize them in folders. The implied structure is misleading, since we
had directives such as `@globalId` which work on both fields and args but
were placed in the `Fields` namespace.

**Breaking changes**

Unless users were extending directive classes directly (which they probably shouldn't anyway), no.


spawnia:
@chrissm79 are you ok with this change?

chrissm79:
@spawnia yeah this needed to happen at some point. Thanks!

PR #702, opened 2019-04-02
Add middleware for subscription endpoints

### LINKED ISSUE #701: Subscription auth endpoints not using middleware
**Describe the bug**

Pusher subscription auth endpoints don't use the route middleware set in the lighthouse config.

**Expected behavior**

When accessing `/graphql/subscriptions/auth` the routes don't have any middleware associated.

**Environment**

Lighthouse Version: v3.0-beta.3

chrisneal:
Resolves #701

### DIFF
--- a/src/Subscriptions/SubscriptionRouter.php
+++ b/src/Subscriptions/SubscriptionRouter.php
@@ -20,7 +20,7 @@ class SubscriptionRouter
             $router->post($routeName. '/subscriptions/auth', [
                 'as' => 'lighthouse.subscriptions.auth',
                 'uses' => SubscriptionController::class.'@authorize',
-            ])->middleware($middleware);
+            ]);
 
             $router->post($routeName. '/subscriptions/webhook', [
                 'as' => 'lighthouse.subscriptions.auth',
--- a/src/Subscriptions/SubscriptionRouter.php
+++ b/src/Subscriptions/SubscriptionRouter.php
@@ -14,16 +14,19 @@ class SubscriptionRouter
      */
     public function pusher($router): void
     {
-        $middleware = config('lighthouse.subscriptions.route.middleware');
+        $router->group(config('lighthouse.subscriptions.route', []), function (): void {
+            $routeName = config('lighthouse.route_name', 'graphql');
 
-        $router->post('graphql/subscriptions/auth', [
-            'as' => 'lighthouse.subscriptions.auth',
-            'uses' => SubscriptionController::class.'@authorize',
-        ])->middleware($middleware);
+            $router->post($routerName. '/subscriptions/auth', [
+                'as' => 'lighthouse.subscriptions.auth',
+                'uses' => SubscriptionController::class.'@authorize',
+            ])->middleware($middleware);
 
-        $router->post('graphql/subscriptions/webhook', [
-            'as' => 'lighthouse.subscriptions.auth',
-            'uses' => SubscriptionController::class.'@webhook',
-        ])->middleware($middleware);
+            $router->post($routerName. '/subscriptions/webhook', [
+                'as' => 'lighthouse.subscriptions.auth',
+                'uses' => SubscriptionController::class.'@webhook',
+            ]);
+
+        });
     }
 }
--- a/src/Subscriptions/SubscriptionRouter.php
+++ b/src/Subscriptions/SubscriptionRouter.php
@@ -17,12 +17,12 @@ class SubscriptionRouter
         $router->group(config('lighthouse.subscriptions.route', []), function (): void {
             $routeName = config('lighthouse.route_name', 'graphql');
 
-            $router->post($routerName. '/subscriptions/auth', [
+            $router->post($routeName. '/subscriptions/auth', [
                 'as' => 'lighthouse.subscriptions.auth',
                 'uses' => SubscriptionController::class.'@authorize',
             ])->middleware($middleware);
 
-            $router->post($routerName. '/subscriptions/webhook', [
+            $router->post($routeName. '/subscriptions/webhook', [
                 'as' => 'lighthouse.subscriptions.auth',
                 'uses' => SubscriptionController::class.'@webhook',
             ]);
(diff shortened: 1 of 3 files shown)


chrisneal:
@spawnia Not added tests yet. What were you thinking?

src/Subscriptions/SubscriptionRouter.php
```diff
@@ -14,14 +14,16 @@ class SubscriptionRouter
      */
     public function pusher($router): void
     {
+        $middleware = config('lighthouse.subscriptions.route.middleware');
+
         $router->post('graphql/subscriptions/auth', [
             'as' => 'lighthouse.subscriptions.auth',
             'uses' => SubscriptionController::class.'@authorize',
-        ]);
+        ])->middleware($middleware);
```
spawnia: I think it is misleading to add an array of route options and only apply one key of it.

Can we just make this a route group, same as in the base config?

spawnia:
> @spawnia Not added tests yet. What were you thinking?

You can add a method in `\Tests\Integration\Subscriptions\SubscriptionTest`, something like:

```php
    /**
     * @test
     */
    public function itAddsRouteOptions(): void
    {
		// Add the config
		// Test that it is applied
	}
```

src/Subscriptions/SubscriptionRouter.php
chrisneal: @spawnia The middleware method accepts an array as the first parameter. So this will work as it does on the `/graphql/ route middleware.

chrisneal:
@spawnia I'm not sure how to test for it adding the route options...

src/Subscriptions/SubscriptionRouter.php
kraffslol: I belive what @spawnia is suggesting is having it share the same behaviour as the config of the regular routes. That way you will be able to apply more than just middleware to the subscription routes. See https://github.com/nuwave/lighthouse/blob/master/src/Support/Http/routes.php

chrisneal: Ah, I miss understood the request at first. I've updated it, and also changed it to use the `route_name` config for the base part of the path.

src/Subscriptions/SubscriptionRouter.php
```diff
@@ -14,16 +14,19 @@ class SubscriptionRouter
...
+            $routeName = config('lighthouse.route_name', 'graphql');
 
-        $router->post('graphql/subscriptions/auth', [
-            'as' => 'lighthouse.subscriptions.auth',
-            'uses' => SubscriptionController::class.'@authorize',
-        ])->middleware($middleware);
+            $router->post($routerName. '/subscriptions/auth', [
+                'as' => 'lighthouse.subscriptions.auth',
+                'uses' => SubscriptionController::class.'@authorize',
+            ])->middleware($middleware);
```
spawnia: Stuff like this is exactly the reason why we require a test for this. `$middleware` is not defined.

src/Subscriptions/SubscriptionRouter.php
```diff
@@ -14,14 +14,19 @@ class SubscriptionRouter
      */
     public function pusher($router): void
     {
-        $router->post('graphql/subscriptions/auth', [
-            'as' => 'lighthouse.subscriptions.auth',
-            'uses' => SubscriptionController::class.'@authorize',
-        ]);
+        $router->group(config('lighthouse.subscriptions.route', []), function (): void {
+            $routeName = config('lighthouse.route_name', 'graphql');
```
spawnia: Nice!

src/Subscriptions/SubscriptionRouter.php
chrisneal: @spawnia I've removed that.

spawnia:
@chrisneal Still need this tested.

It would be sufficient if you can just add a middleware and check if that is being called.

chrisneal:
@spawnia Not quite sure how to do that. Feel free to give me a hand.

src/Subscriptions/SubscriptionRouter.php
```diff
@@ -14,14 +14,19 @@ class SubscriptionRouter
      */
     public function pusher($router): void
     {
-        $router->post('graphql/subscriptions/auth', [
-            'as' => 'lighthouse.subscriptions.auth',
-            'uses' => SubscriptionController::class.'@authorize',
-        ]);
+        $router->group(config('lighthouse.subscriptions.route', []), function (): void {
```
olivernybroe: remember to add `$router` as a `use`, else it's not available in the closure

olivernybroe:
Looking further into this pull request it feels weird tbh.
We are adding in a config for routes which are specific for pusher, but it's not inside the array for pusher settings. 

Either move these settings inside of the pusher settings or you can easily obtain the same support with a custom `SubscriptionRouter` class, as it just passes in the `Router` object to the class.

chrisneal:
@olivernybroe I don't have time to make the changes to this PR anymore, I've used my forked version of this repo. So are you saying you want to close this PR? And we'd need to implement a custom `SubscriptionRouter`?

olivernybroe:
I am saying that that is one possibility which already works with the current code base, however I can see the reason for having an easier way of configuring it. I just wanted to point of that it is actually possible to do what you want to do already.

spawnia:
@olivernybroe i do not quite get how you would achieve a custom route config with the current version. Can you elaborate, maybe we can add info the docs if it is simple enough?

spawnia:
> I don't have time to make the changes to this PR anymore, I've used my forked version of this repo.

@chrisneal Not sure what to make of this. I would wager you won't save time in the long run by maintaining a fork of Lighthouse?

Sorry to see this PR abandoned.

olivernybroe:
@spawnia The way the current system works is by passing a router object to `SubscriptionRouter#pusher`. This method is then responsible for registering every endpoint needed. 
So I can easily just create my own class with a method that accepts a router object and then add the two endpoints with the middlewares i need. I can then register the class in the config `lighthouse.subscriptions.broadcasters.pusher.routes`.

spawnia:
Closing this as the solution by @olivernybroe is fine and the author seems to have abandoned the ship.

PR #708, opened 2019-04-04
Add documentation about the empty Query type in the root schema.

lucasmichot:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Developers often think that they have to provide dummy queries or mutations in the root schema in order to avoid this error: `Error: Query root type must be provided`

Actually adding an empty `type Query` (without curly braces) is enough.

The same applies for `type Mutation`, to avoid the following error: `"Schema is not configured for mutations."`

**Changes**

Add documentation about the empty Query field in the root schema.

**Breaking changes**

None

**TODO**

- [x] Add a test

### DIFF
--- a/docs/3.0/guides/schema-organisation.md
+++ b/docs/3.0/guides/schema-organisation.md
@@ -90,6 +90,17 @@ type Query
 #import post.graphql
 ```
 
+The same applies for mutations: if you want to use them, you can define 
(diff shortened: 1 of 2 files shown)


spawnia:
Thanks @lucasmichot good work!

PR #710, opened 2019-04-04
Fix some docblocks.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

Fix some docblocks.


spawnia:
I wonder how those keep popping up :D

It seems you do have some kind of methodology of spotting those, can you share it?
Maybe in the contributing guidelines?

lucasmichot:
I actually search for these two case-senstitive regular expressions in the comments:
```
@(var|return|throws).*\|[A-Z]
@(var|return|throws)\s*[A-Z]
```

PR #716, opened 2019-04-05
Remove useless key check.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

Remove useless key check. `Arr::get` method already checks this

**Changes**

**Breaking changes**

### DIFF
--- a/src/Subscriptions/Broadcasters/LogBroadcaster.php
+++ b/src/Subscriptions/Broadcasters/LogBroadcaster.php
@@ -85,9 +85,7 @@ class LogBroadcaster implements Broadcaster
      */
     public function broadcasts(?string $key = null): ?array
     {
-        return $key
-            ? Arr::get($this->broadcasts, $key)
-            : $this->broadcasts;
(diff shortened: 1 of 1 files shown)


src/Subscriptions/Broadcasters/LogBroadcaster.php
```diff
@@ -85,9 +85,7 @@ public function broadcast(Subscriber $subscriber, array $data)
      */
     public function broadcasts(?string $key = null): ?array
     {
-        return $key
-            ? Arr::get($this->broadcasts, $key)
-            : $this->broadcasts;
+        return Arr::get($this->broadcasts, $key);
```
spawnia: Are you sure that is the same?

lucasmichot: Yes: https://github.com/laravel/framework/blob/08e43fba6c4208acf04615c77037f186f88b8ba2/src/Illuminate/Support/Arr.php#L284-L286

PR #719, opened 2019-04-05
Add event dispatch + tests.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

The `fire` method has been deprecated for quite some times. It's better to use `dispatch`.
This PR also adds `@event` tests.

**Changes**

**Breaking changes**

### DIFF
--- a/tests/Integration/Schema/Directives/Fields/EventDirectiveTest.php
+++ b/tests/Integration/Schema/Directives/Fields/EventDirectiveTest.php
@@ -8,58 +8,27 @@ use Tests\Integration\Schema\Directives\Fields\Fixtures\CompanyWasCreatedEvent;
 
 class EventDirectiveTest extends DBTestCase
 {
-    /**
-     * @test
-     */
-    public function itDispatchesAnEventWithDispatch(): void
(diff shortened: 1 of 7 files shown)


tests/Integration/Schema/Directives/Fields/EventDirectiveTest.php
```diff
@@ -0,0 +1,136 @@
...
+
+        Event::assertDispatched(CompanyWasCreatedEvent::class, function ($event) {
+            return $event->company->id === 1 && $event->company->name === 'foo';
+        });
+    }
+
+    /**
+     * @test
+     */
+    public function itDispatchesAnEventWithClass(): void
```
spawnia: Since those are just aliases and do the same thing, how about we use a dataProvider for passing in the arg name?

spawnia:
@lucasmichot looks like we simultanously fixed the same thing :)

lucasmichot:
Yep saw that ! We were quite sync @spawnia 👍

chrissm79:
Great work @lucasmichot @spawnia, thank you!!

PR #731, opened 2019-04-12
Add sync case  to MutationExecutor::executeCreateBelongsToMany

megawubs:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

**Changes**

Add `sync` case  to `MutationExecutor::executeCreateBelongsToMany`

**Breaking changes**
None


megawubs:
I have no idea why it contains a lot of commits that are already merged. I've updated with upstream and all. But they keep showing up in the mr.

spawnia:
@megawubs don't worry about the commits, i squash commits anyways when merging a PR.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v3.2.1...master)
+ 
+- Sync existing models in belongsToMany relations using nested mutations https://github.com/nuwave/lighthouse/pull/707
```
```suggestion
### Added

- Sync existing models in belongsToMany relations using nested mutations https://github.com/nuwave/lighthouse/pull/707
```

megawubs:
@spawnia the docs already contained a `sync` key for the store mutation, but no example. I've added the example mutation.

docs/master/guides/relationships.md
```diff
@@ -557,6 +557,26 @@ Lighthouse will detect the relationship and attach/create it.
 }
 ```
 
+It is also possible to use sync instead of connect, to sync te relationship.
```
spawnia:
```suggestion
It is also possible to use the `sync` operation to ensure only the given IDs
will be contained withing the relation.
```

spawnia:
@megawubs just after merging this, a thought popped up:

What does `sync` do here that `connect` does not do? Since the parent is just being created, i guess they are equivalent?

megawubs:
During create, sync an connect are  interchangeable. I added it to provide the same kind of features Laravel does. Also, by allowing sync on create you can re-use input types for create and update.

PR #738, opened 2019-04-17
No need to declare 2 routes, 1 is enough.

lucasmichot:
- [ ] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**
No need to declare 2 routes, 1 is enough.

---

**Changes**
None

---

**Breaking changes**
None


spawnia:
Does this method have an advantage?

lucasmichot:
Well it does, this way we only have one route declared instead of 2 duplicate routes, with 2 different verbs

src/Support/Http/routes.php
```diff
@@ -4,14 +4,11 @@
     $routeName = config('lighthouse.route_name', 'graphql');
     $controller = config('lighthouse.controller');
 
-    if (config('lighthouse.route_enable_get', false)) {
-        app('router')->get($routeName, [
-            'as' => 'lighthouse.graphql',
-            'uses' => $controller,
-        ]);
-    }
+    $methods = config('lighthouse.route_enable_get', false) ?
```
spawnia: We do ternarys with a leading operator.

lucasmichot: Changed in https://github.com/nuwave/lighthouse/pull/738/commits/472f62a7529e7a93a5d0adae7b33e73ec81da8e1

spawnia:
Can you update the changelog please? I think this alters the behaviour of Lighthouse slightly

spawnia:
Thanks @lucasmichot

PR #739, opened 2019-04-17
Enables to use a custom GlobalId resolver

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

We sometimes want to use a specific format for the GlobalID

**Changes**

This PR allows to set the class to use for encoding and decoding globalId.
By default it will use the current base64 implementation


src/Support/Contracts/GlobalId.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public function encode(string $type, $id): string;
+
+    /**
+     * Split a global id into the type and the id it contains.
+     *
+     * @param  string  $globalID
+     * @return array Contains [$type, $id], e.g. ['User', '123']
+     */
+    public function decode(string $globalID): array;
```
lucasmichot: I am not sure that this should really be part of the interface, what do you think @spawnia ?

config/config.php
```diff
@@ -181,6 +181,19 @@
...
+    | Global ID resolver
+    |--------------------------------------------------------------------------
+    |
+    | The Global ID resolver class.
+    | You can set your own here, as long as it implements the following
+    | interface: \Nuwave\Lighthouse\Support\Contracts\GlobalId
+    |
+    */
+
+    'global_id_resolver' => \Nuwave\Lighthouse\Execution\Utils\GlobalId::class,
```
spawnia: We can just let the user rebind this via the container themselves and keep the config slim.

src/Schema/Directives/DeleteDirective.php
```diff
@@ -45,9 +63,11 @@ function ($root, array $args) {
                 if ($this->directiveArgValue('globalId', false)) {
                     // At this point we know the type is at least wrapped in a NonNull type, so we go one deeper
                     if ($argumentDefinition->type->type->kind === NodeKind::LIST_TYPE) {
-                        $idOrIds = array_map([GlobalId::class, 'decodeId'], $idOrIds);
+                        $idOrIds = array_map(function ($id) {
```
spawnia: Let's break this into multiple lines so the arguments align.

src/Support/Contracts/GlobalId.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
+/**
+ * Encode and decode globally unique IDs.
+ *
+ * The way that IDs are generated basically works like this:
+ *
+ * 1. Take the name of a type, e.g. "User" and an ID, e.g. 123
```
spawnia: This part of the description just belongs in our implementation.

tests/Integration/Schema/NodeInterfaceTest.php
```diff
@@ -33,8 +33,8 @@ public function itCanResolveNodes(): void
         }
         '.$this->placeholderQuery();
 
-        $firstGlobalId = GlobalId::encode('User', $this->testTuples[1]['id']);
-        $secondGlobalId = GlobalId::encode('User', $this->testTuples[2]['id']);
+        $firstGlobalId = app(GlobalId::class)->encode('User', $this->testTuples[1]['id']);
```
spawnia: Maybe extract into a variable/instance for autocompletion.

tests/Unit/Execution/Utils/CustomGlobalIdTest.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+namespace Tests\Unit\Execution\Utils;
+
+use Tests\TestCase;
+use Nuwave\Lighthouse\Support\Contracts\GlobalId;
+use Nuwave\Lighthouse\Execution\Utils\GlobalId as BaseGlobalId;
+
+class CustomGlobalIdTest extends TestCase
```
spawnia: I think this class can be removed. All it really does is to ensure the Service Container works, nothing specific about Lighthouse in here.

tests/Unit/Execution/Utils/GlobalIdTest.php
```diff
@@ -12,8 +12,8 @@ class GlobalIdTest extends TestCase
      */
     public function itCanHandleGlobalIds(): void
     {
-        $globalId = GlobalId::encode('User', 'asdf');
-        $idParts = GlobalId::decode($globalId);
+        $globalId = app(GlobalId::class)->encode('User', 'asdf');
```
spawnia: Let's just `new` up an instance here, since this test is about the concrete implementation.

src/Schema/Directives/DeleteDirective.php
lucasmichot: Not sure what you mean here @spawnia ?

chrissm79:
Great job @lucasmichot, thanks!!!

PR #741, opened 2019-04-18
Document custom directives fully, simplify NodeValue handling

spawnia:


chrissm79:
Great work @spawnia!!

PR #743, opened 2019-04-18
Force json on graphql route

spawnia:
This makes it easier to do simple GET requests from clients
with limited HTTP configuration options and always receive
a proper result, even in case of unforeseen errors.


chrissm79:
Thanks @spawnia!

PR #744, opened 2019-04-18
Introduce PHPStan at level 1

### LINKED ISSUE #656: Static analysis
I had a great experience with static analysis tools such as https://github.com/phpstan/phpstan

Do you think we should use it for Lighthouse too? Are there other tools that work better?

spawnia:
Resolves #656


.travis.yml
```diff
@@ -51,6 +52,7 @@ install:
 
 script:
   - if [[ $COVERAGE = 1 ]]; then phpdbg -qrr vendor/bin/phpunit --coverage-clover=coverage.xml; fi
+  - if [[ $STATIC_ANALYSIS = 1 ]]; then composer require nunomaduro/larastan && composer stan; fi
```
lucasmichot: So this means the check will only be done by Travis

composer.json
```diff
@@ -38,6 +38,7 @@
         "mockery/mockery": "^1.0",
         "orchestra/database": "3.5.*|3.6.*|3.7.*|3.8.x-dev",
         "orchestra/testbench": "3.5.*|3.6.*|3.7.*|3.8.*",
+        "phpstan/phpstan": "^0.11.5",
```
lucasmichot: In this case, should this required in travis file only?

spawnia:
@lucasmichot the issue i am working around is the incompatibility of Larastan with Laravel 5.5

composer.json
spawnia: Done

.travis.yml
spawnia: At the moment, yes.

Issue #500, opened 2018-12-11
ModelNotFoundException leads to 404


spawnia:
**Describe the bug**

When firing a query that somehow produces a `ModelNotFoundException`, the app responds with a 404 page.

**Expected behavior**

A spec-compliant GraphQL error is returned.

**Environment**

Lighthouse Version: `master`
Laravel Version: `5.7`
PHP Version: `7.2`

PR #747, opened 2019-04-21
Assert that a 200 is returned when a model is not found.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

When ModelNotFound exception is triggered, assert that an empty object will be returned with a 200 status code.
Closes https://github.com/nuwave/lighthouse/issues/500

**Changes**

**Breaking changes**

### DIFF
--- a/tests/Integration/Schema/Directives/FindDirectiveTest.php
+++ b/tests/Integration/Schema/Directives/FindDirectiveTest.php
@@ -142,4 +142,34 @@ class FindDirectiveTest extends DBTestCase
             ],
         ]);
     }
+
+    /**
+     * @test
+     */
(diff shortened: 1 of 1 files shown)


spawnia:
Hey, thanks for implementing this test - it is a good one.

I am beginning to think that the root cause of #500 is not a problem within Lighthouse. Let's close the issue then, the test covers it just fine.

andershagbard:
@spawnia Why was this test removed? Right now, it returns `500 - Internal server error`

Should `findOrFail` really be used here, or could we just allow the object to return null?
https://github.com/nuwave/lighthouse/blob/9ac11a9a9038a49b1ac0bc82d15175dcb61283fe/src/Schema/Directives/CanDirective.php#L155

Issue #620, opened 2019-02-26
Change default pagination argument from `count` to `first` and add more pagination args


e200:
What I'm suggesting is to rename the @pagination directive argument `count` to `first` since it makes more sense. Count stands more for an operation than a filter argument.

I want also suggest to add more pagination arguments, like `last`:

```graphql
query RECENT_PRODUCTS {
  products(last: 15) {
    data {
      name
    }
  }
}
```

lucasmichot:
I guess ppl are missing many other args too 
Working now on a PR that gives more flexibility

spawnia:
Can you give an usage example for the extra args you are referring too?

lucasmichot:
`before` and `last` for instance 


```gql
# Github
{
  repository(owner: "nuwave", name: "lighthouse") {
    issues(last: 20, before: "Y3Vyc29yOnYyOpHOGMtGew==") {
      nodes {
        id
        title
        createdAt
      }
    }
  }
}
```

e200:
```graphql
query RECENT_PRODUCTS {
  products(last: 15) {
    data {
      name
    }
  }
}

**Comment by spawnia** [comment]
I am not sure i love `first` in combination with `page`.

The examples for `last` and `before` make sense to me, seems like something we might do.
How would that actually look like in terms of implementation? It seems especially tricky to do with cursor-based pagination.

**Comment by e200** [comment]
The `first` make sense you implement the `last` argument and can be used to do the same as a `limit` or `count` arg can do.

The user must be smart enough to understand that `page` is more related to `pagination` while `first` will work using `@paginate` or not.

**Comment by lucasmichot** [comment]
Actually if we take the example of Github GraphQL implementation, there is no page.
There will always be `first`, `last`, `before` and `after` + extra parameters depending on the connection type.

An error message is also returned when using `first` and `last` toegether:

```json
{
  "data": {
    "repository": null
  },
  "errors": [
    {
      "path": [
        "repository",
        "issues"
      ],
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "message": "Passing both `first` and `last` to paginate the `issues` connection is not supported."
    }
  ]
}
```

spawnia:
@chrissm79 are you fine with changing `count` to `first`?

chrissm79:
[...]

PR #752, opened 2019-04-25
Add config option pagination_amount_argument

spawnia:
It sets the name to use for the generated argument on paginated fields
that controls how many results are returned.
This will default to "first" in v4.

- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Resolves https://github.com/nuwave/lighthouse/issues/620

**Changes**

Users can now use `first` instead of `count` when using default pagination.

**Breaking changes**

No, but the default will change in v4.


chrissm79:
Thanks @spawnia !

PR #754, opened 2019-04-25
Fixed order for nested operations

### LINKED ISSUE #728: The order of the definition of an relation input has unexpected side effects

spawnia:
**Related Issue/Intent**

Resolves #728
Closes #732

**Changes**

Apply the nested operations within a nested mutation in a consistent order
that makes sense.

This makes sure we are in line with what the spec requirement that the argument
order given by the client must not matter: https://graphql.github.io/graphql-spec/draft/#sec-Language.Arguments

**Breaking changes**

Nope, the previous behaviour was inconsistent and flaky.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Default the config to always set the `Accept: application/json` header https://github.com/nuwave/lighthouse/pull/743
 - Declare a single named route which handles POST/GET instead of 2 seperate routes https://github.com/nuwave/lighthouse/pull/738
+- Apply the nested operations within a nested mutation in a consistent order
+  that makes sense https://github.com/nuwave/lighthouse/pull/754
 
 ## [3.4.0](https://github.com/nuwave/lighthouse/compare/v3.3.0...v3.4.0) - 2019-04-18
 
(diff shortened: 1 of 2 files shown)


chrissm79:
Thanks @spawnia!

spawnia:
Thanks to @megawubs for spotting this issue and coming up with the fix!

PR #761, opened 2019-04-30
Refresh GraphQLRequest singleton between multiple requests

spawnia:
**Related Issue/Intent**

This prevents a weird error that occurs in tests.

If `GraphQLRequest` is a singleton and test code fires multiple HTTP requests, Lighthouse
always receives the same request.

**Changes**

Forget the  `GraphQLRequest` in the container after resolving each individual HTTP request.

**Breaking changes**

No

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -11,7 +11,6 @@ use Laravel\Scout\ScoutServiceProvider;
 use Tests\Utils\Policies\AuthServiceProvider;
 use Orchestra\Database\ConsoleServiceProvider;
 use Illuminate\Foundation\Testing\TestResponse;
-use Nuwave\Lighthouse\Execution\GraphQLRequest;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Nuwave\Lighthouse\LighthouseServiceProvider;
 use Orchestra\Testbench\TestCase as BaseTestCase;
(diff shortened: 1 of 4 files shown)


mfn:
> If `GraphQLRequest` is a singleton and test code fires multiple HTTP requests, Lighthouse
> always receives the same request.

It feels _slightly_ strange that the change, primarily for support in the tests, is required in non-test code.

Wouldn't it be, for those tests who care, sufficient to perform this call to forget between the instances?

Well I guess this is why you removed it in `postGraphQL` but ... can't say exactly, but doesn't feel right to have this in "production" code, mucking around with the container 🤷‍♀️

spawnia:
@mfn i understand what you are getting at.

Testability is definitely a first-class citizen in Lighthouse and i think it is good to remove *Gotcha!* behaviours such as this.

While the call into the container may not necessary in production, it does not hurt much either. Laravel, and by extension Lighthouse, optimize for a nice dev experience rather then peak performance.

mfn:
Ack, thanks for the explanation 👍

tests/TestCase.php
```diff
@@ -215,11 +214,10 @@ protected function query(string $query): TestResponse
      */
     protected function postGraphQL(array $data, array $headers = []): TestResponse
     {
-        $this->app->forgetInstance(GraphQLRequest::class);
-
         return $this->postJson(
```
lucasmichot: 👍

chrissm79:
Thanks @spawnia, this was a tricky one to catch in tests unless you knew what you were looking for!

PR #762, opened 2019-04-30
Add helper to TestCase to get  a fully qualified reference to a method that is defined on the test class

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

Factorise tests resolvers

**Changes**

**Breaking changes**

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -290,13 +290,13 @@ abstract class TestCase extends BaseTestCase
     }
 
     /**
-     * Get the test resolver.
+     * Get a fully qualified reference to a method that is defined on the test class.
      *
      * @param  string  $method
      * @return string
      */
-    protected function getResolver(string $method = 'resolve'): string
+    protected function qualifyTestResolver(string $method = 'resolve'): string
     {
-        return addslashes(static::class)."@{$method}";
+        return addslashes(static::class).'@'.$method;
     }
 }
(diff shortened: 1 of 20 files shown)


tests/TestCase.php
```diff
@@ -288,4 +288,15 @@ protected function placeholderQuery(): string
...
+
+    /**
+     * Get the test resolver.
+     *
+     * @param  string  $method
+     * @return string
+     */
+    protected function getResolver(string $method = 'resolve'): string
+    {
+        return addslashes(static::class)."@{$method}";
```
spawnia:
```suggestion
        return addslashes(static::class).$method;
```

tests/TestCase.php
```diff
@@ -288,4 +288,15 @@ protected function placeholderQuery(): string
...
         ';
     }
+
+    /**
+     * Get the test resolver.
+     *
+     * @param  string  $method
+     * @return string
+     */
+    protected function getResolver(string $method = 'resolve'): string
```
spawnia: `get` indicicates this returns a property, so maybe we can rename that, ideas:
- qualifyResolverMethod
- qualifiedTestResolver

Open to suggestions.

tests/TestCase.php
spawnia: Oops, let's not do that.

tests/TestCase.php
lucasmichot: 👍

PR #763, opened 2019-04-30
Move subscription exception to Exceptions namespace.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

Move subscription exception to Exceptions namespace.

**Changes**

**Breaking changes**


spawnia:
I am not sure i like that change, having the subscriptions in their own, self-contained namespace is kind of nice.

Issue #751, opened 2019-04-24
Add Caching Ability to Schema Build Process


nrobates:
**Is your feature request related to a problem? Please describe.**

The issue is that when you have larger sized API's that use the package the request duration is actually taking longer than other GraphQL packages. We recently migrated from the `SonarSoftware/laravel-graphql` GraphQL package and now all of our requests are taking twice as long to complete.

**Describe the solution you'd like**

I have done a considerable amount of tracing and profiling and have pinpointed the issue to be inside the `nuwave/lighthouse/src/GraphQL.php` file in the `prepSchema` method. While the documentAST is cached, the real problem is that the following needs to be cached:
```
if (empty($this->executableSchema)) {
            $this->executableSchema = $this->schemaBuilder->build(
                $this->documentAST()
            );
}
```
We really need to have `$this->executableSchema` to be cached as the build process on every request is negatively affecting the request time. However just trying to add in a cache to test this was meant with a great deal of `Serialization of 'Closure' is not allowed` errors.

Any suggestions or thoughts on this would be appreciated.

spawnia:
Thanks for the issue, i am aware of the performance problems.

Generally, Closures should be able to be serialized through https://github.com/opis/closure, if this is easy enough to do, we might employ this as a quick and easy fix.

While this approach could help and should require little work, i think the most effective solution to the performance problem would be to utilize [lazy-loading of types](http://webonyx.github.io/graphql-php/type-system/schema/#lazy-loading-of-types).

We are tracking this over at https://github.com/nuwave/lighthouse/issues/400

Issue #636, opened 2019-03-01
Large schema with many @paginate queries uses up too much memory


atomita:
**Describe the bug**

Does not work a query when the schema has many the queries.  

schema
```graphql
"A datetime string with format 'Y-m-d H:i:s', e.g. '2018-01-01 13:00:00'."
scalar DateTime @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\DateTime")

"A date string with format 'Y-m-d', e.g. '2011-05-23'."
scalar Date @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\Date")

type Query {
    users: [User!]! @paginate(type: "paginator" model: "App\\User")
    user(id: ID @eq): User @find(model: "App\\User")
}

type User {
    id: ID!
    name: String!
    email: String!
    created_at: DateTime!
    updated_at: DateTime!
}

type Company @model(class: "App\\Company") {
  id: Int!

  name: String!

  users: [User!]!
  @belongsToMany(
    type: "paginator"
  )
}

extend type Query {
  companies: [Company!]!
  @paginate(
    type: "paginator"
    model: "App\\Company"
  )

  companies1: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies2: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies3: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies4: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies5: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies6: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies7: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies8: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies9: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies10: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies11: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies12: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies13: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies14: [Company!]! @paginate(type: "paginator" model: "App\\Company")
  companies15: [Company!]! @paginate(type: "paginator" model: "App\\Company")
[...]

PR #768, opened 2019-05-04
Improve performance by lazily unserializing and loading types from the schema

spawnia:
This PR brings a major performance boost to Lighthouse by utilizing [lazy type loading](http://webonyx.github.io/graphql-php/type-system/schema/#lazy-loading-of-types).

Real-world tests in our application have shown a total speed increase of about 4x. We have been using this branch in production for the last two weeks, it has been working great.

Here is the benchmark for how the new unserialization performs in relation to a naive,
full schema scan.

```
+------------------------------+-----------+
| subject                      | mean      |
+------------------------------+-----------+
| deep schema scan             | 149.697μs |
| DocumentAST before           | 46.169μs  |
| DocumentAST after            | 25.858μs  |
+------------------------------+-----------+
```

Resolves https://github.com/nuwave/lighthouse/issues/751
Resolves https://github.com/nuwave/lighthouse/issues/400
~~Probably also~~ Resolves https://github.com/nuwave/lighthouse/issues/636

**Breaking Changes**

Unfortunately, yes.

Even if we were to keep the API compatible, there might still be dependence on the fact that right now, the whole schema including all directives are eagerly evaluated.

The API for AST manipulation was completely reworked. Turns out that the functional paradigms we tried to apply (immutability & pure functions) were resulting in major performance degradations and were needlessly complicating the transformations. Thus, i converted all that to be explicitely mutating, the methods that deal with the AST signify that by using pass-by-reference `&parameters` and are returning `: void`.

### Added

- Add the `@namespace` directive as a replacement for the removed `@group` directive https://github.com/nuwave/lighthouse/pull/768

### Fixed

- Avoid growing the memory extensively when doing complex AST manipulation https://github.com/nuwave/lighthouse/pull/768

### Changed

- Bumped the requirement on `webonyx/graphql-php` to `^0.13.2` https://github.com/nuwave/lighthouse/pull/768
- Rename directive interfaces dealing with types from `Node*` to `Type*` https://github.com/nuwave/lighthouse/pull/768
- Change the signature of the AST manipulating directive interfaces: `TypeManipulator`, `FieldManipulator` and `ArgManipulator` https://github.com/nuwave/lighthouse/pull/768
- Change the API of the `DocumentAST` class to enable a more performant implementation https://github.com/nuwave/lighthouse/pull/768
- Enable the schema caching option `lighthouse.cache.enable` by default https://github.com/nuwave/lighthouse/pull/768 
- Lazily load types from the schema. Directives defined on parts of the schema that are not used within the current
  query are no longer run on every request https://github.com/nuwave/lighthouse/pull/768

### Removed

- Remove `@group` directive in favour of `@middleware` and `@namespace` https://github.com/nuwave/lighthouse/pull/768


chrissm79:
Those benchmarks 😍 🎉

alberthaff:
Awesome. This looks very promising, @spawnia !

I have a bunch of custom directives that i have to rewrite in order to test this. I hope to get it done later today.

spawnia:
@alberthaff can you provide some examples of the changes you had to make to your directives?

alberthaff:
I had to among others, switch out `NodeManipulator` with `FieldManipulator` and change some of the code that generates/ adds types, fields and directives programatically, since `TypeRegistry::register` was removed.

I was still running the beta for version 3, so there has most likely been a lot of changes since then. :-)

composer.json
```diff
@@ -38,6 +38,7 @@
         "mockery/mockery": "^1.0",
         "orchestra/database": "3.5.*|3.6.*|3.7.*|3.8.x-dev",
         "orchestra/testbench": "3.5.*|3.6.*|3.7.*|3.8.*",
+        "phpbench/phpbench": "@dev",
```
olivernybroe: Why do we require dev of phpbench?

docs/master/api-reference/directives.md
```diff
@@ -804,6 +783,31 @@ type User @model {
...
+applies to fields of the type as well. This allows you to specify
+a common namespace for a group of fields.
+
+```graphql
+extend type Query @namespace(field: "App\\Blog") {
+  posts: [Post!]! @field(resolver: "Post@resolveAll")
+}
+```
+
+An `@namespace` directive defined on a field wins in case of a conflict.
```
olivernybroe:
```suggestion
A `@namespace` directive defined on a field directive wins in case of a conflict.
```

src/Defer/Defer.php
```diff
@@ -74,17 +74,17 @@ public function __construct(CanStreamResponse $stream, GraphQL $graphQL)
     /**
      * Set the tracing directive on all fields of the query to enable tracing them.
      *
-     * @param  \Nuwave\Lighthouse\Events\ManipulateAST  $ManipulateAST
+     * @param  \Nuwave\Lighthouse\Events\ManipulateAST  $manipulateAST
      * @return void
      */
-    public function handleManipulateAST(ManipulateAST $ManipulateAST): void
+    public function handleManipulateAST(ManipulateAST $manipulateAST): void
```
olivernybroe: Maybe we should add in our contribution guide how variable naming is. Should this be named `manipulateAst` instead. For example in the `GraphQl.php` file we have a variable called `astBuilder`, if the naming should be the same as in here, it should be named `aSTBuilder`.

src/Pagination/PaginationManipulator.php
```diff
@@ -47,15 +47,15 @@ public static function transformToPaginatedField(
...
-     * @return \Nuwave\Lighthouse\Schema\AST\DocumentAST
+     * @return void
      */
     public static function registerConnection(
-        FieldDefinitionNode $fieldDefinition,
-        ObjectTypeDefinitionNode $parentType,
-        DocumentAST $documentAST,
+        FieldDefinitionNode &$fieldDefinition,
+        ObjectTypeDefinitionNode &$parentType,
+        DocumentAST &$documentAST,
```
olivernybroe: Why is is needed to pass them by reference?

benchmarks/ASTUnserializationBench.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Benchmarks;
+
+use GraphQL\Language\Parser;
+use Nuwave\Lighthouse\Schema\AST\DocumentAST;
+
+/**
+ * @BeforeMethods({"prepareSchema"})
+ */
+class ASTUnserializationBench
```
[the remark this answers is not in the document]
olivernybroe: Could we add phpbench into automated tests? Just so we can see the results in a PR, not necessarily make rules for how it should perform.

src/Pagination/PaginationManipulator.php
alberthaff: If we need to add custom types, they are added to `$documentAST->types`, which makes them available globally. I think that is why it is added with `&`.

src/Defer/Defer.php
spawnia: `aSTBuilder` just looks really weird to me. Variable naming is not that important, we just camelCase and call it a day :)

composer.json
spawnia: Because that is what they recommend https://phpbench.readthedocs.io/en/latest/installing.html#composer-install

benchmarks/ASTUnserializationBench.php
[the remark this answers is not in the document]
spawnia: I don't really want it in travis, as the benchmarks can get quite slow.

src/Pagination/PaginationManipulator.php
spawnia: Trying to shoehorn a non-mutable style into the AST manipulation has caused lots of additional complexity and issues such as #636 

It is not really needed to pass them by reference, since objects are mutable anyway, but i think it is more explicit and obvious that way.

benchmarks/ASTUnserializationBench.php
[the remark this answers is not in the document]
olivernybroe: I have no idea about how to run it async, however travis should just use the return code from the command. So if we make it not return an error code, that might work.

Just thinking that it could be useful for us to see the results of the benchmark whenever a user makes a new PR.

src/Defer/Defer.php
olivernybroe: I agree, aSTBuilder looks really weird and is confusing. It looks like we are saying `a ST Builder`, not `AST Builder`. 
But great, sound perfect :)

src/Pagination/PaginationManipulator.php
olivernybroe: I agree with keeping it mutable. Was just wondering if there were any reason for it, but I guess by reference is faster when we are dealing with an AST, as they are often pretty big. 
Not that the speed impact for that should matter :)

spawnia:
@alberthaff added TypeRegister->register back in. It's quite useful actually, just got around to using it myself.

benchmarks/ASTUnserializationBench.php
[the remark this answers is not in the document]
spawnia: Added it in as a job that is allowed to fail

nrobates:
This is great @spawnia. So far the benchmarks are showing an exact increase in performance equivalent to the decrease we experienced. Very excited for this.

spawnia:
@nrobates decrease in comparison too?

benchmarks/ASTUnserializationBench.php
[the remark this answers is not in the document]
spawnia: https://travis-ci.org/nuwave/lighthouse/jobs/530344137

nrobates:
@spawnia I was just meaning that we noticed an increase in time by 2-4 times what it used to be which I referenced as decrease in performance. So your benchmarks are showing an increase in performance  and speed in offset.

composer.json
```diff
@@ -30,13 +30,12 @@
         "illuminate/routing": "5.5.*|5.6.*|5.7.*|5.8.*",
         "illuminate/support": "5.5.*|5.6.*|5.7.*|5.8.*",
         "illuminate/validation": "5.5.*|5.6.*|5.7.*|5.8.*",
-        "opis/closure": "^3.0",
-        "webonyx/graphql-php": "^0.13"
+        "webonyx/graphql-php": "^0.13.2"
```
olivernybroe: What feature are we requiring from 0.13.2 since you changed it? Isn't it better to have it as loose as possible, composer will automatically install the best version possible.

spawnia: https://github.com/webonyx/graphql-php/pull/475

alberthaff:
@spawnia 
I apparently was not using the schema caching until now.

After the commit that enables caching by default, my endpoint returns an exception when the schema is loaded from cache.

```
(1/1) ErrorExceptionUndefined offset: 0
--
in NodeList.php line 57
at HandleExceptions->handleError(8, 'Undefined offset: 0', '/Users/alberthaff/project/vendor/webonyx/graphql-php/src/Language/AST/NodeList.php', 57, array('offset' => 0))in NodeList.php line 57
at NodeList->offsetGet(0)in NodeList.php line 118
at NodeList->getIterator()in TypeRegistry.php line 133
```

 It works fine for the first request after i clear the cache.

spawnia:
@alberthaff have you updated the dependency to `webonyx/graphql-php` as well? I know this error, it is fixed within `0.13.2`

alberthaff:
@spawnia That fixed it. Thanks!

spawnia:
@chrissm79 now that the docs update has landed, i have given this some final polish and fixed the merge. I think we are good to go here. Let's release this as an alpha right away.

I added a `3.x` branch if we require fixes to be backported while we are readying `master` for a full-blown `4.0`

chrissm79:
Amazing work on this one @spawnia 👍Thank you!!!

mfn:
Wow, congratulations 🎉 Looks like it was almost an Herculean effort!

alberthaff:
Awesome work!

spawnia:
Released this as the first alpha for v4. https://github.com/nuwave/lighthouse/releases/tag/v4.0-alpha.1

PR #773, opened 2019-05-06
Avoid calling ->toArray on collections when we just want to unwrap it

spawnia:
The `toArray()` function actually calls `toArray()` recursively, which is not actually what we want in all of those cases - we just need the underlying items.

This should improve performance a bit.


chrissm79:
Nice, thanks @spawnia!

PR #774, opened 2019-05-07
Allow Query input argument using `@spread` to be optional

### LINKED ISSUE #770: Allow query input using @spread to be optional

I'm currently using the @spread directive on an input to filter a query like so:

```
input UsersFiltersInput{
    id: SimpleIdFilter @eq
    login: SimpleStringFilter @eq
    ...
}

type Query{
    users(
            filters: UsersFiltersInput @spread
            orderBy: [OrderByClause!] @orderBy
        ): [User] 
        @all 
    usersPaginated(
            filters: UsersFiltersInput @spread
            orderBy: [OrderByClause!] @orderBy
        ): [User]
        @paginate(defaultCount: 25)
}
```

Right now, everything works so long as you pass the filters argument into the query.
If you omit the argument, you get the following error:

   • Invalid argument supplied for foreach()

Because of these lines in Schema/Factory/FieldFactory.php

```
                    foreach ($this->spreadPaths as $argumentPath) {
                        $inputValues = $this->argValue($argumentPath);
                        $this->unsetArgValue($argumentPath);

                        array_pop($argumentPath);

                        foreach ($inputValues as $key => $value) {
                            $this->setArgValue(
                                array_merge($argumentPath, [$key]),
                                $value
                            );
                        }
                    }
```

**I'd like to be able to omit my filters input arg and have lighthouse return all results unfiltered**

Run a check that spreadsPath is not empty before running logic.
```
                if (!empty($this->spreadPaths)){
                    foreach ($this->spreadPaths as $argumentPath) {
                        $inputValues = $this->argValue($argumentPath);
                        $this->unsetArgValue($argumentPath);

                        array_pop($argumentPath);

                        foreach ($inputValues as $key => $value) {
                            $this->setArgValue(
                                array_merge($argumentPath, [$key]),
                                $value
                            );
                        }
                    }
                }
```

cradigan:
- fixes issue where a where an error occurs if a query that accepts a spreaded input does not actually pass in the input (even if not set as required)
- tests added to verify fix

- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**
Resolves #770  

**Changes**

- You can now omit an input from your query that uses the spread directive without getting an error

**Breaking changes**

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/nuwave/lighthouse/compare/v3.5.1...master)
+## [Unreleased](https://github.com/nuwave/lighthouse/compare/v3.5.2...master)
+
+## [3.5.2](https://github.com/nuwave/lighthouse/compare/v3.5.1...v3.5.2)
+
+### Fixed
+
+- `input` arguments that were `@spread` into the parent can now be optional https://github.com/nuwave/lighthouse/pull/774
 
 ## [3.5.1](https://github.com/nuwave/lighthouse/compare/v3.5.0...v3.5.1)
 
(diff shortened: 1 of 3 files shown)


spawnia:
Thanks @cradigan looking really good!

PR #777, opened 2019-05-09
Respect the model's connection for database transaction  during `@create` and `@update`

adriangligor:
- [ ] Added or updated tests
not sure how to go about this, no tests exist for these classes
no documentation necessary
- [ ] Updated the changelog
not sure how to go about this

**Related Issue/Intent**

When using Lighthouse on a Lumen or Laravel project with multiple database connections, the `@create` and `@update` directives use the default database without respecting the model's declared connection. This trivial patch fixes this.

**Changes**

When calling `DatabaseManager::connection()`, provide `Model::getConnectionName()` for the optional `name` argument. This will then open a transaction on the main model's connection. The case where nested models define different database connections for themselves is NOT handled, as in that case, transactional behaviour cannot be guaranteed anyway.

**Breaking changes**

This is a trivial fix and should not break any existing behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


spawnia:
Thanks @adriangligor great job

adriangligor:
Cool, thanks @spawnia for the super quick reaction, and the great framework!

PR #781, opened 2019-05-10
Deprecate controller config

spawnia:
This config option is not really necessary and i think nobody is using it, the implementation in the Controller is pretty basic but does enough so that i do not think anyone would want to reimplement it.

It is easy to define another controller thorugh Laravel if needed and call the rest of Lighthouse.

This just marks the config option as deprecated for v4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v3.5.3...master)
 
+### Deprecated
+
+- The `controller` config option will be removed in v4 
+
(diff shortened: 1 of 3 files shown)


chrissm79:
Agreed, thanks @spawnia

stayallive:
@spawnia @chrissm79 I would really appreciate it if some way to disable the registration of routes (without overriding the service provider) could stay.

Because we run multiple domains and to prevent duplicated information we register the routes manually by setting the `controller` option to `false`. I'm perfectly fine with a `'routes' => false` setting or something because I have no need to change the actual controller but I do want to be able to register it from my own routes file to prevent needing to duplicate the (cascading) middleware and domain configuration in the lighthouse configuration file.

spawnia:
@stayallive i thought about the route config quite a bit and came up with a new and simplified config. It allows you to disable route registration completely. https://github.com/nuwave/lighthouse/pull/820/files

stayallive:
@spawnia, that works for me 👍 thanks for that!

Issue #655, opened 2019-03-06
Standardize error messages


spawnia:
Right now the error messages that are thrown are pretty messy.

We should standardize them to make them more predictable and uniform.

- Should we terminate messages with a dot `.`?
- How do we quote things like class names? Backticks, Single or double quotes, square brackets, not at all?

olivernybroe:
Thumbs up on this comment if you think we should terminate with dot.

alberthaff:
@olivernybroe I agree. This definitely makes longer errors messages easier to read.

I think it would also make errors easier to read, if we placed URLs referring to docs, etc. at the very end.
```

// Current Exception message

Error
'Could not find a valid map, be sure to conform to GraphQL multipart request specification: https://github.com/jaydenseric/graphql-multipart-request-spec'


// URL to docs at the end

Error
'Could not find a valid map. Be sure to conform to GraphQL multipart request specification. Instead got: null. (see: https://github.com/jaydenseric/graphql-multipart-request-spec).'
```

JasonTheAdams:
Hi! Jumping in here to see what I can find. 😃

I'm starting with an audit of how exceptions are currently generated, as well as the messages they provide.

### Test-Only Exceptions
The following are exceptions that are only thrown during tests. I suggest creating a `tests/Exceptions` directory for these. Exceptions within `src/Exceptions` should be limited to exceptions used during runtime.

- [AuthenticationException](https://github.com/nuwave/lighthouse/blob/master/src/Exceptions/AuthenticationException.php)
- 

### Unused Error Code
The following are classes/exceptions that do not currently have any use within the codebase.

- [HasErrorBuffer (trait)](https://github.com/nuwave/lighthouse/blob/master/src/Support/Traits/HasErrorBuffer.php)
- [HasErrorBuffer (interface)](https://github.com/nuwave/lighthouse/blob/master/src/Support/Contracts/HasErrorBuffer.php)
- [ErrorBuffer](https://github.com/nuwave/lighthouse/blob/master/src/Execution/ErrorBuffer.php)

Note on `ErrorBuffer`: It is used by the trait, which is not used. It is used by the interface, which is checked for in `FieldFactory` but is never implemented.

---

This is a work in progress. I'm still doing research and will edit this comment over the next couple days with more details.

spawnia:
@JasonTheAdams looks great so far!

[...]

PR #786, opened 2019-05-13
Add compatibility layer to allow @middleware to support Lumen

### LINKED ISSUE #421: Since v2.5, the `@middleware` directive does not work with Lumen
**Describe the bug**
in MiddlewareDirective.php, call getMiddleware() and getMiddlewareGroups().
but Laravel\Lumen\Routing\Router does not have getMiddleware and getMiddlewareGroups function.

Router of lumen: https://github.com/laravel/lumen-framework/blob/5.7/src/Routing/Router.php

**Output/Logs**
`Call to undefined method Laravel\Lumen\Routing\Router::getMiddleware() in /app/lumen/vendor/nuwave/lighthouse/src/Schema/Directives/Fields/MiddlewareDirective.php:167`

**Environment**
Lighthouse Version: from 2.5
Lumen Version: ^5.7
PHP Version: 7.2.11

adriangligor:
- [ ] Added or updated tests
    I tried to add code to MiddlewareDirectiveTest.php but ran into a fundamental dillema: the test suite is made to work with Laravel, not Lumen. The test ultimately extends `Orchestra\Testbench\TestCase` which mocks a `\Illuminate\Foundation\Application` $app, not a `\Laravel\Lumen\Application` $app, which would be needed by the code added in this patch. The patch does not break existing tests though.
    This is a bugfix and doesn't need special mention in the docs.
- [x] Updated the changelog
    It's perhaps better if the maintainer does this to their liking.

**Related Issue/Intent**
Resolves #421 - The `@middleware` directive resolves middleware in a way specific to Laravel. Unfortunately Lumen does not provide public interfaces and methods to resolve middleware at all, so I had to resort to reflection and accessing protected member variables. I took care for the patch to not "invade" the existing code too much, so if Lumen gets better public accessors for middleware at a later point, it will be trivial to remove the code again.

**Changes**
Technically, the patch checks the instance of the router service received from the container. If it's a Lumen router, a replacement router fake is created which wraps around the `\Laravel\Lumen\Application` $app singleton. Unlike in Laravel, where all the information regarding middleware is held by the router service, in Lumen it's held in protected members of the $app. The fake exposes the same interface as the Laravel router (for the purpose of middleware resolving) and allows the rest of the code to work unchanged.

**Breaking changes**
none

### DIFF
--- a/src/LighthouseServiceProvider.php
+++ b/src/LighthouseServiceProvider.php
@@ -151,7 +151,7 @@ class LighthouseServiceProvider extends ServiceProvider
         $this->app->singleton(MiddlewareAdapter::class, function (Container $app): MiddlewareAdapter {
             // prefer using fully-qualified class names here when referring to Laravel-only or Lumen-only classes
             if ($app instanceof \Illuminate\Foundation\Application) {
-                return new LaravelMiddlewareAdapter($app->get('router'));
+                return new LaravelMiddlewareAdapter($app->get(\Illuminate\Routing\Router::class));
             } elseif ($app instanceof \Laravel\Lumen\Application) {
                 return new LumenMiddlewareAdapter($app);
             }
--- a/src/LighthouseServiceProvider.php
+++ b/src/LighthouseServiceProvider.php
@@ -155,7 +155,7 @@ class LighthouseServiceProvider extends ServiceProvider
                 return new LumenMiddlewareBridge($app);
             }
 
-            throw new Exception('Could not correctly determine Laravel framework flavor');
+            throw new Exception('Could not correctly determine Laravel framework flavor.');
         });
 
         if ($this->app->runningInConsole()) {
--- a/src/LighthouseServiceProvider.php
+++ b/src/LighthouseServiceProvider.php
@@ -4,11 +4,14 @@ namespace Nuwave\Lighthouse;
 
 use Closure;
 use Exception;
+use Illuminate\Foundation\Application as LaravelApplication;
+use Illuminate\Routing\Router;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Illuminate\Validation\Validator;
 use Illuminate\Support\ServiceProvider;
 use GraphQL\Type\Definition\ResolveInfo;
+use Laravel\Lumen\Application as LumenApplication;
 use Nuwave\Lighthouse\Schema\NodeRegistry;
 use Nuwave\Lighthouse\Schema\TypeRegistry;
 use Nuwave\Lighthouse\Console\QueryCommand;
@@ -150,13 +153,17 @@ class LighthouseServiceProvider extends ServiceProvider
 
         $this->app->singleton(MiddlewareAdapter::class, function (Container $app): MiddlewareAdapter {
             // prefer using fully-qualified class names here when referring to Laravel-only or Lumen-only classes
-            if ($app instanceof \Illuminate\Foundation\Application) {
-                return new LaravelMiddlewareAdapter($app->get(\Illuminate\Routing\Router::class));
-            } elseif ($app instanceof \Laravel\Lumen\Application) {
+            if ($app instanceof LaravelApplication) {
+                return new LaravelMiddlewareAdapter(
+                    $app->get(Router::class)
+                );
+            } elseif ($app instanceof LumenApplication) {
                 return new LumenMiddlewareAdapter($app);
             }
 
-            throw new Exception('Could not correctly determine Laravel framework flavor.');
+            throw new Exception(
+                'Could not correctly determine Laravel framework flavor, got ' . get_class($app) . '.'
+            );
         });
 
         if ($this->app->runningInConsole()) {
--- a/src/LighthouseServiceProvider.php
+++ b/src/LighthouseServiceProvider.php
@@ -40,9 +40,9 @@ use Nuwave\Lighthouse\Schema\Source\SchemaSourceProvider;
 use Nuwave\Lighthouse\Support\Contracts\ProvidesResolver;
 use Nuwave\Lighthouse\Support\Contracts\CanStreamResponse;
 use Nuwave\Lighthouse\Support\Http\Responses\ResponseStream;
-use Nuwave\Lighthouse\Support\Compatibility\MiddlewareBridge;
-use Nuwave\Lighthouse\Support\Compatibility\LumenMiddlewareBridge;
-use Nuwave\Lighthouse\Support\Compatibility\LaravelMiddlewareBridge;
+use Nuwave\Lighthouse\Support\Compatibility\MiddlewareAdapter;
+use Nuwave\Lighthouse\Support\Compatibility\LumenMiddlewareAdapter;
+use Nuwave\Lighthouse\Support\Compatibility\LaravelMiddlewareAdapter;
 use Nuwave\Lighthouse\Support\Contracts\GlobalId as GlobalIdContract;
 use Nuwave\Lighthouse\Support\Contracts\ProvidesSubscriptionResolver;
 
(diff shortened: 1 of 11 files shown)


adriangligor:
Let me double-check if I understood you correctly:

- I should create an abstract class MiddlewareBridge, which is implemented by two classes, LaravelMiddlewareBridge and LumenMiddlewareBridge
- These two classes should each accept different different arguments to their respective constructors, the Laravel router or the Lumen app, depending on where they get the middleware from
- The MiddlewareBridge class should contain a static factory method ::create(), which builds and returns the correct instance (which is detected by checking the instance of the `router` service)
- The factory method should be used in MiddlewareDirective's constructor to get an instance of the bridge and store it as a member variable. The bridge can then be used repeatedly during the lifetime of the directive without repeating the instanceof check.
- There is then no need to keep the LumenRouterFake class around, most of the code goes into the LumenMiddlewareBridge

Is this approximately what you were imagining? Let me know and I'll rework the patch. The point about unit testing still stands however, I can adapt the code to work with my patch on Laravel, but I can't imagine how to write the test for Lumen.

(All the proposed names are of course up for discussion.)

spawnia:
The proposed plan sounds like a sensible solution.

Later today i will have the time to look into it in more detail. Lumen support would be great.

olivernybroe:
@adriangligor Close to what I meant. Sounds like you have the general idea :) 
A small change would be to not have the static method create and leave everything to Laravel Service Container. 

So make the middleware class have a dependency of your `MiddlewareBridge` and then expect that the Container resolved it to the right type.

Then in our service provider class add in a statement the register method saying if the app is lumen, then resolve `MiddlewareBridge` as `LumenMiddlewareBridge` and vice-versa. You can then choose to bind it to the container as a static, so we don't have to resolve it more than once.

olivernybroe:
Regarding test, this is a big problem we have had in a long time. 
The reason for this is because we are using testbench which resolves a Laravel container.

There are two ways I can see to solve this. 
1. Make a new abstract test case class which creates a Lumen container instead to use when we are testing lumen.
2. Get rid of testbench and do everything from mocks/fakes. testbench is smart for quickly testing a package, but in the end it just adds more problems then the gains. 

I do not expect you at all to make these changes in this PR. But it's definitely something we have to look into.

spawnia:
Directives are always resolved through the container, so you can just typehint the dependency in the constructor, given you registered it in the ServiceProvider.

adriangligor:
Got you both. I'll rework the patch today and hopefully push again by tomorrow.

adriangligor:
Please have a look at my latest commit, I hope it's close to what you were looking for. I tested it locally with Lumen, and also made sure the existing tests still pass, which should cover Laravel.

src/LighthouseServiceProvider.php
```diff
@@ -145,6 +148,16 @@ public function provideSubscriptionResolver(FieldValue $fieldValue): Closure
             };
         });
 
+        $this->app->singleton(MiddlewareBridge::class, function (Container $app): MiddlewareBridge {
+            if ($app instanceof \Illuminate\Foundation\Application) {
+                return new LaravelMiddlewareBridge($app->get('router'));
```
olivernybroe: @spawnia Should we use the interface when getting from the container instead of the string shorthand?

src/LighthouseServiceProvider.php
```diff
@@ -145,6 +148,16 @@ public function provideSubscriptionResolver(FieldValue $fieldValue): Closure
...
+            } elseif ($app instanceof \Laravel\Lumen\Application) {
+                return new LumenMiddlewareBridge($app);
+            }
+
+            throw new Exception('Could not correctly determine Laravel framework flavor');
```
olivernybroe:
```suggestion
            throw new Exception('Could not correctly determine Laravel framework flavor.');
```
#655

src/Support/Compatibility/LumenMiddlewareBridge.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Get the value of a protected member variable of an object.
+     *
+     * @param object $object Object with protected member
+     * @param string $memberName Name of object's protected member
+     * @param null|mixed $default Default value to return in case of access error
+     * @return mixed Value of object's protected member
+     */
+    private static function accessProtected($object, $memberName, $default = null)
```
olivernybroe: Maybe move this method to Utils class. It might be something we need somewhere else?

src/Support/Compatibility/MiddlewareBridge.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Compatibility;
+
+interface MiddlewareBridge
+{
+    /**
+     * Get all of the defined middleware short-hand names.
+     *
+     * @return array
```
olivernybroe: Change the return type in docs to a more specific one. If this is just an array of strings use `string[]`

src/Support/Compatibility/MiddlewareBridge.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Compatibility;
+
+interface MiddlewareBridge
```
olivernybroe: Not sure about the name `MiddlewareBridge`.
Looking at what it does, it is more of a repository, I think I'll prefer the name `MiddlewareRepository`.
@spawnia Any input on this?

adriangligor: Is the term _Repository_ not usually used for a pattern of database access, e.g. as an alternative to the _Active Record_ approach? In any case, I'm rather new to both PHP and Laravel, so I'm not yet acquainted with the Lingo :)

src/Support/Compatibility/LumenMiddlewareBridge.php
adriangligor: Sure, no problem.

src/Support/Compatibility/MiddlewareBridge.php
adriangligor:
```suggestion
     * @return string[]
```

src/Support/Compatibility/MiddlewareBridge.php
```diff
@@ -0,0 +1,20 @@
...
+     * Get all of the defined middleware short-hand names.
+     *
+     * @return array
+     */
+    public function getMiddleware(): array;
+
+    /**
+     * Get all of the defined middleware groups.
+     *
+     * @return array
```
adriangligor:
```suggestion
     * @return string[]
```

src/Support/Compatibility/MiddlewareBridge.php
adriangligor: Wasn't even aware this was a valid type hint in php-doc :)

spawnia:
How about `MiddlewareAdapter`?

src/Support/Compatibility/MiddlewareBridge.php
spawnia: How about `MiddlewareAdapter`?

olivernybroe: Repositories are used to store/retrieve data, while hiding away where the data is actually coming from. For example Laravel uses repositories for Cache and Config and I think they also do for authentication. 
However I think that `MiddlewareAdapter` is just as good.

src/LighthouseServiceProvider.php
adriangligor: I'm going ahead and changing this to use the interface, but you guys should feel free to adapt the code to your own liking.

src/Support/Compatibility/MiddlewareBridge.php
adriangligor: I'm replacing all occurrences of "bridge" with "adapter", if you guys can think of an even more fitting term, feel free to adapt the code later.

adriangligor:
@spawnia @olivernybroe thanks to your detailed feedback, I think the patch is in a much better shape now. If there is any more I can do to get it merged, please let me know. In case of small changes like naming and other strings in order to better match your project's conventions, I don't mind at all if you guys simply change whatever you want.

olivernybroe:
@adriangligor Thank you for the great work.
We have a test there is failing after your changes. It is related to it cannot find the class `Laravel\Lumen\Application` when it is trying to analyse the codebase. Don't know if it gives a conflict, but you might be able to solve this by adding lumen as a dependency in `require-dev` section of `composer.json`.  
You can test this locally by running `phpstan analyse `

adriangligor:
Yes, clearly it can't find the class without declaring Lumen as a dependency. Let me look into it...

spawnia:
@adriangligor The bigger issue of automated Lumen tests is still looming over our heads - your solution to this particular problem seems necessary regardless. I would not even call it a workaround - nicely done!

chrissm79:
@adriangligor Excellent work on this one, thank you!! 👍

ileonardo-c:
I have seen the code of the [class LumenRouterFake](https://github.com/nuwave/lighthouse/commit/761129451d329a1bfac2ffc0c402d449283f10e4) of @adriangligor, could put in version 2.6.4?

spawnia:
@ileonardo-c i would be willing to accept a PR to the `2.x` branch.

PR #792, opened 2019-05-14
Replace sh with bash and remove the leading $

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

**Changes**

Replace sh with bash in docs annotations.

**Breaking changes**


spawnia:
I like the syntax highlighthing way better, but let's remove the leading `$` everywhere we have commadns. Makes copying harder then it needs to be.

image

lucasmichot:
See https://github.com/nuwave/lighthouse/pull/792/commits/ac7fe49dcaa937b11f3d1c3c71e3a8db056001ed @spawnia

spawnia:
Thanks!

PR #797, opened 2019-05-20
The `GraphQL` facade and the container alias `graphql` will be removed in v4

spawnia:
Reduce incidental complexity and overhead by offering only one way to do things.


chrissm79:
Thanks @spawnia

PR #801, opened 2019-05-21
Add config option `cache.ttl` to control expiration time of schema cache

splinter89:
- [x] Updated the changelog

**Related Issue/Intent**

Allow users to customise the expiration time of schema cache.
For example, this is useful when using redis cache with default `maxmemory-policy noeviction`.

**Changes**

Add new config option `cache.ttl`, which mirrors the corresponding parameter of `Cache::remember()`.
The default behaviour of remembering permanently is unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add compatibility layer to allow `@middleware` to support Lumen https://github.com/nuwave/lighthouse/pull/786
 - Add option `decode` to `@globaldId` to control the result of decoding https://github.com/nuwave/lighthouse/pull/796
+- Add config option `cache.ttl` for customizing expiration time of schema cache https://github.com/nuwave/lighthouse/pull/801
 
 ## [3.6.1](https://github.com/nuwave/lighthouse/compare/v3.6.0...v3.6.1)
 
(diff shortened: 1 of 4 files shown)


spawnia:
I don't really get the need for this option. It seems to me that expiring the schema cache is something you can put an arbitrary time value on.

The way i deploy our application results in the schema cache being cleared automatically. If that is not the case in your deployment scenario, you can just call the artisan command:

    php artisan lighthouse:clear-cache

Does that solve the issue you are having?

splinter89:
Unfortunately, it won't help in our case.
We manipulate the schema based on some logic via a custom directive, so we have to use different `cache.key` for different sets of parameters of our logic.

Custom directives are awesome, but having a single non-expiring cache greatly reduces their flexibility.

spawnia:
So you are dynamically manipulating your schema and are using multiple different schemas?
Can you elaborate on your use case a bit and give a concrete example of why this is necessary.

I have trouble imagining a scenario where one would like an arbitrary TTL. In effect that means: "I want my schema to change dynamically at random interval.". Why? To keep clients on their heels?

You can programmatically invalidate the schema cache at well-defined points in time, e.g. a redeployment or basically any other event.

splinter89:
Thank you for this nice suggestion.

I'll try to provide a better explanation. For the sake of simplicity, let's assume we don't use custom directives and schema is the same for every user. In this case we still need `cache.ttl`

Our scenario: rolling deployment with a single Redis instance, where each new version of the app sets its own `cache.key` for possibly updated schema. Of course, for a fixed app version, the schema stays the same and all clients are happy :smile:
The problem is we never know when exactly the instances running older versions are getting removed. We also don't want to depend on Redis eviction policy (it would pointlessly maximize used memory).
Solution: `cache.ttl` will keep our Redis database clean of outdated caches which would never be used as soon as the last instance with old version of the app is removed.

spawnia:
Thanks for the explanation. What you are doing makes sense, i am now convinced this is a nice feature to add.

PR #802, opened 2019-05-21
Extract test helpers into a reusable trait

spawnia:
Extract our common test helpers for making GraphQL requests into a Trait `\Nuwave\Lighthouse\Testing\MakesGraphQLRequests`, similar to `\Illuminate\Foundation\Testing\Concerns\MakesHttpRequests`.

This allows end users to easily integrate them within their test suite.

After some discussion around naming, the helper functions are now called `graphQL`, `postGraphQL` and `multipartGraphQL`


src/Testing/MakesGraphQLRequests.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    abstract protected function transformHeadersToServerVars(array $headers);
+
+    /**
+     * Execute a query as if it was sent as a request to the server.
+     *
+     * @param  string  $query
+     * @return \Illuminate\Foundation\Testing\TestResponse
+     */
+    protected function queryGraphQL(string $query): TestResponse
```
olivernybroe: Could we make this smarter or add another method for batched queries?
Right now the only way to send batched queries is to do 

```
$this->postGraphQL([
            [
                'query' => '
                    {
                        user {
                            email
                        }
                    }
                    ',
            ],
            [
                'query' => '
                    {
                        user {
                            name
                        }
                    }
                    ',
            ],
        ])
```

I think it makes sense to either make queryGraphQL take in a array of query strings and then it adds the `query` key in front of all the values, or make another method called something like `queryGraphQLBatched`.

spawnia: The example call for batched queries seems fine to me. I don't think that testing through batched queries is very common.

olivernybroe: My thoughts was just to remove the need for adding query, so I could write 

```
$this->postGraphQL([
    '{
        user {
            email
        }
    }',
    '{
        user {
            name
        }
    }'
])
```

But I guess you are right that not many will use that feature, and they can easily create the helper if they think the other syntax is too long.

src/Testing/MakesGraphQLRequests.php
```diff
@@ -0,0 +1,101 @@
...
+     * Execute a query as if it was sent as a request to the server.
+     *
+     * @param  mixed[]  $data
+     * @param  mixed[]  $headers
+     * @return \Illuminate\Foundation\Testing\TestResponse
+     */
+    protected function postGraphQL(array $data, array $headers = []): TestResponse
+    {
+        return $this->postJson(
+            config('lighthouse.route_name'),
```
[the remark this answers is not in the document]
spawnia: How would i do that? Can you PR that to the branch?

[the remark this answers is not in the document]
spawnia: Fixed it.

We might rework how routes are specified in the future, the test helper should work with what we have now.

chrissm79:
Nice work on this one @spawnia, should be pretty helpful!

PR #805, opened 2019-05-23
Use Illuminate\Contracts\Events\Dispatcher instead of concrete implementation

killtw:
using contract for events mock in testing
like `Event::fake([BroadcastSubscriptionEvent::class]);`


spawnia:
Thanks @killtw good catch.

PR #812, opened 2019-05-25
Support custom rule classes in `@rules` and `@rulesForArray`

omarusman:
- [x] Added or updated tests
- [x] Updated the changelog

Support or allow custom rules in @rules directive.

**Usage**
```cli
php artisan make:rule MyCustomRule
```
and in your _schema.graphql_
```graphql
input FooBar {
    field: String! @rules(apply: ["required", "App\\Rules\\MyCustomRule"])
}
```


src/Schema/Directives/RulesDirective.php
```diff
@@ -28,6 +28,13 @@ public function getRules(): array
     {
         $rules = $this->directiveArgValue('apply');
 
+        // Resolve custom rule namespace, if possible.
+        foreach ($rules as $key => $rule) {
+            if (strpos($rule, 'App\Rules') === 0) {
```
spawnia: I think we might just use `class_exists()` to determine if we are dealing with a custom rule.

spawnia:
@omarusman hey, thanks for the PR. I like the idea behind this, the approach is elegant.

In order for this to be merged, make sure you can check the boxes in the PR template (i added them back in to the description.). Tests, Docs and the Changelog should be in place.

In addition to the `@rules` directive, the `@rulesForArray` directive should also receive an update.

omarusman:
@spawnia Made some changes. Let me know what do you think. 😊

chrissm79:
This is awesome, thanks @omarusman and @spawnia!

omarusman:
@spawnia  cool 😊

PR #825, opened 2019-05-30
Make nested mutations work with subclassed relationship types

### LINKED ISSUE #572: Nested Mutations do not work with subclassed relationship types
**Describe the bug**

We are using a custom relation class in our code, extending Laravel's built-in `BelongsToMany` class.

When trying to populate this relationship through a nested mutation, Lighthouse fails to detect the underlying relationship, as it does a strict equality check for the typehinted class of the relationship method, see https://github.com/nuwave/lighthouse/blob/master/src/Execution/MutationExecutor.php#L286

**Expected behavior**

Lighthouse should correctly detect such relationships.

We should add a failing test for this case first, and then we can probably solve this issue by using `is_subclass_of` in combination with the strict equality check.

spawnia:
Resolves #572


chrissm79:
Thanks @spawnia!!

PR #826, opened 2019-05-30
Add Client Implementations to Subscriptions menu

enzonotario:
Hi! I think it is missed from the sidebar.

- [ ] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

None

**Changes**

Added a reference into sidebar to the file `subscriptions/client-implementations`.

**Breaking changes**

None


enzonotario:
I think the only one that needs this change is v3.7, because the new structure was introduced from 3.7. The others versions just have a `subscriptions.md` with all sections inside that file (already included in the sidebar).

PR #827, opened 2019-06-02
phpunit: remove file not present anymore

mfn:
- [x] Added or updated tests

**Related Issue/Intent**

n/a

**Changes**

Only updated `phpunit.xml.dist`

**Breaking changes**

None

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -23,7 +23,6 @@
             <directory suffix=".php">src/</directory>
             <exclude>
                 <file>src/LaravelServiceProvider.php</file>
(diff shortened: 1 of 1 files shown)


phpunit.xml.dist
```diff
@@ -23,7 +23,6 @@
             <directory suffix=".php">src/</directory>
             <exclude>
                 <file>src/LaravelServiceProvider.php</file>
```
spawnia: I think we don't need that anymore either.

mfn: Ack, removed the whole `<exclude>` part

PR #830, opened 2019-06-12
Allow empty arrays and other falsy values as input for nested mutation operations like "sync"

thytanium:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

In `belongsToMany` relationships, when using `sync` nested operation with an empty array no updates are executed. The expected behaviour from Eloquent is to disconnect all related models on an empty sync operation.

**Changes**

A change was made in `MutationExecution` class to correctly validate and pass empty arrays when present on sync operations.

**Breaking changes**

It will behave differently if someone relies on the current behaviour of not executing updates on the relationship when sync is called with an empty array.

I believe Lighthouse should follow Eloquent functionality verbatim and that's why I suggest this change.


spawnia:
I expanded the change to cover all other nested operations.

Thanks @thytanium, nice catch!

Issue #831, opened 2019-06-12
Multiple @where directives on a single field


cwhitby:
**Is your feature request related to a problem? Please describe.**

It would be great if we could apply multiple `@where` directives to a single field.
For example:
```graphql
input AuditFilterInput {
    date_after: Date @where(operator: ">=", key: "date") @where(operator: ">=", clause: "orWhere", key: "created_at")
}
```
In this case, only the first directive is applied.

**Describe the solution you'd like**

I can think of 3 options:

A) Allow multiple instances of the `@where` directive. I don't think any other directive allows this either so there might be a limitation in the core of Lighthouse.

B) Allow the `@where` directive to accept an array of conditions, in a similar format to how the Laravel Query Builder does.

C) This behaviour already exists somewhere and I just can't see it in the docs.

Curious to see what others think.

spawnia:
Have you actually tried this in the most recent version of Lighthouse? I agree that this should work.

cwhitby:
It was originally tested using version `3.7.0`, I've just checked again using `4.0-alpha.2` with the same result.

spawnia:
Can you provide a PR with a failing test case in https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Execution/BuilderTest.php please?

spawnia:
We can enable this feature (and possibly other use cases of applying the same directive repeatedly) once the spec allows it:  https://github.com/graphql/graphql-spec/pull/472

spawnia:
Changed handling of the argument filters makes this work for `@where` now, at least https://github.com/nuwave/lighthouse/pull/834 passed for now.

General availability and proper documentation of `repeatable` directives will be tracked in https://github.com/nuwave/lighthouse/issues/1133.

Feel free to reopen if you still experience issues.

PR #834, opened 2019-06-16
Add test for multiple where directives

cwhitby:
- [x] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

References #831 

**Changes**

No code changes, only adding a test for applying multiple `@where` directives on the same field

**Breaking changes**

None


spawnia:
@cwhitby I realized this problem is twofold. While i adapted the handling of `BuilderDirective`'s to allow for multiple builders, the `BaseDirective` does not take into account that there may be more than one directive of the same name.

Thanks for the PR, i am working on a generalized fix.

spawnia:
Actually, the GraphQL specification declares that directives must be unique per location. There is an ongoing RFC to allow repeatable directives: https://github.com/graphql/graphql-spec/pull/472

Jofairden:
@spawnia  you should review this again, as afaik repeatable directives are now officially supported by the spec: https://github.com/graphql/graphql-wg/blob/master/notes/2019-10-10.md#repeatable-directives-ivan-10m

spawnia:
@Jofairden thanks, I am aware. Added support in `graphql-php` for it myself and plan to enable repeatable directives in Lighthouse v5.

spawnia:
@cwhitby the test is working now, since we changed some logic around that makes it function correctly. Also tried your original test [`9a70046` (#834)](https://github.com/nuwave/lighthouse/pull/834/commits/9a7004650c45b2f15c9a576d6a2da7c2f1575159) and got a count of `0`, should be correct?

Merging this for now.

PR #835, opened 2019-06-18
Prefix column references with table name to allow scopes to join other tables

adriangligor:
- [ ] Added or updated tests
  - I added/adapted integration tests for @all and @find, the last one which also includes @eq
- [ ] Updated the changelog

**Related Issue/Intent**

`@find` and `@all` generate finder queries and can be influenced with `@eq`, `@neq`, `@in`, `@notIn` and `@where*` condition directives, while supporting scopes. A scope method receives the builder as an argument, and it may join other tables to it. When it does so, the already selected columns and generated conditions break the query, because they are not prefixing columns with the corresponding table names.

**Changes**

This patch makes sure that condition directions prefix the columns they add to the builder with the model's table name, and that finder directives prefix `*` with the table name as well, otherwise the result may contain multiple values for the same column name (e.g. "id"), which results in mixed-up values.

**Breaking changes**

Should be a pure bugfix change.

Note: the `FindDirectiveTest` can be made to fail by reverting the changes to either `FindDirective` or `EqDirective`. The `AllDirectiveTest` can be made to fail by reverting the changes to `AllDirective`.

### DIFF
--- a/tests/Utils/Models/User.php
+++ b/tests/Utils/Models/User.php
@@ -62,10 +62,5 @@ class User extends Authenticatable
         return $query
             ->join('companies', 'users.company_id', '=', 'companies.id')
             ->where('companies.name', $args['company']);
-        
-        /*return $query->whereHas('company', function (Builder $q) use ($args): void {
-            $q->where('name', $args['company']);
-        })->dd();*/
-        
     }
 }
--- a/tests/Utils/Models/User.php
+++ b/tests/Utils/Models/User.php
@@ -59,8 +59,13 @@ class User extends Authenticatable
 
     public function scopeCompanyName(Builder $query, array $args): Builder
     {
-        return $query->whereHas('company', function (Builder $q) use ($args): void {
+        return $query
+            ->join('companies', 'users.company_id', '=', 'companies.id')
+            ->where('companies.name', $args['company']);
+        
+        /*return $query->whereHas('company', function (Builder $q) use ($args): void {
             $q->where('name', $args['company']);
-        });
+        })->dd();*/
+        
     }
 }
--- a/src/Schema/Directives/AllDirective.php
+++ b/src/Schema/Directives/AllDirective.php
@@ -42,7 +42,7 @@ class AllDirective extends BaseDirective implements FieldResolver
                         $modelClass::query(),
                         $args
                     )
-                    ->get((new $modelClass())->getTable() . '.*');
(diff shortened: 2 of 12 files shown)


tests/Utils/Models/User.php
```diff
@@ -59,8 +59,8 @@ public function roles(): BelongsToMany
 
     public function scopeCompanyName(Builder $query, array $args): Builder
     {
-        return $query->whereHas('company', function (Builder $q) use ($args): void {
-            $q->where('name', $args['company']);
-        });
+        return $query
+            ->join('companies', 'users.company_id', '=', 'companies.id')
```
spawnia: Why would you want to use this over the way it was written initially?

adriangligor: The original scope was using `whereHas()`, which creates an inner query and becomes immune to the bug I'm trying to fix. In our code base, we have cases where scopes join further tables to the builder, which is as far as I know, valid Eloquent usage. I changed the code in order to trigger the buggy behaviour in integration tests.

spawnia: I think that the `whereHas` is idiomatic Eloquent style, the Laravel docs only show that https://laravel.com/docs/5.8/eloquent-relationships#querying-relationship-existence

Manually joining seems unnecessary and having to add on the table name to every column seems extremely tedious.

spawnia:
Lighthouse's built-in directives are designed to work with Eloquent.

adriangligor:
@spawnia I understand your point about consistently applying this logic to all places, and it's fine if you reject the patch.
However, I see nothing non-idiomatic about this usage, as Eloquent builders allow this interaction for a reason (the class sets up call forwarding to `Illuminate\Database\Query\Builder`). We discussed in the team how using the approach presented in the documentation (which you linked) would affect our code, and we agree that the result is not elegant (in a nutshell, we're using scope composition, where scopes on one model may call other scopes on related models in order to add more conditions as joins on their builders, e.g. "this file is visible if the folder it belongs to is visible. that folder is visible if the project it belongs to is visible, etc etc, so that the file is only visible when all its related entities are.").
We're going to fork the directives for now, as this delivers the result we're looking for.
Thanks!

spawnia:
Thanks for your interest and your contribution.

Your use case is valid, but i wager it is not quite common. The complexity of the workaround involved and the fact it is not working in older Laravel versions outweigh the benefit of supporting it.

As mentioned in the previous comments, there are ways to solve what you are trying to do.

PR #837, opened 2019-06-19
Bugfixes/fix multiple models in can directive

ElForastero:
- [x] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

`@can` directive fails if an array of `ID` is passed. For example in the `@delete` directive, as it described in the docs. 

```graphql
deletePosts(id: [ID!]!): [Posts!]! @delete @can(ability: "delete")
```

It happens when `[ID!]!` is passed as an input. In this case, we have to handle Collection.

**Changes**

Check if Collection was provided and run `Gate::check` for each item.

**Breaking changes**

No


ElForastero:
Oops. Wrong branch.

PR #838, opened 2019-06-19
Fix @can directive with multiple models and remove deprecated "if" alias

ElForastero:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

`@can` directive fails if an array of `ID` is passed. For example in the `@delete` directive, as it described in the docs. 

```graphql
deletePosts(id: [ID!]!): [Posts!]! @delete @can(ability: "delete")
```

It happens when `[ID!]!` is passed as an input. In this case, we have to handle Collection.

**Changes**

Check if Collection was provided and run `Gate::check` for each item.

**Breaking changes**

Edit by @spawnia:

While i was at it, i removed the deprecated alias `if` of the `@can` directive.

The migration is straightforward:

```diff
type Mutation {
- deletePost(id: ID!): Post @can(if: "delete")
+ deletePost(id: ID!): Post @can(ability: "delete")
}
```

### DIFF
--- a/tests/Utils/Policies/PostPolicy.php
+++ b/tests/Utils/Policies/PostPolicy.php
@@ -14,6 +14,6 @@ class PostPolicy
 
     public function delete(User $user, Post $post): bool
     {
-        return (int)$post->user_id === (int)$user->getKey();
+        return (int) $post->user_id === (int) $user->getKey();
     }
 }
--- a/tests/Utils/Policies/PostPolicy.php
+++ b/tests/Utils/Policies/PostPolicy.php
@@ -11,4 +11,9 @@ class PostPolicy
     {
         return $post->user_id === $user->getKey();
     }
+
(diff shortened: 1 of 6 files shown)


tests/Utils/Policies/PostPolicy.php
```diff
@@ -11,4 +11,9 @@ public function view(User $user, Post $post): bool
     {
         return $post->user_id === $user->getKey();
     }
+
+    public function delete(User $user, Post $post): bool
+    {
+        return (int)$post->user_id === (int)$user->getKey();
```
ElForastero: The typecasting isn't working for SQLite 🤔

spawnia: It's fine

spawnia:
Thanks for the PR @ElForastero great find

chrissm79:
Thank you @ElForastero!

PR #843, opened 2019-06-28
Fix cache key when the user is not authenticated

hosmelq:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

`@cache(private: true)` fails when the user is not authenticated.

**Changes**

Check if the user is authenticated before create the cache key.

**Breaking changes**

None


spawnia:
@hosmelq i am interested in the use case this solves. Can you give an example of when you would want this behaviour?

hosmelq:
This is my schema.

```graphql
type User {
  id: ID!
  ...
}

type Game {
  id: ID!
  viewerIsAMember: Boolean!
    @field(resolver: "App\\GraphQL\\Types\\Game@viewerIsAMember")
    @cache(maxAge: 300 private: true)
  ...
}
```

The `viewerIsAMember` field identifies if the authenticated user is part of the game but this game is accessible for not authenticated users too, for that reason I'm getting the error.

spawnia:
Makes sense. Can you add a test case?

spawnia:
Thanks @hosmelq nice job.

PR #846, opened 2019-07-03
Simplify complex validation by allowing custom validation directives

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

This change makes it easier to run complex validation through directives.

**Changes**

In some cases, validation rules are more complex and need to use entirely custom logic
or take multiple arguments into account.

To create a reusable validator that can be applied to fields, extend the base validation
directive `\Nuwave\Lighthouse\Schema\Directives\ValidationDirective`. Your custom directive
class should be located in one of the configured default directive namespaces, e.g. `App\GraphQL\Directives`.

```php
<?php

namespace App\GraphQL\Directives;

use Illuminate\Validation\Rule;
use Nuwave\Lighthouse\Schema\Directives\ValidationDirective;

class UpdateUserValidationDirective extends ValidationDirective
{
    /**
     * Name of the directive.
     *
     * @return string
     */
    {
        return 'updateUserValidation';
    }

    /**
     * @return mixed[]
     */
    public function getRules(): array
    {
        return [
            'id' => ['required'],
            'name' => ['sometimes', Rule::unique('users', 'name')->ignore($this->args['id'], 'id')],
        ];
    }

    /**
     * @return string[]
     */
    public function getMessages(): array
    {
        return [
            'name.unique' => 'The chosen username is not available',
        ];
    }
}
```

Use it in your schema upon the field you want to validate.

```graphql
type Mutation {
  updateUser(id: ID, name: String): User @updateUserValidation
}
```

**Breaking changes**

Nope


shaffe-fr:
Hi,
Great PR idea!
To be more consistent with other Laravel classes (FormRequest::rules or custom validation rule message), I would drop the `get` and name the methods `rules` and `messages`.
Are you open to PR?
Best package ever!

spawnia:
@shaffe-fr thanks for your interest. I made it so, good find.

dokicro:
Hi, can this be used at this moment?

spawnia:
It is in the latest `v4` beta.

dokicro:
Hey, @spawnia thank you for all your effort I love this package!

What do you recommend stability and documentation wise should I wait for the v4 release or just use beta?

spawnia:
We are pretty close to a release. The docs are all there, and we continuously test everything. Subscriptions are broken as of now, so we will get in a few bugfixes and then we can release.

dokicro:
Perfect, thank you for all hard work, I will try to contribute in the near feature.

PR #853, opened 2019-07-09
Fix/lumen routes

juusopiikkila:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Lumen's router doesn't have the `match` method used in routes.

**Changes**

Added a check for Lumen when adding routes and choosing the right method to be used.


spawnia:
Thanks, great find!

PR #856, opened 2019-07-09
Make the argument used for finding a model to check @can against configurable

### LINKED ISSUE #850: Add an attribute to @can directive to ignore ID
I'm writing an app currently with lighthouse (which is freaking awesome by the way) and I'm running into a problem caused by the `@can` directive trying to resolve a model by ID when an ID param is passed into the input. The problem with this is I'm creating models using a UUID from the front end so that I can do optimistic updates with the same UUID as the server, so I'm passing in the ID into a create model mutation and trying to run `@can` on that mutation and it's trying to resolve a model that does not exist. So I'm wondering if there's a possibility we can add a param like ignore ID or something to the `@can` directive. 

here's the code that's the culprit in the can directive:  

```php
if ($id = $args['id'] ?? null) {
   /** @var \Illuminate\Database\Eloquent\Model $modelClass */
   $modelClass = $this->getModelClass();
   $gateArguments[0] = $modelClass::findOrFail($id);
}
```

here's the code I'm trying to execute:

```graphql 
createQueryParameter(input: QueryParameterInput! @spread): QueryParameter! @relayCreate @can(ability: "create")
```

Side note... my `@relayCreate` directive is a custom directive I write, because Relay JS likes edges in its payloads when you create a new model... so this directive just creates a payload type and returns an edge an node on that payload type, so I don't have to write that a bunch of times in my schema and resolve it a bunch of times manually.

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Resolves #850

**Changes**

You may specify an argument that is used to find a specific model
instance against which the permissions should be checked.

```graphql
type Query {
    post(id: ID @eq): Post @can(ability: "view", find: "id")
}
```

```php
class PostPolicy
{
    public function update(User $user, Post $post): bool
    {
        return $user->id === $post->author_id;
    }
}
```

**Breaking changes**

The previous behaviour of implicitely using the `id` argument for finding a specific
model to authorize against now no longer works.


alexwhb:
@spawnia You are the best! Thanks for adding that in.

spawnia:
Btw i am open to renaming `find`.

Maybe `identifier` or `key`?

@chrissm79 what do you think?

PR #861, opened 2019-07-12
input directive example updated

elmarce:
For a better understanding of @input directive associated with @spread directive


spawnia:
Now the example code works, nice!

PR #862, opened 2019-07-12
Fix BelongsToMany example in nested mutation docs

elmarce:
isn't it supposed to be @update instead of @create ?


spawnia:
Thanks!

PR #864, opened 2019-07-12
Added disclaimer about mass assignment

Vynlar:
I've heard several examples of new users being confused about why their `@create` and `@update` directives are not working and a common problem is not having put the fields into their `$fillable` list to allow mass assignment.


spawnia:
Thanks @Vynlar good job!

PR #871, opened 2019-07-20
Extend relay relationships to support pivot meta on edge

### LINKED ISSUE #869: Extend connection pagination to support pivot meta on edge
I’m thinking through how pivot data would/could work in Lighthouse. It’s a bit of a do or die for me as I need this to work well or have to use a different framework. I found this issue, which outlines a proposed way: https://github.com/nuwave/lighthouse/issues/113

My primary issue with that suggestion, however, is that I will be using the Relay-style connection wherein pivot meta should be a part of the edge data, not a separate node. So for a user-team many-to-many relationship with `owner` as pivot meta, it would look like:
```
user {
  teams {
    edges {
      cursor
      node
      role
    }
  }
}
```
I’m happy to make a PR for this, but I could use some guidance in how this would be best reflected in the schema-first approach. Since connections and edges are implicit within Lighthouse, it seems like the `@paginate` and `@belongsToMany` would need to support parameters for extending the edge type fields, including their resolvers.

Looks like I would need to modify these methods:
- https://github.com/nuwave/lighthouse/blob/master/src/Pagination/PaginationManipulator.php#L52
- https://github.com/nuwave/lighthouse/blob/master/src/Pagination/ConnectionField.php#L40

---

I'm trying to think through how the schema with directives would best work. Since pivot data is only meaningful in a pagination context, I'm thinking it could be a part of the existing directives. Maybe something like:
```graphql
type User {
    id: ID! @globalId
    name: String!
    email: String!
    teams: [Team!]! @belongsToMany(type: "connection", edgeType: "TeamEdge")
}

type TeamEdge {
  node: Team
  cursor: String!
  role: String!
}
```

The edge resolver would loop through the provided edge and any additional fields would be retrieved using `$item->pivot->$extra_field`.

JasonTheAdams:
**Tasks remaining**
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Resolves #869 

**Changes**
This extends the relate directives — specifically `@belongsToMany` — to include an `edgeType` argument. This allows for supplying a custom edge (as opposed to only the generated one). For example:
```graphql
type User implements Node {
    id: ID! @globalId
    teams: [Team!]! @belongsToMany(type: "connection", edgeType: "TeamEdge")
}

type TeamEdge {
    node: Team,
    cursor: String!
    role: String!
}
```
Then, when the edge is being resolved the node and cursor fields are handled as they are now, but any additional field is retrieved as pivot data.

This takes advantage of the relay edge type which is intended to reflect the relationship meta between entities. In the world of Laravel this is exactly what pivot data is. Going this route also allows for supporting further control of the edge with directives and such.

**Breaking changes**

This should not break anything and be fully backwards compatible.

**Further Thoughts**

I'm considering adding an `Edge` interface so custom edges can implement it. It could even be that the interface is explicitly required.


olivernybroe:
I like to solution, I think it makes a ton of sense to do it this way as it also gives the possibility for directives on the pivot fields. 

I don't think the interface is a good idea as it adds limitations. I think good exception messages are better here. You could consider adding two directives called something like `Node` and `Cursor`. These basically just tells us that this is the field we want to use for cursor and node. 
I feel like that would be expressive when reading and give a lot of flexibility regarding naming.

JasonTheAdams:
@olivernybroe Thanks for chiming in with your vote of confidence!

Respectfully, I disagree with the idea of using directives to identify the node and cursor fields for a few reasons:

1. Lighthouse already handles the resolving of these fields which do end up being a node and string. Allowing the dev to set the cursor as, for example, and int would break things and the dev wouldn't have a way of fixing this.
2. This is within the context of relay, and the current structure _is_ the convention expected for relay. If the dev wants to deviate they shouldn't be using the relay schema structure.
3. That's exactly what interfaces are for: extending an object (type) while remaining confident in the contract's implementation.

If anything, I'd be willing to omit an interface, but I'm not sure why. Lighthouse already has a Node interface, so it seems to make sense to add an Edge one as well to fit this scenario.

Please take a look at the "Help Needed" section in the PR description so I can wrap this out. We need this functionality to use Lighthouse so I'm eager to get this merged, if possible.

JasonTheAdams:
I figured out where the models were so I was able to added a passing test. For some reason the `Tests\Unit\Events\ManipulateASTTest::itCanManipulateTheAST` is failing, but that doesn't seem related to anything I'm doing and is putting out the following error:
```
1) Tests\Unit\Events\ManipulateASTTest::itCanManipulateTheAST
ErrorException: Declaration of Illuminate\Foundation\Testing\Assert::assertArraySubset($subset, $array, bool $checkForObjectIdentity = false, string $message = ''): void should be compatible with PHPUnit\Framework\Assert::assertArraySubset($subset, $array, $strict = false, $message = '')
```
Otherwise all the other tests are passing great!

JasonTheAdams:
One thing I'm debating here is whether the `edgeType` directive argument is actually necessary. It would be possible to just detect within `PaginationManipulator::registerConnection` whether the edge type already exists with the given naming convention. Having the argument does allow for it to be explicitly set, however, and have a different type name. 🤔

src/Pagination/ConnectionField.php
```diff
@@ -35,17 +37,35 @@ public function pageInfoResolver(LengthAwarePaginator $paginator): array
...
+                    case 'cursor':
+                        $data['cursor'] = Cursor::encode($firstItem + $index);
+                        break;
+
+                    case 'node':
+                        $data['node'] = $item;
+                        break;
+
+                    default:
+                        $data[$field->name] = $item->pivot->{$field->name};
```
spawnia: Can paginators be used for resolving something else then models? `->pivot` might be undefined

spawnia:
> One thing I'm debating here is whether the `edgeType` directive argument is actually necessary. It would be possible to just detect within `PaginationManipulator::registerConnection` whether the edge type already exists with the given naming convention. Having the argument does allow for it to be explicitly set, however, and have a different type name. 🤔

I wouldn't mind the extra magic, the `TypeNameEdge` naming seems to be well established.

src/Pagination/ConnectionField.php
JasonTheAdams: I was thinking of checking whether the field has a customer resolver directive. If it does, then it skips it, assuming they’ll handle it. Otherwise, the context (inside an edge) is the middle-space between entities, so I think the pivot table makes sense. Do you have any thoughts on how it should be handled when the dev adds a field that doesn’t have pivot data but doesn’t have a resolve directive, either?

JasonTheAdams:
@spawnia Thanks for the feedback and adjustments!

Going the "magic type" method sounds good to me. We'll just have to make sure it's documented. Should I add that to this PR?

That negates the need to check for an invalid type, but I'm still thinking we should make sure it has the expected base structure. What are your thoughts on adding an Edge interface?
```
interface Edge {
  node: Node
  cursor: String!
}

type TeamEdge implements Edge {
  node: Team
  cursor: String!
  role: String!
}
```
We could then make sure that the edge type implements that interface. If not, the way it's built now it wouldn't break anything if they simply omitted the cursor or node, but then it's not proper relay schema anymore. I'm inclined to be more strict since this section of schema is determined by Relay, not Lighthouse.

JasonTheAdams:
Also, with regards to the magic edge type, should we continue to support the `edgeType` directive argument, or cut it out in favor of the naming convention?

src/Pagination/ConnectionField.php
JasonTheAdams: One simple option here would be to check for `$item->pivot` and silently move along if it doesn't exist, assuming they're handling the field resolving.

spawnia:
> Also, with regards to the magic edge type, should we continue to support the `edgeType` directive argument, or cut it out in favor of the naming convention?

Let's allow both. I can see a need for having different edges pointing towards the same type, e.g. `UserAuthorEdge` and `UserSubscriberEdge`.

spawnia:
> What are your thoughts on adding an Edge interface?

Let's enforce that. Can you add docs and link to the relevant part of the Relay spec?

src/Pagination/ConnectionField.php
spawnia: Checking for `pivot` seems sufficient. Differentiating between what goes in the edge and what goes in the node might get complicated if we go for maximum flexibility, so let's see how far we get with that convention.

src/Schema/AST/ASTHelper.php
```diff
@@ -247,4 +247,23 @@ public static function attachNodeInterfaceToObjectType(ObjectTypeDefinitionNode
 
         return $objectType;
     }
+
+    /**
+     * Checks the given type to see whether it implements the given interface.
+     *
+     * @param ObjectTypeDefinitionNode $type
```
spawnia: Please check the formatting of all PHPDocs in this PR. Two spaces, no alignment

tests/Integration/Schema/Directives/BelongsToManyDirectiveTest.php
```diff
@@ -171,6 +176,200 @@ public function itCanQueryBelongsToManyRelayConnection(): void
...
+                    ],
+                ],
+            ],
+        ])->assertJsonCount(2, 'data.user.roles.edges');
+    }
+
+    /**
+     * @test
+     */
+    public function itThrowsExceptionForEdgeTypeNotImplementingEdge()
```
spawnia:
```suggestion
    public function itThrowsExceptionForEdgeTypeNotImplementingEdge(): void
```

src/Schema/AST/ASTBuilder.php
```diff
@@ -328,6 +334,33 @@ interface Node @interface(resolveType: "Nuwave\\\Lighthouse\\\Schema\\\NodeRegis
...
+    {
+        // Only add the Edge interface if a type actually implements it
+        // Otherwise, a validation error is thrown
+        if (! $this->hasTypeImplementingInterface('Edge')) {
+            return;
+        }
+
+        $this->documentAST->setTypeDefinition(
+            PartialParser::interfaceTypeDefinition('
+                "Edge interface"
```
spawnia: Can you expand those descriptions to be at least a sentence?

JasonTheAdams:
@spawnia Lots of work done here. Here's what's done:

1. Added support for magic edge type: `{$type}Edge`
2. Added Edge interface with implementation checking
3. Added checking for whether an item has pivot data, otherwise relies on a custom resolver
4. Added tests for all of this

There is one test that's failing, presently, and that's because in the `ASTBuilder::addNodeSupport` method checks to see whether the Node interface is ever used and then dynamically loads it. But it's not checking other interfaces. So either we need to expand it to check interfaces or just remove it and add Node support every time. Thoughts?

tests/Integration/Schema/Directives/BelongsToManyDirectiveTest.php
```diff
@@ -171,6 +176,200 @@ public function itCanQueryBelongsToManyRelayConnection(): void
...
+        $this->schema = '
+        type User {
+            roles: [Role!]! @belongsToMany(type: "relay", edgeType: "CustomRoleEdge")
+        }
+        
+        type Role {
+            id: Int!
+            name: String!
+        }
+
```
spawnia: No extra new lines

tests/Unit/Schema/AST/ASTHelperTest.php
```diff
@@ -126,4 +126,14 @@ public function itReturnsNullForNonExistingArgumentOnDirective(): void
             ASTHelper::directiveArgValue($directive, 'bar')
         );
     }
+
+    /**
+     * @test
+     */
+    public function itChecksWhetherTypeImplementsInterface()
```
spawnia:
```suggestion
    public function itChecksWhetherTypeImplementsInterface(): void
```

tests/Unit/Schema/AST/ASTHelperTest.php
```diff
@@ -126,4 +126,14 @@ public function itReturnsNullForNonExistingArgumentOnDirective(): void
...
+    {
+        $type = PartialParser::objectTypeDefinition('type CustomEdge implements Edge { foo: String }');
```
spawnia: Always use multiline strings for SDL

JasonTheAdams:
Thanks for the review, @spawnia! Did you happen to see my question regarding loading the Node interface dynamically? https://github.com/nuwave/lighthouse/pull/871#issuecomment-514757000

JasonTheAdams:
Sorry that I have to keep formatting. I think I've finally tweaked PhpStorm to match what StyleCI is looking for.

spawnia:
> Thanks for the review, @spawnia! Did you happen to see my question regarding loading the Node interface dynamically? [#871 (comment)](https://github.com/nuwave/lighthouse/pull/871#issuecomment-514757000)

I don't get what you want to change there, it worked previously?

spawnia:
> Sorry that I have to keep formatting. I think I've finally tweaked PhpStorm to match what StyleCI is looking for.

No worries, i am going to squash the commits anyway.

JasonTheAdams:
The `Tests\Unit\Schema\Directives\PaginateDirectiveTest::itCanChangePaginationAmountArgument` method is failing its test. The reason is that the new `Edge` interface is being added and is referencing the `Node` interface. The `Node` interface, however, only gets added if a **type** is using it. This means it's not loaded because `Edge` is an interface, not a type.

So we have a couple options:
1. Remove the type checking and load the Node interface every time. The same `addNodeSupport` method also adds a node query, which I feel like should be determined by config option, not whether or not the `Node` interface is used, but that may be outside of the scope of this PR.
2. Add interface checking so `Node` gets loaded if _either_ a type or interface are using it.

Make sense?

spawnia:
Makes complete sense, thanks for the breakdown.

I agree that we might improve upon the point you mentioned in 1.

For this PR, fixing 2. seems appropriate.

JasonTheAdams:
Roger that! I went ahead and made a issue to keep track of this. We can discuss a better solution there and make a new PR for that. 👍

spawnia:
I think we should actually remove the `Edge` interface again, see https://github.com/facebook/relay/blob/master/website/spec/Connections.md#node

Unfortunately, there is no way in the SDL to require a field named `node` of the appropriate type. Restricting it to `Node` is too limiting.

The bug relating to not finding the `Node` interface should still be fixed, i like how you factored out that method.

JasonTheAdams:
Ah, I see what you mean, the `node` field can be more than just an object. That is too limiting. Since it's a required part of the spec, what do you think about doing some validation of our own then? I'm thinking a method that inspects the edge and checks for a `node` field that is one of the accepted types, and a `cursor` field that's a non-nullable String.

That said, if they're omitted the way it's built now it will just not attempt to resolve those fields and act as if there's no problem. So the validation wouldn't be fixing a problem as much as enforcing the Relay spec... hmm. What do you think?

spawnia:
> what do you think about doing some validation of our own then?

... i have not thought of that. Brilliant!

That validation could be run when enabling a config option such as `relay.strict`, see https://github.com/nuwave/lighthouse/issues/878#issuecomment-514805616

JasonTheAdams:
@spawnia I removed the `Edge` interface and usage. Everything is working, now. I'm thinking that by this point this PR has served its purpose. The dev can extend the `BelongsToMany` directive with their own edge type which supports retrieving pivot data.

I like the direction you're going in #878. Let's make another PR that adds the configuration and validation. I really like the idea of having a "Relay mode" in Lighthouse so I'd be happy to help build that out.

If you're cool with that let's merge this PR and then create the next PR that moves on from this code.

spawnia:
Nicely done, thanks @JasonTheAdams

Issue #799, opened 2019-05-21
Unable to search by empty string on where constraints 


zwzn:
**Describe the bug**

When I search for an empty string using where constraints it seems to treats it as null.

``` graphql
{
  assets(
    count: 10,
    where: { column: "asset_number", value: "" }
  ) {
    paginatorInfo {
      total
    }
  }
}
```

The query returns the following error "Did not receive a value to match the WhereConstraints for column asset_number."

**Expected behavior**

I Would expect that searching by an empty string would be the same as searching by a non-empty string.

**Schema**

```graphql
extend type Query {
    assets(
        where: WhereConstraints @whereConstraints
    ): [Asset!]! @paginate(type: "paginator" model: "App\\Asset")
}
```

**Output/Logs**

<details><summary>Click to expand</summary>

``` json
{
  "errors": [
    {
      "message": "Did not receive a value to match the WhereConstraints for column asset_number.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "assets"
      ],
      "trace": [
        {
          "file": "/home/adam/Documents/work/aqmdata2/vendor/nuwave/lighthouse/src/Execution/Builder.php",
          "line": 48,
          "call": "Nuwave\\Lighthouse\\WhereConstraints\\WhereConstraintsDirective::handleBuilder(instance of Illuminate\\Database\\Eloquent\\Builder, array(3))"
        },
        {
          "file": "/home/adam/Documents/work/aqmdata2/vendor/nuwave/lighthouse/src/Schema/Directives/PaginateDirective.php",
          "line": 86,
          "call": "Nuwave\\Lighthouse\\Execution\\Builder::apply(instance of Illuminate\\Database\\Eloquent\\Builder, array(3))"
        },
        {
          "file": "/home/adam/Documents/work/aqmdata2/vendor/nuwave/lighthouse/src/Schema/Factories/FieldFactory.php",
          "line": 221,
          "call": "Nuwave\\Lighthouse\\Schema\\Directives\\PaginateDirective::Nuwave\\Lighthouse\\Schema\\Directives\\{closure}(null, array(3), instance of Nuwave\\Lighthouse\\Schema\\Context, instance of GraphQL\\Type\\Definition\\ResolveInfo)"
        },
        {
[...]

PR #872, opened 2019-07-22
Allow querying for null values using `@whereConstraints`

jakesyl:
Null values are completely valid within graphql

- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

#799 (sort of) and https://lighthouse-php.slack.com/archives/CB28A070S/p1563765389011100:

Screen Shot 2019-07-21 at 11 19 37 PM

**Changes**

Allows `WhereDirective` to query for null values

**Breaking changes**
None


olivernybroe:
Nice work! 
I'll say this is breaking change as maybe somebody is used to it failing on null and actually have code that depends on that. 

I however totally agree with the changes to allow null in it.

jakesyl:
@olivernybroe, I see your point, but the failure on null isn't documented behavior. 

Perhaps we can merge this into the next version? Because this isn't documented behavior I'd consider this a bug

spawnia:
I think the code is fine as is, and will allow you to pass `null` as a value. The `column` can not be null, however.

You can prove me wrong with a failing test ;)

jakesyl:
@spawnia on it :)

jakesyl:
Ah @spawnia, Actually I think your misconception is when a passed in variable is null.

spawnia:
Can you please add a test case to this PR that shows the current faulty behaviour and how it is fixed by this change?

spawnia:
Nice job on the failing test. Removing the check altogether is too much though, added in a test to show why it's there.

jakesyl:
Thanks! Missed that one.

spawnia:
Turns out `isset` returns false when the value is `null`. `array_key_exists` does what i intended.

jakesyl:
image

Need to wait for the CI, but showing green on my end now :).

PR #879, opened 2019-07-25
Implement `connect`, `disconnect` and `delete` operations for nested mutations upon MorphTo relationships

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Allows connecting and disconnecting polymorphic associations for a MorphTo relation using nested mutations.

**Changes**

Implement `connect` and `disconnect`

The GraphQL Specification does not support Input Union types,
for now we are limiting this implementation to `connect` and `disconnect` operations.

Let's see which one of the current RFC's makes it and base the implementation of `create` and `update` upon that. https://github.com/graphql/graphql-spec/issues/114

**Breaking changes**

Nope

### DIFF
--- a/docs/master/eloquent/nested-mutations.md
+++ b/docs/master/eloquent/nested-mutations.md
@@ -476,17 +476,13 @@ input UpdateAuthorInput {
 
 ## MorphTo
 
-```graphql
-type Mutation {
-  createHour(input: CreateHourInput! @spread): Hour @create
-}
+__The GraphQL Specification does not support Input Union types,
+for now we are limiting this implementation to `connect` and `disconnect` operations.__
 
-input CreateHourInput {
-  hourable_type: String!
-  hourable_id: Int!
-  from: String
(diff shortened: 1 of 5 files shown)


docs/master/eloquent/nested-mutations.md
```diff
@@ -476,38 +476,68 @@ input UpdateAuthorInput {
...
+
 ```graphql
 mutation {
   createHour(input: {
-    hourable_type: "App\\\Task"
-    hourable_id: 1
     weekday: 2
+    hourable: {
+      connect: {
+        type: "Tests\\\Utils\\\Models\\\Task"
```
olivernybroe: Can we shorten this type to just `Task`?

spawnia: The Laravel default is to use the full namespace, so let's go with `App\\\Models\\\Task`

PR #883, opened 2019-07-27
Fix issue when using the `@model` directive in a type that has a list field 

hosmelq:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

When I used the `@model` directive in a type with a lists field I get the following error.

`ErrorException: Undefined property: GraphQL\Language\AST\ListTypeNode::$name`

I noticed that this even happens if you don’t use the model directive, if you put the ID field after a list field you get the error.

**Changes**

Add an extra validation.

**Breaking changes**

None.


spawnia:
Thanks @hosmelq

PR #885, opened 2019-07-27
Extra plus symbol

vnponce:
Remove an extra '+' symbols from code examples. Maybe it was the purpose but it feels kind of confuse.


spawnia:
image

It is actually there on purpose, as it marks the two lines that need to be added.

I imagine users most probably have some extra code in their base test classes anyways, so it is not like they could just copy-and-paste the entire thing without the `+` signs.

PR #892, opened 2019-08-01
Show spec-, and user-defined client directives correctly in introspection

### LINKED ISSUE #891: Include the @include, and @skip directives that are part of the GraphQL spec
I'm writing a Relay.js based web app currently with Lighthouse and I love it so much. I ran into a problem the other day that could be easily resolved by the use of the @include(if:$someConditionTrue) directive that is described in the [graphql spec](https://graphql.github.io/graphql-spec/June2018/#sec--include), but I realized this is not currently included in lighthouse for some reason. So I did some research and found out that [webonyx (ref here)](https://webonyx.github.io/graphql-php/type-system/schema/) actually has these directives and you can add them in super easily. I am not working off master currently so I wrote a solution that works on 3.7... but I know you made a lot of changes in 4.0 so wrote this without testing it, but it should do the same thing: 

the 3.7 version I tested and it works in SchemaBuilder.php: 
```php 
      $config = SchemaConfig::create()
            // Always set Query since it is required
            ->setQuery(
                $queryType
            )
            ->setDirectives(
                collect(GraphQL::getStandardDirectives())->merge($this->convertDirectives($documentAST))->all() // just replacing the old code with a marge.  
            );
```

the 4.x version that I did not test, but it should work the same:
```php
        // There is no way to resolve client directives lazily,
        // so we convert them eagerly
        $clientDirectives = [];
        foreach ($documentAST->directives as $directiveDefinition) {
            $clientDirectives [] = $this->clientDirectiveFactory->handle($directiveDefinition);
        }
        $config->setDirectives(array_merge(GraphQL::getStandardDirectives(), $clientDirectives));

        return new Schema($config);
```

Thanks for your time, and the awesome project. 👍

alexwhb:
- [x] Added or updated tests
- [x] Updated the changelog 

Resolves #891

**Changes**

Adds `@include` and `@skip` directives form webonyx. 

> @spawnia: Make the `@include` and `@skip` directives that are part of the GraphQL spec show up in introspection and fix handling of default values in custom client directives

**Breaking changes**

should not break anything. 

> @spawnia: No.


docs/master/api-reference/directives.md
```diff
@@ -1074,6 +1074,39 @@ directive @in(
 ) on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
 ```
 
+## @include
+
+This directive is part of the [GraphQL spec](https://graphql.github.io/graphql-spec/June2018/#sec--include)
```
spawnia: Let's not have this directive listed here and keep this reference limited to schema directives. We should find another place for your writeup though, it is quite good.

src/Schema/SchemaBuilder.php
```diff
@@ -78,7 +79,7 @@ public function build(DocumentAST $documentAST): Schema
         foreach ($documentAST->directives as $directiveDefinition) {
             $clientDirectives [] = $this->clientDirectiveFactory->handle($directiveDefinition);
         }
-        $config->setDirectives($clientDirectives);
+        $config->setDirectives(array_merge(GraphQL::getStandardDirectives(), $clientDirectives));
```
spawnia: Let's split this into multiple lines.

spawnia:
Can you add a test case that ensures the directives are present within introspection?

src/Schema/SchemaBuilder.php
alexwhb: @spawnia something like this? 
```php       
$allDirectives = array_merge(GraphQL::getStandardDirectives(), $clientDirectives);
$config->setDirectives($allDirectives);
```

docs/master/api-reference/directives.md
alexwhb: @spawnia That makes sense. Do you have any ideas on where we should put it? I'd be happy to add that to my PR.

spawnia: `Digging Deeper/Client Directives` seems like a good location. We might add a guide on how to add custom client directives as well.

alexwhb: @spawnia Perfect. I'll try to have that added sometime in the next day or so. That would be really cool to have some docs on client directives. I actually did not realize that was even possible until I came across these.

alexwhb:
@spawnia So I was attempting to write a test like you requested for the introspection, but I ran into a problem... it turns out webonyx has this strange line of code: 
```php
    public function getDirectives()
    {
        return $this->config->directives ?: GraphQL::getStandardDirectives();
    }
``` 
which effectively makes it so that I need to somehow add a directive real of fake into lighhouse to make the test fail first. is there any easy way to do that? 

this is the test that I cam up with. I think it will do the trick assuming there's a way to put in a directive so it fails first: 
```php
    /**
     * @test
     */
    public function itFindsWebnoxDirectives(){
        $this->introspect('
        type Foo {
            bar: Int
        }        
        
        '.$this->placeholderQuery()
        );

        $this->assertTrue(
            $this->isDirectivePresent('include')
        );

        $this->assertTrue(
            $this->isDirectivePresent('skip')
        );

        $this->assertTrue(
            $this->isDirectivePresent('deprecated')
        );
    }

    protected function isDirectivePresent(string $directiveName): bool {
        return (new Collection($this->introspectionResult->jsonGet('data.__schema.directives')))
            ->contains(function (array $directive) use ($directiveName): bool {
                return $directive['name'] === $directiveName;
            });
    }
```

spawnia:
>  is there any easy way to add a fake or real directive to this test?

Check https://github.com/nuwave/lighthouse/blob/master/tests/Unit/Schema/ClientDirectiveTest.php

It should give you an idea of what to add to the docs.

docs/master/api-reference/directives.md
spawnia: Done it

PR #896, opened 2019-08-05
Upgrade to vuepress 1.0.3

robsontenorio:
@spawnia  No luck :(

Just have noticed in the past someone has ejected default theme in order to work on a custom theme.

Most related to https://vuepress.vuejs.org/default-theme-config/#ejecting

Options: 
- Wait next stable release from Vuepress
- Rework docs and sticky to default theme, with minimal customizations..

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -2,7 +2,7 @@
     "license": "MIT",
     "dependencies": {
         "fs-extra": "^7.0.1",
-        "vuepress": "^1.0.0-alpha.22",
+        "vuepress": "^1.0.3",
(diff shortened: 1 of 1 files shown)


robsontenorio:
@spawnia @chrissm79 
If you have admin permission on Netlify Lighthouse`s page you can give a try to changing to NODE 10 image. Currently it is using node 8 on building.

Settings > Build & deploy > Continuous Deployment > Build image selection

spawnia:
@robsontenorio thanks for looking into this. I contacted @chrissm79 to try Node 10.

Can you check if the default theme does indeed work?

PR #898, opened 2019-08-06
Allow for callable classes for resolvers in directives

### LINKED ISSUE #882: Allow for callable classes for resolvers in directives

To keep code clean, we're implementing a single class per resolver. To do that, we use the `__invoke` method so that the class is callable. Right now, to define such a class as a resolver, we need to do this.

```graphql
type Mutation {
    updateUser(
        id: ID!
        email: String
        name: String
    ): Boolean @field(resolver: "UpdateUserMutation@__invoke")
}
```

But there's no reason why it couldn't be just this:

```graphql
type Mutation {
    updateUser(
        id: ID!
        email: String
        name: String
    ): Boolean @field(resolver: "UpdateUserMutation")
}
```

The simplest fix, as I see it, is in the `getMethodArgumentParts` method of the `BaseDirective` class.

```php
protected function getMethodArgumentParts(string $argumentName): array
{
    $argumentParts = explode(
        '@',
        $this->directiveArgValue($argumentName)
    );

    if (
        count($argumentParts) !== 2
        || empty($argumentParts[0])
        || empty($argumentParts[1])
    ) {
        throw new DirectiveException(
            "Directive '{$this->name()}' must have an argument '{$argumentName}' in the form 'ClassName@methodName'"
        );
    }

    return $argumentParts;
}
```

You could just automatically put the `__invoke` if there's no method name in `argumentParts` array like this:

```php
protected function getMethodArgumentParts(string $argumentName): array
{
    $argumentParts = explode(
        '@',
        $this->directiveArgValue($argumentName)
    );

    if (
        empty($argumentParts)
        || count($argumentParts) > 2
        || empty($argumentParts[0])
    ) {
        throw new DirectiveException(
            "Directive '{$this->name()}' must have an argument '{$argumentName}' in the form 'ClassName@methodName' or 'ClassName'"
        );
    }
    
    if (empty($argumentParts[1])) {
        $argumentParts[1] = '__invoke';
    }

    return $argumentParts;
}
```

This fix would allow the rest of the code to behave as expected since it always assumes that `getMethodArgumentParts` returns an array with two values in it.

carlalexander:
This is the PR that fixes #882. 

Instead of adding the `__invoke` method in the `getMethodArgumentParts` method, I moved it to the `Utils` class. I made `methodName` nullable and, if it's `null`, I set it to `__invoke`. We then check if the method name exists.


src/Support/Utils.php
```diff
@@ -48,8 +48,12 @@ public static function namespaceClassname(string $classCandidate, array $namespa
      *
      * @throws \Nuwave\Lighthouse\Exceptions\DefinitionException
      */
-    public static function constructResolver(string $className, string $methodName): Closure
+    public static function constructResolver(string $className, ?string $methodName): Closure
     {
+        if (null === $methodName) {
```
spawnia: Can we just use a default argument here?

spawnia:
The tests are failing.

Can you please stick to the PR template and make sure tests, docs and the changelog are updated?

src/Support/Utils.php
carlalexander: Not right now, no. The `getMethodArgumentPart` method returns an array which you expand right away. So everywhere `methodName` will be `null` when passed to the `constructResolver` method. 

We'd have to rework all the code around that to be able to use a default argument.

carlalexander:
Sorry about that. Working on tests right now. 😃

src/Support/Utils.php
carlalexander: This change wasn't working and wasn't relevant in the end. The original change worked fine for detecting errors.

carlalexander:
I've reworked the PR and added/reworked some of the tests. Do you want all the directives that use resolvers to have tests for it?

What do you want the new text in the docs to be instead of this:
>Consists of two parts: a class name and a method name, separated by an `@` symbol.

src/Schema/Directives/BaseDirective.php
```diff
@@ -181,15 +181,18 @@ protected function getMethodArgumentParts(string $argumentName): array
...
-            || empty($argumentParts[1])
         ) {
             throw new DirectiveException(
-                "Directive '{$this->name()}' must have an argument '{$argumentName}' in the form 'ClassName@methodName'"
+                "Directive '{$this->name()}' must have an argument '{$argumentName}' in the form 'ClassName@methodName' or 'ClassName'"
             );
         }
 
+        if (empty($argumentParts[1])) {
+            $argumentParts[1] = '__invoke';
```
spawnia: Nicely done!

spawnia:
I worked through the docs and changed the necessary parts.

Actually, i like this style so much i decided to expand on it a bit and make it an option for any kind of default resolver location. I want to keep compatibility for now, but in `v5` we can make it the new default.

carlalexander:
Awesome! Glad to help 😄

PR #899, opened 2019-08-06
Compose complex input arguments through nested arg resolvers

### LINKED ISSUE #725: Add a composable, extendable mechanism for nested mutations

As #692 and #691 have shown, the way that nested mutations are structured is not always the most elegant for particular use cases.

While it does have the basic building blocks to deal with any kind of relation,
the way the `MutationExecutor` is structured makes it so the implementation is set in stone.

It would be nice if there was a simple way to extend the resolution process
and to offer an abstraction that makes it easier to compose nested mutations.

Similar to how the nested field resolution process works, we might be able to attach resolvers to specific arguments/input objects.

### LINKED ISSUE #1038: Allow `update` related models on create mutation

In my project I have one-to-one relation `Candidate`-`Person`. They linked via Person with an optional candidate_id column, so if it set, then person is a candidate.

Now I was implementing mutation for creating candidate. One of the situations - create new candidate with an existing person, so create candidate and link it to person.

I am using `@create` directive, so here is small excerpt of definitions:

```graphql
extend type Mutation {
    createCandidate(input: CreateCandidateInput! @spread): Candidate @create
}

input CreateCandidateInput {
    person: InputPersonRelation
}

input InputPersonRelation {
    create: CreatePersonInput
    update: UpdatePersonInput
}
``` 

Then I tried scenario described above:

```
mutation{
  createCandidate(
    input:{
    email: "test@asd.de",
    person:{
      update: {
        id:838891795,
        firstname:"Tester",
        lastname:"Son"
      }
    }
  }) {
    id
    email
    person {
      id
      firstname
      lastname
    }
  }
}
```

Result - new candidate was created, but it wasn't linked to person. Person input was simply ignored.

Then I looked into code and noticed, that `update` properties are recognised and executed ONLY on update mutation executor.  

Actually there are two possible solutions for my problem:

1. (What I was thinking of) - allow `update` in such case. The change is not hard - we should simply copy code that is executed if `update` property is defined from `MutationExecutor::executeUpdateWithLoadedModel()` to `MutationExecutor::executeCreate()`.
2. I have to redefine my mutation with changing the side of view. I would have to start with person. Then my situation will not take place. If Person exists, then I would simply update person and it would work.
3. (Just noticed...) use `upsert` function of `update`/`insert`.

Anyway I wanted to discuss it with you. Do you think this is a good thing to be able to update relations why creation of model?

Do I need  insert/update at all, if we upsert is possible?

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Resolves #725
Resolves #1038

**Changes**

The basic idea behind this is to allow complex nested input to be resolved in a composable fashion, similar to how field resolvers work.

Field arguments may have a resolver attached to them which gets then called with the corresponding slice of the given inputs. It runs after the parent resolver.

Added the capability to run the following directives as nested argument resolvers:

- [x] `@create`
- [x] `@update`
- [x] `@upsert`
- [x] `@delete`

This also adds the `@nest` argument resolver, which is basically a no-op resolver, used to contain multiple other nested argument resolvers in a single input object.

**Breaking changes**

No, but let's test this one thoroughly before releasing it.

The `MutationExecutor` was split up into separate nested argument resolvers under the hood, but works just like before for application developers.


src/Execution/Arguments/ArgPartitioner.php
```diff
@@ -0,0 +1,190 @@
...
+     */
+    public static function nestedArgumentResolvers(ArgumentSet $argumentSet, $root): array
+    {
+        $model = $root instanceof Model
+            ? new \ReflectionClass($root)
+            : null;
+
+        return static::partition(
+            $argumentSet,
+            static function (string $name, Argument $argument) use ($model): bool {
```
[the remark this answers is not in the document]
spawnia: Done, i split it up into two steps.

src/Execution/Arguments/ArgPartitioner.php
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * Returns an array of two new ArgumentSet instances:
+     * - the first one contains all arguments for which the predicate did not match
+     * - the second one contains all arguments for which the predicate matched
+     *
+     * @param  \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet  $argumentSet
+     * @param  \Closure  $predicate
+     * @return \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet[]
+     */
+    protected static function partition(ArgumentSet $argumentSet, \Closure $predicate)
```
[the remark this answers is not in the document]
spawnia: I did make the naming more general and switched the order.

src/Execution/Arguments/ArgResolver.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function __invoke($root, ArgumentSet $args)
+    {
+        [$regularArgs, $nestedArgs] = ArgPartitioner::nestedArgumentResolvers($args, $root);
+
+        $result = ($this->previous)($root, $regularArgs);
+
+        /** @var \Nuwave\Lighthouse\Execution\Arguments\Argument $nested */
+        foreach ($nestedArgs->arguments as $nested) {
+            ($nested->resolver)($result, $nested->value);
```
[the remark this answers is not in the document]
spawnia: That makes it less predictable for the next nested resolver. I think they should all be passed the same exact `$root`, the result of the parent.

spawnia:
@joskfg added the actual implementation that allows using the standard mutation directives as nested argument resolvers.

What do you think of the design and the API?

joskfg:
I hope to finish today the review and fully understand the behaviour.

src/Execution/Arguments/ArgPartitioner.php
```diff
@@ -0,0 +1,204 @@
...
+     * @param  \ReflectionClass|null  $model
+     * @return void
+     */
+    protected static function attachNestedArgumentResolver(string $name, Argument &$argument, ?ReflectionClass $model): void
+    {
+        $resolverDirective = $argument->directives->first(function (Directive $directive): bool {
+            return $directive instanceof ArgumentResolver;
+        });
+
+        if ($resolverDirective) {
```
joskfg: What do you think about to add a ArgumentMiddleware that allows you to execute previous actions before the resolver?

I think this should be easy to implement just checking it here and wrapping the `$argument->resolver`. Actually you are already using the middleware pattern here with thinks like `new ResolveNested(new UpsertModel(new SaveModel($relation)))`. It could/should be another PR, ofc.

src/Execution/Arguments/ArgPartitioner.php
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * Partition the arguments into nested and regular.
+     *
+     * @param  \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet  $argumentSet
+     * @param  mixed  $root
+     * @return \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet[]
+     */
+    public static function nestedArgumentResolvers(ArgumentSet $argumentSet, $root): array
+    {
+        $model = $root instanceof Model
```
joskfg: This is not `$model`, it should be `$modelReflection`

src/Execution/Arguments/ArgPartitioner.php
spawnia: Hey @joskfg i am finally coming around to finishing up this PR. Your idea is a good one, thanks. We can do it in another PR.

src/Execution/Arguments/ArgPartitioner.php
spawnia: Renamed.

spawnia:
After dog-fooding this PR for a few weeks and giving it some final polish, this feature is finally ready. I am excited for people to try it, it has allowed us to compose functionality and DRY up a lot of extra code.

PR #901, opened 2019-08-07
[docs] Upgrade vuepress to v1.0.3

robsontenorio:
@spawnia @chrissm79 

It works even on node 8. Just changed approach of theme cutomization.


spawnia:
Thanks so much @robsontenorio, great work!

PR #904, opened 2019-08-08
[docs] Fix route from home page

robsontenorio:
Add missing route for Eloquent section from home page.


spawnia:
Nice find, thanks!

PR #913, opened 2019-08-12
Call out @model in @node directive in documentations

### LINKED ISSUE #888: Add model to @node directive to simplify resolving
Presently when using the `@node` directive it's necessary to specify a resolver to get the object. This is pretty redundant, however, as most resolvers will simply be `return Model::where('id', '=', $id)->first()`. It would be nice to just specify a model for the 90% use-case.

Let's add a `model: String` argument to the directive so the dev may simply do `@node(model: "App\\User")` to resolve the node. We could even add a `key: String` argument so they can specify which column the where condition should compare against.

The most obvious alternative is to just make a resolver, but that's really the only present solution.

JasonTheAdams:
This resolves #888 by directing the reader to check out the `@model` directive when viewing the `@node` directive.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -1392,6 +1392,8 @@ decoded `id` and resolves to a result.
 function resolveUser($id): \App\User
 ```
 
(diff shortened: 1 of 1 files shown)


spawnia:
Thanks!

Issue #897, opened 2019-08-05
Mutation: validation fails but the mutation still executes


lucacri:
**Describe the bug**

When running a simple mutation that has a field with `@rules`, the validation fails and it returns an error, but the mutation executes anyway.

**Expected behavior/Solution**

According to the docs, "In the case of a validation error, Lighthouse will abort execution and return the validation messages as part of the response.". This is not the case right now

**Steps to reproduce**

1. Create a simple mutation 
```graphql
type Mutation{
TestMutation(message: String @rules(apply: ["required", "min:4"]))
}
```
2. Create the `TestMutation.php`, and just add a `\Log::info("executed")`
3. Run the mutation via the `/graphql` endpoint
4. The log will contain "executed" => wrong

**Environment**

Lighthouse Version: v4.0-beta.1
Laravel Version: 5.8.27

spawnia:
Thanks for the nice and detailed report.

Can you add a PR with a failing test case? That would be a great first step towards fixing this.

emielmolenaar:
Same here. Using Laravel 5.8 and Lighthouse v4.0-beta.2. I was working on a PR so created a simple test on my local machine with my schema. Weird thing is, the test does not fail and the fields are validated properly :smile: 

When running the same mutation via the GraphQL playground, the validation does not work correctly and the mutation gets executed.

I dug around in the validation logic, it has something to do with this:

https://github.com/nuwave/lighthouse/blob/3.x/src/Schema/Factories/FieldFactory.php#L476

Being moved to

https://github.com/nuwave/lighthouse/blob/master/src/Schema/Factories/FieldFactory.php#L234

That happened in https://github.com/nuwave/lighthouse/commit/01ba5157ef701bdbca2054a54f1dfa53c82f1e93 .

If I add...

```
$path = implode(
    '.',
    $this->resolveInfo()->path
);
$this->validationErrorBuffer->flush(
    "Validation failed for the field [$path]."
);
```

.. back to the `validateArgs()` method, everything is working as expected. I can not figure out why the tests are not failing though, the `Closure` given to `setResolver` in https://github.com/nuwave/lighthouse/blob/master/src/Schema/Factories/FieldFactory.php#L176 does not seem to be called outside tests for some reason?

PR #914, opened 2019-08-13
 Ensure validation actually runs before the final resolver

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

Resolves https://github.com/nuwave/lighthouse/issues/897

**Changes**

Ensure validation actually runs before the final resolver

I skipped the test `itCombinesFieldValidationAndArgumentValidation` for now, as it would require a larger rework of how we run validation.

**Breaking changes**

This ensures that `v4` does not cause regressions.

### DIFF
--- a/src/Exceptions/ValidationException.php
+++ b/src/Exceptions/ValidationException.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Nuwave\Lighthouse\Exceptions;
+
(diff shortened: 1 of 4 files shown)


spawnia:
Thanks to @lucacri and @emielmolenaar for bringing up this critical issue!

PR #915, opened 2019-08-15
Fix validators not having access to the given args for the query or mutation

megawubs:
- [x] Added or updated tests
- [x] Updated the changelog

Fixes `$this->args` being `null` inside the `rules()` method of validators extending `ValidationDirective`
**Changes**
Adds $this->setArgs($args) to the `ValidationDirective`. I've updated and added a test to ensure this works.

The test that tested if the `$this->args` was set, did not actually test it. It only tested if the unique rule was being applied. That it did, but it tried to update a different record, which has no effect on the value for the `->ignore()` part. In order to actually test this, I wrote a test that updated the same entity with the same name. Now, it started to complain that it was not unique. This got me to look into to the actual generated rule string, and it turns out the id value was `NULL`, instead of the value we wanted it to be.
image

After this fix, the rule has the expected values
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed the `ValidationDirective` not setting the mutation or query arguments to itself 
+- Fixed the `ValidationDirective` not setting the mutation or query arguments to itself https://github.com/nuwave/lighthouse/pull/915
 
 ## [4.0.0](https://github.com/nuwave/lighthouse/compare/v3.7.0...v4.0.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v3.7.0...master)
 
+### Fixed
+- Fixed the `ValidationDirective` not setting the mutation or query arguments to itself 
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v3.7.0...master)
 
+### Fixed
+- Fixed the `ValidationDirective` not setting the mutation or query arguments to itself
```
spawnia: missing a newline

tests/Integration/ValidationTest.php
```diff
@@ -427,14 +427,70 @@ public function itSetsArgumentsOnCustomValidationDirective(): void
 
         $this->assertSame(
             [
-                'input.name' => [
+                'name' => [
```
spawnia: Optimally, the error would still contain the original path to the argument.

tests/Integration/ValidationTest.php
```diff
@@ -427,14 +427,70 @@ public function itSetsArgumentsOnCustomValidationDirective(): void
...
+        }
+        
+        type User {
+            id: ID
+            name: String
+        }
+        '.$this->placeholderQuery();
+
+        factory(User::class)->create([
+                                         'name' => 'foo',
```
spawnia: Indent less

tests/Integration/ValidationTest.php
megawubs: This is because the tests uses `@spread`

spawnia: Right, i just noticed that. We don't have to fix it in this PR.

PR #919, opened 2019-08-18
Add @whereJsonContains directive

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Related Issue/Intent**

While the `@where` directive offers flexible clauses, it did not work for a `whereJsonContains` query, as the method does not take a second argument `operator`.

**Changes**

Use an input value as a [whereJsonContains filter](https://laravel.com/docs/queries#json-where-clauses).

```graphql
type Query {
    posts(tags: [String]! @whereJsonContains): [Post!]! @all
}
```

You may use the `key` argument to look into the JSON content:

```graphql
type Query {
    posts(tags: [String]! @whereJsonContains(key: "tags->recent")): [Post!]! @all
}
```

### Definition

```graphql
"""
Use an input value as a [whereJsonContains filter](https://laravel.com/docs/queries#json-where-clauses).
"""
directive @whereJsonContains(
  """
  Specify the database column and path inside the JSON to compare. 
  Only required if database column has a different name than the attribute in your schema.
  """
  key: String
) on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
```

**Breaking changes**

Nope

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 
+services:
+  - mysql
+
 cache:
   directories:
     - $HOME/.composer/cache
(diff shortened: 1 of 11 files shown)


spawnia:
Working on getting our test setup running with MySQL.

PR #930, opened 2019-08-23
Remove Lighthouse Utils package from plugins list

maarten00:
Package is no longer maintained


spawnia:
Hey, thanks for the update.

PR #949, opened 2019-09-04
[WIP] Fix model detecting

### LINKED ISSUE #550: @paginate don't work with @can directive
Hi,
when try:
```users: [User!]! @paginate @can(if:"view any user")``` 

I have error:
```No class 'UserPaginator' was found for directive 'can'```

When I delete @can, query works fine. My @can directive is fine, works with other queries and mutations. I expect error when I don't have permissions, and paginated data when permissions is granted.

**Schema**

```graphql
extend type Query {
    "Retrieve all users"
    users: [User!]! @paginate @can(if:"view any user")
}
```

**Output/Logs**

```
No class 'UserPaginator' was found for directive 'can'
```

</details></br>

**Environment**

Lighthouse Version: 2.6.3
Laravel Version: 5.7
PHP Version: 7.1.9

**Additional context**
I use https://github.com/spatie/laravel-permission to manage privileges.

lorado:
- [x] Added or updated tests
- [ ] Updated the changelog

**Related Issue/Intent**

Should resolve #550
 
There is some problem on detecting models, when field type was changed by FieldManipulator directives. We should to rework, how model detection works. I just added a test that currently fails.

I have one Idea how we could probably solve it. May be we should store "original" types, that are defined before FieldManipulator execution. I don't know if it is even possible... 

Or may be we could create an array, that maps before and after types. Each FieldManipulator should then provide before and after types, and store these types into that mapping array. When we then try to resolve a Field type, we would also look into this array, and if there is some entry for discovered type, then change discovered type, and only after this step - search for model class in namespace.

Your ideas?


spawnia:
Hey @lorado can you branch off from the Lighthouse repo directly and not your fork? That has the added bonus of adding the codestyle fixes automatically.

spawnia:
> I have one Idea how we could probably solve it. May be we should store "original" types, that are defined before FieldManipulator execution. I don't know if it is even possible...

We can attach a directive to the intermediary generated definition, or manipulate the `@paginate` directive to have an explicit `model` argument.

spawnia:
Superseded by #974, the issue is now fixed.

Thanks @lorado for getting the ball rolling on this important bugfix.

PR #950, opened 2019-09-05
Register extension directives only through RegisterDirectiveNamespaces event

lorado:
**Changes**

Change the way, how experimental `@whereConstraints` directive is registered. Earlier it was added to directive factory. Thus `@whereConstraints` was not recognised by IDE Helper.

Now directive is registered via adding its namespace to directives namespaces array. Thus IDE helper has additional namespace, when `WhereConstraintsServiceProvider` is enabled.  

I also fixed deprecated notice for `assertArraySubset`. This assertions seems to be deleted in PHPUnit 9, as there were some misunderstood usages over the time. Author gave no suggestions how to replace it, and people just exported this functionality to an extra package. 

**Breaking changes**

No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v4.3.0...master)
 
+### Deprecated
+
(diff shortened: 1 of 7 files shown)


lorado:
hmmm didn't see this package for assertion requires PHP 7.2+ =\

What can we do?

spawnia:
> hmmm didn't see this package for assertion requires PHP 7.2+ =\
> 
> It also freezes tests in some environments. Ok... I remove package

Yeah, i guess we have to live with the deprecation notice for now.

PR #951, opened 2019-09-05
Allow additional route configurations `prefix` and `domain`

ddedic:
- [x] Updated the changelog
- [x] Added or updated tests

**Changes**

Adds support for graphql route group, so we can use router group specific config like domain, prefix, etc.

**Breaking changes**

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Wrap route in group to allow additional configurations such as domain https://github.com/nuwave/lighthouse/pull/951
+- Allow additional route configurations `prefix` and `domain` https://github.com/nuwave/lighthouse/pull/951
 
 ### Deprecated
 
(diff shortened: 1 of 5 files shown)


spawnia:
Can you give a code example of what you are now able to do that was not possible before?

ddedic:
For example, I´m running GraphQL server on it´s own subdomain, and this allows me to assign domain to route group:
`
    'route' => [
        /*
         * The URI the endpoint responds to, e.g. mydomain.com/graphql.
         */
        'uri' => '/',

        /*
         * Lighthouse creates a named route for convenient URL generation and redirects.
         */
        'name' => 'graphql',

        /*
         * Additional options.
         */
        'prefix' => '',
        'domain' => 'graphql.' . env('APP_DOMAIN', 'localhost'),
        'middleware' => [
            \Nuwave\Lighthouse\Support\Http\Middleware\AcceptJson::class,
        ]
    ],
`

spawnia:
Actually, i just realized that we do not need a route group. Changed the PR, can you check if it still works for you? @ddedic

ddedic:
@spawnia  awesome, works as expected!

spawnia:
Thanks for the contribution!

Issue #772, opened 2019-05-06
Avoid Collection for performance gains


spawnia:
```
+--------------------+-------------+--------------+-------+
| subject            | mem_peak    | mean         | diff  |
+--------------------+-------------+--------------+-------+
| foreach            | 13,327,488b | 5,226.149μs  | 1.00x |
| array_map          | 13,327,808b | 5,556.549μs  | 1.06x |
| Collection->map    | 22,000,512b | 12,800.698μs | 2.45x |
+--------------------+-------------+--------------+-------+
```

When we are not using the more advanced features of the Collection class, we should
consider replacing it with a regular `foreach` loop or `array_map` for some easy performance gains.

spawnia:
Closing this, as this is an ongoing work and we should continuously keep it in mind.

When we identify hot paths that need to be optimized, we can always change them as needed.

PR #961, opened 2019-09-12
Performance wins

stayallive:
~Work in progress still!~

Figures below are by no means scientific but measured in a real app with a large schema on a local machine while running in production mode with `php artisan optimize`. Tested and compared with Blackfire.

- e6655df: ~10%

It looks so far like the most gains to be had are in the underlying GraphQL library itself related to the query validator.

image

The relatively small query I am testing with (fieldnames are dummies):

```graphql
query userDiscovery($id: String!) {
  user(id: $id) {
    id
    field1
    field2
    __typename
  }
  viewer {
    id
    tenant {
      id
      settings {
        field1
        field2
        field3
        __typename
      }
      __typename
    }
    field1
    field2
    field3
    __typename
  }
}

```

---

- [x] Added or updated tests
- [ ] Updated the changelog

Related #772

**Changes**

Non-breaking changes that should improve the performance of some code paths resulting in faster response times.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Allow additional route configurations `prefix` and `domain` https://github.com/nuwave/lighthouse/pull/951
 - Enable schema cache only when `APP_ENV` != 'local' https://github.com/nuwave/lighthouse/pull/957
+- Improve performance of `TypeRegistry` resolving fields https://github.com/nuwave/lighthouse/pull/961
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


spawnia:
Looking great so far!

stayallive:
I for now will leave it with this little change which has a significant impact (in my environment at least). Unfortunately the GraphQL is consitently slow even for the smallest of queries and the bulk seems to be in the query validator which is handled upstream.

I'll be back with more later! :)

spawnia:
Can you add a short entry to the Changelog?

spawnia:
Nicely done, thank you!

spawnia:
@stayallive I had to revert this change, as it broke our app.

An issue occurred when running two queries in one request where the first request was using `@spread`. As the same instance of the FieldFactory was used for the second root query, the `$pathsToSpread` remained and wrongly spread the input of the second query.

For the sake of efficiency, we can certainly try to avoid instantiating a new FieldFactory for every field - but we must take care to ensure there is no lingering state in between calls.

stayallive:
Interesting, I thought about that but did not see any issues, obviously missed something here, good one to add a test for sometime :)

PR #972, opened 2019-09-22
composer normalize

spawnia:
Leverage https://github.com/localheinz/composer-normalize

I applied autoformatting and added a check to CI.


spawnia:
Thanks @localheinz i value the consistency this brings.

localheinz:
Thank you, @spawnia! Glad you like it.

PR #973, opened 2019-09-23
Add `\Nuwave\Lighthouse\Execution\DataLoader\BatchLoader#loadMany()` function

balthild:
- [x] Added or updated tests
- [x] Updated the changelog

**Changes**

Adds loadMany() method for BatchLoader


spawnia:
Hey, thanks for your contribution.

How is this supposed to be used? Can you give an example or add something to the docs?

Also, i would like this tested. As of now, there is no guarantee this works or won't suddenly break.

balthild:
Same as the official implementation of dataloader (https://github.com/graphql/dataloader#loadmanykeys). For example, if there's a such schema

```graphql
type User {
  id: ID!
  followedUsers: [User!]! @field(resolver: "User@followedUsers")
}
```

and a query

```graphql
query {
  usersInGroup(groupId: 123) {
    followedUsers {
      id
    }
  }
}
```

Then we might use `$followedUsers = $loader->loadMany($rootValue->followed_users_id)` in the resolver, which is equivalent to

```php
$followedUsers = collect($rootValue->followed_users_id)
  ->map(function ($id) {
    return $loader->load($id);
  })
  ->toArray();
```

spawnia:
That's nice, makes sense.

I am open to including this in Lighthouse, but before that happens we will need 

- tests: You could use a custom resolver that utilizes the function.
- docs: Can you extend this section a bit? https://lighthouse-php.com/master/performance/n-plus-one.html#data-loader
- changelog: that would be just a one-liner to the latest `## Added` section

If you need any help, please let me know!

balthild:
OK, I'll add them later.

PR #974, opened 2019-09-23
Fix default model detection when using other directives combination with `@paginate`

### LINKED ISSUE #550: @paginate don't work with @can directive

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

Resolves #550
Supersedes #949

This now works:

```graphql
type Query {
  users: [User!]!
    @can(ability: "adminOnly")
    @paginate
}
```

When creating the intermediary pagination type the original class of the model is attached through the new `@modelClass` directive.

Since `@modelClass` should be able to replace pretty much all instances where a custom `model` argument was needed, i decided to deprecate the current `@model` directive.

I modified `@node` so it now also handles what `@model` did before and users have an easy migration path.

```diff
-type User @model {
+type User @node {
  id: ID
}
```

The plan is to rename `@modelClass` to `@model` in v5 and make it the new standard for mapping types to models. Let's hold off on pushing it onto users for now.

No. Until v5, `@model` will function as before. When using `@node`, a custom `resolve` argument will take precedence over the new default model resolution that was added.


src/GraphQL.php
```diff
@@ -249,23 +242,11 @@ protected function getValidationRules(): array
     /**
      * Get instance of DocumentAST.
      *
+     * @deprecated use ASTBuilder instead
      * @return \Nuwave\Lighthouse\Schema\AST\DocumentAST
      */
     public function documentAST(): DocumentAST
```
lorado: Do people use this function outside of lighthouse?
Anyway it makes sense to mark it as deprecated first, what about removing in v5?

spawnia: Yes, depreciations will arrive in the next major by default.

spawnia: It might be used for schema manipulation/introspection - for example, in this PR we need to access the DocumentAST from BaseDirective.

src/Schema/Factories/DirectiveFactory.php
```diff
@@ -150,6 +149,7 @@ public function setResolved(string $directiveName, string $className): self
     }
 
     /**
+     * @deprecated
```
lorado: What is this for, and why it got deprecated?

lorado:
Ooooohhhh I just got it... you created a directive, that maps a WHOLE TYPE to a class, not just a return type of a field. Dude, this is really awesome! :D

Should we maybe collect somewhere changes, that has to be done in v5? like renaming `@modelClass` to `@model`, removing `model` from all directives and ignoring it on resolving the class?

tests/Unit/Schema/Directives/BaseDirectiveTest.php
```diff
@@ -22,8 +23,30 @@
...
+        type User @modelClass(class: "Team") {
+            id: ID
+        }
+        ';
+
+        $directive = $this->constructFieldDirective('foo: User @dummy');
+
+        $this->assertSame(
+            Team::class,
+            $directive->getModelClass()
```
lorado: `getModelClass()` is a protected method, but called here from outside. Test works, but IDE claims, that protected method was called. Should we ignore it? Or maybe change to public?

src/Schema/Factories/DirectiveFactory.php
spawnia: I don't know, that is exactly why i deprecated it.

spawnia:
> Should we maybe collect somewhere changes, that has to be done in v5?

What i usually do is to just go through and look for all the places that have something marked as deprecated

> like renaming @modelClass to @model

Plan to do that, yes.

> removing model from all directives and ignoring it on resolving the class?

I think that will have to wait until `v6`. I want to allow for a gradual migration, straight from the `model` argument to `@model`. As the renaming is pending, i would like to wait until `v5` to even suggest that change.

tests/Unit/Schema/Directives/BaseDirectiveTest.php
spawnia: I would rather keep it protected.

There is some magic in the test that allows calling protected methods: https://github.com/nuwave/lighthouse/blob/e6cbda2309b5e74b87fa4ef1c1311529bdd0c049/tests/Unit/Schema/Directives/BaseDirectiveTest.php#L173

As the BaseDirective is such a common class to extend, i think its internal API should be kept stable.

lorado: Ok nice.

lorado:
Ok fine, then let's merge it. **YEY**

spawnia:
> There's a lot of extra refactoring that got snuck in this PR (e.g. changing how the tests work)

True, thanks for digging through. I tend to go a bit overboard with the boy scout rule.

> The only thing that stands out to me is the usage of the DocumentAST class.

Let's continue that discussion in #986

PR #984, opened 2019-10-02
[Feat] Added @count directive

olivernybroe:
- [x] Added or updated tests
- [x] Updated the changelog

You could do something close to this by using the paginator, however then you are forced to fetch at least one model, and also you are exposing data your might not wanna expose. This directive solves this by just having a simple count directive.

**Changes**

Adds a new directive for getting the count of a relation.

**Breaking changes**

New directive added, so shouldn't be breaking.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -579,6 +579,30 @@ class ComplexityAnalyzer {
     }
 ```
 
+## Count
+
+Returns the count of a given relationship
+
+```
+type User  {
+    id: ID!
+    likes: Int! @count(relation: "likes")
+}
+```
+### Definition
+
+```graphql
+"""
+Resolve the field through a count.
+"""
+directive @count(
+  """
+  The relationship which you want to run the count on.
+  """
+  relation: String!
+) on FIELD_DEFINITION
+```
+
 ## @create
 
 Create a new Eloquent model with the given arguments.
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -581,7 +581,7 @@ class ComplexityAnalyzer {
 
 ## Count
 
-Returns the count of a given relationship or a model
+Returns the count of a given relationship or model.
 
 ```graphql
 type User  {
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -581,12 +581,18 @@ class ComplexityAnalyzer {
 
 ## Count
 
-Returns the count of a given relationship
+Returns the count of a given relationship or a model
 
-```
+```graphql
 type User  {
     id: ID!
     likes: Int! @count(relation: "likes")
+}
+```
+
+```graphql
+type Query {
+    categories: Int! @count(model: "Category")
 }
 ```
 ### Definition
@@ -599,7 +605,12 @@ directive @count(
   """
   The relationship which you want to run the count on.
   """
-  relation: String!
+  relation: String
+  
+  """
+  The model to run the count on.
+  """
+  model: String
 ) on FIELD_DEFINITION
 ```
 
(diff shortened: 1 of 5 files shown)


spawnia:
I think this should work on root query fields too.

```graphql
type Query {
  postsCount: Int! @count(model: "Post")
}
```

olivernybroe:
@spawnia Ah, great idea, didn't have that use case myself, but makes sense.

docs/master/api-reference/directives.md
```diff
@@ -579,6 +579,41 @@ class ComplexityAnalyzer {
     }
 ```
 
+## Count
```
spawnia:
```suggestion
## @count
```

docs/master/api-reference/directives.md
```diff
@@ -579,6 +579,41 @@ class ComplexityAnalyzer {
...
+
+Returns the count of a given relationship or a model
```
spawnia:
```suggestion
Returns the count of a given relationship or model.
```

docs/master/api-reference/directives.md
```diff
@@ -579,6 +579,41 @@ class ComplexityAnalyzer {
...
+```graphql
+type Query {
+    categories: Int! @count(model: "Category")
+}
+```
+### Definition
+
+```graphql
+"""
+Resolve the field through a count.
```
spawnia: Returns the count of a given relationship or model.

src/Schema/Directives/CountDirective.php
```diff
@@ -0,0 +1,70 @@
...
+    public function name()
+    {
+        return 'count';
+    }
+
+    public static function definition(): string
+    {
+        return /* @lang GraphQL */ <<<'SDL'
+"""
+Resolve the field through a count.
```
spawnia:
```suggestion
Returns the count of a given relationship or model.
```

src/Schema/Directives/CountDirective.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Makes a simple count query for the relation.
+     *
+     * @param FieldValue $value
+     * @return FieldValue
+     */
+    public function resolveField(FieldValue $value)
+    {
+        return $value->setResolver(
+            function (Model $model) {
```
spawnia: That would fail if used on root field.

```suggestion
            function ($root): int {
```

src/Schema/Directives/CountDirective.php
```diff
@@ -0,0 +1,70 @@
...
+  """
+  The model to run the count on.
+  """
+  model: String
+) on FIELD_DEFINITION
+SDL;
+    }
+
+    /**
+     * Makes a simple count query for the relation.
```
spawnia:
```suggestion
     * Returns the count of a given relationship or model.
```

src/Schema/Directives/CountDirective.php
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param FieldValue $value
+     * @return FieldValue
+     */
+    public function resolveField(FieldValue $value)
+    {
+        return $value->setResolver(
+            function (Model $model) {
+                $relation = $this->directiveArgValue('relation');
+                $modelArg = $this->directiveArgValue('model');
```
spawnia: We only need to get this if we are not resolving a `relation`, so we can pull it down a few lines.

src/Schema/Directives/CountDirective.php
olivernybroe: Yeah I actually forgot to commit some changes. My bad :) I have added tests now.

olivernybroe:
Sorry for the many commits, apparently the batch accept suggestions didn't work.

docs/master/api-reference/directives.md
```diff
@@ -579,6 +579,41 @@ class ComplexityAnalyzer {
...
+    id: ID!
+    likes: Int! @count(relation: "likes")
+}
+```
+
+```graphql
+type Query {
+    categories: Int! @count(model: "Category")
+}
+```
```
spawnia:
```suggestion
```

```

PR #993, opened 2019-10-06
Simplify ASTHelper and add firstByName method

spawnia:
Internal refactoring

### DIFF
--- a/src/Schema/AST/ASTHelper.php
+++ b/src/Schema/AST/ASTHelper.php
@@ -11,10 +11,8 @@ use Illuminate\Support\Collection;
 use GraphQL\Language\AST\ValueNode;
 use GraphQL\Type\Definition\EnumType;
 use GraphQL\Language\AST\ArgumentNode;
-use GraphQL\Language\AST\ListTypeNode;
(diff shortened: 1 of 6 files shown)


src/Schema/AST/ASTHelper.php
```diff
@@ -83,20 +78,21 @@ public static function mergeUniqueNodeList($original, $addition, bool $overwrite
...
     /**
+     * Unwrap lists and non-nulls and get the name of the contained type.
+     *
      * @param  \GraphQL\Language\AST\Node  $definition
      * @return string
      */
     public static function getUnderlyingTypeName(Node $definition): string
     {
-        $type = $definition->type;
-        if ($type instanceof ListTypeNode || $type instanceof NonNullTypeNode) {
```
olivernybroe: Why is doing this not necessary anymore? Or was it never necessary?

spawnia: Somehow my comment got lost, so here it goes:

The `getUnderlyingNamedTypeNode` checks for those wrapping types already.

PR #994, opened 2019-10-07
Docs overhaul: search, back-to-top, medium-zoom, top-level tutorial, revamp index

spawnia:
- Move tutorial to top level navigation bar and remove it from versioned docs

image

- Fix spacing in between code blocks

image

- Remove tailwind and unused custom styles
- Add search (defaults to the latest stable version)

image

- Add back-to-top and medium-zoom

image

- Use plain markdown for the index page

image


docs/.vuepress/config.js
```diff
@@ -19,27 +19,38 @@ module.exports = {
...
+                link: '/users/'
             }
         ],
         sidebar: versioning.sidebars
-    }
+    },
+    plugins: [
+        ['@vuepress/back-to-top', true],
+        ['@vuepress/medium-zoom', true],
+        ['@vuepress/search', {
```
olivernybroe: Yay ❤️

docs/INDEX.md
```diff
@@ -1,94 +1,84 @@
 ---
 home: true
+heroImage: /logo.svg
 actionText: Get Started  →
 actionLink: /docs/latest
-
-footer: Copyright © 2019 Christopher Moore
+features:
+- title: 📜 SDL First
```
olivernybroe: I don't think we should use emojis directly here as they render differently on different operating systems.
Here is an example of mine, which is different from what you sent.
image

spawnia: That might be seen as a strength too. Everybody gets the emoji's that they are used to from their system.

docs/.vuepress/config.js
spawnia: I just added one more minor change so that the latest stable version is searched by default

docs/INDEX.md
olivernybroe: Yeah, but if you look at the image you sent of this, the first emoji did not render on your machine

spawnia: Emojis are universally used, and i don't think it is a big deal if they are not displayed in some rare cases.

You used one in this very comment chain 😉 I think we are good here and really don't feel like including custom icons for every place where we use emojis.

olivernybroe: Sure thing 👍 Just wanted to make you aware of it.

spawnia:
> It looks really good 👍

Thanks for the review!

PR #998, opened 2019-10-09
Add @scope directive

olivernybroe:
- [x] Added or updated tests
- [x] Updated the changelog

I have not written tests and so on yet, because I first want to confirm if this is the approach we would add.

The reason for this scope is because in a lot of my scenarios where I use the `@builder` directive, I basically just call a scope and pass the value down to it directly. I cannot use the `@where` directive as that passes 3 arguments down to my scope which I do not want.

Another solution for this scenario could be to extend the `@builder` directive to support just passing a method it should run on the builder directly as that is essentially what the `@scope` directive does.

**Changes**

This PR adds a new directive called `scope`.
The point of this directive is to make a simple way of adding a scope from your model, without having to wrap it in a class/method.

```
type User {
    id: ID!
    likes(
        liker: ID @scope(name: "whereLiker")
    ): [Like] @hasMany
}
```

**Breaking changes**

None, just a new directive.


src/Schema/Directives/ScopeDirective.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Nuwave\Lighthouse\Schema\Directives;
+
+use Nuwave\Lighthouse\Support\Contracts\ArgBuilderDirective;
+use Nuwave\Lighthouse\Support\Contracts\DefinedDirective;
+
+/**
+ * This directive adds a scope to the builder and supplies the argument value
```
spawnia: Good description, could you move that into the SDL definition below? People are more likely to see that since it is shown in the docs and the autogenerated directives file.

src/Schema/Directives/ScopeDirective.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Add additional constraints to the builder based on the given argument value.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
+     * @param  mixed $value
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     */
+    public function handleBuilder($builder, $value)
+    {
+        $scope = $this->directiveArgValue('name');
```
spawnia: This is really simple - nice! Is there something we can do to make the dev experience better for some edge cases? For example, calling a scope that does not exist or a method that is not a scope.

spawnia:
I like the idea and the design of this, it directly translates a Laravel concept into a directive.

> Another solution for this scenario could be to extend the @builder directive to support just passing a method it should run on the builder directly as that is essentially what the @scope directive does.

I think it is best to have them separate, as that keeps them simple and allows for no confusion.

src/Schema/Directives/ScopeDirective.php
```diff
@@ -0,0 +1,51 @@
...
+  name: String
+) on ARGUMENT_DEFINITION
+SDL;
+    }
+
+    /**
+     * Add additional constraints to the builder based on the given argument value.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
+     * @param  mixed $value
```
spawnia:
```suggestion
     * @param  mixed  $value
```

src/Schema/Directives/ScopeDirective.php
```diff
@@ -0,0 +1,51 @@
...
+  """
+  name: String
+) on ARGUMENT_DEFINITION
+SDL;
+    }
+
+    /**
+     * Add additional constraints to the builder based on the given argument value.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
```
spawnia:
```suggestion
     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder 
 $builder
```

src/Schema/Directives/ScopeDirective.php
olivernybroe: I don't think there is a universal way to check if a method is a scope, as a scope could potentially be added directly to the query builder with a custom query builder.

However I will add a check to see if the method exist, so we can throw a user friendly debug message.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v4.4.0...master)
 
+### Added
```
```suggestion
### Added

```

docs/master/api-reference/directives.md
```diff
@@ -2173,6 +2173,34 @@ scalar DateTime
     @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\DateTime")
 ```
 
+## @scope
+
+Adds a scope to the query builder through a graphql argument.
```
spawnia: To reduce duplication/maintenance, we can remove the `## Definition` section and just pull that up here.

That way, we can put a nice, explicit description in the directive definition and have that available everywhere.

tests/Unit/Schema/Directives/ScopeDirectiveTest.php
```diff
@@ -0,0 +1,118 @@
...
+        $taskWithTag = factory(Task::class)->create();
+        $taskWithTag->tags()->save(
+            factory(Tag::class)->create(['name' => 'Lighthouse'])
+        );
+        $user->tasks()->save($taskWithTag);
+
+        $this->be($user);
+
+        $this->schema = /* @lang GraphQL */
+            '
```
spawnia: Remove one newline before this.

tests/Unit/Schema/Directives/ScopeDirectiveTest.php
```diff
@@ -0,0 +1,118 @@
...
+        $user = factory(User::class)->create();
+
+        $user->tasks()->saveMany(
+            factory(Task::class)->times(2)->create()
+        );
+
+        $this->be($user);
+
+        $this->schema = /* @lang GraphQL */
+            '
```
spawnia: Again, an extra newline

tests/Unit/Schema/Directives/ScopeDirectiveTest.php
```diff
@@ -0,0 +1,118 @@
...
+        /** @var User $user */
+        $user = factory(User::class)->create();
+
+        $user->tasks()->saveMany(
+            factory(Task::class)->times(2)->create()
+        );
+
+        $this->be($user);
+
+        $this->schema = /* @lang GraphQL */
```
spawnia: This test case seems overly complex, i think a schema like this would do, with none of the setup.

```graphql
type Quer {
  tasks(
    name: String @scope(name: "nonExistantScope")
  ): [Task!]! @all
}
```

src/Schema/Directives/ScopeDirective.php
```diff
@@ -0,0 +1,58 @@
...
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     * @throws DirectiveException
+     */
+    public function handleBuilder($builder, $value)
+    {
+        $scope = $this->directiveArgValue('name');
+        try {
+            return $builder->{$scope}($value);
+        } catch (BadMethodCallException $exception) {
+            throw new DirectiveException(
```
spawnia: I am planning to phase that out in favour of the `DefinitionException`, used to universally indicate any kind of schema definition error.

olivernybroe:
Thanks for taking care of the last changes ♥️

spawnia:
> Thanks for taking care of the last changes

All good, great work.

spawnia:
Once we can be sure that repeatable directives will be allowed in GraphQL, we might expand `@scope` to work on a field level too.

tlaverdure:
> Once we can be sure that repeatable directives will be allowed in GraphQL, we might expand @scope to work on a field level too.

@spawnia are you saying that we could possibly apply `@scope` on a field like this?

```graphql
type Post {
  id: ID!
  content: String
  comments_count: Int @scope(name: "withCommentsCount")
}
```

With`scopeWithCommentsCount()` I'd like to perform a subquery on the Builder instance only when `comments_count` is requested.

spawnia:
That's what `@with` does.

tlaverdure:
Thanks @spawnia figured out how I could pull off using `@with`  to eager load a relationship that performs a subquery and `@method` to resolve the value 👍🏾

PR #999, opened 2019-10-09
[Docs] Fix wrong definition in builder directive

olivernybroe:
Looks like the builder directive was using the wrong definition as it is applied on arguments not on fields.


spawnia:
Good find! Can you update the previous docs versions too?

olivernybroe:
Sure thing :)

PR #1005, opened 2019-10-11
Add `@upsert` directive and nested mutation operations to create or update a model   regardless whether it exists

### LINKED ISSUE #995: `@upsert` directive and nested mutations
I'd be able to create a PR about this but I want to know if it is better to do it just for my project or share here with you to incorporate the feature and know what is the best way to do it.

I get information from a source and need to put the information in a destination. Sometimes the data exists in the destination but other times not. Actually I don't care about it so I usually update the current data that are in the destination if it exists or create if it does not exist.

The problem is that with the current mutation directives you need to do a read before doing the action to know if it is an update or a create. If this is done for a lot of data it actually cause performance issues compared to just tell the system to replace the data without taking into account if they exist or not. This is because the read will be done just for the GraphQL server and not the client plus graphql server. 

I'd like to have an idempotent action that could be named "replace" that ensures te final result will be the provided without taking into account if the model already exists or not.

Create it just for my project using a custom directive and extending the MutationExecutor to do a create or update depending on the model state (exists or not).

joskfg:
- [x] Added or updated tests
- [x] Updated the changelog

Resolves #995 

**Changes**

## Upsert

Use the [@upsert](../api-reference/directives.md#upsert) directive to update a model with 
a given `id` or create it if it does not exist. 

```graphql
type Mutation {
  upsertUser(id: ID!, name: String!, email: String): User @upsert
}
```

Since upsert can create or update your data you must have all the minimum fields for a creation as required.
The `id` is always required and must be marked as fillable in the model.

```graphql
mutation {
  upsertUser(id: "123" name: "Hillary"){
    id
    name
    email
  }
}
```

```json
{
  "data": {
    "upsertUser": {
      "id": "123",
      "name": "Hillary",
      "email": null
    }
  }
}
```

## Nested mutations

You can now expose a nested operation `upsert` in all nested mutations that support the `update` operation.

**Breaking changes**

\-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v4.4.0...master)
 
+### Added
+- Add `@upsert` directive for mutations and nested mutations https://github.com/nuwave/lighthouse/pull/1005
+
 ## [4.4.0](https://github.com/nuwave/lighthouse/compare/v4.3.0...v4.4.0)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `@upsert` directive for mutations and nested mutations https://github.com/nuwave/lighthouse/pull/1005
+- Add `@upsert` directive and nested mutation operations to create or update a model
+  regardless whether it exists https://github.com/nuwave/lighthouse/pull/1005
 
 ## [4.4.0](https://github.com/nuwave/lighthouse/compare/v4.3.0...v4.4.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v4.4.0...master)
 
 ### Added
+
 - Add `@upsert` directive for mutations and nested mutations https://github.com/nuwave/lighthouse/pull/1005
 
 ## [4.4.0](https://github.com/nuwave/lighthouse/compare/v4.3.0...v4.4.0)
--- a/docs/4.4/api-reference/directives.md
+++ b/docs/4.4/api-reference/directives.md
@@ -2541,6 +2541,64 @@ type Mutation {
 }
 ```
 
+## @upsert
+
+Create or update an Eloquent model with the input values of the field.
+
+```graphql
+type Mutation {
+    upsertPost(id: ID!, content: String): Post @upsert
+}
+```
+
+### Definition
+
+```graphql
+"""
+Create or update an Eloquent model with the input values of the field.
+"""
+directive @upsert(
+  """
+  Specify the class name of the model to use.
+  This is only needed when the default model resolution does not work.
+  """
+  model: String
+
+  """
+  Set to `true` to use global ids for finding the model.
+  If set to `false`, regular non-global ids are used.
+  """
+  globalId: Boolean = false
+) on FIELD_DEFINITION
+```
+
+### Examples
+
+Lighthouse uses the argument `id` to fetch the model by its primary key if exists.
+If the model doesn't exist it will create it using the primary key indicated.
+This will work even if your model has a differently named primary key,
+so you can keep your schema simple and independent of your database structure.
+
+If you want your schema to directly reflect your database schema,
+you can also use the name of the underlying primary key.
+This is not recommended as it makes client-side caching more difficult
+and couples your schema to the underlying implementation.
+
+```graphql
+type Mutation {
+    upsertPost(post_id: ID!, content: String): Post @upsert
+}
+```
+
+If the name of the Eloquent model does not match the return type of the field,
+or is located in a non-default namespace, set it with the `model` argument.
+
+```graphql
+type Mutation {
+    upsertAuthor(id: ID!, name: String): Author @upsert(model: "App\\User")
+}
+```
+
 ## @where
 
 Use an input value as a [where filter](https://laravel.com/docs/queries#where-clauses).
--- a/docs/4.4/api-reference/directives.md
+++ b/docs/4.4/api-reference/directives.md
@@ -2541,64 +2541,6 @@ type Mutation {
 }
 ```
 
-## @upsert
-
-Create or update an Eloquent model with the input values of the field.
-
-```graphql
-type Mutation {
-    upsertPost(id: ID!, content: String): Post @upsert
-}
-```
-
-### Definition
-
-```graphql
-"""
-Create or update an Eloquent model with the input values of the field.
-"""
-directive @upsert(
-  """
-  Specify the class name of the model to use.
-  This is only needed when the default model resolution does not work.
-  """
-  model: String
-
-  """
-  Set to `true` to use global ids for finding the model.
-  If set to `false`, regular non-global ids are used.
-  """
-  globalId: Boolean = false
-) on FIELD_DEFINITION
-```
-
-### Examples
-
-Lighthouse uses the argument `id` to fetch the model by its primary key if exists.
-If the model doesn't exist it will create it using the primary key indicated.
-This will work even if your model has a differently named primary key,
-so you can keep your schema simple and independent of your database structure.
-
-If you want your schema to directly reflect your database schema,
-you can also use the name of the underlying primary key.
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/nuwave/lighthouse/compare/v4.4.0...master)
 
+### Added
```
```suggestion
### Added

```

docs/4.4/api-reference/directives.md
```diff
@@ -2541,6 +2541,64 @@ type Mutation {
 }
 ```
 
+## @upsert
```
spawnia: We do not put new features in minor releases. The reason for that is that somebody on `4.4.0` might read something in the docs that does not apply to their version.

src/Execution/MutationExecutor.php
```diff
@@ -270,6 +307,101 @@ public static function executeUpdate(Model $model, Collection $args, ?Relation $
...
+     */
+    public static function executeUpsert(Model $model, Collection $args, ?Relation $parentRelation = null): Model
+    {
+        $id = $args->pull('id')
+            ?? $args->pull(
+                $model->getKeyName()
+            );
+
+        try {
+            $model = $model->newQuery()->findOrFail($id);
```
spawnia: Why not just just `find()` and check if a `$model` is returned?

src/Execution/MutationExecutor.php
```diff
@@ -304,6 +436,16 @@ public static function executeUpdate(Model $model, Collection $args, ?Relation $
                 });
             }
 
+            if (isset($nestedOperations['upsert'])) {
+                (new Collection($nestedOperations['upsert']))->each(function ($singleValues) use ($relation): void {
```
spawnia: That iterations has been duplicated a few times, it could be abstracted into its own function, like `handleMultiRelationUpsert`

tests/Integration/Execution/MutationExecutor/HasManyTest.php
```diff
@@ -131,7 +188,58 @@ public function testCanCreateHasMany(): void
...
+    public function actionsOverExistingDataProvider()
+    {
+        yield ['Update action' => 'update'];
+        yield ['Upsert action' => 'upsert'];
+    }
+
+    /**
+     * @dataProvider actionsOverExistingDataProvider
+     */
+    public function testCanCreateHasMany($action): void
```
spawnia: What type is `$action`?

tests/Integration/Execution/MutationExecutor/HasManyTest.php
```diff
@@ -131,7 +188,58 @@ public function testCanCreateHasMany(): void
         ]);
     }
 
-    public function testCanUpdateHasMany(): void
+    public function actionsOverExistingDataProvider()
+    {
+        yield ['Update action' => 'update'];
```
spawnia: I like that pattern, but we do not use it anywhere. Let's keep it consistent.

src/Execution/MutationExecutor.php
joskfg: It is the same result. I used findOrFail because it is more redeable that a find with an `$model !== null` . It is much more clear to read `findOrFail`with a `ModelNotFound`. I really didn't do a performance test to see the difference in performance.

src/Execution/MutationExecutor.php
joskfg: I used the same approach than update. Should I abstract the update too?

tests/Integration/Execution/MutationExecutor/HasManyTest.php
joskfg: Whoops I didn't find any other provider and I see that it was due to a bad filter in que search. I'm changing it to return with arrays.

src/Execution/MutationExecutor.php
joskfg: Do you prefer find with `if $model !== null`? I can change it :-)

spawnia: `if($model){}` should suffice

spawnia: Just reread that part, i think we should rename to `$modelInstance` or something else, to differentiate from the parameter.

src/Execution/MutationExecutor.php
```diff
@@ -295,13 +466,11 @@ public static function executeUpdate(Model $model, Collection $args, ?Relation $
...
 
             if (isset($nestedOperations['update'])) {
-                (new Collection($nestedOperations['update']))->each(function ($singleValues) use ($relation): void {
-                    self::executeUpdate(
-                        $relation->getModel()->newInstance(),
-                        new Collection($singleValues),
-                        $relation
-                    );
-                });
+                self::handleMultiRelationUpdate(new Collection($nestedOperations['update']), $relation);
```
spawnia: That looks really clean, good job!

src/Schema/Directives/UpsertDirective.php
```diff
@@ -0,0 +1,101 @@
...
+use Illuminate\Support\Collection;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\DatabaseManager;
+use Nuwave\Lighthouse\Schema\Values\FieldValue;
+use Nuwave\Lighthouse\Execution\MutationExecutor;
+use Nuwave\Lighthouse\Support\Contracts\GlobalId;
+use Nuwave\Lighthouse\Support\Contracts\FieldResolver;
+use Nuwave\Lighthouse\Support\Contracts\DefinedDirective;
+
+class UpsertDirective extends BaseDirective implements FieldResolver, DefinedDirective
```
spawnia: There is a lot of duplication between this and `@update`, even some with `@create`. I think we should be able to refactor this so only `name()`, `definition()` and a new function like `makeExecutor(Model $model, array $args): \Closure` should be different.

tests/Integration/Execution/MutationExecutor/HasManyTest.php
```diff
@@ -131,7 +188,60 @@ public function testCanCreateHasMany(): void
         ]);
     }
 
-    public function testCanUpdateHasMany(): void
+    public function actionsOverExistingDataProvider()
```
spawnia:
```suggestion
    public function existingModelMutations()
```

tests/Integration/Execution/MutationExecutor/HasManyTest.php
```diff
@@ -131,7 +188,60 @@ public function testCanCreateHasMany(): void
...
+    /**
+     * @dataProvider actionsOverExistingDataProvider
+     */
+    public function testCanCreateHasMany(string $action): void
+    {
+        factory(User::class)->create();
+
+        $this->graphQL("
+        mutation {
+            ${action}User(input: {
```
spawnia: I like how you kept that test DRY. Nice!

spawnia:
I am really impressed with how quickly you were able to dig into this and added a solid implementation with comprehensive tests, while sticking to the existing codestyle.

Apart from the doc entry in the directives reference, there is also https://lighthouse-php.com/master/eloquent/getting-started.html#update and https://lighthouse-php.com/master/eloquent/nested-mutations.html

src/Schema/Directives/UpsertDirective.php
joskfg: There are a lot to refactor from MutationExecutor too, but I don't know if it is the best idea to do it now or it will be a never end refactor. I'll take a look.

spawnia: Let us just do the directives for now, since nested mutations are likely to change soon - making them compose.

joskfg:
> Lets reframe this so we won't have to keep updating the number



> I am really impressed with how quickly you were able to dig into this and added a solid implementation with comprehensive tests, while sticking to the existing codestyle.
> 
> Apart from the doc entry in the directives reference, there is also https://lighthouse-php.com/master/eloquent/getting-started.html#update and https://lighthouse-php.com/master/eloquent/nested-mutations.html

Thanks!

src/Schema/Directives/MutationExecutorDirective.php
```diff
@@ -50,20 +51,38 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
         return $fieldValue->setResolver(
             function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Model {
-                $modelClass = $this->getModelClass();
                 /** @var \Illuminate\Database\Eloquent\Model $model */
-                $model = new $modelClass;
+                $model = new ($this->getModelClass());
 
                 $executeMutation = function () use ($model, $args): Model {
-                    return $this->executeMutation($model, new Collection($args))->refresh();
+                    return $this
```
joskfg: I prefer this indentation

spawnia:
Massive effort and great job on this cool new feature, @joskfg 

Thank you!

PR #1011, opened 2019-10-13
Do not require model class in paginated relations

### LINKED ISSUE #1001: Possible BC 4.4: Failed model detection
I am not 100% sure what is exactly causing the issue but on 4.3 we have the following schema:

```graphql
type SchoolGroup {
    children: [SchoolGroup!] @hasMany(type: "paginator")
}
```

However when we just tried updating to 4.4 the following error shows up:

```
No class 'SchoolGroup' was found for directive 'hasMany'
```

(which is true, `SchoolGroup` is not an actual class/model, it's namespaced in `...\Models\School\Group` in our app)

This can be fixed with adding the `@modelClass` directive it looks like, but this was never a problem before so I'm puzzled what happened and what the exact way to fix this is, it always just worked :tm: :slightly_smiling_face:

As far as I could tell a `4.3`->`4.4` is not supposed to break existing schema's so this might be a bug. I have tried reading through the PR's that seem to be causing them but they're massive, so I'm not able to determine the root cause exactly.

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

Fixes #1001 

**Changes**

Fix regression in 4.4.0 that required matching the type returned from paginated relationship fields with the class name of the model

**Breaking changes**

Actually unbreaking a regression


src/Pagination/PaginationManipulator.php
```diff
@@ -137,7 +153,7 @@ protected function registerPaginator(
         $paginatorFieldClassName = addslashes(PaginatorField::class);
 
         $paginatorType = PartialParser::objectTypeDefinition("
-            type $paginatorTypeName $modelClassDirective {
+            type $paginatorTypeName {$this->modelClassDirective()} {
```
joskfg: Just a tip:

If you use HEREDOC, you won't need to escape all that quotes.

tests/Integration/Schema/Directives/HasManyDirectiveTest.php
```diff
@@ -158,6 +158,52 @@ public function testCanQueryHasManyPaginator(): void
         ])->assertJsonCount(2, 'data.user.tasks.data');
     }
 
+    public function testDoesNotRequireModelClassForPaginatedHasMany(): void
```
joskfg: I don't see the difference between this test and the previous one. Both doesn't have the modelClass set, just the type.

spawnia: The type name does not match the model name. It is not needed, the BC was that 4.4 forced them to line up.

Renamed to `type NotTheModelNameTask` for clarification.

src/Pagination/PaginationManipulator.php
spawnia: Let's discuss this in a seperate PR, that could be useful in a lot of places!

PR #1021, opened 2019-10-21
Handle nullable input objects in TypedArgs conversion

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

**Changes**

Converting client-given args to an `ArgumentSet` fails when `null` is given for an argument that is defined as an `input`.

Added a failing test:

```
1) Tests\Unit\Execution\Arguments\TypedArgsTest::testNullableInputObject
TypeError: Argument 1 passed to Nuwave\Lighthouse\Execution\Arguments\TypedArgs::wrapArgs() must be of the type array, null given, called in /var/www/src/Execution/Arguments/TypedArgs.php on line 193

/var/www/src/Execution/Arguments/TypedArgs.php:91
```

It works now.

**Breaking changes**

No

### DIFF
--- a/src/Execution/Arguments/TypedArgs.php
+++ b/src/Execution/Arguments/TypedArgs.php
@@ -181,6 +181,11 @@ class TypedArgs
      */
     protected function wrapWithNamedType($value, NamedType $namedType)
     {
+        // As GraphQL does not allow empty input objects, we return null as is
+        if ($value === null) {
+            return $value;
+        }
+
         // This might be null if the type is
         // - created outside of the schema string
         // - one of the built in types
@@ -193,9 +198,9 @@ class TypedArgs
             $subArgumentSet->arguments = $this->wrapArgs($value, $typeDef->fields);
 
             return $subArgumentSet;
-        } else {
-            // Otherwise, we just return the value as is and are down with that subtree
-            return $value;
         }
+
+        // Otherwise, we just return the value as is and are down with that subtree
+        return $value;
     }
 }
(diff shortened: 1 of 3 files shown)


src/Execution/Arguments/TypedArgs.php
```diff
@@ -181,6 +181,11 @@ protected function wrapWithType($valueOrValues, $type)
      */
     protected function wrapWithNamedType($value, NamedType $namedType)
     {
+        // As GraphQL does not allow empty input objects, we return null as is
+        if ($value === null) {
+            return $value;
```
olivernybroe: Not sure if it actually have a performance impact, but it should be easier for the code optimizer to optimize it when it's not a variable.

```suggestion
            return null;
```

tests/Unit/Execution/Arguments/TypedArgsTest.php
```diff
@@ -49,6 +38,42 @@ public function testNullableList(): void
...
+
+        $argumentSet = $this->rootQueryArgumentSet([
+            'bar' => null,
+        ]);
+
+        $this->assertCount(1, $argumentSet->arguments);
+
+        $bar = $argumentSet->arguments['bar'];
+        $this->assertInstanceOf(Argument::class, $bar);
+        $this->assertSame(null, $bar->value);
```
joskfg: It is better to use specific assertions instead of generics. Try with [assertNull](https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertNull)

PR #1024, opened 2019-10-24
Change type in example

Dana94:
`User!` was used on line 328, so I assume line 43 was a mistake.

- [ ] Added or updated tests
- [ ] Updated the changelog

**Changes**

Edit in GraphQL example.

**Breaking changes**

None


spawnia:
Thanks!

PR #1027, opened 2019-10-29
Use detailed `$description` property when generating `enum` values from a `BenSampo\Enum\Enum` class

killtw:
- [x] Added or updated tests
- [x] Updated the changelog

**Changes**
Description of LaravelEnumType should using `$enum->desciption` for more detail

### DIFF
--- a/src/Schema/Types/LaravelEnumType.php
+++ b/src/Schema/Types/LaravelEnumType.php
@@ -41,7 +41,7 @@ class LaravelEnumType extends EnumType
                 function (Enum $enum): array {
                     return [
                         'name' => $enum->key,
-                        'value' => $enum->value,
+                        'value' => $enum,
                         'description' => $enum->description,
                     ];
                 },
--- a/src/Schema/Types/LaravelEnumType.php
+++ b/src/Schema/Types/LaravelEnumType.php
@@ -41,8 +41,8 @@ class LaravelEnumType extends EnumType
                 function (Enum $enum): array {
                     return [
                         'name' => $enum->key,
-                        'value' => $enum,
-                        'description' => "$enum->value",
+                        'value' => $enum->value,
+                        'description' => "$enum->description",
                     ];
                 },
                 $enumClass::getInstances()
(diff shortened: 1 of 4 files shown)


src/Schema/Types/LaravelEnumType.php
```diff
@@ -41,8 +41,8 @@ public function __construct(string $enumClass, ?string $name = null)
                 function (Enum $enum): array {
                     return [
                         'name' => $enum->key,
-                        'value' => $enum,
-                        'description' => "$enum->value",
+                        'value' => $enum->value,
```
spawnia: The internal value should stay an instance of the enum class. That makes it work smoothly with type hints and model casts.

src/Schema/Types/LaravelEnumType.php
```diff
@@ -41,8 +41,8 @@ public function __construct(string $enumClass, ?string $name = null)
...
+                        'description' => "$enum->description",
```
spawnia:
```suggestion
                        'description' => $enum->description,
```

tests/Unit/Schema/Types/LaravelEnumTypeTest.php
```diff
@@ -30,6 +31,13 @@ public function testMakeEnumWithCustomName(): void
         $this->assertSame($customName, $enumType->name);
     }
 
+    public function testCustomDescription()
```
spawnia:
```suggestion
    public function testCustomDescription(): void
```

spawnia:
Thanks for the contribution. The test and the change are fine.

Please address the comments i left and add a line to the CHANGELOG.md

src/Schema/Types/LaravelEnumType.php
killtw: But it would record instance as JSON when column type is JSON or TEXT

spawnia: Not if attribute casting is used (https://github.com/BenSampo/laravel-enum#attribute-casting).

Anyways, that has been the current behaviour for quite a while. We can't break it in a minor release. You can always use your own version of `LaravelEnumType` if you need to change that in your project.

Issue #1000, opened 2019-10-09
Calling multiple queries in a test always returns the same result


spawnia:
**Describe the bug**

When issuing a query through `MakesGraphQLRequests#graphQL()`, subsequent requests after the first ones always return the same result.

**Expected behavior/Solution**

Each query to be handled individually.

The `GraphQLController` always receives the same `GraphQLRequest` object. Since that is a singleton (and i think Laravel's `Request` too?), we need to clear that inbetween calls or architect Lighthouse in such a way that prevents this.

**Steps to reproduce**

1. Write a unit test using the `MakesGraphQLRequests` trait
2. Call `graphQL()` multiple times

**Environment**

Lighthouse Version: `master`
Laravel Version: `5.8`

spawnia:
@enzonotario proposed the following fix:

```php
    protected function graphql(string $query)
    {
        $request = new Request(['query' => $query]);
        $lighthouseRequest = new LighthouseRequest($request);
        $response = app(GraphQLController::class, [
            'request' => $lighthouseRequest,
        ])->query($lighthouseRequest);
        return TestResponse::fromBaseResponse($response);
    }
```

I think we have to investigate if this only happens when using Laravel's `MakesHttpRequest` test helpers, or might also happen when issuing multiple queries in a single HTTP request, which might be a valid use case for some people.

enzonotario:
Hi! This issue was solved in this commit: https://github.com/nuwave/lighthouse/commit/f3d4bc7b86f1acf73b4001e9cf4ab8f7ddb0f4d4

I have added this test to `AllDirectiveTest`:
```
public function testGetAllWithoutCache(): void
    {
        factory(User::class, 2)->create();

        factory(Post::class, 2)->create([
            // Do not create those, as they would create more users
            'task_id' => 1,
        ]);

        $this->schema = '
        type User {
            id: ID!
            posts: [Post!]! @all
        }

        type Post {
            id: ID!
        }

        type Query {
            users: [User!]! @all
            posts: [Post!]! @all
        }
        ';

        $this->graphQL('
        {
            users {
                id
            }
        }
        ')->assertJsonCount(4, 'data.users');

        $this->graphQL('
        {
            posts {
                id
            }
        }
        ')->assertJsonCount(2, 'data.posts');

        factory(User::class, 2)->create();

        factory(Post::class, 2)->create([
            // Do not create those, as they would create more users
            'task_id' => 1,
        ]);

        $this->graphQL('
        {
            users {
[...]

PR #1030, opened 2019-10-31
Enable multiple queries in a single request by clearing `BatchLoader` instances after executing each queryCalling multiple mutations with relationship in a test fails

enzonotario:
- [X] Added or updated tests
- [x] Updated the changelog

It's just a failing test related to #1000 so we can investigate it. So far I don't have a proposal solution, sorry. But I hope this PR gets any time the solution.

**Changes**

None.

**Breaking changes**

None.


enzonotario:
Hey @spawnia just want to tell you that this issue is also gone setting the new `batchload_relations` config to `false`.

spawnia:
> Hey @spawnia just want to tell you that this issue is also gone setting the new `batchload_relations` config to `false`.

Found the root cause and fixed it. Can you give this a review?

PR #1031, opened 2019-11-01
Add `syncWithoutDetaching` option for BelongsToMany and MorphToMany relationships

omacranger:
- [x] Added or updated tests
- [x] Updated the changelog

**Changes**

Adds support for `syncWithoutDetaching` method on BelongsToMany relationships.

**Breaking changes**

Should be no breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add `@scope` directive for adding a scope to the query builder https://github.com/nuwave/lighthouse/pull/998
+- Add `syncWithoutDetaching` option for BelongsToMany relationships
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


spawnia:
Thanks for your contribution.

How is that different than using `attach` ?

omacranger:
@spawnia `attach` used without a check to see if the relationship is already connected results in duplicate rows in the pivot table where `syncWithoutDetaching` does not -- atleast as of Laravel 5.8.x. Pseudo code example below: 



Group - Class that contains BelongsToMany to Item
Item - Class that contains BelongsToMany to Group

In an example where someone were to attempt to `attach` the same group multiple times -- either by different users at the same time, network retry, etc:

```
$item->groups()->attach(1);
$item->groups()->attach(1);
```

$item->groups would return an array of two instances of Group with an ID of 1 where `syncWithoutDetaching` would not.

omacranger:
Is there anything else you'd like to see done to this to get the feature merged in?

spawnia:
Hey, thanks for pinging me again.

This is a great enhancement, it makes sense to add that in. Please rebase your fork against the latest master and make sure the CHANGELOG is added at the correct spot and contains a link to this PR.

omacranger:
Should be good to go.

spawnia:
Nicely done 🎉

Issue #1029, opened 2019-10-31
Generator commands can not handle arrays with multiple namespaces


Amsterdamys:
**Describe the bug**

I need to store resolvers split by subdirectories, like:

app
---> GraphQL
-----> Queries
--------> CustomFolder1
----------> MyResolver1.php
--------> CustomFolder2
----------> MyResolver2.php

To do that I had to change namespaces->queries param in lighthouse configs.

It was

`'queries' => 'App\\GraphQL\\Queries',`

It become

`'queries' => ['App\\GraphQL\\Queries', 'App\\GraphQL\\Queries\\CustomFolder2'],`

After I did it all resolvers became visible correctly.

But now when I run following command 

`php artisan lighthouse:query Cards/CallsWithCLM2`

I see the following error:

`QueryCommand::getDefaultNamespace() must be of the type string, array returned`

**Expected behavior/Solution**

I think It's better to support having an array as a value for that field.

**Steps to reproduce**

Described above

**Output/Logs**

<details><summary>Click to expand</summary>

```
  ?[41;1m Symfony\Component\Debug\Exception\FatalThrowableError ?[49;22m : ?[33mReturn value
of Nuwave\Lighthouse\Console\QueryCommand::getDefaultNamespace() must be of the type string,
array returned?[39m

  at ?[32mC:\OSPanel\domains\shaman-shamango-stats-api\vendor\nuwave\lighthouse\src\Console\Q
ueryCommand.php?[39m:?[32m36?[39m
    32|      * @return string
    33|      */
    34|     protected function getDefaultNamespace($rootNamespace): string
    35|     {
  > 36|         return config('lighthouse.namespaces.queries');
    37|     }
    38|
    39|     /**
    40|      * Get the stub file for the generator.

  ?[33mException trace:?[39m

  ?[36m1   ?[39m?[33mNuwave\Lighthouse\Console\QueryCommand::getDefaultNamespace("App")?[39m
      ?[32mC:\OSPanel\domains\shaman-shamango-stats-api\vendor\laravel\framework\src\Illumina
te\Console\GeneratorCommand.php?[39m:?[32m97?[39m

  ?[36m2   ?[39m?[33mIlluminate\Console\GeneratorCommand::qualifyClass("Cards\CallsWithCLM2")
?[39m
      ?[32mC:\OSPanel\domains\shaman-shamango-stats-api\vendor\laravel\framework\src\Illumina
[...]

PR #1033, opened 2019-11-04
Handle arrays of namespaces in generator commands

spawnia:
- [x] Added or updated tests
- [x] Updated the changelog

Resolves https://github.com/nuwave/lighthouse/issues/1029

**Changes**

When the configuration holds an array of namespaces, the generator commands currently fail with:

```
QueryCommand::getDefaultNamespace() must be of the type string, array returned
```

This handles the case of multiple default namespaces by identifying the common part of all configured namespaces and uses that to generate the class.

Just like Laravel generator commands, sub-namespaces can be added:

```
php artisan lighthouse:query Sub\\Foo
```

**Breaking changes**

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Use detailed `$description` property when generating `enum` values from a `BenSampo\Enum\Enum` class  https://github.com/nuwave/lighthouse/pull/1027
 
+### Fixed
+
+- Handle arrays of namespaces in generator commands https://github.com/nuwave/lighthouse/pull/1033
+
 ## [4.5.3](https://github.com/nuwave/lighthouse/compare/v4.5.2...v4.5.3)
 
 ### Fixed
--- a/src/Console/InterfaceCommand.php
+++ b/src/Console/InterfaceCommand.php
@@ -25,15 +25,9 @@ class InterfaceCommand extends LighthouseGeneratorCommand
      */
     protected $type = 'Interface';
 
-    /**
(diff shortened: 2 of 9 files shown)


spawnia:
@Amsterdamys please test and let me know if that works for you.

Amsterdamys:
Now I see no errors when create a new query with 2 namespaces set. But another issue, when I specify 2 namespaces (like that _['App\\GraphQL\\Queries', 'App\\GraphQL\\Queries\\Cards']_) it tries to found my schemas only in the common part (_App\\GraphQL_). So when I call my new card schema it throws the following error in GraphQl client:

`"message": "include(..\\vendor\\composer/../../app/GraphQL/Queries/CallsWithCLM.php): failed to open stream: No such file or directory"`

spawnia:
>  But another issue

That is indeed another issue. Can you open a new bug report so we can deal with that?

To clarify, did the changes in this PR have the expected result for you?

Amsterdamys:
The thing is that it worked before this pull request. My schema was visible in the Cards subdirectory. Now it's not. That's why I'm adding it to here.

But if you're sure it's a separate issue I'll create another bugreport. Should I?

spawnia:
I am a bit confused by what you are saying, you seem to conflate the terms *schema* and *query*.

This PR only changes the generation of *new* query classes. The mechanism that locates schema's as well as the mechanism to locate query classes are entirely untouched.

It would help if you can precisely describe what is going wrong, adding in the necessary code examples. If you are having issues not related to the generation of new query classes, that would go in a separate issue.

Amsterdamys:
Sure

Amsterdamys:
This issue is fixed, yes

PR #1042, opened 2019-11-11
Rework how @can directive fetches model (@softDeletes are not working)

lorado:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Using `@can` on softDeleted models is not working as expected. Within `@can` directive simple query builder gets created without respecting soft deleted models. `@softDeletes` directive ist also doesn't help here, as it is not processed in builder-creation

**Changes**

We should somehow rework, how `@can` fetches model for checking access. Take a look at line 83 in file `Nuwave\Lighthouse\Schema\Directives\CanDirective`. There is the problem. I tried to use following code:

```php
$resolveInfo
->argumentSet
->enhanceBuilder(
    $this->getModelClass()::query(),
)
->findOrFail($args[$find]);
```

This piece is from `@find`, `@all` etc. directives. With given changes it works for my test, but failes the `testThrowsIfNotAuthorized` test with following error: 

```
Illuminate\Database\QueryException : SQLSTATE[42S22]: Column not found: 1054 Unknown column 'foo' in 'where clause' (SQL: select * from `posts` where `foo` = 1 and `posts`.`id` = 1 limit 1)
```

Currently I don't know how to fix it... =\ any ideas are welcome!

**Breaking changes**

There should be no breaking changes

### DIFF
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -101,7 +101,7 @@ class ArgumentSet
                     return $directive instanceof ArgBuilderDirective;
                 });
 
-            if (!empty($directiveFilter)) {
+            if (! empty($directiveFilter)) {
                 $filteredDirectives = $filteredDirectives->filter($directiveFilter);
             }
 
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -83,7 +83,7 @@ class ArgumentSet
      *
      * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
      */
-    public function enhanceBuilder($builder, array $scopes, $directiveFilter = null)
+    public function enhanceBuilder($builder, array $scopes, Closure $directiveFilter = null)
     {
         if (empty($directiveFilter)) {
             $directiveFilter = function (Directive $directive): bool {
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Execution\Arguments;
 
+use Closure;
 use Nuwave\Lighthouse\Schema\Directives\SpreadDirective;
 use Nuwave\Lighthouse\Support\Contracts\ArgBuilderDirective;
 use Nuwave\Lighthouse\Support\Contracts\Directive;
@@ -77,20 +78,14 @@ class ArgumentSet
     /**
      * Apply ArgBuilderDirectives and scopes to the builder.
      *
-     * @param \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
-     * @param string[] $scopes
-     * @param \Closure $directiveFilter
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @param  string[]  $scopes
+     * @param  \Closure  $directiveFilter
      *
      * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
      */
     public function enhanceBuilder($builder, array $scopes, Closure $directiveFilter = null)
     {
-        if (empty($directiveFilter)) {
-            $directiveFilter = function (Directive $directive): bool {
-                return $directive instanceof ArgBuilderDirective;
-            };
-        }
-
         foreach ($this->arguments as $argument) {
             $value = $argument->toPlain();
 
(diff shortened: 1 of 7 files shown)


lorado:
Well... It makes sense why it fails - `@eq` directive defined where part of query it therefore `@can` was not able to fetch the model. The test ist actually a little confusing, as query is not really valid... anyway, I thought it would be good at this point not to respect all the argument directives, except soft deletes (`@trashed` directive). 

So I ended up with defining an additional optional argument on `enchanceBuilder()`, that overrides default directive filter, and in `@can` context I filter all directives and let only `@trashed` execute.

What do you think?

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -77,12 +77,20 @@ function (Directive $directive): bool {
...
+     * @param string[] $scopes
+     * @param \Closure $directiveFilter
+     *
      * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
      */
-    public function enhanceBuilder($builder, array $scopes)
+    public function enhanceBuilder($builder, array $scopes, $directiveFilter = null)
     {
+        if (empty($directiveFilter)) {
+            $directiveFilter = function (Directive $directive): bool {
```
spawnia: I think we should apply the extra filter in addition to filtering for `ArgBuilderDirective`, because the function afterwards can only continue to work with such instances.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -77,12 +77,20 @@ function (Directive $directive): bool {
...
-     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
-     * @param  string[]  $scopes
+     * @param \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
+     * @param string[] $scopes
+     * @param \Closure $directiveFilter
+     *
      * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
      */
-    public function enhanceBuilder($builder, array $scopes)
+    public function enhanceBuilder($builder, array $scopes, $directiveFilter = null)
```
spawnia:
```suggestion
    public function enhanceBuilder($builder, array $scopes, Closure $directiveFilter = null)
```

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -77,12 +77,20 @@ function (Directive $directive): bool {
     /**
      * Apply ArgBuilderDirectives and scopes to the builder.
      *
-     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
-     * @param  string[]  $scopes
+     * @param \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
```
spawnia: Formatting is off, there have to be two spaces inbetween.

src/Schema/Directives/CanDirective.php
```diff
@@ -80,7 +82,16 @@ public function handleField(FieldValue $fieldValue, Closure $next): FieldValue
             $fieldValue->setResolver(
                 function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($previousResolver) {
                     if ($find = $this->directiveArgValue('find')) {
-                        $modelOrModels = $this->getModelClass()::findOrFail($args[$find]);
+                        $modelOrModels = $resolveInfo
+                            ->argumentSet
+                            ->enhanceBuilder(
+                                $this->getModelClass()::query(),
+                                $this->directiveArgValue('scopes', []),
```
spawnia: Allowing scopes here is a hidden new feature. I don't see a concrete use case for it, so let's remove this option for now.

spawnia:
I agree that this should work and apart from some minor changes i requested, the method you used to implement this is fine with me. Well done.

There is an actual user-facing change in here that we should mention in the CHANGELOG. The internals don't really matter there, but you should mention that this particular use-case now works.

The docs for `@can` could use a mention of that behaviour as well, specifically in the part that describes the mechanics of the `find` argument.

src/Schema/Directives/CanDirective.php
lorado: Hm, right

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -100,13 +95,20 @@ public function enhanceBuilder($builder, array $scopes, Closure $directiveFilter
...
                 });
 
+            if (!empty($directiveFilter)) {
+                $filteredDirectives = $filteredDirectives->filter($directiveFilter);
+            }
+
+            $filteredDirectives->each(function (ArgBuilderDirective $argBuilderDirective) use (&$builder, $value) {
+                $builder = $argBuilderDirective->handleBuilder($builder, $value);
+            });
+
```
lorado: > I think we should apply the extra filter in addition to filtering for ArgBuilderDirective, because the function afterwards can only continue to work with such instances.

@spawnia do you mean like this?

spawnia:
Good improvement, thanks @lorado

PR #1043, opened 2019-11-12
Add `injectArgs` option to `@can` directive to pass along client defined   arguments to the policy check

### LINKED ISSUE #1002: Pass arguments fields to policy in @can directive
Let's say we need to read an argument from the query in order to allow/deny something in a Policy. Currently, we have access to the User of course,  and a name of the argument that is used to find a specific model and we can pass extra args according to the definition:
```
"""
Additional arguments that are passed to `Gate::check`. 
"""
args: [String!]
```
So far I can understand, they are just strings. There is no way to send any extra arguments from the query.

My proposal is to add a new field definition (I called 'fields' in the example) in order to send the query arguments ($args):
```
public function handleField(FieldValue $fieldValue, Closure $next): FieldValue
    {
        $previousResolver = $fieldValue->getResolver();

        return $next(
            $fieldValue->setResolver(
                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($previousResolver) {
                    $modelClass = $this->getModelClass();
                    $this->args = $args; // get the args to passthrough the gate
```
```
    protected function getAdditionalArguments(): array
    {
        $args = (array)  $this->directiveArgValue('args');
        $fields = $this->directiveArgValue('fields');
        if($fields) { // check if we need to add the args
            $args = array_merge([$this->args], $args);
        }

        return $args;
    }
```

Now we can do something like this in the policy now:
```
public function foo(User $user, $args)
    {
        $foo = Arr::get($args, 'my-integer');
        return $user->hasPermission('permission') && $foo > 400;
    }
```

### LINKED ISSUE #1004: Union types broken when cache is enabled
Using Unions in cache disabled environment works as expected.
When cache is enabled, same code that worked in previous environment does not work anymore

It should return result same as in cache=disabled environment.

**Steps to reproduce**

1. Create sample query type that returns Union type
2. Disable cache in your .env file LIGHTHOUSE_CACHE_ENABLE=false
3. Run query => Everything is ok.
4. Enable cache in your .env file LIGHTHOUSE_CACHE_ENABLE=true
5. Run query => It breaks

**Output/Logs**

```
{
  "errors": [
    {
      "debugMessage": "array_map(): Argument #2 should be an array",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ],
      "path": [
        "eventsCalendar",
        0,
        "events",
        0
      ],
      "trace": [
        {
          "file": "/var/www/vendor/nuwave/lighthouse/src/Schema/TypeRegistry.php",
          "line": 428,
          "function": "array_map(instance of Closure, instance of GraphQL\\Language\\AST\\NodeList(2))"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Type/Definition/UnionType.php",
          "line": 76,
          "call": "Nuwave\\Lighthouse\\Schema\\TypeRegistry::Nuwave\\Lighthouse\\Schema\\{closure}()"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Type/Definition/UnionType.php",
          "line": 59,
          "call": "GraphQL\\Type\\Definition\\UnionType::getTypes()"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Type/Schema.php",
          "line": 427,
          "call": "GraphQL\\Type\\Definition\\UnionType::isPossibleType(GraphQLType: Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 1343,
          "call": "GraphQL\\Type\\Schema::isPossibleType(GraphQLType: EventCalendarDayDataType, GraphQLType: Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 1038,
          "call": "GraphQL\\Executor\\ReferenceExecutor::ensureValidRuntimeType(GraphQLType: Event, GraphQLType: EventCalendarDayDataType, instance of GraphQL\\Type\\Definition\\ResolveInfo, instance of App\\Models\\Events\\Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 849,
          "call": "GraphQL\\Executor\\ReferenceExecutor::completeAbstractValue(GraphQLType: EventCalendarDayDataType, instance of ArrayObject(1), instance of GraphQL\\Type\\Definition\\ResolveInfo, array(4), instance of App\\Models\\Events\\Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 726,
          "call": "GraphQL\\Executor\\ReferenceExecutor::completeValue(GraphQLType: EventCalendarDayDataType, instance of ArrayObject(1), instance of GraphQL\\Type\\Definition\\ResolveInfo, array(4), instance of App\\Models\\Events\\Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 677,
          "call": "GraphQL\\Executor\\ReferenceExecutor::completeValueWithLocatedError(GraphQLType: EventCalendarDayDataType, instance of ArrayObject(1), instance of GraphQL\\Type\\Definition\\ResolveInfo, array(4), instance of App\\Models\\Events\\Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 952,
          "call": "GraphQL\\Executor\\ReferenceExecutor::completeValueCatchingError(GraphQLType: EventCalendarDayDataType, instance of ArrayObject(1), instance of GraphQL\\Type\\Definition\\ResolveInfo, array(4), instance of App\\Models\\Events\\Event)"
        },
        {
          "file": "/var/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php",
          "line": 821,
          "call": "GraphQL\\Executor\\ReferenceExecutor::completeListValue(GraphQLType: EventCalendarDayDataType, instance of ArrayObject(1), instance of GraphQL\\Type\\Definition\\ResolveInfo, array(3), instance of Illuminate\\Support\\Collection(1))"
        }

.... [rest of the exception]
```

</details></br>

**Environment**

Lighthouse Version: 4.4.0
Laravel Version: 6.2.0

faiverson:
- [x] Added or updated tests

Resolves #1002 

**Changes**
Added a new argument to avoid breaking changes with the current version.
Saying that I think `args` (getAdditionalArguments) is always a plain string, it doesn't have to much sense to me

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -479,7 +479,7 @@ directive @can(
   """
   Pass input variables as arguments that are passed to `Gate::check`. 
   """
-  input: Boolean!
+  injectArgs: Boolean!
 ) on FIELD_DEFINITION
 ```
 
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -531,11 +531,11 @@ type Mutation {
 }
 ```
 
-You can pass the input variables as arguments to the policy setting `input` argument
+You can pass the input variables as arguments to the policy setting `injectArgs` argument
 ```graphql
 type Mutation {
     createPost(input: PostInput): Post
-        @can(ability: "create", input: "true")
+        @can(ability: "create", injectArgs: "true")
 }
 ```
 Now you will have access to `PostInput` values in the policy 
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -476,6 +476,10 @@ directive @can(
   Additional arguments that are passed to `Gate::check`. 
   """
   args: [String!]
+  """
+  Pass input variables as arguments that are passed to `Gate::check`. 
+  """
+  input: Boolean!
 ) on FIELD_DEFINITION
 ```
 
@@ -526,6 +530,15 @@ type Mutation {
         @can(ability: "create", args: ["FROM_GRAPHQL"])
 }
 ```
+
+You can pass the input variables as arguments to the policy setting `input` argument
+```graphql
+type Mutation {
+    createPost(input: PostInput): Post
+        @can(ability: "create", input: "true")
+}
+```
+Now you will have access to `PostInput` values in the policy 
 
 Starting from Laravel 5.7, [authorization of guest users](https://laravel.com/docs/authorization#guest-users) is supported.
 Because of this, Lighthouse does **not** validate that the user is authenticated before passing it along to the policy.
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -459,6 +459,9 @@ class PostPolicy
 ```graphql
 """
 Check a Laravel Policy to ensure the current user is authorized to access a field.
+
+When `injectArgs` and `args` are used together, the client given
+arguments will be passed before the static args.
 """
 directive @can(
   """
@@ -471,22 +474,26 @@ directive @can(
   instance against which the permissions should be checked.
   """
   find: String
+
+  """
+  Pass along the client given input data as arguments to `Gate::check`. 
+  """
+  injectArgs: Boolean = false
+
+  """
+  Statically defined arguments that are passed to `Gate::check`.
   
+  You may pass pass arbitrary GraphQL literals,
+  e.g.: [1, 2, 3] or { foo: "bar" }
   """
-  Additional arguments that are passed to `Gate::check`. 
-  """
-  args: [String!]
-  """
-  Pass input variables as arguments that are passed to `Gate::check`. 
-  """
-  injectArgs: Boolean!
+  args: Mixed
 ) on FIELD_DEFINITION
 ```
 
 ### Examples
 
 In `find` parameter you may specify an input argument which is used to find a specific model
-instance by primary key against which the permissions should be checked.
+instance by primary key against which the permissions should be checked:
 
 ```graphql
 type Query {
@@ -504,7 +511,7 @@ class PostPolicy
 }
 ```
 
-It also works with soft deleted models in combination with `@softDeletes` directive.
+It also works with soft deleted models in combination with `@softDeletes` directive:
 
 ```graphql
 type Query {
@@ -522,7 +529,7 @@ type Mutation {
 }
 ```
 
-You can pass additional arguments to the policy checks by specifying them as `args`.
+You can pass additional arguments to the policy checks by specifying them as `args`:
 
 ```graphql
 type Mutation {
@@ -531,14 +538,17 @@ type Mutation {
 }
 ```
 
-You can pass the input variables as arguments to the policy setting `injectArgs` argument
+You can pass along the client given input data as arguments to the policy checks
+with the `injectArgs` argument:
+
 ```graphql
 type Mutation {
     createPost(input: PostInput): Post
         @can(ability: "create", injectArgs: "true")
 }
 ```
-Now you will have access to `PostInput` values in the policy 
+
+Now you will have access to `PostInput` values in the policy. 
 
 Starting from Laravel 5.7, [authorization of guest users](https://laravel.com/docs/authorization#guest-users) is supported.
 Because of this, Lighthouse does **not** validate that the user is authenticated before passing it along to the policy.
--- a/src/Schema/Directives/CanDirective.php
+++ b/src/Schema/Directives/CanDirective.php
@@ -22,6 +22,11 @@ class CanDirective extends BaseDirective implements FieldMiddleware, DefinedDire
     protected $gate;
 
     /**
+     * @var array
+     */
+    protected $args;
+
+    /**
      * CanDirective constructor.
      * @param  \Illuminate\Contracts\Auth\Access\Gate  $gate
      * @return void
@@ -63,6 +68,11 @@ directive @can(
   Additional arguments that are passed to `Gate::check`. 
   """
   args: [String!]
+    """
+  Send input data as arguments to the policy. 
+  Set false by default
+  """
+  input: Boolean!
 ) on FIELD_DEFINITION
 SDL;
     }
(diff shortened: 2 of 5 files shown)


tests/Utils/Policies/UserPolicy.php
```diff
@@ -32,4 +32,9 @@ public function dependingOnArg($viewer, bool $pass): bool
     {
         return $pass;
     }
+
+    public function severalArgs($user, array $input): bool
+    {
+        return $input['key1'] === 'foo' && $input['key2'] === 'foo2' && $input['key2'] === 'foo3';
```
lorado: you have a typo here. The last equal test should be with key3, not key2

docs/master/api-reference/directives.md
```diff
@@ -476,6 +476,10 @@ directive @can(
   Additional arguments that are passed to `Gate::check`. 
   """
   args: [String!]
+  """
+  Pass input variables as arguments that are passed to `Gate::check`. 
+  """
+  input: Boolean!
```
spawnia: Can we find a name for this that is more clear? How about `injectArgs`?

src/Schema/Directives/CanDirective.php
```diff
@@ -63,6 +68,11 @@ public static function definition(): string
   Additional arguments that are passed to `Gate::check`. 
   """
   args: [String!]
+    """
```
spawnia:
```suggestion
  """
```

src/Schema/Directives/CanDirective.php
```diff
@@ -146,6 +157,9 @@ protected function authorize($user, $model): void
      */
     protected function getAdditionalArguments(): array
     {
-        return (array) $this->directiveArgValue('args');
+        $directiveArgs = (array) $this->directiveArgValue('args');
+        $inputArgs = $this->directiveArgValue('input') === true ? [$this->args] : [];
```
spawnia:
```suggestion
        $inputArgs = $this->directiveArgValue('input')
            ? [$this->args]
            : [];
```

tests/Unit/Schema/Directives/CanDirectiveTest.php
```diff
@@ -156,6 +156,41 @@ public function testProcessesTheArgsArgument(): void
...
+        }
+        ';
+
+        $variables = [
+            'key1' => 'foo',
+            'key2' => 'foo2',
+            'key3' => 'foo3',
+        ];
+
+        $this->postGraphQL(
```
spawnia: Using variables is really verbose.

Please use `->graphQL()` and literal query syntax.

tests/Unit/Schema/Directives/CanDirectiveTest.php
```diff
@@ -156,6 +156,41 @@ public function testProcessesTheArgsArgument(): void
         ')->assertErrorCategory(AuthorizationException::CATEGORY);
     }
 
+    public function testSendInputArgumentToPolicy(): void
+    {
+        $this->be(new User);
+        $this->schema = '
+        type Query {
+            users(key1: String, key2: String, key3: String): [User]!
```
spawnia: Using a single argument is enough for testing, let's keep it simple.

docs/master/api-reference/directives.md
lorado: Yeah, I had the same thought. +1 for `injectArgs`

spawnia:
Can you clarify the semantics of what happens if the statically defined `args` and the injected arguments collide? That should be backed up by a test.

lorado:
> Can you clarify the semantics of what happens if the statically defined args and the injected arguments collide?

If I understand it right, it can not collide. `args` argument directive is an array of strings, so the keys are numeric, but query arguments have string keys. After merging all entries should persist, I think...

tests/Utils/Policies/UserPolicy.php
faiverson: Hmm, then I probably needs to change the assert too

faiverson:
Discussion aside, should we have the argument **args**? So far I can see it is always an array of hardcoded strings. It doesn't have too much sense...
I didn't want to use the same name (I think is the best name) and override the current functionality. But I don't think **args** is very useful. I'd replace **args**  with **injectArgs**

src/Schema/Directives/CanDirective.php
```diff
@@ -63,6 +68,11 @@ public static function definition(): string
   Additional arguments that are passed to `Gate::check`. 
   """
   args: [String!]
+  """
+  Send input data as arguments to the policy. 
+  Set false by default
+  """
+  injectArgs: Boolean!
```
spawnia:
```suggestion
  injectArgs: Boolean = false
```

spawnia:
I am not planning to make a new major version with breaking changes soon.

Adding static values can be useful. The directive definition does lie a bit, actually you can pass a map as well, e.g.: `@can(args: { foo: "bar" })`. That means there can be conflicts, we have to define which is preferred.

I think it makes sense to prefer the static values, since that makes it more predictable that those are always passed and allows overwriting client given values for certain scenarios. Doing it the other way around is basically the same as setting default values, which is already possible in the argument definition itself.

docs/master/api-reference/directives.md
enzonotario: I think it should be `injectArgs`

faiverson:
> I am not planning to make a new major version with breaking changes soon.
> 
> Adding static values can be useful. The directive definition does lie a bit, actually you can pass a map as well, e.g.: `@can(args: { foo: "bar" })`. That means there can be conflicts, we have to define which is preferred.
> 
> I think it makes sense to prefer the static values, since that makes it more predictable that those are always passed and allows overwriting client given values for certain scenarios. Doing it the other way around is basically the same as setting default values, which is already possible in the argument definition itself.

Ok, so we need to keep the args as it is right now.
I made the new test when we have both args and injectArgs.

spawnia:
I did not realize that it is possible to just pass multiple arguments to the checks. That means we do not have to merge, nice.

@faiverson had a pass over the PR, let me know if the changes are fine.

PR #1046, opened 2019-11-16
Add Dark Theme for Docs

omacranger:
- [x] ~~Added or updated tests~~
- [x] Updated CHANGELOG.md

**Changes**

Changes the Vuepress theme to be an extension of the default theme with dark mode support using prefers-color-scheme.

Other changes: 
- Switch to Vuepress theme with prefers-color-scheme / dark mode support
- Add CSS variables for lighthouse color scheme
- Migrate font @import into <head> declaration for performance

**Breaking changes**

Docs theme will change when using an OS & browser with dark mode support. Otherwise not that I'm aware of.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `syncWithoutDetaching` option for BelongsToMany and MorphToMany relationships https://github.com/nuwave/lighthouse/pull/1031
 - Add `injectArgs` option to `@can` directive to pass along client defined
   arguments to the policy check https://github.com/nuwave/lighthouse/pull/1043
+- Dark Mode support for Lighthouse Documentation when user `prefers-color-scheme` is set to `dark` https://github.com/nuwave/lighthouse/pull/1046
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `syncWithoutDetaching` option for BelongsToMany and MorphToMany relationships https://github.com/nuwave/lighthouse/pull/1031
 - Add `injectArgs` option to `@can` directive to pass along client defined
   arguments to the policy check https://github.com/nuwave/lighthouse/pull/1043
+- Dark Mode support for Lighthouse Documentation when user `prefers-color-scheme` is set to `dark` https://github.com/nuwave/lighthouse/pull/1046
```
spawnia: Thanks for taking the time and care to add that in.

The change is great, but it does not belong in `CHANGELOG.md`, only functional changes to the package itself are mentioned there.

docs/.vuepress/config.js
```diff
@@ -12,9 +12,16 @@ module.exports = {
...
+        },],
+        ['link', {
+            rel: 'stylesheet',
+            type: 'text/css',
+            href: 'https://fonts.googleapis.com/css?family=Miriam+Libre:400,700'
+        },],
     ],
+    theme: 'default-prefers-color-scheme',
     themeConfig: {
+        defaultTheme: 'light',
```
spawnia: I would be fine with defaulting to dark 😄

CHANGELOG.md
omacranger: Makes sense. Removed.

docs/.vuepress/config.js
omacranger: I'll let you guys be the one to change that! I think light is a sensible default for those in well-lit environments who might not have access to a browser / OS with `prefers-color-scheme` support.

PR #1050, opened 2019-11-19
Fix text color on code blocks when light color scheme is explicitly set

omacranger:
**Changes**

Fixes `pre` / `code` text color when light color scheme is explicitly set


spawnia:
Thanks!

PR #1053, opened 2019-11-20
Add OBJECT to the @middleware directive

StelianAndrei:
In some IDEs such as PHPStorm, when the middleware directive is used when extending the Query type for example, and error is shown because the directive is only defined on FIELD_DEFINITION.

Example: given the following

```
extend type Query @middleware(checks: ["auth:api"]) {}
```

The error shown is `Query' [@1:1] tried to use a directive 'middleware' in the 'OBJECT' location but that is illegal`

Adding OBJECT to the definition fixes this.

### DIFF
--- a/docs/3.7/api-reference/directives.md
+++ b/docs/3.7/api-reference/directives.md
@@ -1254,18 +1254,11 @@ directive @method(
 
 ## @middleware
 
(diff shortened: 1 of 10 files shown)


spawnia:
That's great, good find.

Can you copy the change over to docs? Directive definitions are mirrored there from versions `3.7` onwards.

PR #1062, opened 2019-11-26
Extend `@rename` directive to work with arguments and input fields

### LINKED ISSUE #521: Use `@rename` on arguments/input object values
Hi All,

I've dug into the source code and can't really find answer, does the rename directive only apply to 'normal' types. IE. 
`type User {...}`

I was hoping I could use it on an Input to convert from camel case (in the schema) to snake case. However my experience using the `@rename` directive in `input`s has been that it doesn't work/rename the fields as it does in `type` sets. Wondering if that is intentional or if I might not be doing something right.  

Here is what I'm doing currently that isn't renaming camel case to snake case

```graphql
input CreateCustomerInput {    
    firstName: String! @rename(attribute: "first_name") @rules(apply: ["required","string"]) 
    lastName: String! @rename(attribute: "last_name") @rules(apply: ["required","string"])
    email: String! @rules(apply: ["required","string","email","unique:customers,email"])
    addresses: [GeneralAddressInput]
}
```

JasonTheAdams:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #521

**Changes**

This extends the `@rename` directive so that it can be used on input fields. Which allows for something like the following:

```graphql
input {
    firstName: String @rename(attribute: "first_name")
}
```

There is likely more to be done, but I wanted to at least take a first stab and get the ball rolling on this. 😄 

**Breaking changes**

This should not break anything as the `@rename` directive has not been changed, only allowed to work in a new context.


JasonTheAdams:
@spawnia Took a first attempt at this. I still need to do some tests. One thing I was thinking about was attempting to do the renaming during the spread method so we're not immutably rebuilding the `ArgumentSet` twice, but it's much easier to perform the changes once the input fields have been spread. Thoughts?

spawnia:
Thanks for starting this, looks good so far.

Let's keep the methods separate. The input objects are probably going to be really small, having a second pass over them won't hurt performance noticably.

src/Schema/Directives/RenameDirective.php
```diff
@@ -28,7 +28,7 @@ public static function definition(): string
   value can be retrieved from.
   """
   attribute: String!
-) on FIELD_DEFINITION
+) on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
```
spawnia: That definition will have to be copied over to the docs.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -75,6 +76,31 @@ function (Directive $directive): bool {
...
+    public function rename(): self
+    {
+        $argumentSet = new self();
+        $argumentSet->directives = $this->directives;
+
+        foreach ($this->arguments as $name => $argument) {
+            $renameDirective = $argument->directives->first(function ($directive) {
+                return $directive instanceof RenameDirective;
+            });
+
```
spawnia: Let's make sure this is applied in nested inputs as well and recurse down here.

src/Schema/Directives/RenameDirective.php
```diff
@@ -41,6 +41,24 @@ public static function definition(): string
...
+    /**
+     * Retrieves the attribute argument for the directive.
+     *
+     * @return string
+     *
+     * @throws DirectiveException
+     */
+    public function getAttribute(): string
     {
         $attribute = $this->directiveArgValue('attribute');
```
spawnia: While we are at it, let's throw a `DefinitionException` here instead of a `DirectiveException`

src/Schema/Factories/FieldFactory.php
```diff
@@ -184,7 +184,7 @@ function () use ($argumentMap, $resolverWithMiddleware) {
                 $this->flushValidationErrorBuffer();
 
                 $argumentSet = $this->typedArgs->fromResolveInfo($this->args, $this->resolveInfo);
-                $spreadArguments = $argumentSet->spread();
+                $spreadArguments = $argumentSet->spread()->rename();
```
spawnia:
```suggestion
                $spreadArguments = $argumentSet->spread();
                $renamedArguments = $spreadArguments->rename();
```

Another option would be to find a generic term for the result of chaining the operations, maybe `$modifiedArguments`?

JasonTheAdams:
I added a unit test for the `ArgumentSet::rename()` method. I still want to add a feature test to make sure that it all works when put together.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -75,12 +76,47 @@ function (Directive $directive): bool {
...
+        $argumentSet = new self();
+        $argumentSet->directives = $this->directives;
+
+        foreach ($this->arguments as $name => $argument) {
+            if ($argument->value instanceof self) {
+                // recurse to nested inputs
+                $argument->value = $argument->value->rename();
+            }
+
+            if (empty($argument->directives)) {
```
spawnia: No need for that check, the property is always initialized as an empty collection.

src/Schema/Directives/RenameDirective.php
```diff
@@ -38,22 +38,36 @@ public static function definition(): string
      * @param  \Nuwave\Lighthouse\Schema\Values\FieldValue  $fieldValue
      * @return \Nuwave\Lighthouse\Schema\Values\FieldValue
      *
-     * @throws \Nuwave\Lighthouse\Exceptions\DirectiveException
+     * @throws \Nuwave\Lighthouse\Exceptions\DefinitionException
```
spawnia: That annotation has to go, we only annotate in the exact method where exceptions are thrown.

tests/Unit/Execution/Arguments/ArgumentSetTest.php
```diff
@@ -120,4 +126,64 @@ public function testListOfInputObjectsToArray(): void
             $argumentSet->toArray()
         );
     }
+
+    public function testRenameInput(): void
+    {
+        $generateRenameDirective = function ($newName) {
```
spawnia: Using the `PartialParser` is much more simple and terse then this.

tests/Unit/Execution/Arguments/ArgumentSetTest.php
```diff
@@ -120,4 +126,64 @@ public function testListOfInputObjectsToArray(): void
...
+        $userSet->arguments = [
+            'firstName' => $firstName,
+            'lastName' => $lastName,
+            'post' => $postArg,
+        ];
+
+        $renamedSet = $userSet->rename();
+
+        $this->assertSame([
+            'first_name' => $firstName,
```
spawnia: What are the three different args testing?

The tests should cover more different cases rather then repeating the same test case.

tests/Unit/Execution/Arguments/ArgumentSetTest.php
JasonTheAdams: @spawnia I think I need some direction on this. I've looked into the `PartialParser` and have attempted to use it. What's confusing is that when I use `PartialParser::directive` I get a `DirectiveNode`, as opposed to a `BaseDirective` descendent. In fact, I can't find the `RenameDirective` anywhere in the result when using that. I also tried using `PartialParser::InputObjectTypeDefinition`, but this also returns nothing but nodes.

The `@spread` test isn't using the `PartialParser`, but is just instantiating the directive, which is likely fine because it doesn't need any arguments. I clearly need to access the `RenameDirective` within the `ArgumentSet::rename()` method, but when I parse inputs with the directive I only get nodes.

I'm clearly missing something, here. Can you please help me to understand where nodes connect with the Lighthouse `BaseDirective` descendents?

enzonotario: Hey @JasonTheAdams I'd like to help you to also continue learning.. may I touch your PR?

JasonTheAdams: @enzonotario I don't mind at all! I look forward to seeing what you change. If you wouldn't mind, please provide some explanation so I can learn for the future. 😄

enzonotario: well I don't promise anything but will try todo my best! :)

enzonotario: Hey folks, I have changed this to using PartialParser, although for now I don't really understand at all about the directive generation, but it works. I hope it is ok. Otherwise, just sorry :)

spawnia: The `BaseDirective` requires a parsed AST definition of a directive under the hood to read the arguments from. The easiest way is to `new` up an instance of whatever directive you want to test and `hydrate` it with the result of a call to `PartialParser`.

spawnia:
Did a pass over the PR, awaiting final feedback and will merge then.

JasonTheAdams:
@spawnia This looks great! Man, I was so close! I wondered if creating a `FieldDefinitionNode` and hydrating it was what I needed to do. Should have tried! Hahah!

Thanks for hopping in and helping! Glad I learned from this and we got a helpful, requested feature in place. 😄

spawnia:
Thanks @JasonTheAdams and @enzonotario for the great collaboration on this PR. Looking forward to more :)

PR #1069, opened 2019-11-30
Add INPUT_FIELD_DEFINITION to orderBy directive location

### LINKED ISSUE #1048: orderBy Directive Location in Documentation / ide-helper
**Describe the bug**

`@orderBy` directive location doesn't seem to include all locations where the directive can be applied. It appears `INPUT_FIELD_DEFINITION` should be included in cases where someone is using an `input` type and the `@spread` directive.

Directive location should be updated to include all valid locations. It would also be nice to have the documentation for this directive updated to include some examples on how to use them in conjunction with input types since filters / inputs are likely a common use case.

**Steps to reproduce**

1. Apply `orderBy` directive on input type where `@spread` is being applied. OrderBy works as expected but isn't listed as a 'valid' location in the schema output of the ide-helper command.

**Environment**

Lighthouse Version: 4.6
Laravel Version: 5.8.35

omacranger:
- [x] Updated CHANGELOG.md

Resolves #1048 

**Changes**

Adds `INPUT_FIELD_DEFINITION` to the directive location for `orderBy` directive and provides example on how to use.

**Breaking changes**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   using `@softDeletes` directive. https://github.com/nuwave/lighthouse/pull/1042
 - Improve the error message for missing field resolvers by offering a solution https://github.com/nuwave/lighthouse/pull/1045
 - Throw `DefinitionException` when missing a type in the type registry https://github.com/nuwave/lighthouse/pull/1066
+- Add `INPUT_FIELD_DEFINITION` to `orderBy` directive location https://github.com/nuwave/lighthouse/pull/1069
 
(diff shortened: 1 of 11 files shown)


omacranger:
@spawnia After working with some other directives to handle functionality, I assume a few of these location declarations are out of date - `@builder` specifically and likely others - that can also be used on `INPUT_FIELD_DEFINITION` when used together with the `@spread` operator. Do you or someone on the team have a potential list for those and I'll have them updated along with this PR?

spawnia:
Great job, i like the additional example.

> Do you or someone on the team have a potential list for those and I'll have them updated along with this PR?

Not that i know of. Let's keep improving the docs iteratively, focused PR's like this one are perfect.

PR #1082, opened 2019-12-11
Enable chained rule provider directives (`ProvidesRules`) to merge the rules before validating

### LINKED ISSUE #1061: Recursively merge rules
I've created a custom directive that extends the rule directive. It is a directive that mutates the incoming data before I validate it. For example:

```graphql

input ManageContractTermsRelation {
  create: [CreatePaymentTerm]  # With percentage attribute
  update: [UpdatePaymentTerm] # With percentage attribute
}

input CreateContract {
  terms: ManageContractTermsRelation @relationRules(apply: ["custom_sum:percentage,100"])
}
```
This `@relationRules` directive combines the `create` and `update` values of the input and merges them with the current state of the database. This makes it easier to validate complicated requests:

```php
public function passes($attribute, Collection $value, string $column, string $expected): bool
{
    return $value->sum($column) === $expected;
}
```

The problem I encountered was when I tried to add normal rules to the same attribute:

```graphql
# ...

input CreateContract {
  terms: ManageContractTermsRelation @rules(apply: ["required"]) @relationRules(apply: ["custom_sum:percentage,100"])
}
```

When I do this, the `custom_sum` rule does not get applied to the rules array. After some research I found the cause of the problem in the `FieldFactory::class`. This class merges new rules with the existing ones. But doesn't do it recursively. This makes it impossible to have multiple instances of a `RulesDirective` on the same attribute.

The fix I suggest is a small one. If we change lines 324 and 325 at `src/Schema/Factories/FieldFactor` to use the `array_merge_recursive` method, my problem is solved.

```php
$rulesA = [
  'terms' => [
    'required',
  ],
];

$rulesB = [
  'terms' => [
    'relation:custom_rule,percentage,100',
  ],
];

// Wrong
array_merge($rulesA, $rulesB);
/*
[
  "terms" => [
    "relation:custom_rule,percentage,100",
  ],
]
*/

// Correct
array_merge_recursive($rulesA, $rulesB);
/*
[
  "terms" => [
    "required",
    "relation:custom_rule,percentage,100",
  ],
]
*/

```
I tried to refactor my custom rules directive to a generic one that I could use to insert all rules at the same time. But because some of my rules require data formatting before validating. This isn't going to work without some workarounds or hacks. The proposed solution is (in my opinion) the most clean one.

I am willing to create a PR when this issue is accepted. I will create one for master, and one for older versions that can have this update without breaking anything. I will also write tests for this.

markwalet:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1061 

Enables chained rule provider directives (`ProvidesRules`) to merge the rules before the validation is executed.


src/Schema/Factories/FieldFactory.php
```diff
@@ -327,6 +332,20 @@ protected function handleArgDirectives(
         }
     }
 
+    protected function gatherValidationDirectives(Collection $directives)
+    {
+        $validators = collect();
+        while ($directive = $directives->first()) {
```
spawnia: If i am not mistaken, all that manual iteration code could be replaced by a single call to `$directives->filter()`

tests/Integration/ValidationTest.php
```diff
@@ -522,6 +522,27 @@ public function testCombinesArgumentValidationByPausingAndResuming(): void
...
+        type Mutation {
+            withMergedRules(bar: String @rules(apply: ["min:40"]) @customRules(apply: ["bool"])): User @create
+        }
+
+        type User {
+            id: ID
+            name: String
+        }
+        ';
+        $this->graphQL('
```
spawnia:
```suggestion

        $this->graphQL('
```

tests/Integration/ValidationTest.php
```diff
@@ -522,6 +522,27 @@ public function testCombinesArgumentValidationByPausingAndResuming(): void
         );
     }
 
+    public function testCombinesArgumentValidationWhenGrouped(): void
+    {
+        $this->schema .= '
```
spawnia:
```suggestion
        $this->schema .= /* @lang GraphQL */ '
```

tests/Integration/ValidationTest.php
spawnia:
```suggestion
        $this->graphQL(/* @lang GraphQL */'
```

tests/Integration/ValidationTest.php
```diff
@@ -522,6 +522,27 @@ public function testCombinesArgumentValidationByPausingAndResuming(): void
...
+        type Mutation {
+            withMergedRules(bar: String @rules(apply: ["min:40"]) @customRules(apply: ["bool"])): User @create
```
spawnia:
```suggestion
            withMergedRules(
                bar: String @rules(apply: ["min:40"]) @customRules(apply: ["bool"])
            ): User @create
```

src/Schema/Factories/FieldFactory.php
markwalet: I wanted to use the approach that was already there. With shifting it off the existing collection. Because I was afraid that a normal filter (my first approach) would do something weird memorywise.

tests/Integration/ValidationTest.php
markwalet: This isn't used anywhere. I'm simply copying codestyle from the code above.

tests/Integration/ValidationTest.php
markwalet: This isn't used anywhere. I'm simply copying codestyle from the code above.

spawnia: We are working towards including those annotations. Please add them in parts of code you touched.

src/Schema/Factories/FieldFactory.php
spawnia: I see. Well i guess that avoids adding the validation rules multiple time.

`partition()` might be a better fit for that case.

markwalet: I first used a partition, but this makes it harder to implement the following usecase. Which was a existing skipped test:
`testCombinesArgumentValidationByPausingAndResuming`:
```
type Mutation {
    createUser(
        foo: String @rules(apply: ["max:5"]) @trim @rules(apply: ["min:4"])
    ): User
        @create
}
```

With a partition, I'd combine all validator directives at once, without making mutations to the data, which should be possible in the future if I'm interpreting this test correctly. This is why I only get the rules directives that occur directly after each other.

The only downside now, is that you cannot retrieve all validation errors in one batch. But that wasn't the case before my PR, and that isn't the bug I'm trying to solve with this PR.

markwalet: Don't get me wrong, I tried to fix it with a partition because I thought that was the way to go haha.

But like your skipped test says:

> This should work once we can reliably depend upon repeatable directives.

This is not yet the case, my PR is a small step towards grouping all validation together, but for now it isn't possible to group them all together without creating some new bugs.

spawnia: That reasoning makes sense. Can you summarize it in a few lines and embed it into the code as an explanatory comment?

markwalet: Done

src/Schema/Factories/FieldFactory.php
```diff
@@ -327,6 +332,23 @@ protected function handleArgDirectives(
         }
     }
 
+    protected function gatherValidationDirectives(Collection $directives)
+    {
+        // We only get the validator directives that are directly following on the latest validator
```
spawnia: Nicely done, even formatted in the canonical Laravel way. Love the attention to detail :)

spawnia:
Great work @markwalet, thank you for this contribution!

PR #1097, opened 2019-12-25
Fix Post's author relation

Ilyes512:
The Post's relation `author` does not work unless you specify the `$foreignKey`. Without it, it will try to guess the foreign key based on the method name (author) so it  will try to find `author_id` instead.

The linked repo does not seem to match the tutorial: https://github.com/nuwave/lighthouse-tutorial/blame/master/app/Post.php#L9-L12

N/A

- ~[ ] Added or updated tests~
- ~[ ] Added Docs for all relevant versions~
- ~[ ] Updated CHANGELOG.md~

**Changes**

Fixed relation issue in tutorial.

**Breaking changes**

N/A

### DIFF
--- a/docs/tutorial/README.md
+++ b/docs/tutorial/README.md
@@ -307,9 +307,9 @@ naming - the type name `Post` is also the name of our Model - and the use of ser
 We add additional type definitions that clearly define the shape of our data: 
 
 ```graphql
-type Query{
+type Query {
     posts: [Post!]! @all
-    post (id: Int! @eq): Post @find
+    post(id: Int! @eq): Post @find
 }
 
 type User {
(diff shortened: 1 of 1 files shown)


docs/tutorial/README.md
```diff
@@ -167,7 +167,7 @@ class Post extends Model
 {
     public function author(): BelongsTo
     {
-        return $this->belongsTo(User::class);
+        return $this->belongsTo(User::class, 'user_id');
```
spawnia: Let's change the migration to name the column `author_id` instead, that way we match the convention that Laravel expects.

spawnia:
Thanks for the PR, great find. I think we should use Laravel conventions if possible, can you please adapt the PR?

docs/tutorial/README.md
Ilyes512: Done

spawnia:
Nicely done, thank you!

Issue #1047, opened 2019-11-18
Order of expressions in OR should not affect SQL query


drtheuns:
**Describe the bug**

When passing an `OR` query, the order of the sub-expressions seem to matter. For example, when passing something like `OR(expr1, AND(expr2, expr3))` it will result in the SQL: `(expr1 AND (expr2 AND expr3))`, while `OR(AND(expr2, expr3), expr1)` will result in the correct SQL: `((expr2 AND expr3) or expr1)`. See the example queries below for a more detailed example.

**Expected behavior/Solution**

The order of the expressions passed to OR should not matter.

**Steps to reproduce**

1. Setup project with some models and a database.
2. Add the query to the graphql file:
```
extend type Query {
    clients(filter: WhereConstraints @whereConstraints orderBy: [OrderByClause!] @orderBy): [Client!]! @paginate(type: "paginator" model: "App\\Models\\Client")
}
```
3. Execute the query:
```graphql
query {
  clients(
    filter: {
      OR: [
        {
          column: "name",
          operator: LIKE,
          value: "Victor%"
        },
        {
          AND: [
            {
              column: "first_name",
              operator: EQ,
              value: "Joy"
            },
            {
              column: "last_name",
              value: EQ,
              value: "Cummings"
            }
          ]
        }
      ]
    }
    count: 10
  ) {
    data {
      id,
      name
    }
  }
}
```

4. Observe the SQL: 
```sql
select count(*) as aggregate from `clients` where (`name` LIKE 'Victor%' and (`first_name` = 'Joy' and `last_name` = 'Cummings'));
--                                            expected this to be "or"   ^^^

```

5. Switch the order of the OR expressions:
```graphql
query {
  clients(
    filter: {
      OR: [
        {
          AND: [
            {
              column: "first_name",
              operator: EQ,
              value: "Joy"
            },
            {
              column: "last_name",
              value: EQ,
              value: "Cummings"
            }
          ]
        },
        {
          column: "name",
          operator: LIKE,
          value: "Victor%"
        }
      ]
    }
    count: 10
  ) {
    data {
      id,
      name
    }
  }
}
```
6. Observe the correct SQL:
```sql
select * from `clients` where ((`first_name` = 'Joy' and `last_name` = 'Cummings') or `name` LIKE 'Victor%') limit 10 offset 0
```

**Environment**

Lighthouse Version: 4.6
Laravel Version: v5.7.28

spawnia:
[...]

PR #1099, opened 2019-12-27
Add support in `@whereConstraints` for IN, IS NULL and BETWEEN operators

### LINKED ISSUE #1093: Add support in @whereConstraints for IN, IS NULL and BETWEEN
Right now in the @whereConstraints directive we have support for this:
```
enum Operator {
  EQ @enum(value: "=")
  NEQ @enum(value: "!=")
  GT @enum(value: ">")
  GTE @enum(value: ">=")
  LT @enum(value: "<")
  LTE @enum(value: "<=")
  LIKE @enum(value: "LIKE")
  NOT_LIKE @enum(value: "NOT_LIKE")
```
I'm proposing to add more operator and be able to have these:
```
enum Operator {
  EQ @enum(value: "=")
  NEQ @enum(value: "!=")
  GT @enum(value: ">")
  GTE @enum(value: ">=")
  LT @enum(value: "<")
  LTE @enum(value: "<=")
  LIKE @enum(value: "LIKE")
  NOT_LIKE @enum(value: "NOT_LIKE")
  IN @enum(value: "In")
  NOT_IN @enum(value: "NotIn")
  BETWEEN @enum(value: "Between")
  NOT_BETWEEN @enum(value: "NotBetween")
  NULL @enum(value: "Null")
  NOT_NULL @enum(value: "NotNull")**
}
```
There is no way to ask if something is NULL or if an ID is IN(1,2,3,4)
it will be a very easy change:
Since Eloquent support shortcuts like whereNot, whereNull, whereIn, whereNotIn, etc
The change should be at the end of the **handleBuilder** method changing the **$where** dynamically to accept these shortcuts and probably the parameters too:
```
$where = 'whereNull';
$builder->{$where}(
                $column,
                $whereConstraints['value']
            );
```

let me know your thoughts!

faiverson:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1093
Resolves https://github.com/nuwave/lighthouse/issues/1047
Resolves https://github.com/nuwave/lighthouse/issues/1115
Resolves https://github.com/nuwave/lighthouse/issues/1120

**Changes**

- Support `IN`, `BETWEEN` and `IS NULL` operators
- Apply nested `OR` conditions in `@whereConstraints` correctly
- Register the operator enum for `@whereConstraints` programmatically and allow  overwriting it through a service provider.

**Breaking changes**

The `Operator` is now included by default, the name defaults to `SQLOperator`.
Users who changed the default `Operator` enum will have to integrate those changes into a new custom class. You may register a custom `\Nuwave\Lighthouse\WhereConstraints\Operator` through a service provider.

`@whereConstraints` is still marked as experimental, so this warrants no major version bump.


chescos:
This looks awesome! Looking forward to trying it out.

docs/master/api-reference/directives.md
```diff
@@ -2867,57 +2864,79 @@ Install the dependency [mll-lab/graphql-php-scalars](https://github.com/mll-lab/
...
-    LIKE @enum(value: "LIKE")
-    NOT_LIKE @enum(value: "NOT_LIKE")
+"Dynamic WHERE constraints for the `where` argument on the query `people`.
+input PeopleWhereWhereConstraints {
+    column: PeopleWhereColumn
+    operator: SQLOperator = EQ
+    value: Mixed
+    AND: [PeopleWhereWhereConstraints!]
+    OR: [PeopleWhereWhereConstraints!]
+    NOT: [PeopleWhereWhereConstraints!]
```
CPSibo: I assumed the duplicated word `Where` was a typo in #1119 and changed it to `PeopleWhereConstraints`. Is that correct?

docs/master/api-reference/directives.md
```diff
@@ -2867,57 +2864,79 @@ Install the dependency [mll-lab/graphql-php-scalars](https://github.com/mll-lab/
...
-enum Operator {
-    EQ @enum(value: "=")
-    NEQ @enum(value: "!=")
-    GT @enum(value: ">")
-    GTE @enum(value: ">=")
-    LT @enum(value: "<")
-    LTE @enum(value: "<=")
-    LIKE @enum(value: "LIKE")
-    NOT_LIKE @enum(value: "NOT_LIKE")
+"Dynamic WHERE constraints for the `where` argument on the query `people`.
```
CPSibo: You still have the unterminated string here that'll throw an error if people copy this code into their own schema.

spawnia:
@faiverson your PR was a great starting point, i went ahead and redid some of the implementation.

The discussion about validating the values of the new operator types (see https://github.com/nuwave/lighthouse/issues/1093) was great, but i realized that would require complicate this feature a lot. I guess we should be fine by just letting Laravel/the database take care of that for now. Making a totally type-safe GraphQL input for that would require polymorphic input types to be somewhat practical.

I think the sanitizing you added is not necessary/potentially harmful, as Laravel is always properly binding the values. Correct me if i am wrong.

docs/master/api-reference/directives.md
chescos: The `WhereConstraints` part at the end is static, `People` is based on the Type, and `Where` is based on the field name. So I assume it's not a typo, it's just that the field is named `where`.

spawnia: We might simplify this by moving `@whereConstraints` to the field level and only ever having a single generated argument, see https://github.com/nuwave/lighthouse/issues/1121

src/WhereConstraints/SQLOperator.php
```diff
@@ -0,0 +1,102 @@
...
+    LT @enum(value: "<")
+    LTE @enum(value: "<=")
+    LIKE @enum(value: "LIKE")
+    NOT_LIKE @enum(value: "NOT_LIKE")
+    IN @enum(value: "In")
+    NOT_IN @enum(value: "NotIn")
+    BETWEEN @enum(value: "Between")
+    NOT_BETWEEN @enum(value: "NotBetween")
+    IS_NULL @enum(value: "Null")
+    IS_NOT_NULL @enum(value: "NotNull")
```
spawnia: @faiverson renamed from `NotNull`, since this matches the actual SQL more closely (`foo IS NOT NULL`). I think clients are more likely to care about that rather then being consistent with the Laravel method that is used.

faiverson:
Thanks for the support! The changes are great and the correction in the documentation is a game changer!

About sanitizing I didn't check what happen if I tried to intentionally make an attempt of sql injection.
That's was the idea behind that, since we are manipulating the DB directly.

I will do some tests but if somebody can create a better way or confirm that's not necessary even better.

src/WhereConstraints/WhereConstraintsDirective.php
```diff
@@ -225,4 +223,21 @@ function (string $columnName): string {
...
+     * Ensure the column name is well formed and prevent SQL injection.
+     *
+     * @param  string  $column
+     * @return void
+     *
+     * @throws \GraphQL\Error\Error
+     */
+    protected static function assertValidColumnName(string $column): void
+    {
+        if (! \Safe\preg_match('/^(?![0-9])[A-Za-z0-9_-]*$/', $column)) {
```
CPSibo: This prevents using columns that start with a number or include a space, both of which I believe many databases allow. They're funky to work with in Laravel (eg. `$model->{'1 fun column'}`) but are possible.

CPSibo:
> since we are manipulating the DB directly.

Can you point me to a line number where you're doing raw DB operations? It looks like everything is going through the Builder objects with normal where* functions, which have their own sanitation of tables, columns, operators, and values according to the DB grammar in use before handing everything off to PDO as a parameterized query.

Unless you guys add in a `whereRaw` option, I believe the Builder will keep the database safe.

src/WhereConstraints/WhereConstraintsDirective.php
spawnia: I don't want to risk SQL injection. There is always the option to define the columns explicitely, that way any name can be used.

spawnia:
The column names are the only unsafe bit here, and i think we got that covered by the regex that validates they are safe too safe. Operators are only defined by the developer, and values are used within bindings.

Issue #1094, opened 2019-12-23
Test in Lumen


rafaelmarcelino19:
I am using lumen 6.0 and php 7.2 added MakesGraphQLRequests to TestCase and when I try $this->postGraphQL() this returns an error: Call to undefined method ExampleTest::postJson()

So I override the methods in TestCase and changed the return to Response and not TestResponse because lumen doesn't have Illuminate\Foundation

In place of $this->postJson() I made $this->post()

And now Work

spawnia:
Using `TestResponse` seems really useful, is it not available at all in Lumen?

spawnia:
`MakesGraphQLRequestsLumen` has been around for a while now.

PR #1100, opened 2019-12-27
Add test trait `\Nuwave\Lighthouse\Testing\MakesGraphQLRequestsLumen` for usage with Lumen

rafaelmarcelino19:
#1094 
TestResponse is not available in lumen I added this class with the same methods of MakesGraphQLRequests to work in Lumen


spawnia:
Thanks for the PR, i really appreciate it.

What's preventing us from using a single class for both?

spawnia:
This PR seems to have no changes. Are you planning any?

rafaelmarcelino19:
sorry, i didn't add the changes

spawnia:
> What's preventing us from using a single class for both?

I was actually asking a question here, not necessarily suggesting we should split it. Thanks for giving it a shot, though.

Considering the two approaches, i actually like having separate classes better. Otherwise the helper kind of becomes "worst of both worlds", with watered down type hints and little workarounds. I tried abstracting further and basing both on a single base trait and found that this complicates things even further.

So, i finally settled on your original approach and added the separate trait back in. Tweaked it a little and added docs and a changelog entry. Let me know if that works for you.

rafaelmarcelino19:
for me ok, I thought you didn't want to keep two classes separate, but that would really become "the worst of both worlds"

PR #1106, opened 2019-12-30
Update README.md

Geoffry304:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**


lorado:
Hi @Geoffry304 
Thank for pointing to a confusing place in docs. I never did this tutorial, but just read it. It seems to me, that all the text is correct, and only image has `user_id` instead of `author_id` listed. It makes more sense to let `author` relation with correct `author_id` foreign key. So I just changed `user_id` to `author_id` on image.

PR #1110, opened 2020-01-02
Nested mutations with pivot data

### LINKED ISSUE #1088: Nested mutations with pivot data

In my project I have some pivot tables with additional data. In Laravel while using sync it is possible to define simply array with IDs, but also optional value for specific key with data, that will be stored in pivot data. As example:

```php
$person->jobs()->sync([
  1,
  2 => [
    'from' => '...',
    'till' => '...'
  ]
]);
```

After executing `$person` becomes linked job ID 1 with empty extra pivot data and job ID 2 with given `from` and `till` additional data, that are stored in pivot tables.

Currently in Lighthouse it is possible to define only simple ID array for sync.

And here is my first question - do we need such a feature in lighthouse core? Solution is actually pretty simple, but we have to define structure, that can be accepted by sync resolver.  

As I know it is not possible to use kind of dictionaries in gql. So we probably need to define an input object, that have kinda "hidden" attribute for ID, like `_id` or kind of. All other values would then be treated like pivot extra data. 

As example, to be able to achive the php code above:

```gql
type Mutation {
  updatePerson(id: ID!, input: UpdatePersonInput! @spread): Person @update
}

input UpdatePersonInput {
  jobs: UpdateJobRelation
}

input UpdateJobRelation {
  sync: [JobSyncInput!]
}

input JobSyncInput {
  _id: ID!
  from: Date
  till: Date
}
```

So mutation would look like this:
```gql
mutation {
  updatePerson(
    id: 1, 
    jobs: {
      sync: [
        {
          _id: 1
        },
        {
          _id: 2,
          from: "...",
          till: "..."
        }
      ]
    }
  ) {
    ...
  }
}
```

Resolver code could look like this:

```php
// $object and $relationName should be calculated before from scheme types. 
// In this example:
// $object = App\Person::find(1);
// $relationName = 'jobs';
$syncArr = [];
foreach ($input['sync'] as $syncInput) {
  $id = $syncInput['_id'];
  unset($syncInput['_id']);
  $syncArr[ $id ] = $syncInput;
}
$object->$relationName()->sync($syncArr);
```

So anyway, just a question - do we need this? Or just don't care, as new input resolver will be possible soon?

lorado:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1088 

**Changes**

The feature is already described in #1088. Shortly - it is now possible to define pivot data on nested mutations using `sync`, `syncWithoutDetach` or `connect`. 

Details how to use will be described in docs.

**NO Breaking changes**

This implementation just adds described functionality on top of existing one. So it is still possible to use simple array of IDs within `sync`, `syncWithoutDetach` and `connect`.


lorado:
@spawnia Please improve docs...  it is so hard for me to write them...

src/Execution/Arguments/NestedManyToMany.php
```diff
@@ -59,4 +59,33 @@ public function __invoke($parent, $args)
             );
         }
     }
+
+    /**
+     * @param  \Nuwave\Lighthouse\Execution\Arguments\Argument $args
+     *
+     * @return array
+     */
+    private function generateRelationArray(Argument $args)
```
spawnia:
```suggestion
    private function generateRelationArray(Argument $args): array
```

src/Execution/Arguments/NestedManyToMany.php
```diff
@@ -59,4 +59,33 @@ public function __invoke($parent, $args)
             );
         }
     }
+
+    /**
```
spawnia: A one-liner that describes the purpose of this function would be great.

src/Execution/Arguments/NestedManyToMany.php
```diff
@@ -59,4 +59,33 @@ public function __invoke($parent, $args)
...
+     */
+    private function generateRelationArray(Argument $args)
+    {
+        $values = $args->toPlain();
+
+        if (empty($values)) {
+            return [];
+        }
+
+        if (! is_array($values[0])) {
```
spawnia: This check might be more straightforward by examining the `Argument` class, it is meant for use cases like these where we need to check the type of arguments during query execution.

src/Execution/Arguments/NestedManyToMany.php
```diff
@@ -59,4 +59,33 @@ public function __invoke($parent, $args)
...
+            // first values isn't array. Assume values are simply IDs, and return them (old behaviour)
+            return $values;
+        } else {
+            // assume values are arrays and contains pivot information
+            $relationArray = [];
+            foreach ($values as $value) {
+                $id = $value['id'];
+                unset($value['id']);
+                $relationArray[$id] = $value;
```
spawnia: What happens if there are no other values?

According to the example you have shown in your issue passing the `id` as the value would be the way to go.

```php
$person->jobs()->sync([
  1,
  2 => [
    'from' => '...',
    'till' => '...'
  ]
]);
```

We should at least cover this use case of mixing values with just `id` and ones with additional data, and have a test for it.

tests/Integration/Execution/MutationExecutor/BelongsToManyTest.php
```diff
@@ -826,4 +846,139 @@ public function testCanDisconnectAllRelatedModelsOnEmptySync(string $action): vo
...
+    public function testCanConnectUserWithRoleAndPivotMetaByUsingSync(): void
+    {
+        $user = factory(User::class)->create();
+        $role = factory(Role::class)->create();
+        $role2 = factory(Role::class)->create();
+        $user->roles()->attach($role2);
+
+        $metaText = Lorem::sentence();
+
+        $this->graphQL('
```
spawnia:
```suggestion
        $this->graphQL(/** @lang GraphQL */ '
```

spawnia:
> @spawnia Please improve docs... it is so hard for me to write them...

They look pretty good already, but i will make a pass over them once we have the code all figured out.

src/Execution/Arguments/NestedManyToMany.php
lorado: Ok, can you say, what I have to compare? `$args->type` can't help there much, as it is in both cases a `ListType`, and the type itself is just a string, like `ID`, but we can't take comparison to `ID`, as I e.g. using `ObfID` in my project as main ID type. 

We could also check the value of `$args->value[0]` if it is an object or not. What do you think?

src/Execution/Arguments/NestedManyToMany.php
lorado: I just read implementation of `sync()` in laravel, if you will pass an array from example above, laravel will convert input to the following array:

```php
[
  1 => [],
  2 => [
    'from' => '...',
    'till' => '...'
  ]
]
```

So in words, it sanitize the input in the way, that final array keys are actually IDs, and the value is an array with values of pivot columns.

I can also add a test that just accepts an id as input without any pivot data

src/Execution/Arguments/NestedManyToMany.php
spawnia: I tried out the other approach and found your initial solution to be simpler. Well done.

PR #1114, opened 2020-01-05
Allow passing `null` or simply no `id` when using `@upsert`

### LINKED ISSUE #1113: @upsert directive fails if there is no id
**Describe the bug**

In `master`, `@upsert` directive is not working without an `id` in the input.

`@upsert` directive should work whether there is an `id` or not in the input. (It was working in this way before the amazing #899 PR).

**Steps to reproduce**

1. Write a mutation with the `@upsert` directive like:
```graphql
type Mutation {
    upsertTask(input: UpsertTaskInput! @spread): Task @upsert
}

input UpsertTaskInput {
    id: ID
    name: String
}
```
2. Execute it like:
```graphql
mutation {
    upsertTask(input: {
        name: "foo"
    }) {
        id
        name
    }
}
```
3. It will fails since there is no `id` in the input. But I think `@upsert` should falls to create a task, instead of try to update one (as it is working in Lighthouse 4.7).

**Output/Logs**

```
ErrorException : Undefined index: id
```

</details></br>

**Environment**

Lighthouse Version: `master`
Laravel Version: `6.4.1`

enzonotario:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1113 

**Changes**

`UpsertModel` tries to get the `id` using `Arr::get()` to avoid `Undefined index` error.

**Breaking changes**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Enable chained rule provider directives (`ProvidesRules`) to merge the rules before validating https://github.com/nuwave/lighthouse/pull/1082
+- Allow passing `null` or simply no `id` when using `@upsert` https://github.com/nuwave/lighthouse/pull/1114
 
 ## [4.7.2](https://github.com/nuwave/lighthouse/compare/v4.7.1...v4.7.2)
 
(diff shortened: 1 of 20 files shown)


enzonotario:
As @lorado says in https://github.com/nuwave/lighthouse/issues/1113#issuecomment-570947006 , I think the docs should be updated to reflect that it's not necessary to pass an `id`. I tried to update docs to reflect that, but please double-check it since my limited english.

Issue #1117, opened 2020-01-06
Add support for specifying column names with `@orderBy`


chescos:
As far as I am aware, it's currently not possible to specify validation rules for the `@orderBy` directive. When the user supplies an invalid column, an exception is thrown because the SQL query results in an error. Shouldn't there be a way to specify an array of allowed columns through the `@orderBy` directive?

enzonotario:
This repository and the issue tracker are used to advance the development of Lighthouse.
Issues that are neither detailed bug reports nor well thought-out feature proposals will be closed.

You can use the following channels to ask support questions:
- [Use the laravel-lighthouse tag on Stackoverflow](https://stackoverflow.com/questions/tagged/laravel-lighthouse) 
- [Hop into Slack](https://join.slack.com/t/lighthouse-php/shared_invite/enQtMzc1NzQwNTUxMjk3LWMyZWRiNWFmZGUxZmRlNDJkMTQ2ZDA1NzQ1YjVkNTdmNWE1OTUyZjZiN2I2ZGQxNTNiZTZiY2JlNmY2MGUyNTQ)

spawnia:
This already works in `@whereConstraints`. I recently ran into this as well and started implementing it for `@orderBy` as well, good catch. Check https://github.com/nuwave/lighthouse/pull/1118 for progress.

chescos:
Awesome, thanks for the quick response!

PR #1118, opened 2020-01-06
Allow restricting the allowed column names in `@orderBy` to a well-defined list.

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/1117

**Changes**

Just like `@whereConstraints`, this automatically generates an `enum` with the correct column names that can be used.

**Breaking changes**

No


docs/master/api-reference/directives.md
```diff
@@ -2877,13 +2891,14 @@ type Person {
 
 type Query {
     people(
-        where: WhereConstraints @whereConstraints(columns: ["age", "type", "haircolour", "height"])
+        where: _ @whereConstraints(columns: ["age", "type", "haircolour", "height"])
```
spawnia: @CPSibo i removed the requirement to set the correct type in the schema and just tell the user to put `_` as a placeholder, do you think that is more clear then what we had before. You mentioned this part was confusing you.

spawnia:
@chescos i would appreciate a review.

PR #1119, opened 2020-01-06
Correct @whereConstraints documentation

### LINKED ISSUE #1115: @whereConstraints Documentation Issues
I want to make sure I'm not just doing something wrong with the @whereConstraints directive before making a pull request to amend the documentation. If you can answer my questions below, I'd be happy to make the documentation for this directive correct and more user-friendly.

**Describe the bug**

The example code gives this for the schema:
```graphql
type Query {
    people(
        where: WhereConstraints @whereConstraints(columns: ["age", "type", "haircolour", "height"])
    ): [Person!]!
}
```
and this for the actual query:
```graphql
{
  people(
    filter: {
      where: [
        {
          AND: [
            { column: AGE, operator: GT value: 37 }
            { column: TYPE, value: "Actor" }
            {
              OR: [
                { column: HAIRCOLOUR, value: "red" }
                { column: HEIGHT, operator: GTE, value: 150 }
              ]
            }
          ]
        }
      ]
    }
  ) {
    name
  }
}
```
However, doing so produces 
`"Unknown argument "filter" on field "people" of type "Query"."`

Having looked at others' code, I'm guessing nesting `filter:` and `where:` is a typo and it should only be 
```graphql
{
  people(
      where: ...
```

Further, I can't find any combination of the default `WhereConstraints` type that allows for `AND` or `OR` arrays, as the example does. A schema like:
```graphql
type Person{
    id: ID!
    name: String!
    age: Int!
}

type Query {
    people(
        where: WhereConstraints @whereConstraints
    ): [Person!]!
}
```
seems to only allow queries like:
```graphql
{
    people(
         where: { column: "name", operator: EQ, value: "whatever" }
    ){
        id
    }
}
```

If I include the `PersonWhereConstraints` and `PersonWhereColumn` types in my schema and change `PersonWhereColumn` to use the `Operator` enum, rather than a `String`, for the `operator` field, the code finally runs using a query more like the example given.

So, is the following correct:
1. The example should not be using `filter: where:` as a nested parameter.
2. The example query will only work if the `PersonWhereConstraints` and `PersonWhereColumn` types are defined.
3. The example query will only work if `PersonWhereColumn` uses the `Operator` enum, rather than a `String` type.
4. The example schema for the query should include something like `@all` at the end.

Also, there appears to be a typo where you're using `PersonWhereWhereConstraints` rather than `PersonWhereConstraints`. As well, there's an unterminated string for the comment above the `PeopleWhereWhereConstraints` definition.

**Environment**

Lighthouse Version: 4.7
Laravel Version: 6.2

CPSibo:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #1115

**Changes**

No behavioral changes. Only documentation changes.

The `@whereConstraints` documentation under `.../api-reference/directives.md` has been corrected and expanded to use functional code examples and to be more thorough.

**Breaking changes**

None.


docs/master/api-reference/directives.md
```diff
@@ -2928,31 +2945,67 @@ that gets actors over age 37 who either have red hair or are at least 150cm.
...
+        where: WhereConstraints @whereConstraints(columns: ["age", "height"])
+    ): [Person!]! @all
+}
+```
+
+When using the default `WhereConstraints` input, Lighthouse allows for filtering on any of the fields in the model, specified as strings in the request. To customize this behavior, you can define your own input and enum:
+
+```graphql
+"Dynamic WHERE constraints for the `where` argument on the query `people`."
+input PeopleWhereConstraints {
```
spawnia: That should not be necessary. Simply specifying the column names in `@whereConstraints` will manipulate the original argument definition to point towards a generated type that holds the correct enum for the allowed columns.

docs/master/api-reference/directives.md
```diff
@@ -2892,32 +2892,49 @@ enum Operator {
 
 ### Usage
 
+The most basic way to use this directive is with the dynamic `WhereConstraints` input and the `@whereConstraints` directive with no arguments.
```
spawnia: I intentionally put this at the very end, as i don't really recommend this approach. The default should be to specify the columns.

docs/master/api-reference/directives.md
```diff
@@ -2892,32 +2892,49 @@ enum Operator {
...
+
 ```graphql
+type Person {
```
spawnia: I like adding that. While reworking a lot of `@whereConstraints`, i included this in https://github.com/nuwave/lighthouse/pull/1099.

docs/master/api-reference/directives.md
```diff
@@ -2892,32 +2892,49 @@ enum Operator {
...
-This is how you can use it to construct a complex query
-that gets actors over age 37 who either have red hair or are at least 150cm.
+The client can then build arbitrarily filtered queries. This will get people who are older than 37:
 
 ```graphql
 {
   people(
-    filter: {
-      where: [
+    where: { column: "age", operator: GT, value: 37 }
```
spawnia: Good idea to add a simple query.

docs/master/api-reference/directives.md
```diff
@@ -2892,32 +2892,49 @@ enum Operator {
...
+  }
+}
+```
+
+This will get actors over the age of 37 who either have red hair or are at least 150cm:
+
+```graphql
+{
+  people(
+    where: {
```
spawnia: Great find fixing those.

spawnia:
Thanks @CPSibo for the great work on the docs.

Since i reworked that feature a great deal, i included some ideas from your changes within https://github.com/nuwave/lighthouse/pull/1099 and am closing this PR in favour of it.

docs/master/api-reference/directives.md
CPSibo: I just tested and I think it's working as you intend:
- If you do not specify the `columns` argument **and** you use the built-in input, columns **must** be queried by string.
- If you do specify the columns **or** you use a custom input like the examples show, columns **must** be queried by the enums.

Ultimately, my PR hopes to clear up some of these fiddly pitfalls that burned so much of my time trying to get this directive working. I think this functionality should be explicitly spelled out in the docs.

spawnia: Using a custom input is not necessary, but i see how the docs could give the wrong impression. Maybe we can hint that the shown code is just autogenerated and not meant to be included in the schema.

PR #1122, opened 2020-01-07
The playground wouldn't be required for production

garhbod:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**


spawnia:
I always have the Playground enabled anywhere, as i don't see any downsides to having it.

It is especially useful in test or staging environments. Since i try to keep those as close as possible to prod, i do not install dev dependencies there and require Playground in the regular dependencies.

lorado:
Hm I am using an external App, currently Altair GraphQL Client. Anyway, I also think it is ok to install playground to normal dependencies. Even if you keep it dev only, it is possible to connect to you graphql endpoint using external clients.

PR #1130, opened 2020-01-10
GitHub Actions

spawnia:
Travis is being really slow to get started on running tests. This goes much faster and allows us to set up different steps more easily.


spawnia:
Hey @chrissm79 i would like to transfer over Codecov, too. Can you give me write  access to codecov? We need to set the `CODECOV_TOKEN` as a secret to enable pushing coverage from a GitHub action.

chrissm79:
@spawnia I added Codecov and you should have the access you need but let me know if not!

spawnia:
@chrissm79 thanks for looking into it. It seems like i do not have write access yet, as i can not see the option to get a token at https://codecov.io/gh/nuwave/lighthouse/settings

It is supposed to be there, judging from the docs: https://docs.codecov.io/docs/frequently-asked-questions#section-where-is-the-repository-upload-token-found

PR #1131, opened 2020-01-11
Add missing descriptions to generated and built-in schema types

### LINKED ISSUE #1123: Some inputs and types are missing descriptions
I've noticed that some inputs and types don't have descriptions, for example, the [PageInfo](https://github.com/nuwave/lighthouse/blob/ced6496c4cdb4aea422c4cca4d875244c5837697/src/Schema/AST/ASTBuilder.php#L286) type or the [OrderBy](https://github.com/nuwave/lighthouse/blob/ced6496c4cdb4aea422c4cca4d875244c5837697/src/Schema/AST/ASTBuilder.php#L392) input.

Is this intentional? Otherwise, I'd be happy to submit a PR that adds the missing descriptions.

chescos:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1123

**Changes**

This PR only adds some missing descriptions to the schema definition. No changes to the behavior of the library.

**Breaking changes**

There are no breaking changes.

### DIFF
--- a/src/Pagination/PaginationManipulator.php
+++ b/src/Pagination/PaginationManipulator.php
@@ -116,7 +116,7 @@ class PaginationManipulator
         $connectionEdge = $edgeType
             ?? $this->documentAST->types[$connectionEdgeName]
             ?? PartialParser::objectTypeDefinition("
-                \"A $fieldTypeName node that contains the node itself and a cursor.\"
+                \"An edge that contains a node of type $fieldTypeName and a cursor.\"
                 type $connectionEdgeName {
                     \"The $fieldTypeName node.\"
                     node: $fieldTypeName
--- a/src/Pagination/PaginationManipulator.php
+++ b/src/Pagination/PaginationManipulator.php
@@ -103,8 +103,12 @@ class PaginationManipulator
         $connectionFieldName = addslashes(ConnectionField::class);
 
         $connectionType = PartialParser::objectTypeDefinition("
+            \"A paginated list of $fieldTypeName edges.\"
             type $connectionTypeName {$this->modelClassDirective()} {
+                \"Pagination information about the list of edges.\"
                 pageInfo: PageInfo! @field(resolver: \"{$connectionFieldName}@pageInfoResolver\")
+
+                \"A list of $fieldTypeName edges.\"
                 edges: [$connectionEdgeName] @field(resolver: \"{$connectionFieldName}@edgeResolver\")
             }
         ");
@@ -112,8 +116,12 @@ class PaginationManipulator
         $connectionEdge = $edgeType
             ?? $this->documentAST->types[$connectionEdgeName]
             ?? PartialParser::objectTypeDefinition("
+                \"A $fieldTypeName node that contains the node itself and a cursor.\"
                 type $connectionEdgeName {
+                    \"The $fieldTypeName node.\"
                     node: $fieldTypeName
+
+                    \"A unique cursor that can be used for pagination.\"
                     cursor: String!
                 }
             ");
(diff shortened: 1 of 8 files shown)


chescos:
Not sure if I caught everything that had no descriptions, I hope so.

- I haven't seen any tests for descriptions, I assume they don't require testing?
- Should the descriptions also be added to schema examples in the docs?

src/Pagination/PaginationManipulator.php
```diff
@@ -103,17 +103,25 @@ protected function registerConnection(
         $connectionFieldName = addslashes(ConnectionField::class);
 
         $connectionType = PartialParser::objectTypeDefinition("
```
spawnia: We can use HEREDOC syntax here to avoid having to escape the double quotes but still be able to interpolate variables.

src/Pagination/PaginationManipulator.php
```diff
@@ -103,17 +103,25 @@ protected function registerConnection(
...
+
+                \"A list of $fieldTypeName edges.\"
                 edges: [$connectionEdgeName] @field(resolver: \"{$connectionFieldName}@edgeResolver\")
             }
         ");
 
         $connectionEdge = $edgeType
             ?? $this->documentAST->types[$connectionEdgeName]
             ?? PartialParser::objectTypeDefinition("
+                \"A $fieldTypeName node that contains the node itself and a cursor.\"
```
spawnia:
```suggestion
                \"An edge that contains a node of type $fieldTypeName and a cursor.\"
```

src/SoftDeletes/SoftDeletesServiceProvider.php
```diff
@@ -48,9 +48,15 @@ function (ManipulateAST $manipulateAST): void {
                 $manipulateAST->documentAST
                     ->setTypeDefinition(
                         PartialParser::enumTypeDefinition('
+                            "The available enumerators that are used to filter trashed query results."
```
spawnia:
```suggestion
                            "Specify if you want to include or exclude trashed results from a query."
```

src/WhereConstraints/SQLOperator.php
```diff
@@ -14,20 +14,48 @@ public static function missingValueForColumn(string $column): string
...
+    "Whether a value is not within a set of values (NOT IN)"
     NOT_IN @enum(value: "NotIn")
+
+    "Whether a value is within a range of values (BETWEEN)"
     BETWEEN @enum(value: "Between")
+
+    "Whether a value is not within a range of values (NOT BETWEEN)"
     NOT_BETWEEN @enum(value: "NotBetween")
+
+    "NULL value test (NULL)"
```
spawnia:
```suggestion
    "Check if the value is null (`IS NULL`)."
```

src/WhereConstraints/SQLOperator.php
```diff
@@ -14,20 +14,48 @@ public static function missingValueForColumn(string $column): string
     public function enumDefinition(): string
     {
         return /** @lang GraphQL */ <<<'GRAPHQL'
+"The available SQL operators that are used to filter query results."
 enum SQLOperator {
+    "Equal operator (=)"
```
spawnia:
```suggestion
    "Equal operator (`=`)."
```

Let's wrap the SQL-equivalents in backticks. GraphQL IDE's such as Playground format descriptions as Markdown.

spawnia:
> Not sure if I caught everything that had no descriptions, I hope so.

If we find more, we can always add it later. I think you got everything.

> * I haven't seen any tests for descriptions, I assume they don't require testing?

Only autogenerated descriptions that contain some logic would require testing.

> * Should the descriptions also be added to schema examples in the docs?

The ones that refer to the generated parts of the schema, yes. Not every schema example needs descriptions, although there are parts where they might help in making the docs more clear.

spawnia:
Thank you for this great effort, this makes the schema way easier to understand and consume for clients. This definitely deserves an update to the changelog, as it is a small but quite meaningful improvement.

src/WhereConstraints/SQLOperator.php
chescos: Thanks, done!

src/WhereConstraints/SQLOperator.php
chescos: Also done :+1:

src/Pagination/PaginationManipulator.php
chescos: Done!

chescos:
@spawnia thanks for the help, I've implemented your suggested changes. It would be awesome if you could give it a last look and tell me if any other changes are required. Otherwise, it should be good to go!

Issue #880, opened 2019-07-26
Don't use request middleware for resolver middleware


pvormittag:
Firstly, I've found Lighthouse to be very productive and it's apparent a lot of thought and work has gone into this project, so a big thanks to all contributors! 🙌

**Is your feature request related to a problem? Please describe.**

Currently, middleware are [request middleware](https://laravel.com/docs/5.8/middleware), which implicitly expect a `Response` object to be returned from `$next($request)`, however when using the `@middleware` directive, Lighthouse will return the result from the field resolver. This means any request middleware with after logic will break when attempting to alter the response. I've experienced this with several of the built-in request middleware classes, such as `Illuminate\Session\Middleware\StartSession`.

Additionally, request middleware expect to be called only once per request and are a global concern. Yet resolver middleware may be called many times on many fields during a single request. Any middleware expecting once-per-request semantics may cause sutble-and-hard-to-find bugs when used in the context of resolver middleware.

I've put together a minimal reproducible example to isolate and discuss the issue:
https://github.com/pvormittag/lighthouse-middleware-issue

**Describe the solution you'd like**

Ultimately request and resolver middleware have different concerns and usage patterns, so I believe the solution is for Lighthouse to provide its own resolver middleware alongside request middleware and describe the differences and usage. This would also align with many other GraphQL server implementations.

I can think of several situations where both are useful, such as using request middleware for authn, which needs the request and can for instance, set `$request->user()` as `$context->currentUser`. Then use resolver middleware to handle authz, which is likely used several times as fields have different access control (i.e. `can`).

pvormittag:
I found a few issues I think illustrate how handling the HTTP request is a similar-yet-different concern from handling the GraphQL query. The subtlety lies in their processing logic sharing several semantics, terms and common usage patterns.

#142, #789, #854

These highlight the importance of documenting the subtle differences and if not handled properly, may become a trap for new users and contributors to the project. The changes needed will likely not be backwards compatible, but I feel are important for Lighthouse moving forward.

It is certainly possible to address with `webonyx/graphql-php`. The handoff between Laravel's HTTP handling logic and graphql-php's query handling logic have a well defined boundary, and unfortunately this has been muddled by `@middleware`, `@auth` and a few other directives.

Stackout:
@pvormittag is correct. Using the request middleware for resolver middleware is an issue. I struggled using sessions for the same reason. His explanation helped me understand it.

SirLamer:
[...]

PR #1135, opened 2020-01-12
Add `@guard` directive and deprecate `@middleware` directive

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/880

**Changes**

The `@middleware` directive will be removed, as it violates the boundary between HTTP and GraphQL
request handling.

Authentication is one of most common use cases for `@middleware`. You can now use
the [`@guard`](docs/master/api-reference/directives.md#guard) on selected fields.

```diff
type Query {
- profile: User! @middlware(checks: ["auth"])
+ profile: User! @guard
}
```

Other functionality can be replaced by a custom [`FieldMiddleware`](docs/master/custom-directives/field-directives.md#fieldmiddleware)
directive. Just like Laravel Middleware, it can wrap around individual field resolvers.

**Breaking changes**

No, `@middleware` is kept functional until v5.


src/Schema/Directives/GuardDirective.php
```diff
@@ -0,0 +1,126 @@
...
+            $fieldValue->setResolver(
+                function () use ($resolver) {
+                    $this->authenticate(
+                        (array) $this->directiveArgValue('with')
+                    );
+
+                    return $resolver(...func_get_args());
+                }
+            )
+        );
```
lorado: TBO, I don't understand this peace of code... I am neither fit with Laravel authorization, nor with some lighthouse core functionality.

src/Schema/AST/ASTHelper.php
```diff
@@ -302,4 +303,36 @@ public static function typeImplementsInterface(ObjectTypeDefinitionNode $type, s
...
+                "The @{$name} directive may only be placed on fields or object types."
+            );
+        }
+
+        /** @var \GraphQL\Language\AST\FieldDefinitionNode $fieldDefinition */
+        foreach ($objectType->fields as $fieldDefinition) {
+            // If the field already has the same directive defined, skip over it.
+            // Field directive are more specific then those defined on a type.
+            if (self::hasDirective($fieldDefinition, $name)) {
+                return;
```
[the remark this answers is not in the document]
lorado: I already changed it in my commit, alongside with the test for the specific situation I thought about

src/Schema/Directives/GuardDirective.php
spawnia: The basic structure here is what all field middleware do: Wrap around the previous resolver and do some steps inbetween. The call to `authenticate` will throw in case the user is not authenticated.

I guess the usage of the magic `func_get_args()` method is a bit confusing. It is in there because we do not care about the resolver arguments at all in this middleware, so we just pass them through without even naming them specifically.

docs/master/api-reference/directives.md
```diff
@@ -1130,6 +1130,25 @@ split into type and ID.
 You may rebind the `\Nuwave\Lighthouse\Support\Contracts\GlobalId` interface to add your
 own mechanism of encoding/decoding global ids.
 
+## @guard
+
+```graphql
+"""
+Run authentication through one more guards.
```
enzonotario: Sorry, shouldn't it be "Run authentication through one **or** more guards."?

src/Schema/AST/ASTHelper.php
```diff
@@ -302,4 +303,36 @@ public static function typeImplementsInterface(ObjectTypeDefinitionNode $type, s
...
+        ) {
+            throw new DefinitionException(
+                "The @{$name} directive may only be placed on fields or object types."
+            );
+        }
+
+        /** @var \GraphQL\Language\AST\FieldDefinitionNode $fieldDefinition */
+        foreach ($objectType->fields as $fieldDefinition) {
+            // If the field already has the same directive defined, skip over it.
+            // Field directive are more specific then those defined on a type.
```
enzonotario: shouldn't it be "Field **directives** are more specific **than** those defined on a type."?

Christophvh:
The big problem with this is that the validation output is in a different structure. It could be helpful to document this when releasing v5. I tried to change this, but it breaks most of our SPA code.

spawnia:
> The big problem with this is that the validation output is in a different structure. It could be helpful to document this when releasing v5. I tried to change this, but it breaks most of our SPA code.

Can you give specific examples?

Christophvh:
Middleware output:
```
{
  "errors": [
    {
      "debugMessage": "Unauthenticated.",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
```

Guard output
```
       "errors": [
    {
      "message": "Unauthenticated.",
      "extensions": {
        "guards": [
          "api"
        ],
        "category": "authentication"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
```

The output is obviously better with guard, but the key changed for where our front-end gets a good error message.

spawnia:
@Christophvh thanks for digging in.

It is not recommended to enable debug messages in production, so i would not expect reliance upon the contents of the `debugMessage` field.

carelloqwerty:
Laravel 7.x . https://laravel.com/docs/master/authentication#closure-request-guards. I've created guard just like in instructions. With @middlware(checks: ["auth:custom"]) I have Auth:user() for @can directive, Auth::viaRequest() gives me Auth:user(), but when I try to use @guard(with: ["custom"]), the code passes through guard, but @can directive returns "You are not authorized to access ...". Seems that @guard only checks if user exist, but it doesn't set auth. How can I set Auth user with @guard directive? @middlware still best option for me.

spawnia:
@carelloqwerty i just released https://github.com/nuwave/lighthouse/releases/tag/4.11.0

Note that `@guard` does not log in users.
To ensure the user is logged in, add the `AttemptAuthenticate` middleware to your `lighthouse.php`
middleware config, see the default config https://github.com/nuwave/lighthouse/blob/master/src/lighthouse.php for an example.

carelloqwerty:
unfortunatly in my 4.10.1 version ther is no such class... I don't have time to update and test it right now, hope it behaves same as @middlevare when you apply it on grope of quries or mutations. One more question, how can I run test and ignore @middleware directive, WithoutMiddleware trait in test doesn't help, and i cant get information about this. And it would be good to explane the same for @guard

spawnia:
I recommend updating and fake login users when testing.

carelloqwerty:
I've finely updated lighthouse, @guard works perfectly, even better then @middleware, i can now use $this->be($user) in test, and all queries and mutations works in test passin through all @can directives. it's amazing

carelloqwerty:
One more question, if @middleware directive is depricated, how can i pass specific query through Laravel 'auth.basic' middleware, without defining this middleware in config file, becouse i need it only for one query.

spawnia:
> One more question, if @Middleware directive is depricated, how can i pass specific query through Laravel 'auth.basic' middleware, without defining this middleware in config file? Becouse i need it only for one query

Using HTTP middleware on the field level is wrong and i do not recommend it. You can use a custom `FieldMiddleware` directive to handle authentication in a specific way if needed.

carelloqwerty:
> > One more question, if @Middleware directive is depricated, how can i pass specific query through Laravel 'auth.basic' middleware, without defining this middleware in config file? Becouse i need it only for one query
> 
> Using HTTP middleware on the field level is wrong and i do not recommend it. You can use a custom `FieldMiddleware` directive to handle authentication in a specific way if needed.

but why then @guard is not wrong on the field level?

spawnia:
`@guard` does not use HTTP middleware.

PR #1138, opened 2020-01-13
Add upgrade guide for `@modelClass` and `@model` in v5

lorado:
Just improving upgrade guide.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -35,22 +35,26 @@ Client queries will have to be changed like this:
 If you absolutely cannot break your clients, you can re-implement `@orderBy` in your
 project - it is a relatively simple `ArgManipulator` directive.
 
-### `@model` directive was repurposed 
+### `@modelClass` and `@model` changed 
 
-`@model` directive in v5 does what `@modelClass` did in v4. To implement similar functionality of `@model` from v4 use `@node` directive instead. 
+The `@model` directive was repurposed to take the place of `@modelClass`, see
+https://github.com/nuwave/lighthouse/pull/974 for details.
+You can adapt to this change in two refactoring steps that must be done in order:
 
-```diff
--type User @model {
-+type User @node {
-    id: ID! @globalId
-}
-```
+1. Rename all usages of `@model` to `@node`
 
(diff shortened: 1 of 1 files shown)


spawnia:
Great idea. I always like to add codeblocks tagged with `diff` and show a concrete example of before/after the upgrade, can you add that here as well?

lorado:
I am not sure about examples... Is it ok?

UPGRADE.md
```diff
@@ -34,3 +34,23 @@ Client queries will have to be changed like this:
...
+    id: ID! @globalId
+}
+```
+
+### `@modelClass` is renamed to `@model`
+
+```diff
+-type PaginatedPost @modelClass(class: "\\App\\Post") {
++type PaginatedPost @model(class: "\\App\\Post") {
+    id: ObfId!
```
spawnia:
```suggestion
    id: ID!
```

lorado: Gosh... sorry, forgot to change ID type

UPGRADE.md
```diff
@@ -34,3 +34,23 @@ Client queries will have to be changed like this:
 
 If you absolutely cannot break your clients, you can re-implement `@orderBy` in your
 project - it is a relatively simple `ArgManipulator` directive.
+
+### `@model` directive was repurposed 
+
+`@model` directive in v5 does what `@modelClass` did in v4. To implement similar functionality of `@model` from v4 use `@node` directive instead.
```
spawnia: I think this part is somewhat confusing. How about we just give the minimal information needed to upgrade and link https://github.com/nuwave/lighthouse/pull/974 to give background information.

The upgrade guide should mention concrete steps, those are to 1. rename `@model` to `@node`  and 2. rename `@modelClass` -> `@model` (in that order).

lorado: Would a comment like "Please apply changes in your project step by step from up to down" be enough? Alternatively we can define number to steps, and they then have to be applied in the incremental order.

spawnia: I tried a bit different approach, what do you think?

lorado: Oh yes, that is much better!

PR #1140, opened 2020-01-13
Add `@whereHasConditions` directive to filter relation existence

### LINKED ISSUE #1112: Add Relationship support for whereConstraints through whereHas and orWhereHas
I believe that adding Relationship support to the `whereConstraints` Directive through Eloquent's `whereHas` and `orWhereHas` methods would be a powerful feature.

I'll use the example from the tutorial to illustrate the use-case. Let's assume this is our database schema:

Database Schema

Currently, the `whereConstraints` Directive allows us to create the following GraphQL schema:

```
type Query {
    users(
        filter: WhereConstraints @whereConstraints(columns: ["id", "name"])
    ): [User!]! @all
}

type User {
    id: ID!
    name: String!
    email: String!
    created_at: DateTime!
    updated_at: DateTime!
    posts: [Post!]! @hasMany
}

type Post {
    id: ID!
    title: String!
    content: String!
    author: User! @belongsTo
}

type Comment {
    id: ID!
    reply: String!
    post: Post! @belongsTo
}
```

We can then query all users which are named `John` like this:

```
{
  users(
    filter: {
      AND: [
        { column: NAME, operator: EQ, value: "John" }
      ]
    }
  ) {
    id
    name
  }
}
```

But what if we wanted to query all users which are named `John` and have made a post which has a comment with the content `thanks`? Currently, that wouldn't be possible with any of the default Directives.

But it should be possible to add support for this kind of query to the `whereConstraints` Directive by leveraging the power of Eloquent's `whereHas` and `orWhereHas` methods.

Here's an idea of how the modified GraphQL schema could look like:

```
type Query {
    users(
        filter: WhereConstraints @whereConstraints(columns: ["id", "name"], relations: {"posts.comments": ["id", "reply"]})
    ): [User!]! @all
}
```

Here's an idea of how the syntax for the query could look like:

```
{
  users(
    filter: {
      AND: [
        { column: NAME, operator: EQ, value: "John" }
      ],
      AND_HAS: [
        { relation: "posts.comments", column: REPLY, operator: EQ, value: "thanks" }
      ]
    }
  ) {
    id
    name
  }
}
```

chescos:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1112

**Changes**

This PR adds a new directive called `@whereHasConditions`. This directive is very similar to the `@whereConstraints` directive, except that the query conditions are applied to an Eloquent relationship.

The `@whereConstraints` directive has also been renamed to `@whereConditions` after [a discussion](https://github.com/nuwave/lighthouse/issues/1112#issuecomment-572511703) in #1112.

Here's how the `@whereHasConditions` directive can be used in the schema:

```graphql
type Query {
    users(
        hasPosts: _ @whereHasConditions(relation: "posts", columns: ["title", "content"])
        hasComments: _ @whereHasConditions(relation: "posts.comments", columns: ["reply"])
    ): [User!]! @all
}
```

And here's an example query:

```graphql
{
    users(
        hasPosts: { column: TITLE, operator: LIKE, value: "%some title%" },
        hasComments: { column: REPLY, operator: EQ, value: "hi" }
    ) {
        id
    }
}
```

**Breaking changes**

The `@whereConstraints` directive has been renamed to `@whereConditions`.


chescos:
@spawnia I'll start working on this PR soon if that's fine with you. Please let me know if there are any other requirements except what we discussed in #1112.

About renaming the `@whereConstraints` directive - do you want to do that in a separate PR or should we do this here as well?

spawnia:
@chescos looking forward to seeing the results, nice!

Let's do the renaming now, i think that will be the least effort overall. It's bettter to make such breaking changes early, before this becomes so great that everybody will use it. We should be able to include it as a stable feature soon.

tests/Integration/WhereHasConditions/WhereHasConditionsDirectiveTest.php
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    type Company {
+        id: ID!
+        name: String
+    }
+
+    type Query {
+        posts(
+            hasUser: _ @whereHasConditions
```
spawnia:
```suggestion
            hasUser: _ @whereHasConditions(relation: "user")
```

chescos: Done!

chescos:
@spawnia The `@whereHasConditions`  directive is pretty much done, I think I covered everything. The code is mostly copied from the `@whereConstraints` directive as it follows the same logic. One question though... the `Operator` and `SQLOperator` classes can be fully shared between both directives - should they just stay in the `WhereConstraints` folder?

A code review would be awesome. After that, I'd add the documentation and add it to the changelog.

src/WhereHasConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Ensure the column name is well formed and prevent SQL injection.
+     *
+     * @param  string  $column
+     * @return void
+     *
+     * @throws \GraphQL\Error\Error
+     */
+    protected static function assertValidColumnName(string $column): void
```
chescos: The `assertValidColumnName` method could be moved to the `SQLOperator` class, so that we don't have to duplicate it in multiple directives.

spawnia:
All related code should be in a namespace `WhereConditions`, including everything for the previous `@whereConstraints`. Having one service provider for both is reasonable.

chescos:
@spawnia great call, thanks! That made things a lot cleaner already. Anything else that could be improved?

src/WhereConditions/WhereConditionsServiceProvider.php
```diff
@@ -43,14 +44,21 @@ function (RegisterDirectiveNamespaces $registerDirectiveNamespaces): string {
...
-                            'Dynamic WHERE constraints for queries.',
+                        static::createWhereConditionsInputType(
+                            static::DEFAULT_WHERE_CONDITIONS,
+                            'Dynamic WHERE conditions for queries.',
+                            'String'
+                        )
+                    )
+                    ->setTypeDefinition(
+                        static::createWhereConditionsInputType(
+                            static::DEFAULT_WHERE_HAS_CONDITIONS,
```
spawnia: I don't think we need that second type, since they are the same. Makes reusage on the client easier when using something like TypeScript.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+
+class WhereHasConditionsDirective extends BaseDirective implements ArgBuilderDirective, ArgManipulator, DefinedDirective
+{
+    /**
+     * @var \Nuwave\Lighthouse\WhereConditions\Operator
+     */
+    protected $operator;
+
+    /**
+     * WhereHasConditionssDirective constructor.
```
spawnia:
```suggestion
     * WhereHasConditionsDirective constructor.
```

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+     * @return void
+     */
+    public function __construct(Operator $operator)
+    {
+        $this->operator = $operator;
+    }
+
+    public static function definition(): string
+    {
+        return /* @lang GraphQL */ <<<'SDL'
```
spawnia:
```suggestion
        return /** @lang GraphQL */ <<<'SDL'
```

@enzonotario :wink:

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+"""
+Add a dynamically client-controlled relationship WHERE condition to a fields query.
+The argument it is defined on may have any name but **must** be
```
spawnia: That is actually no longer true, can you remove that from `@whereConditions` too?

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+    public function __construct(Operator $operator)
+    {
+        $this->operator = $operator;
+    }
+
+    public static function definition(): string
+    {
+        return /* @lang GraphQL */ <<<'SDL'
+"""
+Add a dynamically client-controlled relationship WHERE condition to a fields query.
```
spawnia:
```suggestion
Allows clients to filter a query based on the existence of a related model, using
a dynamically controlled WHERE condition that applies to the relationship.
```

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+The argument it is defined on may have any name but **must** be
+of the input type `WhereHasConditions`.
+"""
+directive @whereHasConditions(
+    """
+    The Eloquent relationship that the conditions will be applied to.
+    This argument can be ommited if the field name and the relationship name are the same.
```
spawnia: I think we should actually go for the `hasXyz` style naming convention. Is is a bit more work in terms of docs and implementation, but i think it encourages much better naming by default.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @param  mixed[]  $whereConditions
+     * @param  bool  $init
+     * @param  string  $boolean
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     */
+    public function handleBuilder($builder, $whereConditions, $init = true, string $boolean = 'and')
```
spawnia:
```suggestion
    public function handleBuilder($builder, array $whereConditions, boolean $init = true, string $boolean = 'and')
```

Always use the stricted types possible.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+    {
+        if ($init) {
```
spawnia: Let's just extract a separate function to handle the inner conditions instead of this extra parameter. I think this extra function could even be the same as the function that is used within `@whereConditions`.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * @param  \GraphQL\Language\AST\InputValueDefinitionNode  $argDefinition
+     * @param  \GraphQL\Language\AST\FieldDefinitionNode  $parentField
+     * @return string
+     */
+    protected function restrictedWhereHasConditionsName(InputValueDefinitionNode &$argDefinition, FieldDefinitionNode &$parentField): string
+    {
+        return Str::studly($parentField->name->value)
+            .Str::studly($argDefinition->name->value)
+            .'WhereHasConditions';
```
spawnia: Let's also just use `WhereConditions` here, since they are in essence the same.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Ensure the column name is well formed and prevent SQL injection.
+     *
+     * @param  string  $column
+     * @return void
+     *
+     * @throws \GraphQL\Error\Error
+     */
+    protected static function assertValidColumnName(string $column): void
```
spawnia: Keep the code DRY. This one seems like a good fit for a `public static` helper function somewhere.

We might let the two directives extend the same base directive, too.

tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
```diff
@@ -0,0 +1,471 @@
...
+                    [
+                        'name' => 'CAMEL_CASE',
+                    ],
+                ],
+            ],
+            $enum
+        );
+    }
+
+    public function testWithoutRelationName(): void
```
spawnia: I think we really only need this test and maybe one or two of the other tests in this class.

Let's make sure the implementation is DRY so we can test it once and know it works anywhere.

src/WhereConditions/WhereConditionsServiceProvider.php
chescos: Done!

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Oh, good catch! Done :)

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Done in all files!

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Thanks, removed it.

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Added this.

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Thanks, added!

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Thanks, changed :)

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Done, I think to go with an abstract class that both directives extend is probably the cleanest way to go.

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Done!

tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
chescos: I abstracted as much as possible, no more repetitions. I'll probably be able to remove most of the duplicate tests now.

src/WhereConditions/WhereHasConditionsDirective.php
chescos: Yeah, I agree that the `hasXyz` naming convention is more descriptive to the client than the `xyz` naming convention. I'll add this.

chescos: Done!

tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
chescos: I removed the redundant tests and added 2 new tests specific to the `@whereHasConditions` behavior.

chescos:
@spawnia could you give it another review?

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @param  mixed[]  $whereConditions
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     */
+    public function handleBuilder($builder, $whereConditions)
+    {
+        // Make sure to ignore empty conditions.
```
spawnia: Why? That might still be a valid filter: Determine if any related model exists.

tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
```diff
@@ -0,0 +1,140 @@
...
+    public function testIgnoreEmptyConditions(): void
+    {
+        factory(User::class)->create([
+            'company_id' => null,
+        ]);
+
+        $this->graphQL(/** @lang GraphQL */ '
+        {
+            users(
+                hasCompany: {}
```
spawnia: I think we could differentiate semantically between `null` (don't use this filter at all) and `{}` (filter but without conditions).

spawnia:
This is starting to come together quite nicely. The refactorings and unification has brought the added line count down from ~800 to ~350, which i am much more comfortable with from a maintenance standpoint.

src/WhereConditions/WhereHasConditionsDirective.php
chescos: I wanted to achieve the same behavior that you get when you supply an empty condition (`{}`) to the `@whereConditions` directive; it just has no effect on the query.

The reason is that when you write a client and a query in the schema has a `@whereConditions` arg and a `@whereHasConditions` arg, it's simpler to write a GraphQL query that uses the same "null" value for both args.

tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
chescos: See [above](https://github.com/nuwave/lighthouse/pull/1140#discussion_r366388616) for my reasoning. I agree that it would be nice to be able to query for relationship existence without a condition though. But it would also be good to have consistent behavior between `@whereConditions` and `@whereHasConditions`.

I could imagine 2 solutions to achieve both:

1. Allow `null` as a value for both `@whereConditions` and `@whereHasConditions`, so that this is the unified value that has no effect
2. Add a separate `@hasRelationships` directive that accepts an array of relationships which are tested through Laravel's [has method](https://laravel.com/docs/5.8/eloquent-relationships#querying-relationship-existence)

I do like the second option more as it is a more explicit and less hacky way to query for relationship existence without a condition, and it could also accept an array of relationships. But I'm not sure if you're okay to add an extra directive just for this purpose.

spawnia: *1.* should already be that case if i am not mistaken.

Having `{}` mean "empty conditions" seems like it is logically consistent and also makes sense from a client perspective.

Imagine you have some kind of UI for the filter and you add the condition that your model has to have a related model that fulfils some complex, nested conditions. To widen the range of results, you remove conditions one by one. At a certain point, you will be left with an empty set `{}` of conditions.

Now, it would be surprising to me if that would cause the relationship restriction to not apply at all now, that should be an additional step: removing the filter altogether by making the argument `null`.

spawnia: We should document and test this behaviour, of course.

chescos: That sounds reasonable too - I'm fine with solution *1.* as well. Do you want to document and test the `null` vs. `{}` behavior or should I do that?

Good job on the documentation separation btw, this is a nice way to unify the docs for both directives.

spawnia: If you can add that to this PR, that would be nice. You have done a great job so far with this.

chescos: Okay, done:

- Using `null` on `@whereConditions`: No effect
- Using `{}` on `@whereConditions`: No effect
- Using `null` on `@whereHasConditions`: No effect
- Using `{}` on `@whereHasConditions`: Query for relationship existence without condition

chescos:
Yes, looking nice so far. Anything else that needs improvement?

spawnia:
This is good to go, awesome work.

We might consider marking this feature as stable in `v5` and including it by default, i really like where it's at now.

Issue #1147, opened 2020-01-15
Paginating related models breaks ability to query model directly


codeclem:
**Describe the bug**
I'm unsure if this is a bug or if I'm doing something wrong, but I asked on Slack and Stack Overflow and have not gotten a working solution. With a schema like this:
```
type Query {
  series: [Series!]! @paginate(defaultCount: 10)
  series(id: ID @eq): Series @find

  episodes: [Episode!]! @paginate(defaultCount: 10)
  episode(id: ID @eq): Episode @find
}

type Series {
  id: ID!
  title: String!
  episodes: [Episode]! @hasMany(type: "paginator")
  plot: String!
}

type Episode {
  id: ID!
  title: String!
  season: Int!
  series: Series! @belongsTo
  plot: String!
}
```
This works fine when querying series like this:
```
query series {
  series {
    paginatorInfo {
      total
      currentPage
      hasMorePages
    }
    data {
      title
      episodes {
        paginatorInfo {
          total
          currentPage
          hasMorePages
        }
        data {
          title
        }
      }
    }
  }
}
```
Series and their episodes each get paginated.

But now I can no longer query episodes directly:
```
query episodes {
  episodes {
    paginatorInfo {
      total
      currentPage
      hasMorePages
    }
    data {
      title
    }
  }
}
```

This returns the error `No class 'EpisodePaginator' was found for directive 'paginate'"` The error does not occur if I remove `(type: "paginator")` from the `episodes` field of `type Series`

**Expected behavior/Solution**
Episodes should be paginated when querying series, and also when querying episodes.

**Output/Logs**

<details><summary>Click to expand</summary>

```
{
  "errors": [
    {
      "debugMessage": "No class 'EpisodePaginator' was found for directive 'paginate'",
      "message": "Internal server error",
      "extensions": {
        "category": "schema"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "episodes"
      ],
      "trace": [
        {
          "file": "/home/clem/projects/mediadb/vendor/nuwave/lighthouse/src/Schema/Directives/BaseDirective.php",
          "line": 229,
          "call": "Nuwave\\Lighthouse\\Schema\\Directives\\BaseDirective::namespaceClassName()"
        },
        {
          "file": "/home/clem/projects/mediadb/vendor/nuwave/lighthouse/src/Schema/Directives/BaseDirective.php",
          "line": 140,
[...]

PR #1149, opened 2020-01-16
Avoid erasing the model information from the wrapping paginated results `type` when defining   a paginated `@hasMany` field after a field with `@paginate`

### LINKED ISSUE #1144: No class 'TodoPaginator' was found for directive 'paginate'
I have the same problem
with that guy

https://stackoverflow.com/questions/59726381/laravel-lighthouse-using-paginate-on-related-models-breaks-ability-to-query-mo
```
type Todo {
    id: ID!
    title: String
}

type User {
    id: ID!
    name: String
    email: String
    todos: [Todo!]! @hasMany(type: "paginator")
}

type Query{
    todo(id: ID! @eq): Todo @find
    todos: [Todo!]! @paginate
    user(id: ID! @eq): User @find
}
```
image

image

MilesWuCode:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1144 

tests/Integration/Schema/Directives/HasManyDirectiveTest.php
- testCanQueryHasManyPaginatorBeforeQuery
- testCanQueryHasManyPaginatorAfterQuery
- testCanQueryHasManyNoTypePaginator


spawnia:
Found a fix: Prepending the type name of the parent object to the generated paginated type does the trick.

Not sure if that is the optimal solution though - it's a breaking change in the schema.

codeclem:
I was having this issue as well (see #1147) and I stumbled upon a workaround by specifying the model in the `Type` schema. Ie:
```
type Query {
  series: [Series!]! @paginate(defaultCount: 10, model: "App\\Series")
  series(id: ID @eq): Series @find

  episodes: [Episode!]! @paginate(defaultCount: 10, model: "App\Episodes")
  episode(id: ID @eq): Episode @find
}
```
That gets rid of the error, but I wouldn't call this a fix either. It seems odd that this would even have an effect when Lighthouse is already configured to search the `App` namespace and isn't having any issues finding models otherwise. But maybe this helps you find the problem.

spawnia:
It seems like i found the proper fix.

What happened is that:
1. The field with `@paginate` generated a `type TaskPaginator` that also has the `@modelClass` directive attached. The functionality in `@paginate` requires that to know the model class later.
2. The field with the `@hasMany` paginator overwrote the `type TaskPaginator` but without the `@modelClass`. That by itself works, as the model class is not required here.
3. Querying for the field with `@paginate` now fails, as the model class is no longer known.

I fixed it by always adding the `@modelClass` directive to the generated paginator type.

Issue #1152, opened 2020-01-16
Makefile requires non-existent composer.lock


chescos:
**Describe the bug**

The [Makefile](https://github.com/nuwave/lighthouse/blob/master/Makefile#L20) requires `composer.lock` for the `vendor` entry, but the `composer.lock` file is not part of the repository. When you follow the [Contribution Guide](https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md), you'll first run:

```
docker-compose up -d
make
```

And the `make` command will fail because of the missing dependency.

**Expected behavior/Solution**

The `vendor` entry should not depend on `composer.lock` as it is not part of the repository.

**Steps to reproduce**

1. Run `git clone https://github.com/nuwave/lighthouse`
2. Run `cd lighthouse`
3. Run `docker-compose up -d`
4. Run `make`
5. You'll receive an error because of the missing dependency

**Output/Logs**

<details><summary>Click to expand</summary>

```
make: *** No rule to make target 'composer.lock', needed by 'vendor'.  Stop.
```

</details></br>

**Environment**

Lighthouse Version: 4.8
Laravel Version: 6.x

PR #1155, opened 2020-01-17
Fix running make before first composer install

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/1152

Thanks @chescos for noticing!


spawnia:
@chescos can you join our Slack channel?

Issue #1151, opened 2020-01-16
Registering custom Operator for @whereConditions is not well documented and probably may be done easier


lorado:
**Is your feature proposal related to a problem? Please describe.**

I was trying out `@whereConditions` also before 4.8 release, and it was actually pretty easy to extend Operator enum. As I use PostgreSQL, I need ILIKE operator for my conditions. 

Currently to extend the list of allowed operators I had to create a class and modify generated enum. I solved it like this:

```php
class PSQLOperator extends SQLOperator
{
    public function enumDefinition(): string
    {
        $originalDefinition = parent::enumDefinition();
        $additionOperators = '
    "Simple pattern matching (`ILIKE`)"
    ILIKE @enum(value: "ILIKE")
';
        $pos = strrpos($originalDefinition, '}');
        return substr_replace($originalDefinition, $additionOperators, $pos, 0);
    }
}
```

And then I bound my implementation in a service provider, that comes after `WhereConditionsServiceProvider` like this:

```php
$this->app->bind(Operator::class, PSQLOperator::class);
```

I am kinda new to Laravel, and this was my first time I came to binding in laravel. And as I didn't knew the concept, it took me some time to achieve the expected result.

**Possible solutions I'd like**

Either let us improve the [documentations custom-operator section](https://lighthouse-php.com/master/eloquent/complex-where-conditions.html#custom-operator) (I could create a PR).

OR maybe let's implement the way, where extending enum with new operator (like in my case) can be made easier? Like use a string from config, that will be injected into `SQLOperator::enumDefinition()`

OR maybe let's include ILIKE by default into enum definition?

chescos:
I wouldn't include the `ILIKE` operator by default as it's specific to PostgreSQL and not part of the SQL spec.

But I agree that adding custom SQL operators could be made easier and better documented. We could define the default operators as an array, like so:

```php
$operators = [
    [
        'description' => 'Simple pattern matching (`LIKE`)',
        'operator' => 'LIKE',
        'value' => 'LIKE',
    ],
    ...
];
```

And then we implement a method that converts the array into the enum definition. Then we could also offer a public method that merges an array of custom operators with the default operators before converting them into the enum.

spawnia:
@chescos The issue with that is that simply defining plain operator values is not enough, the way the query builder is called depends upon their arity, so that would have to be another field in the configuration array. I don't love array based configuration for type safety, we could use a class like this instead:

```php
<?php

[...]

PR #1156, opened 2020-01-17
Improve docs on how to register a custom operator for `@whereConditions`

lorado:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Related to #1151 

**Changes**

Added detailed description how to create and register custom operator implementation for `@whereConditions`

**Breaking changes**

no

### DIFF
--- a/docs/master/eloquent/complex-where-conditions.md
+++ b/docs/master/eloquent/complex-where-conditions.md
@@ -363,7 +363,7 @@ class GraphQLServiceProvider extends ServiceProvider
 }
 ```
 
-Don't forget to register you just created service provider in `config/app.php`
+Don't forget to register your new service provider in `config/app.php`:
 
 ```php
 'providers' => [
--- a/docs/master/eloquent/complex-where-conditions.md
+++ b/docs/master/eloquent/complex-where-conditions.md
@@ -323,29 +323,9 @@ You may register a custom `\Nuwave\Lighthouse\WhereConditions\Operator` through 
 This may be necessary if your database uses different SQL operators then Lighthouse's default or you
 want to extend/restrict the allowed operators.
 
-First create a class that implements `\Nuwave\Lighthouse\WhereConditions\Operator`, e.g. `\App\GraphQL\CustomSQLOperator`.
-In following example we will extend `\Nuwave\Lighthouse\WhereConditions\SQLOperator` that actually implements 
-`\Nuwave\Lighthouse\WhereConditions\Operator` and modify result of `enumDefinition()` to include `ILIKE` operator to the
-end of the `SQLOperator` enum. 
-
-```php
-use Nuwave\Lighthouse\WhereConditions\SQLOperator;
-
-class CustomSQLOperator extends SQLOperator
-{
-    public function enumDefinition(): string
-    {
-        $originalDefinition = parent::enumDefinition();
-        $additionOperators = '
-    "Simple pattern matching (`ILIKE`)"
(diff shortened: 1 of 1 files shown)


docs/master/eloquent/complex-where-conditions.md
```diff
@@ -322,3 +322,52 @@ This query would retrieve all persons, no matter if they have a role or not:
...
+
+First create a class that implements `\Nuwave\Lighthouse\WhereConditions\Operator`, e.g. `\App\GraphQL\CustomSQLOperator`.
+In following example we will extend `\Nuwave\Lighthouse\WhereConditions\SQLOperator` that actually implements 
+`\Nuwave\Lighthouse\WhereConditions\Operator` and modify result of `enumDefinition()` to include `ILIKE` operator to the
+end of the `SQLOperator` enum. 
+
+```php
+use Nuwave\Lighthouse\WhereConditions\SQLOperator;
+
+class CustomSQLOperator extends SQLOperator
```
spawnia: I think this is quite fragile, so i would not recommend doing this to users.

docs/master/eloquent/complex-where-conditions.md
```diff
@@ -322,3 +322,52 @@ This query would retrieve all persons, no matter if they have a role or not:
...
+after `\Nuwave\Lighthouse\WhereConditions\WhereConditionsServiceProvider::class` in your `config/app.php`. E.g. create `\App\Providers\GraphQLServiceProvider`:
+
+```php
+use App\GraphQL\CustomSQLOperator;
+use Illuminate\Support\ServiceProvider;
+use Nuwave\Lighthouse\WhereConditions\Operator;
+
+class GraphQLServiceProvider extends ServiceProvider
+{
+    public function register()
```
spawnia:
```suggestion
    public function register(): void
```

docs/master/eloquent/complex-where-conditions.md
lorado: Ye... ok. do you have an example for me? :D

docs/master/eloquent/complex-where-conditions.md
```diff
@@ -322,3 +322,52 @@ This query would retrieve all persons, no matter if they have a role or not:
...
+    {
+        $this->app->bind(Operator::class, CustomSQLOperator::class);
+    }
+}
+```
+
+Don't forget to register you just created service provider in `config/app.php`
```
spawnia:
```suggestion
Don't forget to register your new service provider in `config/app.php`:
```

docs/master/eloquent/complex-where-conditions.md
spawnia: We have been discussing some ways that allow extending the `SQLOperator` more cleanly, let's hold off on our recommended solution until we implemented something better.

Doing string manipulation on the definition string is really hacky.

spawnia:
I think the part that shows how to register custom `Operator` implementations is great, but let's not open the door on extending the `SQLOperator` until we have a nicer solution in place.

lorado:
@spawnia how about this text?

spawnia:
Did one pass over it, you can add final tweaks and merge then.

lorado:
It's good. I was a little confused of packing service provider into `\App\GraphQL` namespace, as I like to put all service providers in one place, like config. But I noticed, that in the enum registering example you also used `\App\GraphQL` namespace. So I'll let it as is to stay consistent.

PR #1160, opened 2020-01-18
Add `model` argument to `@can` directive definition

jampack:
model field is not generated in the definition causing the IDE to raise alerts

- [x ] Added or updated tests
- [x ] Added Docs for all relevant versions
- [x ] Updated CHANGELOG.md

Resolves non

**Changes**
Added model directive to definition area

**Breaking changes**
None

### DIFF
--- a/src/Schema/Directives/CanDirective.php
+++ b/src/Schema/Directives/CanDirective.php
@@ -52,10 +52,10 @@ directive @can(
   instance against which the permissions should be checked.
   """
   find: String
-  
+
   """
   Specify the class name of the model to use.
-  This is only needed when the default model resolution does not work.
+  This is only needed when the default model detection does not work.
   """
   model: String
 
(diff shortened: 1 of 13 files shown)


src/Schema/Directives/CanDirective.php
```diff
@@ -52,6 +52,12 @@ public static function definition(): string
   instance against which the permissions should be checked.
   """
   find: String
+  
+  """
+  Specify the class name of the model to use.
```
jampack: This can be more descriptive..

src/Schema/Directives/CanDirective.php
```diff
@@ -52,6 +52,12 @@ public static function definition(): string
...
+  This is only needed when the default model resolution does not work.
```
jampack: 'model resolution' can be be 'model detection'

src/Schema/Directives/CanDirective.php
spawnia: Open for suggestions.

Issue #1163, opened 2020-01-20
Pagination on relation fields does not respect the relationship


Nohac:
**Describe the bug**
If the `@paginate` directive is applied to a relation field, the relation is not being used, and all data of that model is returned in paginated form.
example:
```graphql
type User {
  id: ID!
  email: String!
  todos: [Todo!]! @paginate(defaultCount: 10)
}
```

**Expected behavior/Solution**
The paginated result should be based on the relation

**Environment**
Lighthouse Version: v4.8.1
Laravel Version: v5.8.35

PR #1164, opened 2020-01-20
Improve `@paginate` docs, mention distinction to `@hasMany`

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/1163 by stating that `@paginate` is not meant for relationships.

Also, add some structure to the docs.


spawnia:
@Nohac do you think the added hint is sufficient? Should we mention the distinction between `@paginate` and `@hasMany` somewhere else?

Nohac:
> @Nohac do you think the added hint is sufficient? Should we mention the distinction between `@paginate` and `@hasMany` somewhere else?

From a noobie perspective I think it's generally confusing that the relation scopes are doing pagination, but now that I know I don't really mind it. I think the note you added to the `@paginate` doc helps a lot. Thanks!

PR #1165, opened 2020-01-21
Support exclude_if and exclude_unless rule

atomita:
- [x] Added or updated tests
~~- [ ] Added Docs for all relevant versions~~
- [x] Updated CHANGELOG.md

**Changes**

Fixed an issue where `exclude_if` or `exclude_unless` rules would not be excluded from args when applied.

**Breaking changes**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix eager-loading relations where the parent type is an `interface` or `union` and
   may correspond to multiple different models https://github.com/nuwave/lighthouse/pull/1035
+- Support `exclude_if` rule and `exclude_unless` rule
 
 ## [4.8.1](https://github.com/nuwave/lighthouse/compare/v4.8.1...4.8.0)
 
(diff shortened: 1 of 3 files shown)


spawnia:
It looks like the `exclude_if` and `exclude_unless` rules do not only validate the input but also change it.

That seems like a really bad design decision from Laravel, are there any other validation rules that work like this?

atomita:
@spawnia 
There seems to be no other than `exclude_if` and `exclude_unless` at laravel v6.11

spawnia:
I think we might not support this feature of Laravel.

First of all, it just seems wrong to use validation to sanitize incoming data. It should be a discrete mechanism in a different step.

Also, i don't think that masking client errors is a great idea in general. This behaviour is not transparent to clients. Without a description, there is no way for them to know that the argument they passed is going to be ignored. If the backend ever changes, there is no way for the client to know.

In terms of implementation, changing the incoming data in a validation step breaks some assumptions that Lighthouse currently makes. We would have to go out of our way to support this, for instance, we would also need to mutate the `ArgumentSet` - a data structure that holds the passed argument values and their types.

atomita:
@spawnia 
I'm sad, but I understand.

spawnia:
Thank you for your contribution and your understanding. Please don't let this discourage you, your PR certainly was well done and i very much welcome future contributions.

I mentioned the fact that those rules are not supported in the directive definition, i am glad you brought it up: https://github.com/nuwave/lighthouse/pull/1168. I hope the added section about sanitization helps you.

PR #1166, opened 2020-01-21
Fix renaming input fields that are nested within lists using `@rename`

enzonotario:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/1167

**Changes**

`@rename` could be used within nested inputs, but failed if the input was nested within a list.

The following example was broken but now works:

```graphql
type Query {
    foo(
        input: [FooInput]
    ): Boolean @mock
}

input FooInput {
    baz: String @rename(attribute: "bar")
}
```

**Breaking changes**

None.


spawnia:
Thanks @enzonotario for bringing up this bug and adding the failing test case. That is always a great starting point.

PR #1169, opened 2020-01-23
Fix nested mutation `upsert` across two levels of BelongsTo relations

enzonotario:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

It's a failing test that shows a weird issue:

Nested mutation works with a `HasMany` relationship that has two `BelongsTo` relationships, but just when using `connect`, but not with `upsert`. I know that the example is maybe confussing and not a real-case, but I did it as is just trying to re-use the existing Lighthouse's test suite models (I could add a real-word example if needed, but it just will change the models and relationships names).

The werird here is that it works with Lighthouse 4.7, so I think that the amazing #899 PR has been involved here.

It may be the expected result now, but I think we lose something (as it was working in Lighthouse 4.7).

**Changes**

None so far.

**Breaking changes**

None.


spawnia:
@enzonotario great report, as always.

I found the issue was actually rooted within the combination of `upsert` and two nested BelongsTo relationships. What happened is that within `upsert`, a call on `->newInstance()` erased all knowledge of already-attached relations. I found no good reason for that to be there, so i removed it.

It seems like the order of results within `BelongsToMany` changed as a result of the now fixed order of operations. I think that should be fine, as the tests show that the result is basically still the same.

src/Execution/Arguments/UpsertModel.php
```diff
@@ -27,16 +27,24 @@ public function __construct($previous)
...
+            $id = $args->arguments['id']
+                ?? $args->arguments[$model->getKeyName()]
+                ?? null
+        ) {
+            if (
+                $existingModel = $model
+                    ->newQuery()
+                    ->find($id->value)
+            ) {
+                // TODO there is a slight lingering bug here. In case the $model already
```
enzonotario: If I understand you correctly, I think this is issue is not present, but even before this change, considering the previous code:
```
$id = $args->arguments['id']
    ?? $args->arguments[$model->getKeyName()]
    ?? null;

if ($id) {
    $model = $model->newQuery()->find($id->value)
        ?? $model->newInstance();
} else {
    $model = $model->newInstance();
}
```
When there was an `id` input, and the model exists, `$model = $model->newQuery()->find($id->value)`, just like now.

That said, I don't really know how can we deal with it, since I think it needs a design decision. I even think that this could lead us to undesired effects, like allowing to the client to change the relationships of some models, jus think that an `User` could create a `Comment`, then it can update it and change the `user` relationship to another `User`, passing all responsabilities to that User.

I have added a test hoping to explain it a bit more.

enzonotario:
Thanks a lot @spawnia ! Now I understand what have been happened.

I'm with you about the `BelongsToManyTest`, I also think it's fine.

src/Execution/Arguments/UpsertModel.php
spawnia: Well done, yeah this works fine.

PR #1174, opened 2020-01-29
Fix handling of nested mutation operations that receive `null`

spawnia:
(cherry picked from commit 03a6d6bf8bf52217a05d9308a899c437275cadb6)

- [x] Added or updated tests
~~- [ ] Added Docs for all relevant versions~~
- [x] Updated CHANGELOG.md

**Changes**

When passing `null` into a nested mutation operation, such as `create` or `sync`, Lighthouse throws an error.

Since `null` is an acceptable value, it should be handled gracefully and simply do nothing.

**Breaking changes**

None.


spawnia:
@sp3259 please review.

PR #1186, opened 2020-02-10
Fix `\Nuwave\Lighthouse\Testing\MakesGraphQLRequestsLumen` test helper

rafaelmarcelino19:
Hello this pull request resolve new error caused after the new changes in MakesGraphQLRequestsLumen.

**The error**:
Return value of TestCase::postGraphQL() must be an instance of MakesGraphQLRequestsLumen, instance of UserTest returned

**Solution**:
a trait cannot be instantiated, so there is no way to return, change the return of the methods to 'self' in MakesGraphQLRequestsLumen. And then, this will work.

Also changing in $this->introspectionResult->decodeResponseJson() the $introspectionResult is not a instance of TestResponse then don't have the method decodeResponseJson()


src/Testing/MakesGraphQLRequestsLumen.php
```diff
@@ -30,15 +30,15 @@ trait MakesGraphQLRequestsLumen
...
      * @return $this
      */
-    protected function graphQL(string $query, array $variables = null, array $extraParams = []): MakesGraphQLRequestsLumen
+    protected function graphQL(string $query, array $variables = null, array $extraParams = []): self
     {
         $params = ['query' => $query];
 
         if ($variables) {
-            $params += ['variables' => $variables];
+            $params = array_merge($params, ['variables' => $variables]);
```
spawnia: Why did you change this?

rafaelmarcelino19: That was unnecessary, but i don't see problems

spawnia: Let's not diverge the test classes, they should be kept as close together as possible.

rafaelmarcelino19: OK

spawnia:
Thanks @rafaelmarcelino19 

A Lumen test conformance suite would be great to have in the future, i certainly want to try and support it.

PR #1194, opened 2020-02-17
WIP: Ensure `toResponse` is not called in Pipeline

### LINKED ISSUE #1193: Returning a `Responsable` object from a resolver causes `toResponse` to be called
**Describe the bug**

When a single instance of a class implementing the `Responsable` interface is returned from a resolver, it's `toResponse` method is called.

The offending piece of code is this bit of Laravel: https://github.com/laravel/framework/blob/5.8/src/Illuminate/Pipeline/Pipeline.php#L166

Later versions of the Pipeline class do not exhibit this behaviour, they make no assumptions about what the pipeline is used for.

The object should be processed by GraphQL as is.

**Steps to reproduce**

1. Set up Lighthouse in Laravel 5.8
1. `@find` a model implementing `Responsable`

spawnia:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #1193

**Changes**

**Breaking changes**


spawnia:
@SirLamer i actually could not reproduce the issue, can you try and dig further?

In your case, `toResponse` seems to be getting called somewhere - but the Pipeline does not seem to do it. Can you try looking into a stacktrace and identify where exactly the call is coming from?

composer.json
```diff
@@ -25,7 +25,7 @@
     "require": {
         "php": ">= 7.1",
         "ext-json": "*",
-        "illuminate/contracts": "5.5.* || 5.6.* || 5.7.* || 5.8.* || ^6.0",
+        "illuminate/contracts": "5.7.*",
```
half2me: why constrict to 5.7.x?

spawnia: This PR is not going to land in `master`, it is only meant to find an issue specific to this Laravel version.

spawnia:
Closing this, @SirLamer feel free to investigate further.

PR #1200, opened 2020-02-19
Add `@hash` directive and deprecate `@bcrypt`

Chrissi2812:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds a directive for Laravel's `Hash::make` facade so Lighthouse uses the defined driver of laravel for password hashing, instead of bcrypt via `@bcrypt` directive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
+- Add `@hash` directive which uses Laravel's `Hash::make` to create an appropriate hash.
 
 ## 4.9.0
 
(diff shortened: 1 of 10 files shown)


spawnia:
Can this be done without facades? We generally don't use them because we want to support Lumen, which does not enable them by default.

Chrissi2812:
@spawnia I guess something like this https://stackoverflow.com/a/23574080/6370595 could work. I didn't work with Lumen so I don't know exactly whats supported. 

But what what speaks against adding a hash directive, you already have a bcrypt directive? 
Basically it could just replace the `@bcrypt` directive and you could then configurate the hash directive to use the driver you want (Bcrypt, Argon2i or Argon2id).

spawnia:
I think that adding this directive is a good idea, i just want to see it implement without the use of facades.

spawnia:
Thank you @Chrissi2812 

This will land in the next release, and i will actually replace `@bcrypt` with it in the upcoming major version.

PR #1203, opened 2020-02-19
Add support to extend `input`, `interface` and `enum` types

### LINKED ISSUE #1202: Extending input types
Following this requirement from Apollo:
https://github.com/apollographql/graphql-tools/issues/641

At this moment, we can extend a type (including Query/Mutations). It'd be great to add the same functionality for input types.
```
type Query {
  getFoo: Foo
}
type Foo {
  foo: String
}
extend type Foo {
  bar: String
}

type Mutation {
  updateFoo(input: FooInput): Foo
}
input FooInput {
  foo: String
}
extend input FooInput {
  bar: String
}
```
Bonus track, add exteding for everything interface, enums, etc. When we use schema stitching it is a great feature to have.

faiverson:
- [x] Added or updated tests

Resolves #1202

### DIFF
--- a/src/Schema/AST/ASTBuilder.php
+++ b/src/Schema/AST/ASTBuilder.php
@@ -2,16 +2,22 @@
 
 namespace Nuwave\Lighthouse\Schema\AST;
 
+use GraphQL\Language\AST\EnumTypeDefinitionNode;
 use GraphQL\Language\AST\EnumTypeExtensionNode;
+use GraphQL\Language\AST\InputObjectTypeDefinitionNode;
 use GraphQL\Language\AST\InputObjectTypeExtensionNode;
+use GraphQL\Language\AST\InterfaceTypeDefinitionNode;
 use GraphQL\Language\AST\InterfaceTypeExtensionNode;
 use GraphQL\Language\AST\ObjectTypeDefinitionNode;
 use GraphQL\Language\AST\ObjectTypeExtensionNode;
+use GraphQL\Language\AST\TypeDefinitionNode;
+use GraphQL\Language\AST\TypeExtensionNode;
 use Illuminate\Contracts\Config\Repository as ConfigRepository;
 use Illuminate\Contracts\Events\Dispatcher as EventDispatcher;
 use Illuminate\Support\Arr;
 use Nuwave\Lighthouse\Events\BuildSchemaString;
 use Nuwave\Lighthouse\Events\ManipulateAST;
+use Nuwave\Lighthouse\Exceptions\DefinitionException;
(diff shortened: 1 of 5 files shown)


tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -43,13 +43,80 @@ public function testCanMergeTypeExtensionFields(): void
         );
     }
 
+    public function testCanMergeInputExtensionFields(): void
+    {
+        $this->schema = '
```
spawnia:
```suggestion
        $this->schema = /** @lang GraphQL */ '
```

src/Schema/AST/ASTBuilder.php
```diff
@@ -180,13 +184,23 @@ protected function applyTypeExtensionManipulators(): void
...
+                    $typeExtension instanceof InterfaceTypeExtensionNode
+                ) {
                     $relatedObjectType = $this->documentAST->types[$typeName];
 
                     $relatedObjectType->fields = ASTHelper::mergeUniqueNodeList(
                         $relatedObjectType->fields,
                         $typeExtension->fields
                     );
+                } elseif ($typeExtension instanceof EnumTypeExtensionNode) {
+                    $relatedObjectType = $this->documentAST->types[$typeName];
```
spawnia:
```suggestion
                    $extendedEnum = $this->documentAST->types[$typeName];
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -58,4 +125,54 @@ public function testDoesNotAllowDuplicateFieldsOnTypeExtensions(): void
...
+        input Inputs {
+            foo: String
+        }
+
+        extend input Inputs {
+            foo: Int
+        }
+        ';
+
+        $this->expectException(DefinitionException::class);
```
spawnia: What are the actual produced error messages? Are they somewhat understandable?

spawnia:
Thank you for this PR, looks pretty good overall. With some polish, it should be good to go.

An entry in the `CHANGELOG` would be great, too.

tests/Unit/Schema/AST/ASTBuilderTest.php
faiverson: I think so, it will show something like:
```
Nuwave\Lighthouse\Exceptions\DefinitionException: Duplicate definition TWO found when merging
```

src/Schema/AST/ASTBuilder.php
faiverson: Changed!

faiverson:
> An entry in the `CHANGELOG` would be great, too.
Done in the new commit

PR #1208, opened 2020-02-25
Add option `passOrdered` to `@method` to pass just the arguments as ordered parameters

### LINKED ISSUE #1052: Method directive without resolver arguments
I often have methods on my models which I want to use in graphql, like this.

```
    /**
     * Checks if the job is paused.
     *
     * @param bool $isPaused
     * @return bool
     */
    public function isPaused(bool $isPaused = true): bool
    {
        return ($this->is_paused ?? false) === $isPaused;
    }
```
With an argument on a type like this
```
paused: Boolean! @method(name: "isPaused")
```

Looking at it, there shouldn't really be a reason for it to fail, as it is not obvious that data is passed to my method. 

However having the option to get the resolver arguments are still really useful and should still be there.
My solution for this is to add a new argument to the `method` directive called `withResolver` which is an boolean.
```
paused: Boolean! @method(name: "isPaused" withResolver: true)
```

And this argument defaults to `false`, so if it is not set, then the arguments won't be supplied to the method.

This will be a breaking change as it would require all users who uses the arguments to actually update their schema.

JasonTheAdams:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1052 

**Changes**

This introduces the "pass" directive argument to the `@method` directive. This makes it possible to pass field arguments to the model method, as well as controlling the order of the parameters.

**Breaking changes**

This should not break anything. If the `pass` argument is not used then it will default to working exactly as it did before.

The only note here is that I didn't find any previously existing `MethodDirective` tests, so I'm not sure how we'd like to test this (if we do).


JasonTheAdams:
@spawnia How would you like me to add this to the `CHANGELOG.md`? Are you planning a 4.10 release?

docs/master/api-reference/directives.md
```diff
@@ -1439,9 +1439,32 @@ directive @method(
...
+  """
+  The field arguments to pass (in order) to the underlying method. Each string in the array
+  should correspond to an argument of the field.
+  """
+  pass: [String!]
 ) on FIELD_DEFINITION
 ```
 
+### Examples
+Simply call the method and pass the resolver arguments to the method:
```
spawnia:
```suggestion

Simply call the method and pass the resolver arguments to the method:

```

JasonTheAdams:
I'm honestly not sure why tests are failing, especially the ones I wrote. They're passing locally.

enzonotario:
There is a `MethodDirectiveTest` but at `Unit` namespace. I just have fixed the namespace of the test you have created, but maybe we have to join it in one Test.

spawnia:
While writing tests for validating that `pass` matches the arguments, i found that its definition is quite redundant. We might just replace it with a boolean flag and use the lexical scope of the argument definitions to reorder and spread the incoming arguments.

spawnia:
Added an implementation for an argument `passOrdered`, do you think we can just use that instead of the explicit argument naming?

I think that the added flexibility of explicit arguments is only that it allows to skip arguments. That seems unnecessary as the argument could just be removed instead. I would rather go for a simple and boilerplate-free solution that covers 99% of all use cases.

JasonTheAdams:
@enzonotario Got it, thanks for fixing that!

@spawnia Hmm... at that point why not just check to see if there are any arguments and pass them? I suppose `passOrdered` could default `true`, so it could be set `false` to prevent it — but that's less backwards compatible.

JasonTheAdams:
I do agree, though, that I can't think of a scenario where I would want some field arguments that aren't passed to the method. This directive is really a terminal field resolver, so it wouldn't make sense to have further resolving or anything else that would use extra arguments. As such, the presence of arguments (and their order) could be enough to be passed to the method — otherwise, why would the arguments exist?

spawnia:
@JasonTheAdams thanks so much for getting this going, i think we arrived at a really nice solution now.

PR #1212, opened 2020-02-27
Correct link to FieldMiddleware interface in docs

andershagbard:


spawnia:
Thanks @andershagbard good catch

PR #1216, opened 2020-03-01
Access nested inputs with dot notation in `find` option of `@can` #1216

### LINKED ISSUE #1162: @can(find: String) can't access to nested values
Given the following schema:

```graphql
extend type Mutation {
    sendConversationMessage(input: SendConversationMessageInput!): ConversationMessage!
        @can(ability: "message", find: "input.conversationId" model: "App\\Conversation")
        @middleware(checks: ["auth:api"])
}

input SendConversationMessageInput {
    conversationId: ID! @rules(apply: ["required", "exists:conversations,id"])
    message: String @rules(apply: ["required", "max:1000"])
}
```

The `CanDirective` fails as it can't find an index in the arguments called `input.conversationId`.

I'd expect `@can` to access to my input's conversationId value in order to check against the policy and the given model.

As we can see in https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/CanDirective.php#L146 this could probably easily be achieved if we instead used dot notation to access the arguments array to be queried (as we do with config files in Laravel, for example). That said, I'm not familiar with the in & outs of the library to know if this is an acceptable solution. If it's an acceptable solution, I'll be happy to send a PR and tests for it.

**Output/Logs**

```
[{
    "errors": [
        {
            "debugMessage": "Undefined index: input.conversationId",
            "extensions": {
                "category": "internal"
            },
            "locations": [
                {
                    "line": 2,
                    "column": 21
                }
            ],
            "path": [
                "sendConversationMessage"
            ],
            "trace": [
                {
                    "file": "/Users/ipalaus/Projects/xyz/xyz-api/vendor/nuwave/lighthouse/src/Schema/Directives/CanDirective.php",
                    "line": 128,
                    "call": "Illuminate\\Foundation\\Bootstrap\\HandleExceptions::handleError(8, 'Undefined index: input.conversationId', '/Users/ipalaus/Projects/xyz/xyz-api/vendor/nuwave/lighthouse/src/Schema/Directives/CanDirective.php', 128, array(3))"
                },
                {
                    "file": "/Users/ipalaus/Projects/xyz/xyz-api/vendor/nuwave/lighthouse/src/Schema/Directives/CanDirective.php",
                    "line": 100,
                    "call": "Nuwave\\Lighthouse\\Schema\\Directives\\CanDirective::modelsToCheck(instance of Nuwave\\Lighthouse\\Execution\\Arguments\\ArgumentSet, array(1))"
                },
                {
...
```

</details></br>

**Environment**

Lighthouse Version: 4.7.2 - but for the looks of the CanDirective in master, this is still happening.
Laravel Version: 6.6.0

andershagbard:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1162

**Changes**
The `find` method of the can directive now uses [Arr::get](https://laravel.com/docs/6.x/helpers#method-array-get), to search the model. This allows to pass a string in a dot notation like `data.foo.bar`.

**Breaking changes**
Previously it would throw an `Undefined Index Error` if the string did not map to a value. Now it will throw a `DefinitionException`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Added
 
-- Updated option `find` to `@can` to allow dot notation https://github.com/nuwave/lighthouse/pull/1216
+- Access nested inputs with dot notation using the `find` option of `@can` https://github.com/nuwave/lighthouse/pull/1216
 - Add `@hash` directive which uses Laravel's hashing configuration https://github.com/nuwave/lighthouse/pull/1200
 - Add option `passOrdered` to `@method` to pass just the arguments as ordered parameters https://github.com/nuwave/lighthouse/pull/1208
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Added
 
+- Updated option `find` to `@can` to allow dot notation https://github.com/nuwave/lighthouse/pull/1216
 - Add `@hash` directive which uses Laravel's hashing configuration https://github.com/nuwave/lighthouse/pull/1200
 - Add option `passOrdered` to `@method` to pass just the arguments as ordered parameters https://github.com/nuwave/lighthouse/pull/1208
 
--- a/docs/4.9/api-reference/directives.md
+++ b/docs/4.9/api-reference/directives.md
@@ -452,6 +452,8 @@ directive @can(
   """
   The name of the argument that is used to find a specific model
   instance against which the permissions should be checked.
+
+  You may pass the string as a dot notation to search in a array.
   """
   find: String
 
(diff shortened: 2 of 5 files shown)


spawnia:
This looks really clean, good work.

> Previously it would throw an Undefined Index Error if the string did not map to a value. Now the value will be defaulted to null, and throw a ModelNotFoundException.

This kind of breakage is fine in a new minor version. What do you think about adding an explicit exception for that case? It seems like it is distinct from not finding a model.

andershagbard:
> What do you think about adding an explicit exception for that case? It seems like it is distinct from not finding a model.

I actually like the possibility that it can "fail" silently. This adds more flexibility as the policy accepts `null` in place of `$user`.

```graphql
type CreateCommentData {
  user_id: ID # Not required
  title: String!
  comment: String!
}
```

andershagbard:
> What do you think about adding an explicit exception for that case? It seems like it is distinct from not finding a model.

Good idea. Just a generic server exception?

```php
throw new \Exception("Undefined index. Could not find $find");
```

spawnia:
I think that the onus is on the developer, they should just make the argument required. 

We usually utilize the `*Manipulator` directive interfaces for this. They contain functions that are run during schema validation, so we could just throw a `DefinitionException` there to enforce the argument is defined as non-null `!`.

andershagbard:
Updated

spawnia:
Nicely done, thank you

PR #1225, opened 2020-03-07
Throw GraphQL error when ModelNotFoundException is thrown in CanDirective

andershagbard:
- [x] Added or updated tests
- [x] Added Docs for all relevant versions N/A
- [x] Updated CHANGELOG.md

**Changes**
Catch `ModelNotFoundException` in `CanDirective` and instead throw `GraphQL\Error\Error`.

Also improved a few tests with GraphQL errors.

**Breaking changes**
Using `@can` no longer throws `ModelNotFoundException` when model is not found. Instead the object returns as null, and a GraphQL error is sent. Status code is also changed to 200.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
-### Changes
+
+### Changed
 
 - Throw user readable `Error` instead of `ModelNotFoundException` when model is not found in `@can` https://github.com/nuwave/lighthouse/pull/1225
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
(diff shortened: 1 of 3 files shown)


tests/Integration/Schema/Directives/CanDirectiveDBTest.php
```diff
@@ -77,14 +76,23 @@ public function testFailsToFindSpecificModel(): void
...
+            'errors' => [
+                [
+                    'message' => 'No query results for model [Tests\Utils\Models\User] not-present',
+                ],
+            ],
+            'data' => [
+                'user' => null,
+            ],
+        ])->assertErrorCategory('graphql')
+            ->assertStatus(200);
```
spawnia: That should be the case anyhow, regardless of what error may have occurred. 

I added one more test for this [here](https://github.com/nuwave/lighthouse/pull/1228), please remove it from this test class so it stays focused.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
+- Using `@can` no longer throws `ModelNotFoundException` when model is not found. Instead the object returns as null, and a GraphQL error is sent. Status code is also changed to 200.
```
```suggestion
### Changes

- Throw user readable `Error` instead of `ModelNotFoundException` when model is not found in `@can` https://github.com/nuwave/lighthouse/pull/1225
```

tests/Integration/Schema/Directives/CanDirectiveDBTest.php
andershagbard: Do we still want to assert the `errors.message` or do you just wan't me to remove `assertErrorCategory` and `assertStatus`?

Issue #1236, opened 2020-03-17
Add Custom NameSpace, then generation command failed.


mikakane:
**Describe the bug**

add custom namespace , like`lighthouse.namespace.mutations` and so on , then generation commands failed.

I run command below 

```
 php artisan lighthouse:mutation Message -v
```

i can get log below

```
 Maximum function nesting level of '256' reached, aborting! {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Maximum function nesting level of '256' reached, aborting! 

[stacktrace]
... 
#13 ..../vendor/laravel/framework/src/Illuminate/Console/GeneratorCommand.php(102): Illuminate\\Console\\GeneratorCommand->qualifyClass('Message')
#14 ..../vendor/laravel/framework/src/Illuminate/Console/GeneratorCommand.php(102): Illuminate\\Console\\GeneratorCommand->qualifyClass('Message')
#15 ..../vendor/laravel/framework/src/Illuminate/Console/GeneratorCommand.php(102): Illuminate\\Console\\GeneratorCommand->qualifyClass('Message')
#16 ..../vendor/laravel/framework/src/Illuminate/Console/GeneratorCommand.php(102): Illuminate\\Console\\GeneratorCommand->qualifyClass('Message')
#17 ..../vendor/laravel/framework/src/Illuminate/Console/GeneratorCommand.php(102): Illuminate\\Console\\GeneratorCommand->qualifyClass('Message')
....
```

this is valid configuration. this will work fine.

```
    'namespaces' => [
        'mutations' => [
            'App\\GraphQL\\Mutations',
            'App\\GraphQL\\Mutations2',
        ],
    ],
```

But configuration below, command wont work fine.

```
    'namespaces' => [
        'mutations' => [
            'App\\GraphQL\\Mutations',
            'Acme\\NewService\\GraphQL\\Mutations',
        ],
    ],
```

**Expected behavior/Solution**

**Steps to reproduce**

1.
2.
3.

**Output/Logs**

<details><summary>Click to expand</summary>

```
# Add in log output/error messages here
```

</details></br>

**Environment**

Lighthouse Version:
Laravel Version:

mikakane:
this will happen because `$matching` is empty in LighthouseGeneratorCommand::commonNamespace.

https://github.com/nuwave/lighthouse/blob/master/src/Console/LighthouseGeneratorCommand.php#L51

spawnia:
[...]

PR #1237, opened 2020-03-17
Update LighthouseGeneratorCommand.php

mikakane:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

ref #1236

**Changes**

When `$matching` is empty, command get infinit loop, and failed.

add `return end($namespaces) ` for fallback when `$matching` is empty.

**Breaking changes**


spawnia:
I decided to handle the issue a bit differently, thanks for bringing it up.

https://github.com/nuwave/lighthouse/pull/1245

PR #1240, opened 2020-03-18
Add artisan command `lighthouse:directive` to add directive class #1240

WoutervdBrink:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

Note: I could not find any tests for the other generator command and don't really know what to test or how, so if that's required / wanted I'd definitely like some help on that.

**Changes**

Adds a `lighthouse:directive` command which creates an empty directive class in the appropriate namespace according to the `lighthouse.php` config file.

**Breaking changes**

None.

**Final remarks**

I initially wanted to let the command also implement / include the various directive interfaces, but could not find a way to make that developer-friendly without creating a lot of options for the command. Thus I stuck with an empty class.

### DIFF
--- a/src/Console/stubs/directive.stub
+++ b/src/Console/stubs/directive.stub
@@ -1,0 +1,10 @@
+<?php
+
+namespace DummyNamespace;
+
+use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
+
+class DummyClass
+{
+    // TODO implement the directive
+}
--- a/src/Console/stubs/directive.stub
+++ b/src/Console/stubs/directive.stub
@@ -4,7 +4,7 @@ namespace DummyNamespace;
 
 use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
 
-class DummyClass
+class DummyClass extends BaseDirective
 {
     // TODO implement the directive
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Add `AttemptAuthentication` middleware to optionally log in users and delegate access guards
   to the field level https://github.com/nuwave/lighthouse/pull/1197
 
+- Add `lighthouse:directive` Artisan command to create an empty directive class
+
 ### Fixed
 
 - Eager load nested relations using the `@with` directive https://github.com/nuwave/lighthouse/pull/1068 
--- a/docs/master/api-reference/commands.md
+++ b/docs/master/api-reference/commands.md
@@ -8,6 +8,12 @@ are namespaced under `lighthouse`.
 Clear the cache for the GraphQL AST.
 
     php artisan lighthouse:clear-cache
+
+## directive
+
+Create a class for a GraphQL directive.
+
+    php artisan lighthouse:directive
    
 ## ide-helper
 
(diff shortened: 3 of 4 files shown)


lorado:
As inspiration to make this command more useful:

- maybe take one argument for "directive name" and translate it to valid php class name (appending `Directive` suffix)
- maybe use use [prompting for input](https://laravel.com/docs/master/artisan#prompting-for-input) with "Multiple Choice Questions". We may asking "what interfaces should this directive implement" or kind of.

WoutervdBrink:
Thanks for your suggestions, @lorado. You're quite right about the class name; I forgot that they need the `Directive` suffix.

About the input prompts: agreed, that might be a solution, perhaps only letting the command ask question(s) with an option.

I don't have much experience with directives; how often will a directive be implementing more than one interface?

src/Console/stubs/directive.stub
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace DummyNamespace;
+
+use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
+
+class DummyClass
```
spawnia: This should extend `BaseDirective`, which makes implementation much simpler.

spawnia:
Thanks for this PR, i quite like the idea of adding that in.

@lorado makes a great point, i think the command should take a short directive name and make the necessary modifications `ucfirst($name) . 'Directive'`.

Interactivity for choosing interfaces to implement would be great, i don't think that automation is a concern here.

> I don't have much experience with directives; how often will a directive be implementing more than one interface?

Quite often, most commonly one step to manipulate/validate the schema and another for resolve time.

spawnia:
If you can figure out a convenient way to test the generator's, that would be great. I must admit they have accidentally been broken, so i see value in having that. Not a hard requirement, though :)

src/Console/stubs/directive.stub
WoutervdBrink: Whoops, forgot about that. Thank you!

WoutervdBrink:
Hi @spawnia, first of all thank you very much for your feedback and interest!

I have implemented some of your suggestions and pushed them; take a look if you'd like and let me know what you think. The command accepts three options `--type`, `--field` and `--argument` (multiple are possible) and depending on these options the command `confirm`s if you would like to implement certain associated interfaces.

Finally there is a `--validator` option which directly generates a validator directive, but does not implement any interfaces, as I suspected you probably will not be using those.

The command is becoming quite complex so I definitely wish to create some tests, which are a work-in-progress. I'll also take a look at creating tests for the other generator commands, which will probably be trivial after tackling the tests for this command.

Would you like to receive the tests for the other generator commands in a separate PR or is it okay if I add them here?

WoutervdBrink:
Oh, apparently GitHub didn't show me yet that you closed this PR... should have marked it as WIP. The command is not actually in the service provider yet in the commit you just merged :see_no_evil:. Yet more proof it needs tests! :wink:

spawnia:
Did not catch that one, ups. Looking forward to an additional PR that adds the changes you described. The proposed DX seems pretty good.

> Finally there is a --validator option which directly generates a validator directive

Validator directives are going to be replaced by validator classes as a first-class concept in the upcoming v5, no need to implement anything with them now. Not trying to encourage you too much, but if you need to break something to improve commands in general - now is the time ;)

I don't mind having the changes and additional tests in a bigger PR, feel free to go ahead.

WoutervdBrink:
Alright, I will continue working on the tests and create a new PR when they're ready. Thank you.

PR #1254, opened 2020-03-24
WIP: Where conditions and order by relation

StelianAndrei:
**Disclosure**

This is a proposal that might or might not be in the scope or on the roadmap for this package. The reason of this PR is to determine if there is any point of pursuing this further.

**Proposal**
What this PR attempts to accomplish is:

- allow the `whereConditions` directive to filter based on relation attributes using whereHas()
- allow the `orderBy` directive to work with relation attributes

Considering the following schema:

```
type County {
    id: ID!
    name: String!
    cities: [City] @hasMany
}

type City {
    id: ID!
    county: County! @belongsTo
    name: String!
}

type Query {
    counties(
        filter: _ @whereConditions, 
        orderBy: [OrderByClause!] @orderBy
    ): [County!]!
}
```
This PR aims to make queries similar to the following possible:

```
{
    counties(
        filter:{
            column: "cities.name", operator: LIKE, value: "%Test%"
        }
    ) {
        id
    }
}
```

The same would also be possible for the orderBy directive: ordering by a property of a related model.

As I said, this is just a proposal and nowhere near ready to be merged as documentation and tests are missing. Also I'm more than certain not every scenario has been covered by this PR, especially since I'm fairly new to GraphQL and Lighthouse.

Any feedback is greatly appreciated. Thank you

### DIFF
--- a/src/OrderBy/OrderByDirective.php
+++ b/src/OrderBy/OrderByDirective.php
@@ -5,6 +5,7 @@ namespace Nuwave\Lighthouse\OrderBy;
 use GraphQL\Language\AST\FieldDefinitionNode;
 use GraphQL\Language\AST\InputValueDefinitionNode;
 use GraphQL\Language\AST\ObjectTypeDefinitionNode;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Str;
 use Nuwave\Lighthouse\Schema\AST\DocumentAST;
 use Nuwave\Lighthouse\Schema\AST\PartialParser;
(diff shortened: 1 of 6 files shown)


spawnia:
@StelianAndrei thanks for the proposal, just got around to reviewing it. Have you gotten any further with your experiments.

I do see how adding more power to those directives can be useful, but am also wary of making them to complex. There comes a point at which they are just reimplementing SQL.

In their current form, specifying the allowed columns and having them generate well-typed arguments is great for clients that don't know about the internal structure of the application. Adding flexibility generally lessens the ability to make the API self-describing.

There is also a degree of control that i think is valuable. Allowing clients to construct arbitrary queries might enable them to construct really expensive queries that can put a big strain on your database.

StelianAndrei:
Hi @spawnia . Sorry for the very late reply. For some reasons I'm not getting notifications from GitHub.

I see your concerns regarding the complexity of the queries and the implications this might have on the database. The project for which this PR was created was actually a closed access one, so it wasn't really an issue for us, however in the bigger context, I see how this could be a huge problem.

Sadly, I haven't managed to get any more progress on this, mostly because under the current global situation, the project that it was meant to be implemented on has been put on hold and attention has been diverted to other projects.

As an alternative maybe an additional package that could complement Lighthouse and provide this kind of functionality would be a better solution. With the proper information for the developer regarding the implications and possible risks, of course.

That being said, I'll close this PR for now. Thank you

spawnia:
@StelianAndrei if you can trust the client, you can be much more liberal in allowing manipulation of the query. Injecting fragments of raw SQL might be a reasonable choice. That only works when you are in a very controlled environment where you can be sure no one is messing with your endpoint.

Issue #1266, opened 2020-04-07
CORS Error in Laravel 7 using Laravel Lighthouse


crazyabdul:
**Describe the bug**
I keep getting a CORS error despite following this [solution](https://github.com/nuwave/lighthouse/issues/445#issuecomment-453401025) where `\Fruitcake\Cors\HandleCors::class` was added to middlewares in the lighthouse config file.

**Expected behavior/Solution**
I expected the CORS error to be gone after adding laravel-cors to the lighthouse config but the probem persists.

**Steps to reproduce**

1. install laravel 7.2.2 (It comes with laravel-cors by default)
2. Install lighthouse
3. add `\Fruitcake\Cors\HandleCors::class` to `route->middleware` in `config\lighthouse`
4. make a request from the browser

**Output/Logs**

<details><summary>Click to expand</summary>

```
Access to fetch at 'http://localhost:8000/graphql' from origin 'http://localhost:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

</details></br>

**Environment**

Lighthouse Version: 4.10.2
Laravel Version: 7.2.2

spawnia:
[This comment](https://github.com/nuwave/lighthouse/issues/445#issuecomment-575885544) suggests that the `OPTIONS` method may be the culprit here. Can you check which exact request is triggering the error?

lorado:
Ye, I also tip for the `OPTIONS` request to fail here. Try to add `\Fruitcake\Cors\HandleCors::class` to `$middleware` in `App\Http\Kernel`.

Don't forget to add your graphql route to `pahts` in `config/cors.php`. Also configure allowed origins, for testing you can take `['*']`. Here is my working example of `config/cors.php`:

```php
return [
    'paths' => ['api/*', 'graphql*'],
    'allowed_methods' => ['*'],
    'allowed_origins' => explode(',', env('ALLOWED_CORS_ORIGINS', [])),
    'allowed_origins_patterns' => [],
    'allowed_headers' => ['*'],
    'exposed_headers' => [],
    'max_age' => 0,
    'supports_credentials' => false,
];
```

spawnia:
@lorado can you review https://github.com/nuwave/lighthouse/pull/1267 ?

crazyabdul:
[...]

PR #1267, opened 2020-04-07
Document how to configure CORS

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/1266

### DIFF
--- a/docs/2.6/getting-started/configuration.md
+++ b/docs/2.6/getting-started/configuration.md
@@ -200,3 +200,21 @@ return [
     'batched_queries' => true,
 ];
 ```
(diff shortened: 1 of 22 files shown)


lorado:
Yes, that is correct. One question - maybe should we mention, that this solution works with Laravel 7 upwards? I think some people may still use on Laravel 6, and there they have to install that CORS package separately.

crazyabdul:
@lorado I agree with you. I think that will make things much clearer, especially for lighthouse beginners like myself

spawnia:
Thanks for the feedback!

PR #1273, opened 2020-04-12
Add more return type hints

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Always set the return type hint.

**Breaking changes**


src/Schema/Types/GraphQLSubscription.php
```diff
@@ -16,7 +16,7 @@ abstract class GraphQLSubscription
      * @param  \Nuwave\Lighthouse\Subscriptions\Subscriber  $subscriber
      * @return bool
      */
-    public function can(Subscriber $subscriber)
+    public function can(Subscriber $subscriber): bool
```
spawnia, line 19: Those are breaking changes, since the class is meant to be extended. Let's add a TODO and change that in the next major version.

src/Support/Contracts/TypeResolver.php
```diff
@@ -12,5 +13,5 @@ interface TypeResolver extends Directive
      * @param  \Nuwave\Lighthouse\Schema\Values\TypeValue  $value
      * @return \GraphQL\Type\Definition\Type
      */
-    public function resolveNode(TypeValue $value);
+    public function resolveNode(TypeValue $value): Type;
```
spawnia, line 16: Breaking change, let's add a TODO instead.

src/Support/Traits/HasArgumentPath.php
```diff
@@ -29,7 +29,7 @@ public function argumentPathAsDotNotation(): string
      * @param  mixed[]  $argumentPath
      * @return $this
      */
-    public function setArgumentPath(array $argumentPath): self
+    public function setArgumentPath(array $argumentPath)
```
spawnia, line 32: Why?

spawnia:
Hey @lucasmichot, good to hear from you. I totally agree with adding strict types, but let's hold off breaking interfaces and abstract classes until the next major version.

src/Support/Http/Middleware/AcceptJson.php
```diff
@@ -24,7 +23,7 @@ class AcceptJson
      * @param  \Closure  $next
      * @return \Illuminate\Http\JsonResponse
      */
-    public function handle(Request $request, Closure $next): JsonResponse
+    public function handle(Request $request, Closure $next)
```
spawnia: What else can that be?

PR #1278, opened 2020-04-12
Add missing method docblocks.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Add missing method docblocks.

**Breaking changes**

### DIFF
--- a/src/ClientDirectives/ClientDirective.php
+++ b/src/ClientDirectives/ClientDirective.php
@@ -26,6 +26,12 @@ class ClientDirective
      */
     private $definition;
 
+    /**
+     * ClientDirective constructor.
+     *
+     * @param  string  $name
+     * @return void
+     */
     public function __construct(string $name)
     {
         $this->name = $name;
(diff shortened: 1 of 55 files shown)


lucasmichot:
This test fails for some random reason: https://github.com/nuwave/lighthouse/pull/1278/checks?check_run_id=581346219

@spawnia, I feel there are too many tests run for a single PR.

It might make some sense to extract some of them to run daily as a corn, instead of at each new PR.

spawnia:
I am actually considering adding https://docs.styleci.io/fixers#no_superfluous_phpdoc_tags to remove redundant PHPDocs.

> I feel there are too many tests run for a single PR.
> It might make some sense to extract some of them to run daily as a corn, instead of at each new PR.

Which ones in particular?

The amount of test runs is going down soon as we drop support for older PHP/Laravel versions in a new major version of Lighthouse.

lucasmichot:
> I am actually considering adding https://docs.styleci.io/fixers#no_superfluous_phpdoc_tags to remove redundant PHPDocs. I don't think they add anything useful to the code, unless extra information such as refined type hints or a description is added.

Yes totally up for it 👍 

> The amount of test runs is going down soon as we drop support for older PHP/Laravel versions in a new major version of Lighthouse.

👌 looking forward

I guess the coverage specific  could be dropped for each PR, and run on master instead.

PR #1279, opened 2020-04-12
Optimize conditions.

lucasmichot:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Optimize conditions.

**Breaking changes**

### DIFF
--- a/src/Defer/Defer.php
+++ b/src/Defer/Defer.php
@@ -118,7 +118,7 @@ directive @defer(if: Boolean = true) on FIELD
             return $data;
         }
 
-        if ($this->isDeferred($path) || ! $this->acceptFurtherDeferring) {
+        if (! $this->acceptFurtherDeferring || $this->isDeferred($path)) {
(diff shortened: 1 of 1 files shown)


src/Defer/Defer.php
```diff
@@ -118,7 +118,7 @@ public function defer(Closure $resolver, string $path)
             return $data;
         }
 
-        if ($this->isDeferred($path) || ! $this->acceptFurtherDeferring) {
+        if (! $this->acceptFurtherDeferring || $this->isDeferred($path)) {
```
spawnia, line 121: How is that better?

lucasmichot: Yes, you only have to check for a value, then use a method.
In such a situation we should always run the most expensive code only if necessary.

spawnia: I see your point.

In this particular case, `defer()` is not expensive and is much more likely to be true, so it makes sense to put it first.

PR #1281, opened 2020-04-13
Add StyleCI setting no_superfluous_phpdoc_tags

spawnia:


spawnia:
@lucasmichot what do you think?

lucasmichot:
I am totally 👍 

Hopefully with some addition type-hints for params and for return types, the code will be way thinner.

If a new release were to require PHP7.4 (?), even class properties could benefit typehinting and their docblocks could therefore also be dropped.

spawnia:
Yeah, really looking forward to `7.4`. We are not going to make the jump quite yet, but are dropping `7.1` support.

PR #1283, opened 2020-04-13
Improve subscription context serializer

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The current way the subscription serializer works is to serialize the `user` from the context and unserialize it on the other end. However this generates a lot of data to be stored and also could cause the user model to be outdated when unserialized.

This changes the subscription serializer to make use of the way Laravel's own queue component serializes models to serialize the context user and deserializes it on the other end.

I started working on this because I had some instances where the unserialization of the user model never worked (`unserialize(): Error at offset 11758 of 25737 bytes`) and prevented the subscription from even fireing.

In short:

- Less data stored in the subscription storage
- Always up-to-date user model in the context

**Breaking changes**

None.


src/Subscriptions/Serializer.php
```diff
@@ -63,7 +66,16 @@ public function unserialize(string $context): GraphQLContext
 
         $request->setUserResolver(
             function () use ($rawUser) {
-                return unserialize($rawUser);
+                $user = $this->getRestoredPropertyValue($rawUser);
+
+                // This is here for backwards compatibility, before the Laravel
```
spawnia, line 71: To clarify, this is only needed so that current deployments can be upgraded seamlessly?

stayallive: Correct.

tests/Integration/Subscriptions/SerializerTest.php
```diff
@@ -0,0 +1,96 @@
...
+            $retrievedFromDatabase = true;
+        });
+
+        $unserialized = $serializer->unserialize(
+            $serializer->serialize($context)
+        );
+
+        $this->assertTrue($retrievedFromDatabase);
+        $this->assertInstanceOf(User::class, $unserialized->user());
+        $this->assertEquals($user->getKey(), $unserialized->user()->getKey());
```
spawnia, line 51:
```suggestion
        $this->assertSame($user->getKey(), $unserialized->user()->getKey());
```

src/Subscriptions/Serializer.php
spawnia: Cool, nicely done. Can you change the wording to reflect that this enables seamless upgrades to existing deployments and add a TODO to remove this in the next major version?

tests/Integration/Subscriptions/SerializerTest.php
```diff
@@ -0,0 +1,96 @@
...
+                'content' => '',
+            ],
+            'user' => serialize($context->user()),
+        ]);
+
+        $unserialized = $serializer->unserialize($serialized);
+
+        $this->assertFalse($retrievedFromDatabase);
+        $this->assertInstanceOf(User::class, $unserialized->user());
+        $this->assertEquals($user->getKey(), $unserialized->user()->getKey());
```
spawnia, line 94:
```suggestion
        $this->assertSame($user->getKey(), $unserialized->user()->getKey());
```

spawnia:
Great job, thank you :+1:

PR #1284, opened 2020-04-13
Add TTL option for subscriptions storage

stayallive:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds an optional TTL to the subscription storage manager that allows auto-cleanup of subscriptions.

If for some reason a webhook is missed it is nice to have a way to cleanup stale subscription to prevent broadcasting unneeded. Currently topics and subscribers are stored forever (and topics are never cleanup up, but that's for another day), this allows setting a TTL that will let the cache driver know when an item can be removed from storage.

**Breaking changes**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,6 @@ You can find and compare releases at the [GitHub release page](https://github.co
 ### Added
 
 - Add flag `--json` to `print-schema` to output JSON instead of GraphQL SDL https://github.com/nuwave/lighthouse/pull/1268
-
-### Changed
-
 - Add TTL option for subscriptions storage https://github.com/nuwave/lighthouse/pull/1284
 
 ## 4.11.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Add flag `--json` to `print-schema` to output JSON instead of GraphQL SDL https://github.com/nuwave/lighthouse/pull/1268
 
+### Changed
+
+- Add TTL option for subscriptions storage https://github.com/nuwave/lighthouse/pull/1284
+
 ## 4.11.0
 
 ### Added
--- a/src/Subscriptions/StorageManager.php
+++ b/src/Subscriptions/StorageManager.php
@@ -28,11 +28,17 @@ class StorageManager implements StoresSubscriptions
      */
     protected $cache;
 
+    /**
+     * @var \DateInterval|int|null
+     */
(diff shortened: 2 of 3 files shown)


stayallive:
Looks like this causes unintended side effect, investigating...

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Add flag `--json` to `print-schema` to output JSON instead of GraphQL SDL https://github.com/nuwave/lighthouse/pull/1268
 
+### Changed
```
spawnia, line 16:
```suggestion
### Added
```

src/Subscriptions/StorageManager.php
```diff
@@ -100,8 +106,13 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
 
         $topic[] = $subscriber->channel;
 
-        $this->cache->forever($topicKey, json_encode($topic));
-        $this->cache->forever($subscriberKey, $subscriber);
+        if ($this->ttl === null) {
```
spawnia, line 109: If `$this->ttl === null`, Laravel will store the item in the cache forever anyways. The explicit approach you chose is more obvious though, so i am fine with it.

src/lighthouse.php
```diff
@@ -253,6 +253,13 @@
          */
         'storage' => env('LIGHTHOUSE_SUBSCRIPTION_STORAGE', 'redis'),
 
+        /*
+         * Default subscription storage TTL.
```
spawnia, line 257:
```suggestion
         * Default subscription storage time to live.
```

src/lighthouse.php
```diff
@@ -253,6 +253,13 @@
...
+         *
+         * Indicates how long a subscription can be active before it's automatically removed from storage.
```
spawnia, line 259:
```suggestion
         * Indicates how long a subscription can be active before it's automatically removed from storage.
         * Setting this to `null` means the subscriptions are stored forever. This may cause
         * stale subscriptions to linger indefinitely in case cleanup fails for any reason.
```

src/Subscriptions/StorageManager.php
stayallive: This only works since 5.8, I had the third argument set to `null` but on 5.7 and below that won't work since the argument there is not nullable and would result in a ttl of `0` meaning it expires immediately (https://github.com/laravel/framework/blob/5.5/src/Illuminate/Cache/RedisStore.php#L90).

PR #1285, opened 2020-04-13
Fix nested mutations with multiple belongsTo relations at the same level

### LINKED ISSUE #1213: Only one nested belongsTo create fires when more than one is present
**Describe the bug**

When a type has multiple belongsTo relationships, and a mutation triggers a create on both of them, only the one that appears first in the schema is fired and the rest are ignored.

I've been able to trace it as far as [SaveModel](https://github.com/nuwave/lighthouse/blob/8a2317e496e92f888fe575b4ed8291bb04a17f36/src/Execution/Arguments/SaveModel.php#L47). The ArgPartitioner finds both of the belongsTo relations fine, but the foreach is only looping once?

Every belongsTo entity is created when requested

**Steps to reproduce**

Schema:
```
type Mutation {
    createProduct(input: CreateProductInput! @spread): Product @create
}

"TYPES"

type Product {
    id: ID!
    name: String!
    sku: String!

    manufacturer: ProductManufacturer! @belongsTo
    category: ProductCategory! @belongsTo
}

type ProductManufacturer {
    id: ID!
    name: String!
    products: [Product!]! @hasMany
}

type ProductCategory {
    id: ID!
    name: String!
    products: [Product!]! @hasMany
}

"INPUTS"

input CreateProductInput {
    name: String!
    sku: String!
    manufacturer: CreateProductManufacturerRelation!
    category: CreateProductCategoryRelation!
}

input CreateProductManufacturerRelation {
    connect: ID
    create: CreateProductManufacturerInput
}

input CreateProductCategoryRelation {
    connect: ID
    create: CreateProductCategoryInput
}

input CreateProductManufacturerInput {
    name: String!
}

input CreateProductCategoryInput {
    name: String!
}
```

Query:
```
mutation {
  createProduct(input: {
    name:"My amazing product"
    sku:"SKU0001"
    manufacturer: {
      create: {
        name: "Awesome Corp"
      }
    }
    category: {
      create: {
        name: "Doodads"
      }
    }
  }){
    id
  }
}
```

**Output/Logs**

```
"debugMessage": "SQLSTATE[HY000]: General error: 1364 Field 'category_id' doesn't have a default value (SQL: insert into `products` (`name`, `sku`, `manufacturer_id`) values (My amazing product, SKU0001, 21))"
```

</details></br>

**Environment**

Lighthouse Version: 4.9
Laravel Version: 6.4.1

vitorsemeano:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1213.

**Changes**

This resolves a bug when inserting a model that relates to multiple belongs to relations that also are being created (using create or upsert wihtout id).

When SaveModel detects a BelongsTo relation, he associates with the relation, and also persist the relation. Since belongsTo means the relation is related with a model that contains the foreign key,
the code will try to persist. What happens is that the model have multiple belongsTo relations, all required and without default value. When this happens, and the code tries to save the model, it's still missing the remaining belongsTo relations id, generating a query error. 

This PR adds a check so when the parent model, related by a BelongsTo relation, doesn't exist, it will not call the save method in that level. The save will then be called after the for each applied for the parent model level. 

**Breaking changes**

No breaking changes are introduced in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
-- `Fixed` Fix query error when using @create/@upsert directives with multiple belongsTo at the same level https://github.com/nuwave/lighthouse/pull/1285
+### Fixed
+
+- Fix nested mutations with multiple `belongsTo` relations at the same level https://github.com/nuwave/lighthouse/pull/1285
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+- `Fixed` Fix query error when using @create/@upsert directives with multiple belongsTo at the same level https://github.com/nuwave/lighthouse/pull/1285
+
 ### Added
 
 - Add flag `--json` to `print-schema` to output JSON instead of GraphQL SDL https://github.com/nuwave/lighthouse/pull/1268
@@ -27,7 +29,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Fixed
 
-- Eager load nested relations using the `@with` directive https://github.com/nuwave/lighthouse/pull/1068 
+- Eager load nested relations using the `@with` directive https://github.com/nuwave/lighthouse/pull/1068
 - Avoid infinite loop with empty namespace in generator commands https://github.com/nuwave/lighthouse/pull/1245
 - Automatically register `TestingServiceProvider` for `@mock` when running unit tests https://github.com/nuwave/lighthouse/pull/1244
 
@@ -235,7 +237,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Fixed
 
-- Fix broken behaviour when using union types with schema caching https://github.com/nuwave/lighthouse/pull/1015 
+- Fix broken behaviour when using union types with schema caching https://github.com/nuwave/lighthouse/pull/1015
 
 ## 4.4.2
 
@@ -380,7 +382,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Change the signature of the AST manipulating directive interfaces:
   `TypeManipulator`, `FieldManipulator` and `ArgManipulator` https://github.com/nuwave/lighthouse/pull/768
 - Change the API of the `DocumentAST` class to enable a more performant implementation https://github.com/nuwave/lighthouse/pull/768
-- Enable the schema caching option `lighthouse.cache.enable` by default https://github.com/nuwave/lighthouse/pull/768 
+- Enable the schema caching option `lighthouse.cache.enable` by default https://github.com/nuwave/lighthouse/pull/768
 - Lazily load types from the schema. Directives defined on parts of the schema that are not used within the current
   query are no longer run on every request https://github.com/nuwave/lighthouse/pull/768
 - Simplify the default route configuration.
@@ -429,7 +431,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Deprecated
 
-- The `GraphQL` facade and the container alias `graphql` will be removed in v4 
+- The `GraphQL` facade and the container alias `graphql` will be removed in v4
 
 ## 3.6.1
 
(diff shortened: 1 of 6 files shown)


src/Execution/Arguments/SaveModel.php
```diff
@@ -66,9 +66,15 @@ public function __invoke($model, $args)
...
-                ->associate($model)
-                ->save();
+            $parentModel = $this->parentRelation
+                                ->associate($model);
+
+            // If the parent Model does not exist (still to be saved),
+            // a save could break any pending belongsTo relations that still
+            // needs to be created and associated with the parent model
+            if ($parentModel->exists()) {
+                $parentModel->save();
```
spawnia, line 76: That would trigger multiple saves, right? Is there a simple way through which we can save just once?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+- `Fixed` Fix query error when using @create/@upsert directives with multiple belongsTo at the same level https://github.com/nuwave/lighthouse/pull/1285
```
```suggestion
### Fixed

- Fix nested mutations with multiple `belongsTo` relations at the same level https://github.com/nuwave/lighthouse/pull/1285
```

tests/Integration/Execution/MutationExecutor/BelongsToTest.php
```diff
@@ -906,4 +906,98 @@ public function testUpsertAcrossTwoNestedBelongsToRelationsAndOverrideExistingMo
 
         $this->assertEquals([2], $role->users()->pluck('users.id')->toArray());
     }
+
+    public function testCreateMultipleBelongsTothatDontExist()
+    {
+        $this->schema = /** @lang GraphQL */ '
+
```
spawnia, line 913:
```suggestion
```

src/Execution/Arguments/SaveModel.php
vitorsemeano: Hi, this actually reduces the number of saves being done before this change. Before it will always save the parentModel. I tried to remove the save method, but that breaks a test case for deep nested mutations when using a sync. The test that broke was testUpsertAcrossTwoNestedBelongsToRelationsAndOverrideExistingModel on line 907. I found that test a little strange, since it changes the id of the related user to the role, using a hasMany, when you are triggering that change in a mutation to the user.

vitorsemeano: I looked into that testcase, and what it does is run a deep nested sync after a many to many relation being used to create a role. If i remove the save, in that case the sync would be overwritten by the first many to many relation:

User -> hasMany -> role_user -> belongsTo -> role -> belongsToMany -> users
S1                                  S4                                     S2                                       S3

Where Sn is the order of saves that the actual code does.

I am not sure the usefulness of this case, where you want to create a model related to an existing one, and then change that relation. Could be simpler to just create the Role and then associated the Users?

vitorsemeano: The save is only needed to guarantee that a operation like that persists before applying changes to the same relation in deeper levels of the mutation. Once again, i am not sure the usefulness of this specific case. If such redeems unnecessary, the save could be removed. The problem is that nothing tell us that this specific case is not used by other users.

spawnia: That test case is indeed a bit far fetched, but i think it was added by someone who had that exact use case. The other test that fails when removing `save()` here is one where `@create` is used as a nested arg resolver.

The check you implemented makes the most use cases possible and work smoothly, so i am fine with adding it.

spawnia:
Thanks @vitorsemeano, nicely done.

alancwoo:
Sorry to jump into this issue but I noticed this error message when trying to execute a mutation: `Class Illuminate\Database\Eloquent\Relations\belongsToMany does not exist, did you forget to import the Eloquent relation class?`

Oddly, it works locally and not on a remotely deployed app, I'm assuming due to [the recent push](https://github.com/nuwave/lighthouse/pull/1338/commits/ee9694c7b1444b967984cfbc42aeba2d2e37f242)

All my relevant models implement `use Illuminate\Database\Eloquent\Model` or `use Illuminate\Database\Eloquent\Relaions\Pivot` accordingly

spawnia:
@alancwoo i assume you are working on Windows?

alancwoo:
@spawnia (thankfully) I'm not actually haha. I'm working locally on a mac, deploying to an Ubuntu server.

spawnia:
Well, it still looks like you have an issue with the casing (`belongsToMany`).

PR #1287, opened 2020-04-13
Allow setting test database connection data from env vars

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Read test database connection data from env variables. This makes it easier to run tests locally without docker.

**Breaking changes**

None.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -10,6 +10,8 @@ on:
 
 env:
   REQUIRED_PHP_EXTENSIONS: "mbstring, mysqli, pdo_mysql"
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -10,6 +10,8 @@ on:
 
 env:
   REQUIRED_PHP_EXTENSIONS: "mbstring, mysqli, pdo_mysql"
+  LIGHTHOUSE_TEST_DB_HOST: "127.0.0.1"
+  LIGHTHOUSE_TEST_DB_PASSWORD: "root"
```
stayallive, line 14: Even better, should've thought about that 👌

PR #1286, opened 2020-04-13
Type hint StoresSubscription instead of implementation

stayallive:
- [x] Added or updated tests
- [x] Added Docs for all relevant versions
- [x] Updated CHANGELOG.md

**Changes**

Instead of concrete implementation typehint the contract allowing us to replace it.

**Breaking changes**

None, unless `SubscriptionRegistry->__construct` was extended by the user.

PR #1288, opened 2020-04-13
Remove subscriber reference from topic when deleted

tlaverdure:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

When subscribers are deleted their subscription key is not removed from the topic they are subscribed to. This should reduce the amount of calls to `$this->cache->get()` to check if the subscriber exists before broadcasting.

**Breaking changes**

N/A


stayallive:
I have been thinking on how to solve this problem to and #1284 solved it a tiny bit, but not really... so when building a Redis optimized version I came up with wrapping the subscriber so we can store it with the topic it subscribed to.

https://gist.github.com/stayallive/9b9fe8556faf9c8ba6f09ca54610e40f#file-topicsubscriber-php-L7-L28

This way on deleting the subscriber we can remove the channel from the topic (in my case using `srem` to remove it from the Redis set).

https://gist.github.com/stayallive/9b9fe8556faf9c8ba6f09ca54610e40f#file-redissubscriptionstorage-php-L104-L111

This requires wrapping the topic with the subscriber when setting the subscriber info.

https://gist.github.com/stayallive/9b9fe8556faf9c8ba6f09ca54610e40f#file-redissubscriptionstorage-php-L90-L94

This might be a good direction to go in to prevent [recreating the topic key](https://github.com/nuwave/lighthouse/pull/1288/files#diff-dfc9592f8e72c1b1aecdd1409da1944cR147) because that just won't work since a subscription can have a custom topic encoder/decoder.

_Note: Everything in that gist is experimental and cannot be used until #1286 & #1289 are merged and released otherwise you cannot swap out the subscriber storage class._

src/Subscriptions/StorageManager.php
```diff
@@ -118,19 +118,42 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+    
+    /**
+     * Remove the subscriber from the topic they are subscribed to.
+     *
+     * @param  Subscriber $subscriber
+     * @param  string $subscriberKey
+     */
+    public function removeSubscriberFromTopic(Subscriber $subscriber, string $subscriberKey)
+    {
+        $key = strtoupper(Str::snake($subscriber->operationName));
```
stayallive, line 146: This won't work because a subscription can have a custom topic encoding/decoding scheme.

https://github.com/nuwave/lighthouse/blob/6552b798ab99790ca3344dad1b46046f9880b194/src/Schema/Types/GraphQLSubscription.php#L23-L45

src/Subscriptions/StorageManager.php
```diff
@@ -118,19 +118,42 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+
+        $topic = Collection::make(
+            $this->cache->has($topicKey)
+                ? json_decode($this->cache->get($topicKey), true)
+                : []
+        )->filter(function ($key) use ($subscriberKey) {
+            return $key === $subscriberKey;
+        });
+
+        $this->cache->forever($topicKey, json_encode($topic->all()));
```
stayallive, line 157: This is incorrect to with the recent TTL changes (#1284).

https://github.com/nuwave/lighthouse/blob/6552b798ab99790ca3344dad1b46046f9880b194/src/Subscriptions/StorageManager.php#L109-L115

src/Subscriptions/StorageManager.php
```diff
@@ -118,19 +118,42 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+        $key = strtoupper(Str::snake($subscriber->operationName));
+        $topicKey = self::TOPIC_KEY.".{$key}";
+
+        $topic = Collection::make(
+            $this->cache->has($topicKey)
+                ? json_decode($this->cache->get($topicKey), true)
+                : []
+        )->filter(function ($key) use ($subscriberKey) {
+            return $key === $subscriberKey;
+        });
```
stayallive, line 155:
```suggestion
        )->reject(function ($key) use ($subscriberKey) {
            return $key === $subscriberKey;
        });
```

Otherwise the only thing that remains is the one you wanted to remove :)

src/Subscriptions/StorageManager.php
tlaverdure: Thanks will update.

src/Subscriptions/StorageManager.php
tlaverdure: 👍🏾

src/Subscriptions/StorageManager.php
tlaverdure: I thought this would be the case. Can dig into this more later.

tlaverdure: Storing the topic the subscriber before storing it now. Think that should do it?

spawnia:
Can you merge the latest changes from `master`? I changed a few calls related to `Cache::has()`.

src/Subscriptions/StorageManager.php
```diff
@@ -110,8 +111,45 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+     *
+     * @param  string  $channel
+     * @return \Nuwave\Lighthouse\Subscriptions\Subscriber|null
+     */
     public function deleteSubscriber(string $channel): ?Subscriber
     {
-        return $this->cache->pull(self::SUBSCRIBER_KEY.".{$channel}");
+        $key = self::SUBSCRIBER_KEY.".{$channel}";
+
+        if ($this->cache->has($key)) {
```
spawnia, line 124: Do not use `has()`, see https://github.com/nuwave/lighthouse/pull/1290

tlaverdure: Got it.

src/Subscriptions/StorageManager.php
```diff
@@ -110,8 +111,42 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+     * @param  string  $channel
+     * @return \Nuwave\Lighthouse\Subscriptions\Subscriber|null
+     */
     public function deleteSubscriber(string $channel): ?Subscriber
     {
-        return $this->cache->pull(self::SUBSCRIBER_KEY.".{$channel}");
+        $key = self::SUBSCRIBER_KEY.".{$channel}";
+        $this->removeSubscriberFromTopic($key);
+
+        return $this->cache->pull($key);
```
spawnia, line 125: This still incurs an extra cache lookup, `->get()` is already called within `removeSubscriberFromTopic()`. We should keep IO to a minimum.

stayallive, line 125:
```suggestion
        $subscriber = $this->cache->pull(self::SUBSCRIBER_KEY.".{$channel}");
        
        if ($subscriber !== null) {
	        $this->removeSubscriberFromTopic($subscriber);
        }
        
        return $subscriber;
```

This will also require changes to `removeSubscriberFromTopic` to accept a `Subscriber`.

src/Subscriptions/StorageManager.php
```diff
@@ -110,8 +111,42 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+     *
+     * @param  string $subscriberKey
+     */
+    protected function removeSubscriberFromTopic(string $subscriberKey)
+    {
+        $subscriber = $this->cache->get($subscriberKey);
+
+        if (!$subscriber || !$subscriber->topic) {
+            return;
+        }
```
stayallive, line 139:
```suggestion
    protected function removeSubscriberFromTopic(Subscriber $subscriber)
    {
        if (empty($subscriber->topic)) {
            return;
        }
```

Considering my changes in the `deleteSubscriber` method.

src/Subscriptions/StorageManager.php
```diff
@@ -110,8 +111,42 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+
+        $topic = Collection::make(json_decode($this->cache->get($subscriber->topic), true))
+            ->reject(function ($key) use ($subscriberKey) {
+                return self::SUBSCRIBER_KEY.".{$key}" === $subscriberKey;
+            });
```
stayallive, line 144:
```suggestion
            ->reject(function ($key) use ($subscriber) {
                return self::SUBSCRIBER_KEY.".{$key}" === self::SUBSCRIBER_KEY.".$subscriber->channel;
            });
```

tlaverdure:
Thanks @stayallive, your suggestions were coming in realtime as I was on the same lines, haha.

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Add flag `--json` to `print-schema` to output JSON instead of GraphQL SDL https://github.com/nuwave/lighthouse/pull/1268
 - Add TTL option for subscriptions storage https://github.com/nuwave/lighthouse/pull/1284
-
+- `StorageManager` will remove subscribers from topics now when they are deleted.
 ### Changed
```
stayallive, line 18:
```suggestion

### Changed

- When a subscription subscriber is deleted the subscriber reference will also be removed from the topic https://github.com/nuwave/lighthouse/pull/1288
```

It should be under `Changed`, and what about the wording, that might better describe what is happening?

src/Subscriptions/StorageManager.php
```diff
@@ -110,8 +111,43 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
+    {
+        if (empty($subscriber->topic)) {
+            return;
+        }
+
+        $topic = Collection::make(json_decode($this->cache->get($subscriber->topic), true))
+            ->reject(function ($key) use ($subscriber) {
+                return self::SUBSCRIBER_KEY.".{$key}" === self::SUBSCRIBER_KEY.".{$subscriber->channel}";
+            });
+
```
stayallive, line 146:
```suggestion

        if ($topic->isEmpty()) {
            $this->cache->forget($subscriber->topic);
            
            return;
        }

```

To prevent storing empty topic, we might want to `forget` empty topics.

stayallive:
@tlaverdure perfect timing, thanks for keeping up with them and integrating them in your PR, this is starting to look real good 😎

src/Subscriptions/StorageManager.php
tlaverdure: Good call.

spawnia:
The PR would not have worked as provided, serialization for `$subscriber->topic` was missing. I reworked the whole thing quite a bit and ensure we got it covered through tests. Sorry for blowing up this PR 🤯 

@tlaverdure @stayallive If you could review the changes, that would be great. Great work so far.

tlaverdure:
All good! Thanks for the additions, it makes things a lot clearer. Looks good to me.

src/Subscriptions/Broadcasters/PusherBroadcaster.php
```diff
@@ -62,13 +61,11 @@ public function unauthorized(Request $request): JsonResponse
     public function hook(Request $request): JsonResponse
     {
         (new Collection($request->input('events', [])))
-            ->filter(function ($event): bool {
-                return Arr::get($event, 'name') === 'channel_vacated';
+            ->filter(function (array $event): bool {
+                return $event['name'] === 'channel_vacated';
```
stayallive, line 65:
```suggestion
                return isset($event['name']) && $event['name'] === 'channel_vacated';
```

I would still add a guard just in case Pusher (or someone testing) sends an invalid payload "breaking" the code (warnings/notices are converted to fatal errors in Laaravel).

spawnia: In case of an invalid payload, i would rather crash then to silently fail.

PR #1290, opened 2020-04-13
Avoid race condition that may occur when using Cache::has()

spawnia:
As documented in `CacheInterface`:

> NOTE: It is recommended that has() is only to be used for cache warming type purposes
and not to be used within your live applications operations for get/set, as this method
is subject to a race condition where your has() will return true and immediately after,
another script can remove it making the state of your app out of date.

### DIFF
--- a/src/Subscriptions/StorageManager.php
+++ b/src/Subscriptions/StorageManager.php
@@ -112,14 +112,6 @@ class StorageManager implements StoresSubscriptions
 
     public function deleteSubscriber(string $channel): ?Subscriber
     {
-        $key = self::SUBSCRIBER_KEY.".{$channel}";
-
-        $subscriber = $this->cache->get($key);
-
-        if ($subscriber) {
-            $this->cache->forget($key);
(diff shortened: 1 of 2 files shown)


src/Subscriptions/StorageManager.php
```diff
@@ -115,19 +110,13 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
...
     {
         $key = self::SUBSCRIBER_KEY.".{$channel}";
-        $hasSubscriber = $this->cache->has($key);
 
         $subscriber = $this->cache->get($key);
 
-        if ($hasSubscriber) {
+        if ($subscriber) {
             $this->cache->forget($key);
         }
```
stayallive, line 121: You might want to consider using `return $this->cache->pull(self::SUBSCRIBER_KEY.".{$channel}")` for a one-liner implementaton.

spawnia:
Thanks for the review @stayallive

Issue #1293, opened 2020-04-14
@can directive cannot be used to hide attributes


cecilymiller:
`@can` directive cannot be used to hide attributes

Usign `@can` directive on non-relationship attributes throws an error.

```javascript
type User {
    secret_attr: Int @can('view-secret-attr')
}
```

if `@can` directive is not designed to hide attributes, how to do so then?

**Output/Logs**

`Type 'Int' on 'secret_attr' can not be found in the schema.`

**Environment**

Lighthouse Version: ^4.5
Laravel Version: 6.18.3

spawnia:
> @can directive cannot be used to hide attributes

Correct, it enables access control.

> `Type 'Int' on 'secret_attr' can not be found in the schema.`

Does that error occur without `@can`, too?

cecilymiller:
> Does that error occur without @can, too?

No

> Correct, it enables access control.

So, how we can hide attributes for specific users? Isn't access control pretty much the same thing?

spawnia:
> So, how we can hide attributes for specific users?

How does Laravel do it? `@can` works with policies, on the Model level.

cecilymiller:
Ok, I was wondering is there a `lighthouse` way to do it.

Do you know a better way to do it than writing `getSecretAttribute` getters one by on on each model?

spawnia:
I would look for a method that Laravel already uses and thus can be reused for different kinds of Model access. If there is nothing else, the getters you describe seem fine.

A custom field middleware directive is another option.

cecilymiller:
Thanks for your suggestions!

> A custom field middleware directive is another option.

That's something I was looking for. Perfect

lorado:
@spawnia I think this is a common problem. I currently also have to implement protection of some fields, depending on user role. As user roles may be implemented in tons of ways, the logic should be defined on the app-side. 

Should I write a guide (with example), how people can achieve filed/attribute protection (without throwing an error, and instead simply return null, e.g.)?

spawnia:
@lorado That seems like a great idea, maybe an additional section to `Security -> Authorization`.

spawnia:
> @spawnia Why exactly can't `@can` be use on fields on a type?
> From a GQL perspective, this seems odd as access control on query fields is no different than for type fields…

It is just not implemented. What is GQL?

trm217:
[...]

PR #1297, opened 2020-04-15
Add HowTo to docs: define custom field middleware to protect attributes

lorado:
- [ ] ~~Added or updated tests~~
- [ ] ~~Updated CHANGELOG.md~~

Gives an example on how to solve #1293

**Changes**

Added an additional section to `Security`->`Authorization`, that describes with an example, how to protect particular fields (in most cases model attributes) and return `null`, if user has no access to the field.

**Breaking changes**

None

### DIFF
--- a/docs/master/security/authorization.md
+++ b/docs/master/security/authorization.md
@@ -164,15 +164,20 @@ class PostPolicy
 }
 ```
 
-## Custom fields/attributes restriction
+## Custom field restrictions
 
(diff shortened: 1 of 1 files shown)


lorado:
@spawnia Could you check the example and correct the text? I hope it will help people =)

spawnia:
Nice example @lorado, i reworked it a bit. Feel free to merge.

PR #1298, opened 2020-04-15
Replace deprecated `assertContains` by `assertStringContainsString`

lorado:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
Replaced deprecated `assertContains` by `assertStringContainsString`.

There is also `assertArraySubset`, that was also deprecated, but as long we have php 7.1 in the test matrix, we can't replace it. I added a TODO note above the line with a solution, so we don't forget it. Here is a discussion about this function https://github.com/nuwave/lighthouse/pull/950#issuecomment-528354592 

**Breaking changes**
None


lorado:
puhh ok, it doesn't work with PHPUnit 6. Ok then doesn't matter atm

spawnia:
This is on the agenda for the next major release for sure. Those warnings are really annoying 😃

PR #1301, opened 2020-04-18
Allow the subscription broadcast job queue to be configured

### LINKED ISSUE #1300: Configuring the subscription broadcast name
So right now it's "impossible" (nothing is of course) to configure the queue on which the subscription is broadcasted (if you choose to use the queue).

This is because an event/listener is used, because Laravel handles reading the `$queue` property very poorly and you cannot set it in the constructor or implement a getter method ([bit more context](https://laracasts.com/discuss/channels/laravel/set-queue-name-dynamically-for-event-listener?page=1#reply=429197)).

There are a few options I see and I wanted to discuss it before creating a PR. Hope to discuss the solutions below and/or read better options.

---

#### 1. Remove the use of a event/listener

And use a proper queued job on which we can run `onQueue` and use other methods to set the queue name read from the config file.

#### 2. Read the event listener class from the container

Use an interface to read the event listener implementation from the container so the user can implement it's own (extend the current one and set the `$queue` property).

#### 3. Use a "hack"

Defining this function on the event listener would work:

```php
    public function __get($name)
    {
        if ($name === 'queue') {
            return env('QUEUE_NAME_AUDIO');
        }

        return $this->{$name};
    }
```

---

I'm partial to 1 because it allows us to just add a config option to set the queue name and use an actual queue job instead of queued event allowing us to set tags for Horizon too (see #1301).

But 2 sounds the most simple from a code perspective seeing as everything uses an interface already.

However 1 is probably a breaking change (removing the event) although we could trigger the new job from the event listener and remove the `ShouldQueue` from the event listener to solve that "issue" and add it to the 4.12 release.

3 is possible but seems like a terrible plan 😄

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1300.

**Changes**

Uses a "proper" queue job so we can define a custom queue from the configuration file to queue all broadcast jobs on.

**Breaking changes**

None.


spawnia:
Looks reasonable. Is there any reason to keep the event around apart from preserving backwards compatibility? Can you imagine any, possibly obscure, use cases for it?

stayallive:
No I thought about it but the only reason I didn’t remove it was BC to keep the in 4.x. I think the event was a bad choice in this case since I can’t think of any reason a user might want to listen to this even except for maybe analytical purposes, but that doesn’t really sound like a valid reason either to be honest.

src/Subscriptions/Job/BroadcastSubscriptionJob.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Nuwave\Lighthouse\Subscriptions\Job;
+
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Nuwave\Lighthouse\Schema\Types\GraphQLSubscription;
+use Nuwave\Lighthouse\Subscriptions\Contracts\BroadcastsSubscriptions;
+
+class BroadcastSubscriptionJob implements ShouldQueue
+{
+    use Queueable;
```
spawnia, line 12: I decoupled this from the event class in anticipation of its removal in v5. I guess we need to add `SerializesModels` in here?

spawnia:
@stayallive please add an entry both in `CHANGELOG.md` as well as a section in `UPGRADE.md`.

src/Subscriptions/Job/BroadcastSubscriptionJob.php
stayallive: Yes, because of the `$root` argument probably being a model. I've added that trait.

stayallive:
@spawnia I've added the changelog and upgrade docs but I'm not sure about the `UPGRADE.md` part, maybe you can flesh that out a bit more if needed.

PR #1303, opened 2020-04-18
Report non client-safe errors through Laravel's `ExceptionHandler` #1303

spawnia:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/1026

**Changes**

Propagate the reporting of error that occur within GraphQL execution to Laravel's default `ExceptionHandler`.

Thanks to @olivernybroe for this idea in https://github.com/nuwave/lighthouse/issues/1026#issuecomment-554254333

**Breaking changes**

No


src/Execution/ReportErrorHandler.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static function handle(Error $error, Closure $next): array
+    {
+        $previous = $error->getPrevious();
+
+        // If the Error does not wrap another Error, it is related to a client error
+        // and shown in the error response anyway, we don't really need to report it
+        if ($previous) {
+            /** @var \Illuminate\Contracts\Debug\ExceptionHandler $reporter */
+            $reporter = app(ExceptionHandler::class);
```
stayallive, line 22: This could be moved to dependency injection through the `__constructor` right? 

The should work because the pipeline should resolve the error handlers from the container.

src/Execution/ReportErrorHandler.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Nuwave\Lighthouse\Execution;
+
+use Closure;
+use GraphQL\Error\Error;
+use Illuminate\Contracts\Debug\ExceptionHandler;
+
+/**
+ * Report errors through the default exception handler configured in Laravel.
+ */
+class ReportErrorHandler implements ErrorHandler
```
stayallive, line 12:
```suggestion
class ReportingErrorHandler implements ErrorHandler
```

Could possibly be a more descriptive name for it.

src/Execution/ReportErrorHandler.php
spawnia: Not right now, as the interface for the `ErrorHandler` requires the method to be `static`. I plan to change that in v5.

spawnia:
I am not sure which errors should be reported by default and how we can differentiate between them.

It does seem unnecessary to report syntax errors in client queries and validation errors. On the other hand, it might be useful to report errors where someone might try to attack the server, such as authorization errors or when somebody triggers the complexity limitation repeatedly.

src/Execution/ReportErrorHandler.php
stayallive: Woops, missed the static keyword 🤦‍♂️, good that is planned on chaning in v5, might cleanup the handlers a bit 👍

stayallive:
Could a possible option be to only report errors that are not client safe? If they're client safe we can assume the client should be able to handle them, however that would mean that complexity errors for example are not reported since they're considered client safe. But there might not be a good criteria on what to do or not report, so going for client safe could be a safe one and if you need more you can add your own error handler reporting those (or replace this reporting handler altogether with one with custom logic).

PR #1306, opened 2020-04-19
Log in subscribers when broadcasting a subscription update

### LINKED ISSUE #1302: User authentication context not restored in subscription broadcast
When broadcasting a subscription the GraphQL request is not performed with the same user authentication context, so if you use `auth()->user()` anywhere in your code this will fail in a subscription broadcast (or use the incorrect user if you don't use a queue).

We can argue it is bad to depend on the `auth()->user()` in code flowing through Lighthouse because it has it's own context with the authenticated user (which is properly set/restored on a subscription broadcast), and you might be right, but that's not the point of this 😄 Some getters in models will make use of the authenticated user (think `getCanDeleteAttribute()` for example) or do some stuff based on the authenticated status/role/config. 

So depending on where you stand on the above, this is a non-issue or a big issue.

The biggest problem is actually when you are not queueing your broadcasts because then the subscriptions are executed as the currently authenticated user and might allow you to get broadcasted information retrieved for the wrong user. If you run a actual queue worker this is probably not an issue since your code will just crash / handle it on `auth()->user()` being null.

After some research I concluded there might only be one option to correctly "restore" the user context with 1 caveat.

Since the Laravel Guard has no real way of setting and clearing the current user (without sending all kinds of login/logout events) we would have to register a "fake" guard that allows us to do so and we have to store the current authenticated user ID and the used user repository so we can register the faked guard and set that as the default guard to use (and restore the previous guard after we're done).

The only cavaet is if the user would use `auth('api')->user()` in their code, because we cannot replace a named guard without some real hocus-pocus I'm not comfortable implementing. But this might be an acceptable trade-off seeing how the current behaviour is broken and/or even dangerous (when not using queued subscription broadcasts).

I would like to hear your thoughts on this issue and/or if a fix should be explored or this is considered a feature instead of a bug and `auth()->user()` should never be used in a GraphQL request (only `$context->user()`).

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Fixes #1302.

**Changes**

Add a new interator used to iterator over subscription subscribers to restore the guard context when executing the GraphQL queries to broadcast to the subscribers.

**Breaking changes**

Might be considered a bugfix instead of a breaking change because `auth()->user()` is now set correctly to the subscriber user context when resolving subscription broadcasts.


stayallive:
@spawnia this is my current plan for dealing with #1302, I'm missing tests and docs but I wanted to show this to see if this is a direction we can take or you'd like me to structure it differently.

src/Subscriptions/Contracts/SubscriptionIterator.php
```diff
@@ -8,9 +8,18 @@
...
+     * @param  \Closure  $handleSubscriber
+     * Receives each subscriber in the passed in collection.
+     * function(\Nuwave\Lighthouse\Subscriptions\Subscriber $subscriber): void
+     *
+     * @param  \Closure|null  $onError
+     * Is called when $handleSubscriber throws.
      * @return void
      */
-    public function process(Collection $items, Closure $cb, Closure $error = null);
+    public function process(Collection $subscribers, Closure $handleSubscriber, Closure $onError = null);
```
spawnia, line 24: Feel free to improve the naming and description of this interface and the parameters.

spawnia:
@stayallive that approach looks solid. I especially like that it is pluggable, so there is little risk in experimenting a little. I am totally fine with shipping it once we add tests, docs and some polish.

stayallive:
Do you want to add a configuration option to enable this behaviour so we can ship it without enabling it at all and if there are no issues we can enable it by default on v5 and remove the config options for it?

spawnia:
> Do you want to add a configuration option to enable this behaviour so we can ship it without enabling it at all and if there are no issues we can enable it by default on v5 and remove the config options for it?

I don't see a need for that, since this should work out of the box. If we need to add a bugfix, we can add that. In case someone needs to revert to the old behaviour, they can rebind the interface.

tests/Unit/Subscriptions/Iterators/GuardContextSyncIteratorTest.php
```diff
@@ -0,0 +1,173 @@
...
+        return new Collection([
+            $this->generateSubscriber(),
+            $this->generateSubscriber(),
+            $this->generateSubscriber(),
+        ]);
+    }
+
+    private function generateSubscriber(): Subscriber
+    {
+        $resolveInfo = $this->getMockBuilder(ResolveInfo::class)
```
spawnia, line 123:
```suggestion
        $resolveInfo = $this->createMock(ResolveInfo::class)
```

...and delete the following lines.

tests/Unit/Subscriptions/Iterators/GuardContextSyncIteratorTest.php
```diff
@@ -0,0 +1,173 @@
...
+            $this->subscribers(),
+            static function (): void {
+                throw new Exception(self::EXCEPTION_MESSAGE);
+            },
+            static function (Exception $e) use (&$exception): void {
+                $exception = $e;
+            }
+        );
+
+        $this->assertSame(self::EXCEPTION_MESSAGE, $exception->getMessage());
```
spawnia, line 66: Nit: This does not validate the exact same exception is passed. You can store it in a variable before throwing and compare object equality.

tests/Unit/Subscriptions/Iterators/SyncIteratorTest.php
```diff
@@ -47,10 +47,10 @@ public function testCanPassExceptionToHandler(): void
 
         $this->iterator->process(
```
spawnia, line 48: That test seems to cover a property that iterators should generally fulfill.

stayallive: Not exactly sure what you mean with this, this is an pre-existing test, what would you change here?

src/Subscriptions/Iterators/GuardContextSyncIterator.php
```diff
@@ -0,0 +1,66 @@
...
+use Closure;
+use Exception;
+use Illuminate\Contracts\Auth\Factory;
+use Illuminate\Contracts\Config\Repository;
+use Illuminate\Support\Collection;
+use Nuwave\Lighthouse\Subscriptions\Contracts\SubscriptionIterator;
+use Nuwave\Lighthouse\Subscriptions\Subscriber;
+use Nuwave\Lighthouse\Subscriptions\SubscriptionGuard;
+
+class GuardContextSyncIterator implements SubscriptionIterator
```
spawnia, line 14: I would like a more descriptive name for that. How about `LoginSubscriberIterator`?

stayallive: I had a hard timing thinking of a good name. How about `AuthenticatedSubscriberIterator`?

spawnia: `AuthenticatingSubscriberIterator`?

spawnia: I will change it.

stayallive: Sounds good to me!

tests/Unit/Subscriptions/Iterators/SyncIteratorTest.php
spawnia: I abstracted the common tests that check if iterators are well behaved.

PR #1307, opened 2020-04-19
Fix subscriptions storage manager TTL

stayallive:
- [ ] ~~Added or updated tests~~
- [ ] ~~Updated CHANGELOG.md~~ (not needed since changes are already described in #1284)

**Changes**

This fixes the inconsistent cache TTL for Laravel `5.5-5.7` / `5.8+` in the subscription storage manager introduced in #1284. 

We already use this code to do this in other cache related code in Lighthouse.

Also added the dependency on Carbon since we use that in this package directly, got the version constraints from the 5.5 and 7.x branch of `laravel/framework`.

**Breaking changes**

None. Changes are not released yet.


spawnia:
How are they inconsistent?

composer.json
```diff
@@ -25,6 +25,7 @@
     "require": {
         "php": ">= 7.1",
         "ext-json": "*",
+        "nesbot/carbon": "^1.26.0 || ^2.17",
```
spawnia, line 28: Although Carbon should be a dependency for most if not all Laravel projects, i would like to avoid adding it as a dependency. Can we do without it?

stayallive: We can ofcourse leave it out, but since we use Carbon (Tracing, @cache directive & Date(Time) scalar) I though it might be a good practice to do mention it as a direct dependency.

But we never had it before so I'm fine with leaving it out since we can be almost 100% sure Carbon is available.

stayallive:
In Laravel `5.5-5.7` the TTL argument was in minutes (or Carbon).

But in `5.8` that changed to seconds (or Carbon).

So this way it is always in seconds (we document it is in seconds in the config file).

> https://laravel-news.com/cache-ttl-change-coming-to-laravel-5-8

composer.json
spawnia: Right, we are even using it in core parts of Lighthouse. We should loosen the constraints a bit, though, to match what the corresponding Laravel versions minimally need.

```suggestion
        "nesbot/carbon": "^1.26.0 || ^2.0",
```

stayallive: Seems reasonable, we don't have a real constraint on the version 👍

Issue #1190, opened 2020-02-14
Laravel Airlock support


ipalaus:
**Describe the bug**

I've been trying out Laravel Airlock instead of using Passport for its simplicity. But I'm either not doing correctly, or there is something I'm missing out as I can't get it to authenticate the user in the Lighthouse context. All the `@can` directives will fail too.

I installed Airlock as per their README, and then I switched all my protected endpoints to use `@guard(with: ["airlock"])`:

```graphql
extend type Query {
    getConversationMessages(conversationId: ID!): [ConversationMessage!]!
        @can(ability: "view", find: "conversationId", model: "App\\Conversation")
        @guard(with: ["airlock"])
}
```

The first thing I noticed is that I couldn't use `$context->user()` in my queries/mutations. Just using `Auth::user()` did the trick. I switched my tests to use `Airlock::actingAs(factory(User::class)->create());` for authentication, and all my tests passed. I assumed everything was working as expected, to my surprise one of our developers told me the endpoints that were protected with `@can` were all failing.

Digging into the code I found that the problem was, in the end, the first one that I noticed: the context doesn't have the user. The line that is making `@can` not working is the one where we call for `$context->user()` which is null, so the queries will never be authorized.
https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/CanDirective.php#L94

**Expected behavior/Solution**

I would expect that using `@guard(with: ["airlock"])` it would inject the user to the context.

**Steps to reproduce**

1. Install Laravel Airlock
2. Add User and get a login to authenticate the request https://github.com/laravel/airlock#authenticating-mobile-applications
3. Protect a query using a Policy with a `@can` directive
4. Request the query using the token from point 2

**Environment**

Lighthouse Version: 6.13.1
Laravel Version: 4.8

ipalaus:
This might be related to #1191

sicsol:
@ipalaus i'm using airlock however i'm using it with the `@middleware` directive. Try switching it from `airlock` to `auth:airlock`

Here is a sample of what I'm using

`@middleware(checks: ["auth:airlock"])`

ipalaus:
@sicsol hmm! that's a change I did because I read in the documentation that they were deprecating @middleware and we should be using @guard.

https://github.com/nuwave/lighthouse/blob/master/CHANGELOG.md#deprecated-1
https://github.com/nuwave/lighthouse/blob/master/UPGRADE.md#v4-to-v5

[...]

PR #1317, opened 2020-04-23
Add docs on using `@guard` with Sanctum

acro5piano:
Hi, thank you for the wonderful library.

Based on the following thread, it is highly common usecase to use lighthouse with Laravel Sanctum, so this section would be helpful especially for beginners.

https://github.com/nuwave/lighthouse/issues/1190

Hope this helps.

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**

### DIFF
--- a/docs/master/security/authentication.md
+++ b/docs/master/security/authentication.md
@@ -56,7 +56,9 @@ extend type Query @guard(with: ["api:admin"]) {
 
 ### Using Laravel Sanctum
 
-If you are using [Laravel Sanctum](https://laravel.com/docs/master/sanctum) to authenticate users, specify `sanctum` middleware.
+Lighthouse will use your application's default guard, unless specified otherwise.
+If you are using [Laravel Sanctum](https://laravel.com/docs/master/sanctum) only for your API,
+you can set it when using `@guard`:
 
 ```graphql
 type Query {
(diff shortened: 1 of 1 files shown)


docs/master/security/authentication.md
```diff
@@ -54,6 +54,16 @@ extend type Query @guard(with: ["api:admin"]) {
 }
 ```
 
+### Using Laravel Sanctum
+
+If you are using [Laravel Sanctum](https://laravel.com/docs/master/sanctum) to authenticate users, specify `sanctum` middleware.
```
spawnia, line 59:
```suggestion
Lighthouse will use your application's default guard, unless specified otherwise.
If you are using [Laravel Sanctum](https://laravel.com/docs/master/sanctum) only for your API,
you can set it when using `@guard`:
```

spawnia: It is not necessary to specify `sanctum` if it is the default.

acro5piano: Sorry, I misunderstood the authentication mechanism.
In `config/auth.php`, setting `guards.api.driver` to `sanctum` solved.

Anyway thank you  for the suggestion.

spawnia:
Thanks @acro5piano for improving the docs :)

acro5piano:
@spawnia 
Thank you for merging!

Issue #1255, opened 2020-03-24
DateTime scalar Laravel 7


Christophvh:
Hi,

Since Laravel 7 the standard output of DateTime values is in Iso8601 format. Would you accept a pr which changes the output/Input to this format?

spawnia:
We might introduce a new scalar `DateTimeTz` for that purpose.

andrewmclagan:
Awesome..!

andrewmclagan:
I'm sure everyone has implemented their own. But for the sake of the community until there is an offical DTZ support:

```php
<?php

namespace App\Http\Scalars;

use Illuminate\Support\Carbon;
use Exception;
use GraphQL\Error\Error;
use GraphQL\Error\InvariantViolation;
use GraphQL\Language\AST\StringValueNode;
use GraphQL\Type\Definition\ScalarType;
use GraphQL\Utils\Utils;

class DateTime extends ScalarType
{
    /**
     * Serialize an internal value, ensuring it is a valid datetime string.
     *
     * @param  \Carbon\Carbon|string  $value
     * @return string
     */
    public function serialize($value): string
    {
        if ($value instanceof Carbon) {
            return $value->toJSON();
        }

        return $this
            ->tryParsingDateTime($value, InvariantViolation::class)
            ->toJSON();
    }

    /**
     * Parse a externally provided variable value into a Carbon instance.
     *
     * @param  mixed  $value
     * @return \Carbon\Carbon
     */
    public function parseValue($value): Carbon
    {
        return $this->tryParsingDateTime($value, Error::class);
    }

    /**
     * Parse a literal provided as part of a GraphQL query string into a Carbon instance.
     *
     * @param  \GraphQL\Language\AST\Node  $valueNode
     * @param  mixed[]|null  $variables
     * @return \Carbon\Carbon
     *
     * @throws \GraphQL\Error\Error
     */
    public function parseLiteral($valueNode, ?array $variables = null): Carbon
    {
        if (! $valueNode instanceof StringValueNode) {
            throw new Error(
                'Query error: Can only parse strings got: '.$valueNode->kind,
                [$valueNode]
            );
        }

        return $this->tryParsingDateTime($valueNode->value, Error::class);
    }

    /**
     * Try to parse the given value into a Carbon instance, throw if it does not work.
     *
     * @param  mixed  $value
     * @param  string|\Exception  $exceptionClass
     * @return \Carbon\Carbon
     *
[...]

PR #1320, opened 2020-04-24
Add DateTimeUtc scalar

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

#1255 

**Changes**

@Christophvh

**Breaking changes**


spawnia:
@Christophvh i had trouble with parsing that format, i am not sure if i got it right.

Christophvh:
> @Christophvh i had trouble with parsing that format, i am not sure if i got it right.

Yes, but I think the parse method should not enforce a format. Any valid date is fine imo, because there are many variations for the ISO-format and when we receive those they should also work. Also a front-end can send a date without a time-zone, in that case, we can assume this follows the timezone of the application.

Carbon::parse(); will work.

spawnia:
`Carbon::parse()` will accept just about anything you throw at it, which i don't think is too great of an idea. Garbage in -> Garbage out.

I do think we should enforce UTC, but might have to allow some degree of freedom in regards to sub-second measurements. Browsers have restricted the precision you can get in order to prevent spectre, meltdown.

```
(new Date).toJSON()
"2020-04-24T08:21:17.625Z"
```

Christophvh:
> `Carbon::parse()` will accept just about anything you throw at it (e.g. `Monday next week`), which i don't think is too great of an idea. Garbage in -> Garbage out.
> 
> I do think we should enforce UTC, but might have to allow some degree of freedom in regards to sub-second measurements. Browsers have restricted the precision you can get in order to prevent spectre, meltdown.
> 
> ```
> (new Date).toJSON()
> "2020-04-24T08:21:17.625Z"
> ```

Yes true, maybe something like this:  Carbon::createFromTimeString('17:00:00', 'UTC');

src/Schema/Types/Scalars/DateTimeTz.php
```diff
@@ -8,11 +8,11 @@ class DateTimeTz extends DateScalar
...
     protected function format(Carbon $carbon): string
     {
-        return $carbon->toIso8601String();
+        return $carbon->toJSON();
     }
 
     protected function parse($value): Carbon
     {
-        return Carbon::createFromFormat(Carbon::ISO8601, $value);
+        return Carbon::createFromIsoFormat('YYYY-MM-DDTHH:mm:ss.SSSSSSZ', $value);
```
Christophvh, line 16:
```suggestion
        return Carbon::createFromTimeString($value, 'UTC');
```

PR #1321, opened 2020-04-24
Multiple belongs to in the same level with existing records

vitorsemeano:
- [ x ] Added or updated tests
- [x] Updated CHANGELOG.md

Related with #1285 

**Changes**

This solves a problem that I detected related with multiple belongsTo at the same level.

The method used to solve this problem, was the exists method of the class Model. This method makes a query with the current attributes of the model instance and check if such exists. This could lead to a true result even if such instance don't exist on the database. 

To solve this, the use of the exists property of the model should be used instead of the method. This property tells if the instance is associated to a record of the database. 

I added tests to test this specific case.

**Breaking changes**

This breaks the test *testUpsertAcrossTwoNestedBelongsToRelationsAndOverrideExistingModel*. 

I believe this test will create a duplication for the role_user, before the sync is being made. In the way the test is written, it will create a second relationship between user and role, even if we are referencing the same user and role of the first creation. 

I think there are other ways of achieving what this test is achieving, by two ways:

- specify the id of the relation role_user
- use the belongsToMany relationship to access the related role

I modified this test, specifying the role_user id, as also i created a new test for the belongsToMany case. 

I added this in the breaking changes because i don't know if this was the intended behaviour of lighthouse or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## 4.12.1
+
+### Fixed
+
+- Fix creating multiple nested BelongsTo relationships on the same level when previous records
+  with matching attributes exist https://github.com/nuwave/lighthouse/pull/1321
+
 ## 4.12.0
 
 ### Added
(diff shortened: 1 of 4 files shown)


spawnia:
Thanks for your great work @vitorsemeano, just released the fix in v4.12.1

PR #1338, opened 2020-04-28
Throw an exception if the return type declaration class for a relation does not exist

tjbp:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Throws an exception if the return type declaration for a model relation is a class that doesn't exist. PHP doesn't check if the return type for a method exists, so if an alias for `Illuminate\Database\Eloquent\Relations\BelongsTo` was accidentally omitted and `BelongsTo` was used as the return type (as an example), the return type would be `App\BelongsTo`. Since Lighthouse only compares this value with valid Eloquent relation classes, it silently fails to recognise the relation method.


src/Execution/Arguments/ArgPartitioner.php
```diff
@@ -187,6 +188,10 @@ public static function methodReturnsRelation(
             return false;
         }
 
+        if (! class_exists($returnType->getName())) {
+            throw new Exception('Class '.$returnType->getName().' does not exist, did you forget to alias the Eloquent relation class?');
```
spawnia, line 192:
```suggestion
            throw new Exception('Class '.$returnType->getName().' does not exist, did you forget to import the Eloquent relation class?');
```

Aliasing would be `use App\Foo as Bar;`

tests/Utils/Models/WithoutRelationAlias.php
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @property int $id
+ * @property string $name
+ */
+class WithoutRelationAlias extends Model
+{
+    protected $guarded = [];
+
+    public function users(): HasMany
```
spawnia, line 15:
```suggestion
    public function users(): HasMany // @phpstan-ignore-line Missing the import on purpose
```

spawnia:
Great change, thank you! I like it when we can provide useful error messages on obvious programmer errors.

src/Execution/Arguments/ArgPartitioner.php
spawnia, line 192: I think we should use a more particular exception type, maybe `DefinitionException`?

tjbp: @spawnia fine by me - I was going to use `DefinitionException` but thought it might be reserved for exceptions related directly to GraphQL language.

spawnia: The current Docblock said so, good call. It is however meant to be used a bit more broadly than that, cases where the PHP code does not match the schema are also covered by it.

spawnia:
Thanks @tjbp

PR #1343, opened 2020-05-01
Configure a default `guard` for all authentication functionality

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Allow configuring a default guard for all of Lighthouse to use.

**Breaking changes**

Not yet, but the default guard will change to `api` in v5.

### DIFF
--- a/src/Schema/Directives/GuardDirective.php
+++ b/src/Schema/Directives/GuardDirective.php
@@ -80,7 +80,7 @@ SDL;
     protected function authenticate(array $guards): void
     {
         if (empty($guards)) {
-            $guards = [config('lighthouse.guard', null)];
+            $guards = [config('lighthouse.guard')];
         }
 
         foreach ($guards as $guard) {
--- a/src/Schema/Directives/GuardDirective.php
+++ b/src/Schema/Directives/GuardDirective.php
@@ -42,8 +42,8 @@ users to still receive partial results.
 """
 directive @guard(
   """
-  Specify which guards to use, e.g. "api".
-  When not defined, the default driver is used.
+  Specify which guards to use, e.g. ["api"].
+  When not defined, the default from `lighthouse.php` is used.
   """
   with: [String!]
 ) on FIELD_DEFINITION | OBJECT
--- a/src/Schema/Directives/GuardDirective.php
+++ b/src/Schema/Directives/GuardDirective.php
@@ -80,7 +80,7 @@ SDL;
     protected function authenticate(array $guards): void
     {
         if (empty($guards)) {
-            $guards = [null];
+            $guards = [config('lighthouse.guard', null)];
         }
 
         foreach ($guards as $guard) {
--- a/src/Support/Http/Middleware/AttemptAuthentication.php
+++ b/src/Support/Http/Middleware/AttemptAuthentication.php
@@ -40,7 +40,7 @@ class AttemptAuthentication
     protected function attemptAuthentication(array $guards): void
     {
         if (empty($guards)) {
-            $guards = [config('lighthouse.guard', null)];
+            $guards = [config('lighthouse.guard')];
         }
 
         foreach ($guards as $guard) {
--- a/src/Support/Http/Middleware/AttemptAuthentication.php
+++ b/src/Support/Http/Middleware/AttemptAuthentication.php
@@ -40,7 +40,7 @@ class AttemptAuthentication
     protected function attemptAuthentication(array $guards): void
     {
         if (empty($guards)) {
-            $guards = [null];
+            $guards = [config('lighthouse.guard', null)];
         }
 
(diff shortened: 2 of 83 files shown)


spawnia:
We might deprecate the ability to configure other guards in the directives `@guard` and `@auth`, as i don't believe it is wise to use different guards for different fields.

src/Schema/Directives/GuardDirective.php
```diff
@@ -80,7 +80,7 @@ function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
     protected function authenticate(array $guards): void
     {
         if (empty($guards)) {
-            $guards = [null];
+            $guards = [config('lighthouse.guard', null)];
```
stayallive, line 83:
```suggestion
            $guards = (array)config('lighthouse.guard', []);
```

This would allow passing either an array or string to the `lighthouse.guard` config to still keep the option for multiple default guards (think `['api', 'web']` for example).

src/Support/Http/Middleware/AttemptAuthentication.php
```diff
@@ -40,7 +40,7 @@ public function handle($request, Closure $next, ...$guards)
     protected function attemptAuthentication(array $guards): void
     {
         if (empty($guards)) {
-            $guards = [null];
+            $guards = [config('lighthouse.guard', null)];
```
stayallive, line 43:
```suggestion
            $guards = (array)config('lighthouse.guard', []);
```

This would allow passing either an array or string to the `lighthouse.guard` config to still keep the option for multiple default guards (think `['api', 'web']` for example).

src/Schema/Directives/GuardDirective.php
spawnia: The issue with specifying multiple guards is that the call to `$request->user()`, inside the Context, does only accept a single string.

We would have to change that part.

stayallive: When passing an array of guards it will check each and the first "valid" one is set as the default (`$this->auth->shouldUse($guard)`) so `$request->user()` will retrieve the correct user, so I don't think any changes are needed there?

spawnia: I came up with this PR while trying Lighthouse in a Laravel install where Sessions are completely ripped out. The call to `$request->user()` still fails when the overall default guard is set to `web`, it tries session based auth.

We can prevent that by passing the guard explictly, but calling `$request->user(['api', 'fallback'])` is not valid.

src/Support/Http/Middleware/AttemptAuthentication.php
spawnia: An empty array is not a good default here, it would not check any guards in that case.

src/Schema/Directives/GuardDirective.php
stayallive: If you rip out sessions and still have the default guard set to `web` you are configuring it wrong somewhere, no? Assuming [this](https://github.com/laravel/laravel/blob/0f133c1e8eff44d765a40fc1934172e7bf046e56/config/auth.php#L17) is set to `api`. You should not have to pass any argument to `$request->user()` since it will use the "default" to `api` (which can still return `null`), then you should also not have to have a `fallback` guard, although I might misunderstand your actual app setup.

And when you can set `lighthouse.guard` to `['api', 'web']` it will call `$auth->shouldUse($guardName)` changing the "default" guard to whatever one is authenticated (or default back to whatever `auth.defaults.guard` has set and will return the correct result without needing to pass the `$guard`.

src/Support/Http/Middleware/AttemptAuthentication.php
stayallive: You are correct it should be `[null]` as default indeed, I missed that.

src/Schema/Directives/GuardDirective.php
spawnia: I agree that the particular error i ran into will not be an issue. Still, for users that default the guard to `web`, unnecessary work is being done on every request.

The changes from https://github.com/nuwave/lighthouse/pull/1343/commits/1b9ae16821c52e57a4d60c020cae5d1a21a09d1d prevent that, but setting `lighthouse.guard = []` triggers an error when calling `$request->user()`.

Do you think using multiple guards is all that common?

stayallive: Hmm, okay yeah I understand why you might want to pass the guard now.

I know I do it, probably not that common though. I can leave it set to `null` and add my own middleware to achieve the same thing (like I'm doing now) so no harm in this, had not considered the possibility of optimising the `$request->user()` call.

spawnia:
Thanks @stayallive

PR #1348, opened 2020-05-03
Use `forceFill()` when mutating models

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The schema already prevents clients from passing attributes they are not allowed to set.
This avoids forcing the user to mark all their attributes as fillable, allowing
mass-assignment protection to be used for other cases, where it makes sense.

**Breaking changes**

No


lorado:
VERY useful change! I had situations, where I was asking myself, why the data don't get stored via mutation... and every time it was mass-protection... I LIKE IT!

PR #1351, opened 2020-05-04
Update PusherLink example to handle unsubscribes

stayallive:
- [ ] ~Added or updated tests~
- [ ] ~Documented user facing changes~
- [ ] ~Updated CHANGELOG.md~

I don't think this requires any of the above unless we want to mention it in the changelog, although it is not really a change... more a doc update :)

**Changes**

I noticed the PusherLink for Apollo example does not unsubscribe correctly, this PR aims to resolve that and have a "proper" PusherLink that unsubscribes when Apollo closes the subscription.

**Breaking changes**

None. User paste this code in their project.


spawnia:
Nicely done, i especially appreciate putting the reference to the original in there.

Issue #1339, opened 2020-04-28
Add global defaultCount for pagination directive


BenHeimberg:
### Add global defaultCount value for pagination directive

When working on big projects with a lot of queries that returns paginated results, it would be nice to be able to set the default count outside of the paginate directive.

It would save time, and simplify development of systems that requires the same pagination counts on almost every query.

### Current implementation
```
type Query {
     queryOne: @paginate(defaultCount: 15)
     queryTwo: @paginate(defaultCount: 15)
}
```

### Proposed change
```
type Query {
     queryOne: @paginate
     queryTwo: @paginate
}
```

### Solution
Add `pagination_default_count` property to the config file.

I would love to implement it, if you guys think its reasonable.

_Ps. This is my first ever proposal :)
Great library!_

spawnia:
That sounds very reasonable. In order to be backwards compatible, the setting will have to default to `null`, we might add another default in v5

See https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md for how you can prepare a PR.

PR #1352, opened 2020-05-05
Configure the default amount of items in paginated lists with `pagination.default_count`

BenHeimberg:
- [x]  Test created
- [x] CHANGELOG updated

**Changes**

Added config for defaultCount argument in @paginate directive as defined in #1339  .

**Breaking changes**

None.

### DIFF
--- a/src/lighthouse.php
+++ b/src/lighthouse.php
@@ -118,41 +118,33 @@ return [
     |--------------------------------------------------------------------------
     | Pagination
     |--------------------------------------------------------------------------
-    |
-    | Limits the maximum "count" that users may pass as an argument
-    | to fields that are paginated with the @paginate directive.
-    | A setting of "null" means the count is unrestricted.
+    | Pagination related configuration settings.
     |
     */
 
-    'paginate_max_count' => null,
+    'pagination' => [
+        /**
+         * Set the default count value for the @paginate directive.
+         * A setting of "null" mean that "defaultCount" argument needs to be set
+         * for each paginate directive, or to pass the "first" argument in the request.
+         */
+        'default_count' => null,
 
-    /*
-    |--------------------------------------------------------------------------
-    | Pagination Default Count
-    |--------------------------------------------------------------------------
-    |
-    | Set the default count value for the @paginate directive.
-    | A setting of "null" mean that "defaultCount" argument needs to be set
-    | for each paginate directive, or to pass the "first" argument in the request.
-    |
-    */
+        /**
+         * Limits the maximum "count" that users may pass as an argument
+         * to fields that are paginated with the @paginate directive.
+         * A setting of "null" means the count is unrestricted.
+         */
+        'max_count' => null,
 
-    'pagination_default_count' => null,
-
-    /*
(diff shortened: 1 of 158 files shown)


spawnia:
You can add tests to https://github.com/nuwave/lighthouse/blob/master/tests/Unit/Schema/Directives/PaginateDirectiveTest.php

BenHeimberg:
Test added

tests/Unit/Schema/Directives/PaginateDirectiveTest.php
```diff
@@ -297,4 +297,31 @@ public function testThrowsIfBuilderIsNotPresent(): void
         }
         ');
     }
+
+    public function testDoesNotRequireDefaultCountArgIfDefinedInConfig() : void
+    {
+        $defaultCount = 5;
+        config(['lighthouse.paginate_default_count' => $defaultCount]);
+
+        $this->buildSchema(/** @lang GraphQL */'
```
spawnia, line 306:
```suggestion
        $this->schema = /** @lang GraphQL */ '
```

BenHeimberg: Thanks!

src/lighthouse.php
```diff
@@ -127,6 +127,19 @@
...
+    | Pagination Default Count
+    |--------------------------------------------------------------------------
+    |
+    | Set the default count value for the @paginate directive.
+    | A setting of "null" mean that "defaultCount" argument needs to be set
+    | for each paginate directive, or to pass the "first" argument in the request.
+    |
+    */
+
+    'pagination_default_count' => null,
```
spawnia, line 141: How about we namespace the configuration options for pagination? This would be `pagination.default_count` then.

tests/Unit/Schema/Directives/PaginateDirectiveTest.php
```diff
@@ -297,4 +297,31 @@ public function testThrowsIfBuilderIsNotPresent(): void
...
         ');
     }
+
+    public function testDoesNotRequireDefaultCountArgIfDefinedInConfig() : void
+    {
+        $defaultCount = 5;
+        config(['lighthouse.paginate_default_count' => $defaultCount]);
+
+        $this->schema = /** @lang GraphQL */ '
+            type User {
```
spawnia, line 307: Formatting nit: We don't indent those strings.

spawnia:
The tests are failing because the test class provides no database access. Since you are using the DB, the test must be moved to https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Schema/Directives/PaginateDirectiveDBTest.php instead. Thank you!

src/lighthouse.php
BenHeimberg: Sound good!

BenHeimberg:
Ok, so I created the 'pagination' array in config & converted all references of the props.
Also the test is moved to the correct place.

spawnia:
Nice job @BenHeimberg, thank you!

PR #1360, opened 2020-05-13
Use @rename directive in validation

tjbp:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

When a validation rule utilises a different attribute it should respect the `@rename` directive.

**Breaking changes**

Nothing I can think of.

### DIFF
--- a/tests/Integration/ValidationTest.php
+++ b/tests/Integration/ValidationTest.php
@@ -21,6 +21,9 @@ class ValidationTest extends DBTestCase
             list: [String]
                 @rules(apply: ["required", "email"])
                 @rulesForArray(apply: ["max:2"])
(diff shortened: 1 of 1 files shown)


tjbp:
I have only added tests since I'm interested in input on this before proceeding with any implementation.

spawnia:
This change would actually result in severe breakage for many scenarios that rely on the current interaction between validation and `@rename`.

Applying validation first is especially crucial for clients. If renaming were to occur first, they might get a nonsensical validation error with an argument name they haven't sent.

The `confirmed` validation won't work, unless you switch to snake_case naming. You can always replace it with `same:myConfirmation`, though.

tjbp:
Fair enough! I did get onto this through the confirmed validation and the snake_case issue, so thanks for the tip.

PR #1371, opened 2020-05-20
Write definitions for programmatically registered types to file with `lighthouse:ide-helper`

### LINKED ISSUE #1369: Add programmatically registered types to generated file via 'ide-helper'

At the moment I have only one small "issue" - I use enums with `BenSampo\Enum\Enum`, and they are sure added to the schema with `$typeRegistry->register(new LaravelEnumType($enum))`.

But lighthouse is focused more in defining SDL directly, so the most time I am defining schema in `.graphql`-files. But my IDE don't know about existing of enums, that are defined in PHP, like I described above. The code works, schema is valid, but while writing schema IDE says something like: "The type '...' is not present ...". 

I tried to add enum definition to autogenerated `schema-directives.graphql`, and it works. IDE knows enum, and it isn't a part of schema, so schema stays unchanged, valid, without double definitions of enums. 

I think it would be good, if we could add at least enum definitions to autogenerated file via `lighthouse:ide-helper`.

The better solution - add all of types to that file, which are registered with `$typeRegistry->register()`

So let us discuss, which solution you think is good, and do we need it at all? I would like to write a PR, when the direction of solution is clear.

lorado:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1369

**Changes**
`lighthouse:ide-helper` artisan command will also print all types registered with `TypeRegistry::register()` to the generated graphql file.

**Breaking changes**
NO


lorado:
This is really a tiny solution, but I hope it will help a lot =) for my purpose it definitely works well

src/Schema/TypeRegistry.php
```diff
@@ -41,7 +41,7 @@ class TypeRegistry
      *
      * @var array<string, \GraphQL\Type\Definition\Type>
      */
-    protected $types;
+    protected $types = [];
```
lorado, line 44: FYI: I had to add a default value, because when I tried to call `$typesRegistry->possibleTypes()` without to register any type at all, it was throwing an error, because `possibleTypes()` expects to return an array, but null was returned, because `$types` were not defined and not manipulated.

spawnia: Good idea regardless.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+- Improve `lighthouse:ide-helper` https://github.com/nuwave/lighthouse/pull/1371
```
spawnia, line 13:
```suggestion
- Add programmatically registered types to schema when running `lighthouse:ide-helper` https://github.com/nuwave/lighthouse/pull/1371
```

src/Console/IdeHelperCommand.php
```diff
@@ -108,6 +111,14 @@ protected function buildSchemaString(array $directiveClasses): string
                 .$definition."\n";
         }
 
+        // append programmatically registered types
+        $typeRegistry = resolve(TypeRegistry::class);
```
spawnia, line 115:
```suggestion
        $typeRegistry = app(TypeRegistry::class);
```

This removes one layer of indirection. Injection through `handle()` would be even better though, it is good to see the dependencies upfront.

src/Console/IdeHelperCommand.php
```diff
@@ -108,6 +111,14 @@ protected function buildSchemaString(array $directiveClasses): string
...
+        $typeRegistry->setDocumentAST(new DocumentAST);
+        $programmaticallyRegisteredTypes = $typeRegistry->possibleTypes();
```
spawnia, line 117: This actually loads SDL defined types, too.

lorado: I don't think so. I pass an empty `DocumentAST` instance, with an empty schema. When `possibleTypes()` gets executed, `$this->documentAST->types` in line 182 in `TypeRegistry.php` is an empty array, so the function simply returns `$this->types`, and there are only types, that were registered with `$typeRegistry->register()`

lorado: I tried the command in my live project - only programmatically registered types are considered and printed into output file

spawnia: Nice! I appreciate the creativity, but found that adding an empty `DocumentAST` is a bit of a hacky workaround. I added `resolvedTypes` to `TypeRegistry` to make this more straightforward.

spawnia:
Changed it so it writes to `programmatic-types.graphql` instead of munging them into `schema-directives.graphql`.

If we come up with a good name that covers both, i am open to putting them into one file in v5.

lorado:
Well ok, it may be now less hacky, but I am not sure, if we will get only programmatically registered types via `TypeRegistry::resolvedTypes()`, as the function name already says, it will return resolved types. What if later, we make a change in code, and `possibleTypes()` get's called by core code earlier, before we call `resolvedTypes()` within `IdeHelperCommand`, it will break. 

Or do you mean, it will not happen? 

Before I startet this PR, my first thought was, that programmatically registered types are stored in a different/dedicated array, so they never get mixed with types from SDL. May be it would be a clearer solution?

spawnia:
I totally get what you mean, the way the `TypeRegistry` is organized makes me feel a bit uneasy. Splitting the types would incur a performance cost on every type lookup, not sure if it is significant.

Actually, it might be useful to print generated types too, since one might want to reference them too in some cases. How about we gather the type names defined within the user schemas and diff that with the final schema?

lorado:
Well yes, currently solution is working, and I think it is enough for now. We can keep your suggested solution to SDL types from all types, it sounds to be most robust. If currently command breaks, we modify it then.

P.S. generated PHP file was broken, as you added autogenerated comment-lines before opening `<?php`, so php file was not valid. I made a simple but dirty solution for it. Take a look, and if its ok for you, then go for merge

lorado:
Oh well... it got a little messy :D you can modify it as you wish. I thought it is good to have same comment notice about autogenerated file...

lorado:
By the way - we could combine `programmatic-types.graphql` and `schema-directives.graphql` into `_lighthouse_ide_helper.graphql` in v5, if you wish. It would be at least straightforward and clear.

spawnia:
@lorado Great catch, works great now!

spawnia:
@lorado Great catch, works great now!

lorado:
Hmm... now generated graphql files don't contain autogenerated notice... :D

spawnia:
Love string manipulation ❤️

Issue #1372, opened 2020-05-20
Broken @update directive


Rockson:
**Describe the bug**

The last version v4.13.0 breaks  the @update directive when not using "id" as primary key in the model.
While the version v4.12.4 works fine
The documentation states: 

> Lighthouse uses the argument id to fetch the model by its primary key. **This will work even if your model has a differently named primary key**, so you can keep your schema simple and independent of your database structure.

So having a primary key other than "id" should be supported.

As of now the directive is setting the column ID even though there is no ID Column in the database

**Steps to reproduce**

> type MyModel  {
    id: ID! @rename(attribute: "my_different_id")
    content: String
}

> myedit(id: ID!, content: String!): MyModel @update

> class MyModel {
        protected $table = "my_table";
        protected $primaryKey = "my_different_id";
        protected $fillable = ["content"];
}

**Output/Logs**
The generated SQL is:

 (SQL: update "my_table" set "CONTENT" = CONTENT, "ID" = 1022 where "MY_DIFFERENT_ID" = 1022)

So basically, the new version is adding ID = ? to the generated query, even though the column is not the primary key

**Lighthouse Version**

v4.13.0

webard:
Same problem in my application, but I think it is Eloquent's forceFill method problem, not lighthouse.

Rockson:
I think Eloquent's forceFill behaviour is intended, that's why it's called **force**. If you pass it a property that doesn't exist on the table the model has no way to know it so it creates the wrong sql. The mistake is on the lighthouse side

webard:
You may be right @Rockson  but when you change primaryKey name, forceFill method should respect your setting. On the other side, you may have `$primaryKey = 'primary_id'` column and `id` column for something else. It is messy, but possible. One of solutions is "in the middle" method - forceFill with respecting $primaryKey.

Rockson:
> You may be right @Rockson but when you change primaryKey name, forceFill method should respect your setting. On the other side, you may have `$primaryKey = 'primary_id'` column and `id` column for something else. It is messy, but possible. One of solutions is "in the middle" method - forceFill with respecting $primaryKey.

Yeah it should never happen that in your db the actual primary key is called "primary_id" and then in the same table you have another column named "id", but I don't think Eloquent should assert this rule in his code, it may happen on some legacy code, who knows.

spawnia:
I see how switching to using `forceFill()` caused your code to break. In hindsight, i wish i considered that case and been more cautious.

[...]

PR #1373, opened 2020-05-21
Add failing test for @update with a custom primary key

andershagbard:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Adds a failing test for #1372.

This error is caused by `forceFill` in https://github.com/nuwave/lighthouse/pull/1348. By using `forcefill` all attributes gets updated, including `id`, even when the `id` shouldn't get updated.


spawnia:
@andershagbard it works out fine when `@rename` is applied to the argument.

PR #1390, opened 2020-05-28
Add @withCount directive and eager load in @count directive

tlaverdure:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md

**Changes**

Adds a directive to eager load relationship counts. The `@count` directive is ok for simple counts, but when loading many models `@withCount` will leverage Eloquent's `withCount()` method to prevent n+1 issues. This PR is takes inspiration from `@with`.

**Breaking changes**

None

```graphql
type User {
  tasks_count: Int! @withCount
}
```


spawnia:
Looking pretty good so far. Just by glancing over the implementation, it seems like we might be able to abstract some shared parts between `@with` and this directive?

tlaverdure:
Updated, let me know if we need anything else.

spawnia:
> Updated, let me know if we need anything else.

Passing CI would be a great next step.

tlaverdure:
All good now. Thanks!

tests/Integration/Schema/Directives/WithCountDirectiveTest.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Tests\Integration\Schema\Directives;
+
+use Nuwave\Lighthouse\Support\AppVersion;
+use Tests\DBTestCase;
+use Tests\Utils\Models\Task;
+use Tests\Utils\Models\User;
+
+class WithCountDirectiveTest extends DBTestCase
```
spawnia, line 10: The tests in this class need to be more thorough. I don't see how they validate the directive actually eager-loads the count and saves database queries.

Also, the semantics for the `_count` convention should be defined (and tested) more clearly.

tests/Integration/Schema/Directives/WithCountDirectiveTest.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            user: User @first
+        }
+
+        type User {
+            tasks_count: Int! @withCount
+            tasks_relation_count: Int! @withCount(relation: "tasks")
```
spawnia, line 25: I removed the `$relation as $field` part of the directive, so this fails for now. The semantics of this are a bit unclear to me. It is not described as part of the `@withCount` directive definition.

For example, what would happen in a case like this, assuming the user has a column `name`:

```graphql
type User {
  name: Int! @withCount(relation: "tasks")
}
```

Would the eager-load supersede the column?

tlaverdure: > For example, what would happen in a case like this, assuming the user has a column name:
> ```graphql
> type User {
>   name: Int! @withCount(relation: "tasks")
> }
>```
>Would the eager-load supersede the column?

Yes, I think it should supersede it since the field may or may not follow the *_count* conventions. For example, `completed_tasks` could return how many tasks are completed on a task list. It uses the underlying `tasks` relation but should be aliased to `completed_tasks` when fetched from the database and returned as so allowing it to be resolved properly. This follows Laravel's `withCount()` aliasing capabilities:

https://github.com/laravel/framework/blob/7d30c1b564ddcd87a0fb6b59bc7b8d3cde7432e8/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L368-L378

>Ultimately, i think we are better off by not conflating the resolving of the field with eager >loading. They might be completely unrelated, e.g.:
> ```graphql
> type User {
>   statistics: UserStatistics @method @withCount(relation: "tasks")
> }
> ```

In this case, the `statistics()` method on `User::class` would use `$rootValue->tasks_count` to resolve the `UserStatistics`. I don't see an issue with that.

tests/Integration/Schema/Directives/WithCountDirectiveTest.php
tlaverdure: I've added some more tests to ensure to show queries being saved.

Trying to make the semantics clearer as well. Although it would be most common, there are going to be cases where you may not even use the `_count` convention. For example:


```graphql
{
  tasks_count: Int @withCount #This is a typical case
  unread_notifications: Int @withCount(relation: "notifications", scopes: ["unread"]) # I think this would be a common case
  balance: Int @withCount(relation: "transactions", scopes: ["withBalanceSum"]) #An abstract case but plausible – altering the query constraint to get a different type of aggregate like SUM(column). I actually used something of the sort yesterday!

}
```

spawnia:
@tlaverdure we need to clear up the distinction between a `FieldMiddleware` and a `FieldResolver`. You seem to conflate their responsiblities.

Just like `@with`, a `FieldMiddleware` is explicitly **not** about producing a value for the field. It may merely be used to assist in allowing the resolver to work more efficiently, but must not assume that it provides the actual value.

On the other hand, `@count` actually produces a value for the field it sits on. As you noted, it does so quite inefficiently. I think what should actually happen in this PR is that `@count` is optimized to use eager loading, similar to the other relation directives such as `@hasMany`.

This eliminates the use for using the `as` style aliasing and allows for the following use cases:

```graphql
type User {
  # We can optimize this through eager-loading
  tasks_count: Int @count(relation: "tasks")

  # Allowing extra scopes seems reasonable
  unread_notifications: Int @count(relation: "notifications", scopes: ["unread"])

  # Assuming statistics() requires the count, improves efficiency
  statistics: UserStatistics @withCount(relation: "tasks") @method

  # This becomes possible soon when repeatable directives are allowed
  statistics: UserStatistics @withCount(relation: "tasks") @withCount(relation: "posts") @method
}
```

spawnia:
Btw, thanks for your work on this PR so far. I just want to make sure we are implementing the right thing.

Utilizing eager-loading in `@count`, and adding `@withCount` to enable additional use cases is a great improvement to Lighthouse.

tlaverdure:
Ah yes, thanks for the clarification!

Yes I agree with both. We should add eager loading to `@count` and `@withCount` shouldn't produce a value but provide the eager loading the count through the field middleware. I can try to take a stab at that later on when I get a chance.

Lighthouse is awesome, thanks for maintaining it!

tlaverdure:
Made some updates. Not sure what to do about this: https://github.com/nuwave/lighthouse/pull/1390/checks?check_run_id=750500773

spawnia:
> Made some updates. Not sure what to do about this: [#1390 (checks)](https://github.com/nuwave/lighthouse/pull/1390/checks?check_run_id=750500773)

I will take a look, doesn't seem like it has something to do with your PR.

src/Schema/Directives/WithRelationDirective.php
```diff
@@ -0,0 +1,92 @@
...
+            );
+        };
+    }
+
+    /**
+     * Return a new deferred resolver.
+     *
+     * @param  callable  $resolver
+     */
+    protected function deferredRelationResolver($resolver): Closure
```
spawnia, line 59:
```suggestion
    protected function deferredRelationResolver(callable $resolver): Closure
```

Really like how you have broken this apart, good stuff.

src/Schema/Directives/WithCountDirective.php
```diff
@@ -0,0 +1,50 @@
...
+
+class WithCountDirective extends WithRelationDirective implements FieldMiddleware, DefinedDirective
+{
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'SDL'
+"""
+Eager-load the count of an Eloquent relation.
+
+The name of the field is assumed to be `${RELATION}_count`,
```
spawnia, line 18: We should remove that part, as `@count` should be used when the field should actually resolve to the count.

tlaverdure: Got it.

spawnia:
@tlaverdure are you fine with the changes i made?

tlaverdure:
@spawnia yep! Looks good to me.

spawnia:
Thanks for the collaboration @tlaverdure, nicely done

tlaverdure:
Alright! 🙌🏾

P-Shakibafar:
#1817

Issue #1409, opened 2020-06-07
Creating where condition for nested relations


alexey-krukover:
I will gladly provide the necessary changes for this issue

**What problem does this feature proposal attempt to solve?**

Currently the `whereCondition` directive does not support filtering for nested relations

**Which possible solutions should be considered?**

Adding support for relation path with delimiter, that simply checks the provided condition with the Laravel function `whereHas`. Which in my option can be integrated with the `handleWhereConditions` function inside the `whereConditionsBaseDirective` class. In addition, some of the `whereHasConditionsBaseDirective` code can be removed and use the new options of the `whereConditionsBaseDirective` instead. Also to further improve future support. beside the relation path, we can pass an integer to filter out results based on how many items match the `has` condition, which is also natively supported in Laravel.

**Example**
```graphql{
user(
    where: {
       HAS: {
           relation  : 'posts.comments'
           amount    : 3
           operator  : EQ
           condition : {
               column : 'text'
               value  : 'test'
           }
       }
     }) {
    name
  }
}

**Comment by spawnia** [comment]
How would the generated input types look like then compared to now: https://lighthouse-php.com/master/eloquent/complex-where-conditions.html#wherehasconditions

What about nesting the `HAS` conditions within other logical operators such as `AND`?

**Comment by Necroblight** [comment]
In addition maybe also add property for `operator' to be able to set `has` condition with `equal` or `less than`  comparison conditions.

**Comment by Necroblight** [comment]
> What about nesting the `HAS` conditions within other logical operators such as `AND`?

I encountered the conflict of getting relation in builder versus in model. I have some rough ideas how it could be possible to overcome, that I can try out. If the idea itself is fine with you, then the problem itself I'll try to solve in the pull request.

**Comment by Necroblight** [comment]
After researching the issue more, I understand your questions better. So here's my revised response;

> How would the generated input types look like then compared to now: https://lighthouse-php.com/master/eloquent/complex-where-conditions.html#wherehasconditions

```graphql
input UsersWhereConditions
{
  column: UsersColumn

  operator: SQLOperator = EQ

  value: Mixed

  AND: [PeopleHasRoleWhereConditions!]

  OR: [PeopleHasRoleWhereConditions!]

  HAS: UsersRelationWhereConditions
}

input UsersRelationWhereConditions
{
  relation: String!

  operator: SQLOperator = GTE

  amount: Int = 1

  condition: PeopleHasRoleWhereConditions
}
```

[...]

Issue #1401, opened 2020-06-03
Cannot use blank types _ inside objects


v1narth:
Directives like `@orderBy` and `@whereConditions` can accept a blank type defined by `_`, which is working perfectly in the queries root, but not inside an object.

Working:

``` bash
type Query {
    users(where: _ @whereConditions) : [User]
}
```

Not working:

``` bash
type Query {
    users(input: UsersQueryInput @spread) : [User]
}

Input UsersQueryInput {
    where: _ @whereConditions
}
```

Im not sure if its a bug, or it is intended, are there any reasons for it not to be implemented?

Version 4.13.1

spawnia:
Categorizing this as a bug because the definition for `@whereConditions` indicates that this should work on both places: `ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION`.

Lighthouse applies a process we call *schema manipulation* for some directives, allowing them to transform the GraphQL AST: https://github.com/nuwave/lighthouse/blob/301930b356faa63da681d564cbef41c7d312f2aa/src/Schema/AST/ASTBuilder.php#L133-L137

The directives `@whereConditions` and `@orderBy` implement the [`ArgManipulator`](https://github.com/nuwave/lighthouse/blob/master/src/Support/Contracts/ArgManipulator.php) interface. This is only called when they are defined directly on the args of a field.

While there is a directive interface for definitions that are directly applied on types (e.g. `input Foo @example`), we currently have no mechanism for traversing down into the fields of `input` types and apply transformation there.

An implementation for this would entail:
- Adding a new interface `InputFieldManipulator`
- Add a method to `ASTBuilder` that applies it to `input` fields
- Implement the interface for the `ArgManipulator` directives where it makes sense

BenHeimberg:
Thanks for the clarification and fix.

BenHeimberg:
Also, there are ways of using nested where conditions?

spawnia:
#1412 is nice, but does not solve the problem you were alluding to in this issue. I created #1417 as a follow up to track the usage of directives on input fields.

v1narth:
Thanks!

PR #1412, opened 2020-06-09
Added ability to handle relations in base where conditions

alexey-krukover:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Implemented: https://github.com/nuwave/lighthouse/issues/1409

**Breaking changes**

none


Necroblight:
> Great solution for #1401 thank you, awesome work : )

Thank you, hopefully it will be implemented quickly!

src/WhereConditions/WhereConditionsServiceProvider.php
```diff
@@ -11,8 +11,14 @@
 
 class WhereConditionsServiceProvider extends ServiceProvider
 {
+    public const DEFAULT_HAS_OPERATOR = 'GTE';
```
spawnia, line 14: This hardcodes a value that is internal to the `SQLOperator`, which is just one possible implementation of the `Operator` interface.

src/WhereConditions/WhereConditionsServiceProvider.php
```diff
@@ -91,8 +100,50 @@ public static function createWhereConditionsInputType(string $name, string $desc
 
                 "A set of conditions that requires at least one condition to match."
                 OR: [$name!]
+
+                "A condition for chacking relations."
```
spawnia, line 104: That has a typo and could be more descriptive.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -52,17 +52,15 @@ public function handleBuilder($builder, $whereConditions)
...
-                        $this->handleWhereConditions($builder, $whereConditions);
-                    }
-                );
-            }
-        );
+        return $this->handleWhereConditions($builder, [
+            'HAS' => [
+                'relation'  => $this->getRelationName(),
+                // TODO: There should be a way to convert operator to its value from the code; WhereConditionsServiceProvider::DEFAULT_HAS_OPERATOR
+                'operator'  => '>=',
```
spawnia, line 59: This value is an implementation detail of the `Operator` interface.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -52,17 +52,15 @@ public function handleBuilder($builder, $whereConditions)
...
-            function ($builder) use ($whereConditions): void {
-                // This extra nesting is required for the `OR` condition to work correctly.
-                $builder->whereNested(
-                    function ($builder) use ($whereConditions): void {
-                        $this->handleWhereConditions($builder, $whereConditions);
-                    }
-                );
-            }
-        );
+        return $this->handleWhereConditions($builder, [
```
spawnia, line 55: It should be possible to abstract this call more cleanly.

src/WhereConditions/WhereHasConditionsDirective.php
```diff
@@ -43,7 +43,7 @@ public static function definition(): string
     /**
      * @param  \Illuminate\Database\Eloquent\Builder  $builder  The builder used to resolve the field.
      * @param  mixed  $whereConditions The client given conditions
-     * @return \Illuminate\Database\Eloquent\Builder  The modified builder.
+     * @return \Illuminate\Database\Eloquent\Builder | \Illuminate\Database\Query\Builder  The modified builder.
```
spawnia, line 46: I think this will not work with a non-eloquent query builder.

alexey-krukover:
> I fixed a bunch of formatting and naming. Please stick to the existing coding conventions.

Sorry it took so long to get back to it, was too busy with other stuff.

src/WhereConditions/WhereHasConditionsDirective.php
alexey-krukover: I left that option on, in case someone would want to use that method manually, or from different place in code, but would only have a raw builder. If that shouldn't be the case, then yeah, it's unnecessary there. And it would break only if someone would try to use it manually and use `HAS`, which we can fix by checking if `model` before applying the `HAS`. Tell me which way you prefer.

src/WhereConditions/WhereConditionsServiceProvider.php
alexey-krukover: Should I just use the real SQL operator directly, or is there some method to get an enum name of `SQLOperator`, using the operator itself? Or should I just create a function like `default` that just return the name directly?

alexey-krukover:
So should we use the generated input I recommended in here; https://github.com/nuwave/lighthouse/issues/1409#issuecomment-641866097

src/WhereConditions/WhereHasConditionsDirective.php
alexey-krukover: At the risk of sounding like a total noob. What do you exactly mean by abstracting the call more cleanly? You mean creating a separate method that accept the model and `HAS` parameters, and using that instead? Or something else?

src/WhereConditions/WhereConditionsServiceProvider.php
alexey-krukover: Or maybe create a `GREATER_THAN_EQUALS` const in `SQLOperator`

src/WhereConditions/WhereHasConditionsDirective.php
alexey-krukover: Sorry if it's another silly question. But how exactly I use it?

src/WhereConditions/WhereHasConditionsDirective.php
alexey-krukover: The Static Code Analysis complains if I remove it, because the `handleWhereConditions` returns both of builders.

spawnia: You can use a `/** @var */` PHPDoc to assert the result of `handleWhereConditions` is a specific type.

src/WhereConditions/WhereConditionsServiceProvider.php
```diff
@@ -91,8 +100,50 @@ public static function createWhereConditionsInputType(string $name, string $desc
 
                 "A set of conditions that requires at least one condition to match."
                 OR: [$name!]
+
+                "A nested ord direct relation condition that create an amount criteria, and optionaly nested condition criteria."
```
spawnia, line 104: I do not understand that description, grammar seems wrong too.

src/WhereConditions/WhereHasConditionsDirective.php
spawnia: You call this function in a way that emulates what happens when being passed a GraphQL input, so you have to reconstruct this array shape. In terms of the layers of abstractions, that is moving the wrong way.

What should happen instead is:
- there is one underlying method to call from PHP
- another method converts the inputs from GraphQL to call that method

src/WhereConditions/WhereHasConditionsDirective.php
spawnia: Users are free to implement their own version of the `Operator` interface, with `SQLOperator` being just one possible implementation. The customs implementation can be bound through a service provider.

The implementation of the where directives must not assume that `SQLOperator` is used, that may not be the case. Thus, we must make no assumptions about the internal values of the operators.

alexey-krukover: > Note that this is a breaking change and will have to target the upcoming `v5` release.

Found a better alternative without it becoming a breaking change,

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -67,6 +85,43 @@ function ($builder) use ($orConnectedConditions): void {
...
 
+    /**
+     * @param QueryBuilder|EloquentBuilder $builder
+     * @param array<string, mixed>|null $condition
+     */
+    public function handleHasCondition($builder, Model $model, string $relation, ?array $condition = null, ?int $amount = null, ?string $operator = null): void
+    {
+        $additionalArguments = [];
+
+        if ($operator) {
```
spawnia, line 96: What if `$operator` is `''`?

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -67,6 +85,43 @@ function ($builder) use ($orConnectedConditions): void {
...
+            $additionalArguments[] = $operator;
+        }
+
+        if ($amount) {
```
spawnia, line 100: What if `$amount` is `0`?

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -67,6 +85,43 @@ function ($builder) use ($orConnectedConditions): void {
         return $builder;
     }
 
+    /**
+     * @param QueryBuilder|EloquentBuilder $builder
```
spawnia, line 89: Always use FQCN in PHPDoc.

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
```diff
@@ -333,6 +335,338 @@ public function testAddsNestedAndOr(): void
...
     }
 
+    public function testHasMixed(): void
+    {
+        factory(User::class, 9)->create()->each(function ($user) {
+            $user->posts()->saveMany(factory(Post::class, 2)->create());
+        });
+
+        factory(Comment::class, 1)->create([
+            'post_id' => 3,
```
spawnia, line 345: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#test-data-setup

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
```diff
@@ -333,6 +335,338 @@ public function testAddsNestedAndOr(): void
...
+                    column: "id"
+                    operator: LT
+                    value: 8
+                    AND: [{
+                        column: "id"
+                        operator: GT
+                        value: 2
+                        HAS: {
+                            relation  : "posts.comments"
+                            operator  : GT
```
spawnia, line 389: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#alignment

src/WhereConditions/WhereConditionsServiceProvider.php
```diff
@@ -91,8 +98,50 @@ public static function createWhereConditionsInputType(string $name, string $desc
 
                 "A set of conditions that requires at least one condition to match."
                 OR: [$name!]
+
+                "Check whether a relation exists, or if it matches a count criteria. And with option for a where clause."
```
spawnia, line 102:
```suggestion
                "Check whether a relation exists. Extra conditions or a minimum amount can be applied."
```

src/WhereConditions/Operator.php
```diff
@@ -20,6 +20,22 @@ public function enumDefinition(): string;
      */
     public function default(): string;
 
+    /**
+     * The default value if no has operator is specified.
+     *
+     * @example "GTE"
+     */
+    public function defaultHasOperator(): string;
```
spawnia, line 28: Those are breaking changes, so this branch will have to target the upcoming v5.

spawnia:
This is going in the right direction, looking much better now.

src/WhereConditions/Operator.php
```diff
@@ -20,6 +20,22 @@ public function enumDefinition(): string;
...
+
+    /**
+     * The value of the default has operator.
+     *
+     * TODO: Find or create an easy way to get the real enum value from the enum options
```
spawnia, line 33: This can be solved by looking at the parsed result of the entire definition (or maybe the executable enum type). In principle:

```php
\GraphQL\Type\Definition\EnumType#getValue($defaultHasOperator)
```

spawnia:
@alexey-krukover I cut the last release of v4, so we are now ready to merge this PR once finished.

src/WhereConditions/WhereConditionsBaseDirective.php
alexey-krukover: Then it is the same as false, and just uses the default operation of Laravel.

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
alexey-krukover: The ones where I make multiple creates, do I need to stick to that convention as well?

alexey-krukover: Also If I create just a large batch of comments, and just assign specific ids to other comments. Does it goes against the convention to use the  ids directly?

spawnia: We rely heavily on static analysis. Using plain arrays with string keys totally circumvents that and should never be used.

src/WhereConditions/Operator.php
spawnia: Please mention this in a short paragraph in `UPGRADE.md`.

src/WhereConditions/WhereConditionsBaseDirective.php
spawnia: Let's change this to `!== null` as well. Passing `''` would then likely cause an error, which is preferrable to silently discarding it.

spawnia:
@alexey-krukover please resolve conflicts.

spawnia:
Tests are failing because of:

> Error: Class 'Nuwave\Lighthouse\WhereConditions\PartialParser' not found

Use `GraphQL\Language\Parser` instead. Make sure to run `composer update` to have the current version of `webonyx/graphql-php`.

stephenjason89:
Are there any documentation on this available?
Thank you for all the work here

PR #1415, opened 2020-06-11
Add failing test for variables + subscriptions

spawnia:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**

### DIFF
--- a/tests/Integration/Subscriptions/SubscriptionTest.php
+++ b/tests/Integration/Subscriptions/SubscriptionTest.php
@@ -28,14 +28,24 @@ class SubscriptionTest extends TestCase
             body: String
         }
 
+        enum PostStatus {
+            PUBLISHED @enum(value: "published")
+            DELETED @enum(value: "deleted")
+        }
+
         type Subscription {
             onPostCreated: Post
+            onPostUpdated(status: PostStatus!): Post
         }
 
         type Mutation {
             createPost(post: String!): Post
                 @field(resolver: "{$this->qualifyTestResolver()}")
                 @broadcast(subscription: "onPostCreated")
+
+            updatePost(post: String!): Post
+                @field(resolver: "{$this->qualifyTestResolver()}")
+                @broadcast(subscription: "onPostUpdated")
         }
 
         type Query {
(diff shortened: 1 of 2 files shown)


spawnia:
@stayallive opened the PR to see the results of the test run.

spawnia:
That's a tricky one.

We currently store the internal value of the `ENUM` when serializing the subscription, which causes the rehydrated execution during the broadcast to fail. This might also be problematic when using custom scalar types whose internal representation might not be serializable.

I think we should try and get a hold of the raw args/variables that are passed during the initial subscription request and serialize exactly those. That will ensure the rehydrated requests function exactly the same.

stayallive:
Yeah was thinking the same, will brew a bit on this, thanks for the pointers 👍

stayallive:
That prettify docs shouldn't have ran here and or had any effect... that will mess up some PR's 😉

spawnia:
@stayallive a tricky part is that the raw request may contain literal argument values as well as variables. We don't ever see an intermediary representation where those are merged, but not cast to their internal values (e.g. Enum classes).

I will dig in to `graphql-php` and see if they expose this structure, or at least a method to build it from the raw request.

src/Subscriptions/SubscriptionBroadcaster.php
```diff
@@ -93,6 +93,7 @@ function (Subscriber $subscriber) use ($root): void {
                     $subscriber->args,
                     $subscriber->setRoot($root)
                 );
+                dump($data);
```
tiko-star, line 96: do we need this here?

tiko-star (changes requested):

src/Subscriptions/SubscriptionBroadcaster.php
stayallive: This PR is a work of progress and currently not in a working state. So no the dump is not needed in the final PR, but while in this state it's fine, it's a draft PR after all.

src/Schema/Types/LaravelEnumType.php
```diff
@@ -151,4 +151,13 @@ public function serialize($value): string
...
     }
+
+    public function parseValue($value)
+    {
+        if ($value instanceof $this->enumClass) {
+            return $value;
+        }
+
+        return parent::parseValue($value);
+    }
```
spawnia, line 162: Hey @stayallive, I think this might fix the issue at least for enums from `bensampo/laravel-enum`. I don't know what we can do about the enums modified by `@enum` tnough.

PR #1416, opened 2020-06-11
Added local scope example for the @find directive

SuddenlyRust:


docs/4.12/api-reference/directives.md
```diff
@@ -845,6 +847,35 @@ type Query {
 }
 ```
 
+#### Adding Local Scopes
```
spawnia, line 850: Since many directives allow adding scopes, this section might be a better fit in https://lighthouse-php.com/master/eloquent/getting-started.html

SuddenlyRust: You are right. I am gonna open a new PR later :)

PR #1418, opened 2020-06-11
Document how to use Model scopes from directives

SuddenlyRust:
Example added to demonstrate how to use local scopes with a query


spawnia:
Thanks, nicely done

PR #1419, opened 2020-06-11
Fix PHPStan types for AtttemptAuthentication middleware

spawnia:


spawnia:
Fixed in https://github.com/nuwave/lighthouse/pull/1426

PR #1420, opened 2020-06-11
Throw partial errors when failing to delete, forceDelete or restore a model

### LINKED ISSUE #1204: @delete directive does not return error when delete has failed through observer.
**Describe the bug**

@delete directive does not return error when delete has failed through observer.

My mutation in graphql-schema

```graphql
extend Type Mutation {
  deleteUser(id: Int! @eq): User @delete
}
```
Note: Im using `Int! @eq` instead of `ID!` as my database has id as a incrementing integer.

My observer

```php
    public function deleting(User $user)
    {
        return false;
        ...
        try {
            ...
        } catch (Exception $e) {
            ...
            return false;
        }
    }
```
Note: The first `return false` in the `deleting` method is there for testing purposes.

Expected behaviour.
1. When observer's deleting method(event-handler) returns false, delete actions fails.
2. GraphQL should consider that as a fail and return a error message.

First step is happening as in my test-case delete is not happening as expected.
But GraphQL is not considering that be a failure and is behaving as if it was a successful deletion.

**Steps to reproduce**

1.
2.
3.

**Output/Logs**

```
# Add in log output/error messages here
```

</details></br>

**Environment**

Lighthouse Version: v4.8.1
Laravel Version: v6.9.0

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1204 

**Changes**

- Throw partial errors when failing to delete, forceDelete or restore a model https://github.com/nuwave/lighthouse/pull/1420
- Add `\Nuwave\Lighthouse\Execution\ErrorPool` to allow collection of partial errors https://github.com/nuwave/lighthouse/pull/1420

**Breaking changes**

No


lorado:
I like it 👍

PR #1422, opened 2020-06-11
Only invoke default resolver

spawnia:
### Rename `resolve` to `__invoke`

Field resolver classes now only support the method name `__invoke`, using
the name `resolve` no longer works.

```diff
namespace App\GraphQL\Queries;
class SomeField
{
-   public function resolve(...
+   public function __invoke(...
```


rushairer:
Big change...

PR #1425, opened 2020-06-11
Introduce the concept of "Public" Subscriptions

jakedowns:
this feature solves a need we have on a project to not loop through all of our subscribers and broadcast to them individually, but to broadcast to them all using a single pusher event.

Usage:
1. set `public $IS_PUBLIC  = true` on your GraphQLSubscription class definition to flag that it should use a shared public channel for it's subscriptions/broadcasts
2. add a `public function getChannelName(array $args)` method must return a string name for the shared / public channel. It accepts args that are passed to the query at subscription-initialization-time, and can be used to make a semi-dynamic channel name key (in our case we group public channels by a certain model id that is passed in)

What it does:
- when **subscribing** to a "public" channel, the `extensions.lighthouse_subscriptions.channels` array will return the public/shared channel name as returned by `getChannelName` rather than the subscriber's private channel name
- when **broadcasting** on a public subscription, **SubscriptionBroadcaster** will pick up on the fact that it's a public subscription and rather than iterating over all Subscribers subscribed to the topic, it will trigger a single broadcastPublic to the Broadcaster driver using the "public" channel name.

Technical Notes:
- the Subscription Query the client sends during the initial subscription request, will be the query that is used to resolve all future broadcasts on the public channel for all subscribers. (until that subscriber is removed from Redis)
  - this might be a vulnerability, seeing as a ne'er-do-well could technically change the query for all subscribers, breaking the app, if they were able to subscribe first before the "official" primary subscription occurs, so maybe having a mirrored definition on the server's GraphQLSubscription class under "getQueryString" would help thwart that by prescribing what the query should be (which, goes against GraphQL's entire nature, so 🤷 )
  - we could also have something where when the GraphQLSubscription is booted/registered, it injects it's own "primary subscriber" into redis if it's not already there, defining the Query for the public channel, and all subscriber queries would effectively be ignored. Outside of $args which are needed to be passed to getChannelName to enable semi-dynamic channel names
  - or, a user-id or something like that on the GraphQLSubscription that defines that only a particular user is allowed to be saved as the **main** subscriber for a public channel, but then that'd require some extra end-user work to be sure to perform that initial subscription at a specific point in time
- if you try to broadcast on a public channel that hasn't had at least one subscriber, an error will be thrown `"no subscribers for public channel $topic $fieldName"`
- the first time a subscriber subscribes to a public channel, a single redis entry is created mapping the public Topic to that subscriber. Subsequent broadcast request for the public topic refer to that public_topic.subscriber for the Query,Context,Args,etc...
- when users unsubscribe, the public_topic.subscriber persists.
- clearing redis is the only way currently to "close" a public_topic/public_channel
- could add a public_channel_subscribers:<int> field in redis, and increment/decrement it accordingly, closing after last subscriber vacates/unsubscribes.

- [_**Working on getting testing setup on my local**_ ] Added or updated tests
- [Will do if there's interest] Documented user facing changes
- [Will do if there's interest] Updated CHANGELOG.md

### DIFF
--- a/src/Schema/Types/GraphQLSubscription.php
+++ b/src/Schema/Types/GraphQLSubscription.php
@@ -77,14 +77,6 @@ abstract class GraphQLSubscription
     public $IS_PUBLIC = false;
 
     /**
-     * @return ?string
-     */
-    public function getQueryString()
-    {
-        return '';
-    }
-
-    /**
      * @param array<mixed> $args
      *
      * @return ?string
--- a/src/Schema/Types/GraphQLSubscription.php
+++ b/src/Schema/Types/GraphQLSubscription.php
@@ -89,7 +89,7 @@ abstract class GraphQLSubscription
      *
      * @return ?string
      */
-    public function getChannelName(array $args)
+    public function getChannelName(?array $args = [])
     {
         return null;
     }
--- a/src/Schema/Types/GraphQLSubscription.php
+++ b/src/Schema/Types/GraphQLSubscription.php
@@ -70,7 +70,7 @@ abstract class GraphQLSubscription
     abstract public function filter(Subscriber $subscriber, $root);
 
     /**
-     * Boolean flag for public subscriptions
+     * Boolean flag for public subscriptions.
      *
      * @var bool
      */
@@ -81,7 +81,6 @@ abstract class GraphQLSubscription
      */
     public function getQueryString()
     {
-        return null;
     }
 
     /**
--- a/src/Schema/Types/GraphQLSubscription.php
+++ b/src/Schema/Types/GraphQLSubscription.php
@@ -81,6 +81,7 @@ abstract class GraphQLSubscription
      */
     public function getQueryString()
     {
+        return '';
     }
 
     /**
@@ -90,6 +91,6 @@ abstract class GraphQLSubscription
      */
     public function getChannelName(?array $args = [])
     {
-        return null;
+        return '';
     }
 }
--- a/src/Schema/Types/GraphQLSubscription.php
+++ b/src/Schema/Types/GraphQLSubscription.php
@@ -70,9 +70,26 @@ abstract class GraphQLSubscription
     abstract public function filter(Subscriber $subscriber, $root);
 
     /**
+     * Boolean flag for public subscriptions
+     *
+     * @var bool
+     */
+    public $IS_PUBLIC = false;
+
+    /**
      * @return ?string
      */
     public function getQueryString()
+    {
+        return null;
+    }
+
+    /**
+     * @param array<mixed> $args
+     *
+     * @return ?string
+     */
+    public function getChannelName(array $args)
     {
         return null;
     }
--- a/src/Schema/Types/GraphQLSubscription.php
+++ b/src/Schema/Types/GraphQLSubscription.php
@@ -68,4 +68,12 @@ abstract class GraphQLSubscription
      * @return bool
      */
     abstract public function filter(Subscriber $subscriber, $root);
+
+    /**
+     * @return ?string
+     */
+    public function getQueryString()
+    {
+        return null;
+    }
 }
--- a/src/Subscriptions/Contracts/StoresSubscriptions.php
+++ b/src/Subscriptions/Contracts/StoresSubscriptions.php
@@ -42,4 +42,18 @@ interface StoresSubscriptions
      * @return \Nuwave\Lighthouse\Subscriptions\Subscriber|null
      */
     public function deleteSubscriber(string $channel);
+
+    /**
+     * Store main / single subscriber on a public topic
+     *
+     * @return void
+     */
+    public function storeSubscriberPublic(Subscriber $subscriber, string $public_topic_name);
+
+    /**
+     * Get the one and only subscriber for a public topic, if any
+     *
+     * @return \Nuwave\Lighthouse\Subscriptions\Subscriber|null
+     */
+    public function publicSubscriberForTopic(string $channel);
 }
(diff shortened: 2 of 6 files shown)


spawnia:
How would that work from a client perspective?

jakedowns:
> How would that work from a client perspective?

So, the client still subscribes the exact same way as default 😄 !

The only thing to be aware of, is that the Subscription Query the client sends during the initial subscription request, should match the pre-defined Subscription Query string that is defined in the GraphQLSubscription class

so, for my example, i'm still using the same PusherLink/Apollo connection as before when my Subscription was using per-subscriber-private-channels.

it's up to the server to recognize that the requested subscription maps to a $IS_PUBLIC  GraphQLSubscription class and to do what it needs to do accordingly.

spawnia:
Matching based on the query string seems very brittle. At the very least, some kind of normalization should be done so that superficial differences (e.g. whitespace) don't require different queries.

Writing a separate class for every possible query also seems very tedious.

What actual problem are you trying to solve?

jakedowns:
Great point, and you're right, I should introduce something like that.

The idea is, we have a couple subscriptions where we know that all of our subscribers will always be getting the same payload, and we will have lots of users subscribing, potentially hundreds of users simultaneously.

we would rather Pusher handle the scale of the broadcast after we send it a single event, rather than relying on our Server iterating over each client channel, triggering the multitude of events.

I'm going to do some testing, but it might be possible that the client could just send the args, and an dummy Subscription that is empty, that is, it won't need to match 1:1 with the server definition. I never strictly compare them

spawnia:
I am happy to optimize Lighthouse for uses that are actually GraphQL.

If your goal is simply to broadcast the result of a static query, why do you even need GraphQL?

spawnia:
That said, i can imagine optimizations to group queries that are functionally identical into the same channel. We can look at the parsed query + variables to determine equality in a safe manner.

jakedowns:
First off, thank you for taking the time to respond and consider this PR. Responsive package maintainers are very rare. 🙏 

&, you know what, i forgot, I no longer reference getQueryString, so I will remove that. 
what it does is uses the query as passed by the first subscriber to the public channel, stored in redis.

As to _Why??_ You're right, that's the same point my co-worker made. I just figured since every other bit of our API was GraphQL, it made sense to keep it under the Apollo/GQL-Subscription Umbrella

It also allows quick flexibility, should we change that query in the future, using GQL schema, like all the other parts of our SPAs data model, rather than having it use something unique & different (Ala Echo/Laravel Broadcast, what have you) 

I really like the declarative / self-documenting nature of GQL

jakedowns:
Added an update to my original post:

> Technical Notes:
> - the Subscription Query the client sends during the initial subscription request, will be the query that is used to resolve all future broadcasts on the public channel for all subscribers. (until that subscriber is removed from Redis)
>   - this might be a vulnerability, seeing as a ne'er-do-well could technically change the query for all subscribers, breaking the apps, so maybe having a mirrored definition on the server's GraphQLSubscription class under "getQueryString" would help thwart that by prescribing what the query should be (which, goes against GraphQL's entire nature, so 🤷 )
>   - we could also have something where when the GraphQLSubscription is booted/registered, it injects it's own "primary subscriber" into redis if it's not already there, defining the Query for the public channel, and all subscriber queries would effectively be ignored. Outside of $args which are needed to be passed to getChannelName to enable semi-dynamic channel names
>   - or, a user-id or something like that on the GraphQLSubscription that defines that only a particular user is allowed to be saved as the **main** subscriber for a public channel, but then that'd require some extra end-user work to be sure to perform that initial subscription at a specific point in time

spawnia:
I recommend you familiarize yourselves with the GraphQL specification, specifically http://spec.graphql.org/draft/#sec-Subscription

A compliant GraphQL server - which Lighthouse tries to be - MUST respond to subscriptions or other operations with a proper payload. That means the returned fields match what the query asked for. Nothing more, nothing less.

Your proposed approach breaks that in a multitude of ways. Consider aliasing, the following is a valid subscription query:

```graphql
subscription Foo {
  anyClientGivenName: actualFieldName
}
```

A client can rightfully expect this response:

```json
{
  "data": {
    "anyClientGivenName": ...
  }
}
```

jakedowns:
Yeah, you're absolutely right. I think the idea of grouping identical subscription queries into the same channel is the more appropriate solution. I think I can modify my PR to follow along the lines of that idea, which would also solve the "first subscriber is a jerk" issue

spawnia:
All of the following are valid subscription queries.

```graphql
type Subscription {
  foo: String
  bar: String
}

subscription {
  foo: foo
}

subscription {
  bar: foo
}

subscription {
  foo
}

subscription {
  foo: bar
  bar
}
```

I am not sure if there is much optimization to be had, given the extremely dynamic nature of GraphQL queries. Getting this right seems very tricky.

Have you looked at the way that other implementations have optimized subscriptions?

jakedowns:
I have not, i'll do some investigation

jakedowns:
I'm gonna close this for now, thanks for taking the time to talk through it with me.

spawnia:
Thank you for getting involved and giving this a go.

Opened an issue over at https://github.com/webonyx/graphql-php to work on determining query equality, let's see if something comes out of it.

palypster:
@jakedowns have you sorted it out? We're dealing probably with the same issue here:
https://github.com/nuwave/lighthouse/issues/1651
we're working on a solution. I'm curious if you succeeded with your situation, maybe you have some good points to share. Thx

jakedowns:
> @jakedowns have you sorted it out? We're dealing probably with the same issue here:
> #1651
> we're working on a solution. I'm curious if you succeeded with your situation, maybe you have some good points to share. Thx

Hey there, we ended up just using the Pusher API directly for our particular use case, rather than using the GQL subscription mechanism

PR #1434, opened 2020-06-17
Simplify field generator

spawnia:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

In our projects, we have only used `$resolveInfo` a couply of times. Most often it is just not needed.

This simplifies the default generated field to be much less noisy and reduces it to the minimum.

**Breaking changes**

No, existing code is unaffected. The new classes still work.


src/Console/FieldGeneratorCommand.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
joskfg, line 1: Why do we need this abstract class instead of implement it in QueryCommand?

src/Console/stubs/field_simple.stub
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace DummyNamespace;
+
+class DummyClass
+{
+    /**
+     * @param  null  $_ Always `null`.
```
joskfg, line 8: Shouldn't this comment be the same than in field_full.stub?
```
 @param  null  $rootValue Usually contains the result returned from the parent field. In this case, it is always `null`.
```

src/Console/FieldGeneratorCommand.php
spawnia: Good catch. Generating mutations uses the same stub and should work the same, i forget to change the command class. 5a0f1137ac144a69c5dda6c28d4255ad9bbd5157

src/Console/stubs/field_simple.stub
spawnia: We might shorten both to:

```suggestion
     * @param  null  $_
```

joskfg: I agree

spawnia: I decided to strip all descriptions from the simple version and include them in the full version.

PR #1443, opened 2020-06-22
fix the key name for column for orderBy queries

Mecit:
posts(orderBy: [{ column: POSTED_AT, order: ASC }]) changed to posts(orderBy: [{ field: POSTED_AT, order: ASC }])

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**
No

### DIFF
--- a/docs/4.14/api-reference/directives.md
+++ b/docs/4.14/api-reference/directives.md
@@ -1975,7 +1975,7 @@ Querying a field that has an `orderBy` argument looks like this:
 
 ```graphql
 {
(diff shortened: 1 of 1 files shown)


spawnia:
I recommend to use `column`. Change the `orderBy` configuration in `lighthouse.php` to prefer for the upcoming change in v5.

Clarified the docs in https://github.com/nuwave/lighthouse/pull/1444

Mecit:
My bad, I thought it was the other way around since 'column' wouldn't work. Thanks for making the docs clearer about this.

Issue #1389, opened 2020-05-28
Schema cache not working


tushar-inno:
**Describe the bug**
When I set true for cache schema in lighthouse.php file. After updating some changes in my code when I try to deploy new code. It still uses cache schema but not updated. Not even working use of **php artisan lighthouse:clear-cache**

**Expected behavior/Solution**

```
When I run php artisan lighthouse:clear-cache it needs to clear cache and again store as cache
```

</details></br>

**Lighthouse Version**
4.13.0

spawnia:
Please list the exact steps you are doing.

tushar-inno:
Thanks for the quick reply.

Below is my lighthouse.php file:
`<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Route Configuration
    |--------------------------------------------------------------------------
    |
    | Controls the HTTP route that your GraphQL server responds to.
    | You may set `route` => false, to disable the default route
    | registration and take full control.
    |
    */

    'route' => [
        /*
         * The URI the endpoint responds to, e.g. mydomain.com/graphql.
         */
        'uri' => '/endpoint',

        /*
         * Lighthouse creates a named route for convenient URL generation and redirects.
         */
        'name' => 'endpoint',

        /*
         *
         * Beware that middleware defined here runs before the GraphQL execution phase,
         * so you have to take extra care to return spec-compliant error responses.
         * To apply middleware on a field level, use the @middleware directive.
         */
        'middleware' => [
            "api",
            \Nuwave\Lighthouse\Support\Http\Middleware\AcceptJson::class,
        ],

        /*
         * The `prefix` and `domain` configuration options are optional.
         */
        //'prefix' => '',
        //'domain' => '',
    ],

    /*
    |--------------------------------------------------------------------------
    | Schema Location
    |--------------------------------------------------------------------------
    |
    | This is a path that points to where your GraphQL schema is located
    | relative to the app path. You should define your entire GraphQL
    | schema in this file (additional files may be imported).
    |
    */

    'schema' => [
        'register' => base_path('graphql/schema.graphql'),
    ],

    /*
    |--------------------------------------------------------------------------
    | Schema Cache
    |--------------------------------------------------------------------------
    |
    | A large part of schema generation consists of parsing and AST manipulation.
    | This operation is very expensive, so it is highly recommended to enable
    | caching of the final schema to optimize performance of large schemas.
    |
    */

    'cache' => [
[...]

PR #1446, opened 2020-06-22
Add cache store option to config

tlaverdure:
- [ ] Added or updated tests
- [X] Updated CHANGELOG.md

Need to use a different cache store for AST due to cache item size limitations.

Somewhat related to https://github.com/nuwave/lighthouse/issues/1389

**Changes**

Adds an option to use a different cache store from the default. 

**Breaking changes**

N/A


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Add `@withCount` directive to eager load relationship counts on field access https://github.com/nuwave/lighthouse/pull/1390
 - Extend `lighthouse:directive` artisan command to allow choosing interfaces https://github.com/nuwave/lighthouse/pull/1251
+- Add `lighthouse.cache.store` configuration option to set the cache store to use for shema caching https://github.com/nuwave/lighthouse/pull/1446
```
spawnia, line 16:
```suggestion
- Add `lighthouse.cache.store` configuration option to set the cache store to use for schema caching https://github.com/nuwave/lighthouse/pull/1446
```

tlaverdure: 😅

src/lighthouse.php
```diff
@@ -85,6 +85,7 @@
     'cache' => [
         'enable' => env('LIGHTHOUSE_CACHE_ENABLE', env('APP_ENV') !== 'local'),
         'key' => env('LIGHTHOUSE_CACHE_KEY', 'lighthouse-schema'),
+        'store' => env('LIGHTHOUSE_CACHE_STORE', null),
```
spawnia, line 88: Can you add a small docblock to each option?

This one might be:

```suggestion

        /*
         * Allows using a specific cache store, uses the app's default if set to null.
         */
        'store' => env('LIGHTHOUSE_CACHE_STORE', null),
```

spawnia:
We might include a hint that file-based caching works well when deploying immutable container images, maybe in the docs? Our app currently has no need for other caches, but if we were to add them, i would still keep up that separation.

src/lighthouse.php
```diff
@@ -83,8 +83,25 @@
...
+         */
         'key' => env('LIGHTHOUSE_CACHE_KEY', 'lighthouse-schema'),
+
+        /**
+         * Allows using a specific cache store, uses the app's default if set to null.
+         */
+        'store' => env('LIGHTHOUSE_CACHE_STORE', null),
+
+        /**
+         * Indicates how long the schema should remain cached. A null value
```
spawnia, line 102:
```suggestion
         * Duration in seconds the schema should remain cached, null means forever.
```

src/lighthouse.php
```diff
@@ -83,8 +83,25 @@
     */
 
     'cache' => [
+        /**
```
spawnia, line 86:
```suggestion
        /*
```

Because they are not actual PHPDocs.

src/lighthouse.php
```diff
@@ -83,8 +83,25 @@
...
         'key' => env('LIGHTHOUSE_CACHE_KEY', 'lighthouse-schema'),
+
+        /*
+         * Allows using a specific cache store, uses the app's default if set to null.
+         */
+        'store' => env('LIGHTHOUSE_CACHE_STORE', null),
+
+        /**
+         * Duration in seconds the schema should remain cached, null means forever.
+         * ensures the schema is cached forever or until cleared.
```
spawnia, line 103:
```suggestion
```

tlaverdure:
Yeah adding that hint in the docs would be good.

Issue #1406, opened 2020-06-04
Authentication issue with Laravel Sanctum


junioro12:
We're having an issue using Lighthouse with Sanctum's SPA Authentication.  It seems that if we pass a referer that is in the list of SANCTUM_STATEFUL_DOMAINS we get returned a CSRF Token Mismatch even if we are attempting to access a query that doesn't require authentication.

If we are actually authenticated and pass the referer and the correct auth info we do get back our requests successfully.

If we aren't authenticated and the referer is passed, the EnsureFrontendRequestsAreStateful seems to trigger a CSRF issue...even though no authentication has been provided or is required.

If we try to authenticate and then call a normal API route that is protected by the "auth:sanctum" middleware, debugging the request shows Sanctum arrives at an unauthenticated conclusion and then the response back is correct...in the case of Lighthouse, it seems like we're being thrown out at the EnsureFrontendRequestsAreStateful middleware state even though we aren't actually attempting to get data that requires authentication, or passing any data that would indicate we are authenticated.

Switching to token based authentication works fine...and in our current development state we're fine to do that in the short term, not sure if anyone else seems to be experiencing this?

I believe what should happen is that if we are accessing data that doesn't require authentication, the data should be returned, but it doesn't seem to make it that far.

**Steps to reproduce**

1. Trying to access a query that requires no authentication throws the CSRF token mismatch if not authenticated...if we then
2. Authenticate using a React app on one of Sanctum's allowed domains, we get an authenticated session...then
3. We can access the query in #1 now that Sanctum seems to authenticate us properly

spawnia:
I would expect this to happen. Laravel middleware does not know anything about GraphQL and if fields that require authentication are requested.

The recommended approach is to use a middleware that optionally authenticates users, such as [`AttemptAuthentication`](https://github.com/nuwave/lighthouse/blob/master/src/Support/Http/Middleware/AttemptAuthentication.php).

[`EnsureFrontendRequestsAreStateful`](https://github.com/laravel/sanctum/blob/2.x/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php) seems to require authentication for any request and does not allow it to be absent. You can try and create a version of it that optionally requires authentication. If you can come up with a nice solution, feel free to add a PR to Lighthouse or post it in this issue.

dlindenkreuz:
I'm dealing with the same issue and – CSRF protection aside – solved it like this:

```php
// config/lighthouse.php
return [
  // ...
[...]

PR #1458, opened 2020-06-26
docs: clarify usage with laravel sanctum, login / logout mutations

dlindenkreuz:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

It's not obvious how to use [Laravel Sanctum](https://laravel.com/docs/7.x/sanctum) as a guard. This PR adds docs to clarify usage. Additionally, this PR covers login / logout mutations with an example, also in conjunction with Laravel Sanctum.

As discussed in #1406.

**Breaking changes**

none

### DIFF
--- a/docs/master/security/authentication.md
+++ b/docs/master/security/authentication.md
@@ -120,6 +120,24 @@ You can create or destroy a session with mutations instead of separate API endpo
 **Note that this only works when Lighthouse's guard uses a session driver.** Laravel's token based authentication
 does not allow logging in or out on the server side.
 
+This requires the following middlewares to be added to `config/Lighthouse.php`:
+
+```php
+    'route' => [
+        // ...
+        'middleware' => [
+            \Illuminate\Cookie\Middleware\EncryptCookies::class,
+            \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
+            \Illuminate\Session\Middleware\StartSession::class,
+            
+            // When using Laravel Sanctum, replace the middleware above with only:
+            // \Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::class
+            
+            // ... other middleware
+        ]             
+    ],
+```
+
 ```graphql
 type Mutation {
     login(email: String!, password: String!): User
@@ -128,7 +146,7 @@ type Mutation {
 ```
 
 ```php
-// Generate skeleton with `php artisan lighthouse:mutation login`
+// Example mutation resolver
 
 class Login
 {
@@ -154,7 +172,7 @@ class Login
 ```
 
 ```php
-// Generate skeleton with `php artisan lighthouse:mutation logout`
+// Example mutation resolver
 
 class Logout
 {
@@ -172,6 +190,6 @@ class Logout
 
 ```
 
-If you are using [Laravel Sanctum](https://laravel.com/docs/master/sanctum), you should use Sanctum's guard
+If you are using [Laravel Sanctum](https://laravel.com/docs/master/sanctum), you should use Sanctum's guard in your mutation resolvers
 to handle sessions. Sanctum supports both session-based and token-based authentication. The example implementations above
 only make sense when using session-based authentication.
--- a/docs/master/security/authentication.md
+++ b/docs/master/security/authentication.md
@@ -162,7 +162,7 @@ class Login
     public function __invoke($_, array $args): User
     {
         // Plain Laravel: Auth::guard() 
-        // Laravel sanctum: Auth::guard(config('sanctum.guard', 'web')) 
+        // Laravel Sanctum: Auth::guard(config('sanctum.guard', 'web')) 
         $guard = ?;
 
         if( ! $guard->attempt($args)) {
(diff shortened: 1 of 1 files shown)


docs/master/security/authentication.md
```diff
@@ -45,12 +45,31 @@ This setting is used whenever Lighthouse looks for an authenticated user, for ex
...
+        // ...
+        'middleware' => [
+            \Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::class,
+            // ... other middlewares
+        ]             
+    ],
     'guard' => 'sanctum',
 ```
 
+Note that Sanctum requires you to send an CSRF token as [header](https://laravel.com/docs/7.x/csrf#csrf-x-csrf-token)
```
spawnia, line 65:
```suggestion
Note that Sanctum requires you to send an CSRF token as [header](https://laravel.com/docs/csrf#csrf-x-csrf-token)
```

spawnia:
@dlindenkreuz i will try out the suggested changes in https://github.com/nuwave/lighthouse-example and review once i find the time.

ODelibalta:
Thank you for this, sincerely. 

But 

I did the changes based on the proposed doc waiting for review and I get page 419 page expired for my login mutation. It is a fresh install

ODelibalta:
I've done my login according to the updated doc. It does the login mutation fine but then the default ```User``` query returns ```null``` or if I add the ```@guard``` directive to the hello world query it says unauthenticated. 

When I inspect in my browser, the X-CSRF-COOKIE  in response headers is changed on each request.

I made a button that fires a guarded hello world query and on each click lighthouse returns a different cookie value even though in my header I am sending the first cookie value I get in login page. 

```
type Query {
    users: [User!]! @paginate(defaultCount: 10)
    user(id: ID @eq): User @find
    hello: String! @guard
    me: User @auth
}

type Mutation {
    login(email: String!, password: String!): User @field(resolver: "App\\GraphQL\\Mutations\\Auth\\Login")
}

type User {
    id: ID!
    name: String!
    email: String!
    created_at: DateTime!
    updated_at: DateTime!
}
```
login resolver
```
class Login
{
    /**
     * @param  null  $_
     */
    public function __invoke($_, array $args)
    {
        // $guard = Auth::guard();
        // When using Laravel Sanctum, always use Sanctum's guard to create a session:
        $guard = Auth::guard(config('sanctum.guard', 'web'));

        if ($guard->user()) {
            return $guard->user();
        }

        $credentials = collect($args)->only('email', 'password')->all();

        throw_unless(
            $guard->attempt($credentials),
            AuthenticationException::class
        );

        return $guard->user();
    }
}
```
cors config
```
return [ 

    'paths' => ['api/*',  'graphql', 'sanctum/csrf-cookie'],

    'allowed_methods' => ['*'],

    'allowed_origins' => ['*'],

    'allowed_origins_patterns' => [],

    'allowed_headers' => ['*'],

    'exposed_headers' => [],

    'max_age' => 0,

    'supports_credentials' => true,

];
```
client side login page
```
useEffect(() => {
    console.log(`${server}/sanctum/csrf-cookie`);
    axios.get(`${server}/sanctum/csrf-cookie`);
  }, []);

  const [loginUser, { loading }] = useMutation(LOGIN_USER, {
    errorPolicy: 'all',
    onCompleted: (res) => {
      if (res.login) {
        Router.push('/');
      } else {
        setState((prevState) => ({
          ...prevState,
          loginError: true,
        }));
      }
    },
  });
```

and in my apollo client 
```
export default function createApolloClient(initialState, ctx) {
  // The `ctx` (NextPageContext) will only be present on the server.
  // use it to extract auth headers (ctx.req) or similar.

  const authLink = setContext((_, { headers  }) => {
    // get the authentication token from local storage if it exists
    const token = Cookies.get("XSRF-TOKEN");
    // return the headers to the context so httpLink can read them
   
    return {
      headers: {
        ...headers,
        "Access-Control-Allow-Credentials": true,
        ...(token ? { "X-CSRF-TOKEN": `${token}` } : {}),
        ...(token ? { "authorization": `Bearer ${token}` } : {}),
      },
    };
  });

  const httpLink = createHttpLink({
    uri: serverUrl,
    credentials: 'same-origin',
  });

  return new ApolloClient({
    ssrMode: Boolean(ctx),
    link: authLink.concat(httpLink),
    connectToDevTools: true,
    cache: new InMemoryCache().restore(initialState),
  });
}
```

at this point I am not sure how the login works but hell world does not. I get the cookie via axios first then do login but with the same cookie I can not do a guarded hello world.

ODelibalta:
I've implemented this based on the docs here and I can login but, when I try to access a guarded query after, it says I am unauthenticated. I inspected my browser and I am sending the cookies and I have tried the same thing with graphql playground with headers added.

spawnia:
There is substantial amounts of setup that needs to be in place when starting from a bare-bones GraphQL only configuration.

Can you give me some feedback on https://github.com/nuwave/lighthouse-example/pull/9, how can that be improved?

spawnia:
Thanks @dlindenkreuz

Issue #1470, opened 2020-07-13
Complex where queries: Integrity constraint violation: 1052 Column "XXX" in where clause is ambiguous


Jofairden:
**Describe the bug**

Querying a certain field on a relation of the resulting type of the query and applying a complex where query to filter the relation causes the SQL error:
`SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'ID' in where clause is ambiguous`

**Expected behavior/Solution**

I feel like selections in the query need to be prefixed, or `addSelect` should be used (if it isn't already)

**Steps to reproduce**

1. Query some data and request a field on a relation of the data
2. Have a `hasX` complex where filter to filter the queried field, e.g. ID or name etc.

**Output/Logs**

Example error: (see example code below)
<details>      

```
"debugMessage": "SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'ID' in where clause is ambiguous (SQL: select count(*) as aggregate from `teams` where exists (select * from `players` inner join `player_team` on `player_team`.`player_id` = `players`.`id` where `teams`.`id` = `player_team`.`team_id` and (`ID` = 120)))",
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
```

</details></br>

**Lighthouse Version**
`"nuwave/lighthouse": "^4.12",`

# Example code:
Query:
(Requesting teams of players, request players of the teams (paginated) and filter the players relation for an ID)
```query {
  teams(
    hasPlayers:
    {
      column: ID value:120
    }
  ) {
    data {
      id
      players {
        data {
          id
        }
      }
    }
  }
}
```

Example GraphQL query file:
```
extend type Query @guard(with: ["api"]) {
    teams(
        hasPlayers: _ @whereHasConditions(columnsEnum: PlayerColumn)      
    ): [Team!]
    @paginate
```

Team type:
```
type Team {
    "The ID of this team"
    id: ID!
    "Players of this team"
    players: [Player!] @hasMany
}
```

Playercolumn example:
```
num PlayerColumn {
    ID
}
```

spawnia:
Please add a PR with a failing test case in https://github.com/nuwave/lighthouse/blob/master/tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php in order to investigate this further.

By the way (not invalidating your issue), but have you considered using the following query instead:

```graphql
{
  player(id: 120) {
    teams {
      players { ... }
    }
  }
}
```

[...]

PR #1472, opened 2020-07-14
Add testConditionOnQueriedField for WhereHasConditionsDirective

Jofairden:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

See #1470 

**Changes**

Added a test for this issue

**Breaking changes**
None

**Details**
I cannot seem to run any tests at all, I just keep getting a bunch of errors. Care to tell me how the tests can be setup locally?

### DIFF
--- a/tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
+++ b/tests/Integration/WhereConditions/WhereHasConditionsDirectiveTest.php
@@ -4,6 +4,7 @@ namespace Tests\Integration\WhereConditions;
 
 use Nuwave\Lighthouse\WhereConditions\WhereConditionsServiceProvider;
 use Tests\DBTestCase;
+use Tests\Utils\Models\Post;
 use Tests\Utils\Models\User;
 
 class WhereHasConditionsDirectiveTest extends DBTestCase
@@ -168,6 +169,31 @@ class WhereHasConditionsDirectiveTest extends DBTestCase
                     ],
                 ],
             ],
+        ]);
+    }
+
+    public function testConditionOnQueriedField() : void {
+        factory(Post::class)->create();
+
+        $this->graphQL(/** @lang GraphQL */'
+        {
+            posts(
+                hasUser: {
+                    column: "id"
+                    value: 1
+                }
+            ) {
+                id
+            }
+        }
+        ')->assertExactJson([
+            'data' => [
+                'posts' => [
+                    [
+                        'id' => '1'
+                    ]
+                ]
+            ]
         ]);
     }
 }


spawnia:
> I cannot seem to run any tests at all, I just keep getting a bunch of errors. Care to tell me how the tests can be setup locally?

https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#how-to-contribute-to-lighthouse

Jofairden:
Ok well I seem to have identified where the issue comes from, also explaining why there's no issue right now in Lighthouse after adding the test.

We are coupling players to teams with an intermittent table called player_team, the resulting query when using the whereCondition looks like:
```sql
select count(*) as aggregate 
from `teams` 
where EXISTS (
	select * from `players` 
	inner join `player_team` 
	on `player_team`.`player_id` = `players`.`id` 
	where `teams`.`id` = `player_team`.`team_id` 
	AND (`ID`  = 120)
)
```
Since it's literally just taking the enum value as the column for the condition we get an ambiguous error, because what it should be checking is `AND (players.id = 120)`. I tried prefixing the column but the engine doesn't allow that, it does seem that this should be supported by default though. If I prefix the final AND clause myself and run it in my SQL editor it works fine.

spawnia:
@Jofairden so you are not using joins instead of plain Eloquent relations? Automatically prefixing column names when joins are used is a long standing issue within Laravel.

I got my hands dirty and tried to fix this before, but found that the rabbit hole goes quite deep: https://github.com/spawnia/laravel-framework/tree/prefer-original-table-columns-on-join

Jofairden:
> 
> 
> @Jofairden so you are not using joins instead of plain Eloquent relations? Automatically prefixing column names when joins are used is a long standing issue within Laravel.
> 
> I got my hands dirty and tried to fix this before, but found that the rabbit hole goes quite deep: https://github.com/spawnia/laravel-framework/tree/prefer-original-table-columns-on-join

The relationship that's being queried here is a `hasManyThrough` relationship, specifically if we look at players on team:
```php
    public function players()
    {
        return $this->hasManyThrough(
            Player::class,
            PlayerTeam::class,
            "team_id",
            "id",
            "id",
            "player_id");
    }
```
Anyways, so are you saying this is more or less an issue in Laravel and less so in Lighthouse / our app ? That'd be quite an issue for us

spawnia:
@Jofairden you can try and replicate this issue without Lighthouse using the query builder yourself. I believe that this would produce a similar error:

```php
$team->players()->where('id', 123)
```

Jofairden:
> ```
> $team->players()->where('id', 123)
> ```

Yup I can confirm this now:
```
=> Illuminate\Database\Eloquent\Relations\HasManyThrough {#4223}
>>> Team::query()->find(1)->players()->where('id', 1)->first()
Illuminate/Database/QueryException with message 'SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select `players`.*, `player_team`.`team_id` as `laravel_through_key` from `players` inner join `player_team` on `player_team`.`player_id` = `players`.`id` where `player_team`.`team_id` = 1 and `id` = 1 limit 1)'
```

Jofairden:
Having the same for a polymorphic relation:
```
    public function players()
    {
        return $this->morphedByMany(Player::class,
            'participant',
            'participants',
            'fixture_id')
            ->withPivot('location');
    }
```

When I prefix the first one it works, but the filters aren't doing this:
```Team::query()->find(1)->players()->where(players.id', 1)->first()```

Any ideas on that?

PR #1511, opened 2020-08-05
Remove deprecations

spawnia:
**Breaking changes**

Loads.


src/Subscriptions/Subscriber.php
```diff
@@ -38,14 +37,14 @@ class Subscriber implements Serializable
     public $query;
 
     /**
-     * The name of the queried operation.
+     * The name of the queried field.
      *
-     * TODO replace with subscription field name, is guaranteed be be unique because of
```
spawnia, line 43: @stayallive do you think that basing subscriptions on the field name makes sense?

stayallive: Yeah, since the client can send any field name they want (so many operation names could reference the same subscription), it seems more logical to use the developer defined field name which should be coming from the GraphQL schema.

I am thinking but I'm not sure why you'd want the operation name in the Subscriber. Although looking at the code this was probably done to fix the missing validation in graphql-php which should be all correct now so we don't need to do our own checks.

spawnia: Thanks for your input.

The more I think about it, it makes much more sense to base channels on the field name, since it allows grouping them more efficiently. As you said, the operation name can be set by clients, and is not guaranteed to even match the queried fields.

PR #1466, opened 2020-07-09
Fix @count on polymorphic relations

tlaverdure:
- [X] Added or updated tests
- [ ] Documented user facing changes
- [X] Updated CHANGELOG.md

When using `@count` on a model that morphsTo different models, the batch loader may confuse keys between the different models.

Related: https://github.com/laravel/framework/pull/25997#issue-221049454

**Changes**

Separate models that need to be counted by their class name before attempting to load counts.

**Breaking changes**

N/A

PR #1517, opened 2020-08-08
Fix polymorphic key collisions when using @with

tlaverdure:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Similar to https://github.com/nuwave/lighthouse/pull/1466

**Changes**

Avoid polymorphic key collisions when using @with

**Breaking changes**
N/A


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
-
+- Ensure the `@with` directive works properly with polymorphic relations https://github.com/nuwave/lighthouse/pull/1517
```
spawnia, line 11: Can we still target this fix towards 4.x, too?

src/Execution/DataLoader/RelationBatchLoader.php
```diff
@@ -55,12 +56,12 @@ public function resolve(): array
 
         if ($this->paginationArgs !== null) {
             $modelRelationFetcher = new ModelRelationFetcher(
-                $this->getParentModels(),
+                $this->getParentModels(false),
```
spawnia, line 59: Without reading further, I have no idea what `false` does. Maybe use a separate function instead.

src/Execution/DataLoader/RelationBatchLoader.php
```diff
@@ -77,17 +78,52 @@ function (Model $model): array {
 
     /**
      * Get the parents from the keys that are present on the BatchLoader.
+     *
+     * @return EloquentCollection<mixed>
```
spawnia, line 82: FQCN in PHPDocs.

tests/Integration/Schema/Directives/WithDirectiveTest.php
```diff
@@ -95,4 +97,142 @@ public function testEagerLoadsNestedRelation(): void
...
+        $task->activity()->save($activity3);
+
+        $task->images()
+            ->saveMany(
+                factory(Image::class, 4)->make()
+            );
+
+        $this->graphQL(
+        /** @lang GraphQL */
+        '
```
spawnia, line 175:
```suggestion
        $this->graphQL(/** @lang GraphQL */ '
```

src/Execution/DataLoader/RelationBatchLoader.php
```diff
@@ -77,17 +78,52 @@ function (Model $model): array {
...
      */
-    protected function getParentModels(): EloquentCollection
+    protected function getParentModels(bool $shouldLoadRelation = true): EloquentCollection
```
spawnia, line 84: This is pretty much duplicated from the `RelationCountBatchLoader`, right? Let's dedupe so we don't miss bugfixes when changing either one of them.

This could be a `public static function` on a utility class, right?

CHANGELOG.md
tlaverdure: Let me know if I placed it in the right place now.

spawnia: Looks better now. As this is a bug fix, do you think we can target 4.x with it?

tlaverdure: Yes I think we should.

src/Execution/DataLoader/RelationBatchLoader.php
tlaverdure: Updated. Should clean things up. Thanks for the suggestion.

src/Execution/DataLoader/RelationBatchLoader.php
tlaverdure: Should be clearer now.

PR #1522, opened 2020-08-10
docs: correct the lighthouse-schema command

danielroe:
This simply corrects the command in the docs to work.


spawnia:
Good catch, thank you

PR #1527, opened 2020-08-12
Fixed the link corresponding to the MorphMany relationship

crThiago:
**Changes**

changed **#one-to-one-polymorphic-relations** to **#manyto-one-polymorphic-relations**


spawnia:
Great catch, thanks for taking the time to fix all docs versions :)

PR #1530, opened 2020-08-14
Prefix complex conditions with table name to avoid ambiguous SQL

daniel-de-wit:
Relates to: #1470

Test whereHas on belongsToMany relationship results in QueryException.

```
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select * from `users` where (exists (select * from `roles` inner join `role_user` on `roles`.`id` = `role_user`.`role_id` where `users`.`id` = `role_user`.`user_id` and `id` = 1)))
```

- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

### DIFF
--- a/src/WhereConditions/WhereConditionsBaseDirective.php
+++ b/src/WhereConditions/WhereConditionsBaseDirective.php
@@ -115,10 +115,15 @@ abstract class WhereConditionsBaseDirective extends BaseDirective implements Arg
                     $relation,
                     function ($builder) use ($relation, $model, $condition): void {
                         if ($condition) {
+                            $relatedModel = $this->nestedRelatedModel($model, $relation);
+
                             $this->handleWhereConditions(
                                 $builder,
-                                $condition,
-                                $this->nestedRelatedModel($model, $relation)
+                                $this->prefixConditionWithTableName(
+                                    $condition,
+                                    $relatedModel
+                                ),
+                                $relatedModel
                             );
                         }
                     },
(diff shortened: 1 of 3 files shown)


spawnia:
Thanks for adding a reproducible example, will investigate later. If you can come up with a solution or some direction on where we could go, that would be even better.

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -174,7 +174,7 @@ public function manipulateArgDefinition(
      */
     protected static function assertValidColumnName(string $column): void
     {
-        $match = \Safe\preg_match('/^(?![0-9])[A-Za-z0-9_-]*$/', $column);
+        $match = \Safe\preg_match('/^(?![0-9])[.A-Za-z0-9_-]*$/', $column);
```
daniel-de-wit, line 177: In order to prefix the table name, a `.` character should be allowed

spawnia:
Nested `HAS` on the `@whereConditions` should have the same issue, I think the solution can be extended to fix that too.

spawnia:
This fix will be included in the next major release v5. If you need quick access to it in a stable release, you can send a backported fix towards https://github.com/nuwave/lighthouse/tree/4.x.

I tried cherry-picking the changes, but there are some non-trivial changes within the where conditions directives which enable this change in the first place. I did not pursue it further.

Issue #1535, opened 2020-08-19
[V5] @globalId not working with @can


hosmelq:
**Describe the bug**

In the following query I get this error `No query results for model [App\Models\Student] U3R1ZGVudDoy` located in `CanDirective.php`, this was working this morning before update the package with the latest changes.

If I remove the @can directive works well.

```gql
student(id: ID! @globalId(decode: "ID") @eq): Student
    @can(ability: "view", find: "id")
    @find
```

**Expected behavior/Solution**

Decode the ID field before pass it to @can

**Lighthouse Version**

dev-master

PR #1536, opened 2020-08-20
Fix middleware directive order

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/1535


spawnia:
@hosmelq this should restore the previous execution order. Please review.

hosmelq:
Everything working beautifully @spawnia, thank you so much. And thank you for all the effort in V5 will be a great release.

PR #1539, opened 2020-08-24
Remove `@deprecated` directive class

spawnia:
Not needed since it is a built-in directive.


spawnia:
This does not really improve Lighthouse or make it easier.

PR #1543, opened 2020-08-28
Fix typo in upgrade guide from @middlware to @middleware

jonwaldstein:
**Changes**

fixed typo in `@middlware` to `@middleware`


spawnia (approved):
Thanks, good catch

PR #1562, opened 2020-09-13
Change column to field.

TheROPFather:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

When querying an order by directive the documentation specifies you to use `column:`, however `field:` should be used instead.

Using `column:` results in the following error. 

```
Field "column" is not defined by type FooOrderByOrderByClause.
```

**Breaking changes**

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -1777,7 +1777,7 @@ Querying a field that has an `orderBy` argument looks like this:
 
 ```graphql
 {
-  posts(orderBy: [{ column: POSTED_AT, order: ASC }]) {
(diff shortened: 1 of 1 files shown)


spawnia:
Thanks for the PR. The docs you are trying to change are correct though. In previous versions, it was called `field` first - then it become configurable - the upcoming v5 will have this argument named `column`.

https://github.com/nuwave/lighthouse/pull/1337

TheROPFather:
My apologies, but I may be confused - I'm on v4.16 and the docs also reference it as being column there, however this does not work for me? Am I missing something that is documented elsewhere?

PR #1575, opened 2020-09-25
prevents duplicated global scope binding

eriksulymosi:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

remove global scopes when build whereHasConditions query to prevent duplicated bindings

**Example**

schema.graphql
```graphql
type Query {
    contents(
        relationOne: _ @whereHasConditions(
            relation: "relationOne"
            columns: ["field"]
        )
        relationTwo: _ @whereHasConditions(
            relation: "relationTwo"
            columns: ["field"]
        )
    ): [Content!]! @paginate
}
```

Model.php
```php
class Content extends Model
{
    protected static function booted()
    {
        static::addGlobalScope('age', function (Builder $builder) {
            $builder->where('age', '>', 200);
        });
    }

    public function relationOne(): Relation // e.g: HasMany
    {
        // ...
    }

    public function relationOne(): Relation // e.g: HasMany
    {
        // ...
    }
}
```

GraphQL query
```
query {
  contents(
    relationOne: {
      column: FIELD
      value: "field-value"
    }
    relationOne: {
      column: FIELD
      value: "field-value"
    }
  ) {
    fields...
  }
}
```

### DIFF
--- a/src/WhereConditions/WhereConditionsBaseDirective.php
+++ b/src/WhereConditions/WhereConditionsBaseDirective.php
@@ -111,7 +111,7 @@ abstract class WhereConditionsBaseDirective extends BaseDirective implements Arg
 
         $builder->whereNested(
             function ($builder) use ($model, $relation, $condition, $additionalArguments): void {
-                $whereHasQuery = $model->whereHas(
+                $whereHasQuery = $model->newQueryWithoutScopes()->whereHas(
                     $relation,
                     function ($builder) use ($relation, $model, $condition): void {
                         if ($condition) {


spawnia:
That seems to defeat the purpose of global scopes. They are supposed to be always on, barring cases where they are explicitly excluded.

eriksulymosi:
If there are multiple "whereHasConditions" in a query at once and the global scope adds "where" conditions (eg `where('published_at', '>=', now())`) to the builders, then merges this query into the original query, you can slip the bingings array.

spawnia:
@eriksulymosi can you provide a test case that fails with the current code? That way we can make sure to solve an actual problem and also avoid regressions in the future.

eriksulymosi:
I started doing it, but in the meantime I realized that for some reason two binding are swapped out.

PR #1577, opened 2020-10-05
fix namespace directive definition

hosmelq:
- [x] Updated CHANGELOG.md

**Changes**

Add the `field` argument to the `@ namespace` directive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 ### Fixed
 
 - Prefix complex conditions with table name to avoid ambiguous SQL https://github.com/nuwave/lighthouse/pull/1530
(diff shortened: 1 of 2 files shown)


spawnia:
Thanks for the PR, but I will have to reject it.

As the description says, the arguments of this directive are a map from directive names to namespaces. Describing such a structure in terms of SDL is not possible, we would have to list all possible directives. Which again, is not possible because there might be user-defined directives.

hosmelq:
This is mostly for IDE support that throws an error.

CleanShot 2020-10-09 at 10 51 31

spawnia:
@hosmelq I got that, but having just `field` is just not accurate. Other argument names can be used.

An overall more appropriate definition would be this, but it is a breaking change:

```
directive @namespace(mappings: DirectiveNamespaceMapping)

type DirectiveNamespaceMapping {
  directive: String!
  namespace: String!
}
```

PR #1581, opened 2020-10-12
Fix typo in @can docs

SuddenlyRust:
Otherwise PhpStorm complains `Argument value is not a valid value of scalar "Boolean".`


spawnia:
Thank you, nice find

PR #1592, opened 2020-10-20
Set `lighthouse.debug` config through env `LIGHTHOUSE_DEBUG'

EvilFreelancer:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

Hello Lighthouse team! This PR is not related to any issues.

**Changes**

Hardcoded `debug` option in `lighthouse.php` config has been replaced to `env('LIGHTHOUSE_DEBUG')` with default value: `\GraphQL\Error\DebugFlag::INCLUDE_DEBUG_MESSAGE | \GraphQL\Error\DebugFlag::INCLUDE_TRACE`

This will allow changing the logging mode through environment variables without changing the parameters of the configuration file.

**Breaking changes**

There is no breaking changes, just a small tune for usability in k8s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Change `ErrorHandler` method `handle()` to non-static `__invoke()` and allow discarding
   errors by returning `null`
 - Allow subscriptions without named operations, base channels on the field name
+- Hardcoded `debug` option in `lighthouse.php` config has been replaced to `env('LIGHTHOUSE_DEBUG')` 
 
 ### Removed
 
--- a/src/lighthouse.php
+++ b/src/lighthouse.php
@@ -177,6 +177,26 @@ return [
     | Control the debug level as described in http://webonyx.github.io/graphql-php/error-handling/
     | Debugging is only applied if the global Laravel debug config is set to true.
     |
+    | Instead of using the bitwise "OR" operator https://www.php.net/manual/en/language.operators.bitwise.php
+    | you can specify an integer inside LIGHTHOUSE_DEBUG environment variable.
+    |
+    |    0 => INCLUDE_NONE
+    |    1 => INCLUDE_DEBUG_MESSAGE
+    |    2 => INCLUDE_TRACE
+    |    3 => INCLUDE_TRACE | INCLUDE_DEBUG_MESSAGE
+    |    4 => RETHROW_INTERNAL_EXCEPTIONS
+    |    5 => RETHROW_INTERNAL_EXCEPTIONS | INCLUDE_DEBUG_MESSAGE
+    |    6 => RETHROW_INTERNAL_EXCEPTIONS | INCLUDE_TRACE
+    |    7 => RETHROW_INTERNAL_EXCEPTIONS | INCLUDE_TRACE | INCLUDE_DEBUG_MESSAGE
+    |    8 => RETHROW_UNSAFE_EXCEPTIONS
+    |    9 => RETHROW_UNSAFE_EXCEPTIONS | INCLUDE_DEBUG_MESSAGE
+    |   10 => RETHROW_UNSAFE_EXCEPTIONS | INCLUDE_TRACE
+    |   11 => RETHROW_UNSAFE_EXCEPTIONS | INCLUDE_TRACE | INCLUDE_DEBUG_MESSAGE
+    |   12 => RETHROW_UNSAFE_EXCEPTIONS | RETHROW_INTERNAL_EXCEPTIONS
+    |   13 => RETHROW_UNSAFE_EXCEPTIONS | RETHROW_INTERNAL_EXCEPTIONS | INCLUDE_DEBUG_MESSAGE
+    |   14 => RETHROW_UNSAFE_EXCEPTIONS | RETHROW_INTERNAL_EXCEPTIONS | INCLUDE_TRACE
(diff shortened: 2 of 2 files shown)


spawnia:
Given the magic nature of the bitwise-OR connected flags, an explaining comment in the section above seems fitting.

EvilFreelancer:
@spawnia hello! You mean need to add somewhere information about bitmask OR operator?

Btw, i've checked documentation and did not found block with explanation of ENV parameters (in past confuration.md was for v4.0), in any other cases I would add information about this.

spawnia:
The `config.php` is supposed to be self-documenting, that is why there is no external documentation about it. Each configuration option has a block of description above it.

The usage of those flags is already a bit tricky, but even moreso when extracting to an ENV variable. It would have to contain essentially a magic number, and be cast to an integer before usage, too.

EvilFreelancer:
@spawnia list of all possible combinations added to description of debug option.

EvilFreelancer:
Simple checker.

```php
<?php

use GraphQL\Error\DebugFlag;

require_once __DIR__ . '/vendor/autoload.php';

echo (DebugFlag::NONE) . PHP_EOL;
echo (DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::INCLUDE_TRACE) . PHP_EOL;
echo (DebugFlag::INCLUDE_TRACE | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::RETHROW_INTERNAL_EXCEPTIONS) . PHP_EOL;
echo (DebugFlag::RETHROW_INTERNAL_EXCEPTIONS | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::RETHROW_INTERNAL_EXCEPTIONS | DebugFlag::INCLUDE_TRACE) . PHP_EOL;
echo (DebugFlag::RETHROW_INTERNAL_EXCEPTIONS | DebugFlag::INCLUDE_TRACE | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::INCLUDE_TRACE) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::INCLUDE_TRACE | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::RETHROW_INTERNAL_EXCEPTIONS) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::RETHROW_INTERNAL_EXCEPTIONS | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::RETHROW_INTERNAL_EXCEPTIONS | DebugFlag::INCLUDE_TRACE) . PHP_EOL;
echo (DebugFlag::RETHROW_UNSAFE_EXCEPTIONS | DebugFlag::RETHROW_INTERNAL_EXCEPTIONS | DebugFlag::INCLUDE_TRACE | DebugFlag::INCLUDE_DEBUG_MESSAGE) . PHP_EOL;
```

spawnia:
Is the value of the env variable actually an integer or rather `'3'`?

EvilFreelancer:
Is more like 0011 because we need execute bitwise operation.

0001 | 0010 = 0011 (3)

EvilFreelancer:
Plus here as you can see used integers https://github.com/webonyx/graphql-php/blob/master/src/Error/DebugFlag.php

EvilFreelancer:
Good explanation of how hex system is works https://www.electronics-tutorials.ws/binary/bin_3.html

spawnia:
The issue is that the value of `env()` is always a string. You cannot do bitwise operations with a string.

EvilFreelancer:
@spawnia hello! Yes, but it will be converted to int in `toArray` method
```
    // src/GraphQL.php
    //
    public function applyDebugSettings(ExecutionResult $result): array
    {
        // If debugging is set to false globally, do not add GraphQL specific
        // debugging info either. If it is true, then we fetch the debug
        // level from the Lighthouse configuration.
        return $result->toArray(
            config('app.debug')
                ? config('lighthouse.debug')
                : false
        );
    }
```
of `src/Executor/ExecutionResult.php` class
```
public function toArray(int $debug = DebugFlag::NONE) : array
```

EvilFreelancer:
It easy to check, just need change in config debug string to `'debug' => env('LIGHTHOUSE_DEBUG'),` then add to `.env` string like `LIGHTHOUSE_DEBUG=3`, where 3 is `DebugFlag::INCLUDE_TRACE | DebugFlag::INCLUDE_DEBUG_MESSAGE`.

EvilFreelancer:
By the way, I just noticed that the library passes false if debugging is disabled and nevertheless it works, and all because if convert to integer a boolean value false, you get 0, that is INCLUDE_NONE.

spawnia:
Thanks @EvilFreelancer

EvilFreelancer:
Cool, was glad to help :)

PR #1593, opened 2020-10-22
Fix docs: Laravel's function is `syncWithoutDetaching`

alireza0x0000001:
Laravel's function is syncWithoutDetaching but it is written syncWithoutDetach in current documentations for nested mutations.

Changed syncWithoutDetach function mentioned in docs to syncWithoutDetaching.


spawnia:
Thanks, great catch.

Issue #1594, opened 2020-10-22
@morphTo doesn't work when relationship method/name differ


FreekVR:
When using `@morphTo` on a relationship where the method name in the model differs from the actual relationship name, the relationship fails to load (returns `null`). Using a matching method/relationship name or removing the `@morphTo` directive resolves the issue.

**Steps to reproduce**

1. Create a Comments table as follows:
```
comments
    id - integer
    body - text
    commentable_id - integer
    commentable_type - string
```
2. Create a Comment model with the following relationship methods:
```php
    /**
     * This loads fine in tinker/elsewhere in Laravel but NOT in lighthouse
     * @return MorphTo
     */
    public function parent(): MorphTo
    {
        return $this->morphTo('commentable');
    }

    /**
    *  This works in lighthouse
     * @return MorphTo
     */
    public function commentable(): MorphTo
    {
        return $this->morphTo();
    }
```
3. Create the following schema for Comments:
```graphql
union CommentableType = Update | Organisation
type Query {
    comments(
        commentable_id: Int @eq,
        commentable_type: AvailableCommentableTypes @eq
        orderBy: _ @orderBy(columns: ["created_at", "updated_at"])
    ): Comment @paginate
}

type Comment {
    id: ID!
    content: String!
    author: User @belongsTo
    created_at: DateTimeTz
    updated_at: DateTimeTz
    parent: CommentableType @morphTo
    commentable: CommentableType @morphTo
    children: [Comment]! @morphMany
}
```
4. Query:
```graphql
{
  comments(first: 250) {
    data {
      id,
      parent {
        __typename
        ... on Update {
          id
        }
      }
      commentable {
        __typename
        ... on Update {
          id
        }
      }
    }
  }
}
```

5. Results in an empty "parent" field but filled "commentable", expect them to be the same:

```json
{
  "data": {
    "comments": {
      "data": [
        {
          "id": "1",
          "parent": null, // should be the same as commentable here
          "commentable": {
            "__typename": "Update",
            "id": "1864"
          }
        },

```

**Lighthouse Version**
4.17.0

spawnia:
Can you provide a PR with a failing test case?

FreekVR:
Done! You can further verify the issue by removing the "@morphTo" directive from the new test case, in which case the test case will pass.

enzonotario:
It seems to be an issue related to larave. See https://github.com/laravel/framework/issues/24787#issuecomment-403871817

[...]

PR #1595, opened 2020-10-22
Add a failing test case for custom relationship method names (#1594)

FreekVR:
- [x]  Added or updated tests
- [ ] Updated CHANGELOG.md

Failing test case as per request in  #1594

**Changes**

Adds a failing test case to prove `@morphTo` directive breaks returned values for "MorphTo"-relationships which use an different method and morph/relationship name in the model.

**Breaking changes**

None

### DIFF
--- a/tests/Integration/Schema/Directives/MorphToDirectiveTest.php
+++ b/tests/Integration/Schema/Directives/MorphToDirectiveTest.php
@@ -135,6 +135,57 @@ class MorphToDirectiveTest extends DBTestCase
         ]);
     }
 
+    public function testCanResolveMorphToWithAlternateMethodName(): void
+    {
+        $this->schema =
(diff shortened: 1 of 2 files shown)


spawnia:
This is a known issue in Laravel, see https://github.com/laravel/framework/issues/24787#issuecomment-403871817. Lighthouse just uses good old `EloquentCollection::load()` here.

```php
$parents->load([$relationName => $this->decorateBuilder]);
dump($parents->first());

Tests\Utils\Models\Image #3146
   ...
   relations: array (2)
   |  'imageableAlias' => null
   |  'imageable' => Tests\Utils\Models\Task #3329
```

Redefining the relation as suggested in the above thread leads to the intended result:

```php
return $this->morphTo(null, 'imageable_type', 'imageable_id');
```

PR #1601, opened 2020-10-30
Use addNestedWhereQuery instead of "manually" merge

eriksulymosi:
**Changes**

use addNestedWhereQuery instead of "manually" merge

**Breaking changes**

no breaking changes


spawnia:
Can you take a look at the PHPStan error? I think your code is correct, but you may have to do some extra work to convince PHPStan of it.

eriksulymosi:
i'll check it

eriksulymosi:
what do you think, might be a bug in larastan? :confused:

spawnia:
If the definitions were more accurate, this could be avoidable. Until then - given you want this merged - we can work around the issue by extracting into variables and asserting the correct type through an annotation.

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -109,29 +109,26 @@ public function handleHasCondition(
...
-                        if ($condition) {
-                            $relatedModel = $this->nestedRelatedModel($model, $relation);
-
-                            $this->handleWhereConditions(
-                                $builder,
-                                $this->prefixConditionWithTableName(
-                                    $condition,
-                                    $relatedModel
-                                ),
+        /**@phpstan-ignore-next-line */
```
spawnia, line 112:
```suggestion
        // @phpstan-ignore-next-line
```

spawnia:
Thank you, nicely done

PR #1605, opened 2020-10-31
Remove unnecessary check

enzonotario:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Larastan was complaining about that. It's not necessary since, if `$resolverClass` is not null, it goes to a return statement.

**Changes**

None.

**Breaking changes**

None.


spawnia:
Nicely done, thanks

PR #1615, opened 2020-11-19
Test helper `multipartGraphQL` now accepts arrays instead of JSON strings

jandrodev:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

When it comes to testing mutations with files using the multipartGraphQL method, it is quite tedious to work with json (especially when formatting and viewing the different operations and map blocks). That is why it becomes easier to work with arrays. Under the hood, the operation is the same, but at the time of writing the test it changes a bit to, in my opinion, make it more understandable.

I add the base method and 3 tests to check its operation in which you can see the idea that I comment. If you like, I would need to add the corresponding part in the documentation.

No breaking changes.

### DIFF
--- a/src/Testing/MakesGraphQLRequests.php
+++ b/src/Testing/MakesGraphQLRequests.php
@@ -79,40 +79,13 @@ trait MakesGraphQLRequests
      * This is used for file uploads conforming to the specification:
      * https://github.com/jaydenseric/graphql-multipart-request-spec
      *
-     * @param  array<string, mixed>  $parameters
+     * @param  array<string, mixed>|array<int, array<string, mixed>> $operations
+     * @param  array<string, string> $map
      * @param  array<int, \Illuminate\Http\Testing\File>  $files
      * @param  array<string, string>  $headers  Will be merged with Content-Type: multipart/form-data
      * @return \Illuminate\Testing\TestResponse
      */
-    protected function multipartGraphQL(array $parameters, array $files, array $headers = [])
-    {
-        return $this->call(
-            'POST',
-            $this->graphQLEndpointUrl(),
-            $parameters,
-            [],
-            $files,
-            $this->transformHeadersToServerVars(array_merge(
-                [
-                    'Content-Type' => 'multipart/form-data',
-                ],
-                $headers
-            ))
-        );
-    }
-
-    /**
-     * Send a multipart form request to GraphQL in array format. For mutations that are
-     * very large or contain many files, it becomes much easier to work with array instead of json.
-     *
-     * @param array<string, mixed> $operations
-     * @param array<int, string> $map
-     * @param array<int, \Illuminate\Http\Testing\File> $files
-     * @param array<string, string> $headers Will be merged with Content-Type: multipart/form-data
-     *
-     * @return \Illuminate\Testing\TestResponse
-     */
-    public function multipartArrayGraphQL(array $operations, array $map, array $files, array $headers = [])
+    protected function multipartGraphQL(array $operations, array $map, array $files, array $headers = [])
     {
         $data = [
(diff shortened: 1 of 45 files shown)


src/Testing/MakesGraphQLRequests.php
```diff
@@ -101,6 +101,39 @@ protected function multipartGraphQL(array $parameters, array $files, array $head
         );
     }
 
+    /**
+     * Send a multipart form request to GraphQL in array format. For mutations that are
+     * very large or contain many files, it becomes much easier to work with array instead of json.
```
spawnia, line 106:
```suggestion
     * Send a multipart form request to GraphQL.
```

I agree with what is written, but prefer to let the usefulness of the method speak for itself ;)

An actual usage example would be more useful.

src/Testing/MakesGraphQLRequests.php
```diff
@@ -101,6 +101,39 @@ protected function multipartGraphQL(array $parameters, array $files, array $head
...
+     *
+     * @param array<string, mixed> $operations
+     * @param array<int, string> $map
```
spawnia, line 109:
```suggestion
     * @param array<string, string> $map
```

src/Testing/MakesGraphQLRequests.php
```diff
@@ -101,6 +101,39 @@ protected function multipartGraphQL(array $parameters, array $files, array $head
         );
     }
 
+    /**
+     * Send a multipart form request to GraphQL in array format. For mutations that are
+     * very large or contain many files, it becomes much easier to work with array instead of json.
+     *
+     * @param array<string, mixed> $operations
```
spawnia, line 108:
```suggestion
     * @param array<string, mixed>|array<int, array<string, mixed>> $operations
```

spawnia (changes requested):
I don't want to have two methods to do the same thing. Yours seems better - working with arrays is indeed more pleasant than JSON - so it should replace the previous one.

Given we are working towards a new major version, such breakage is fine. You can prove that your method holds up by converting existing uses of the old multipart method.

jandrodev:
Perfect, so I modify the old method with the new idea.

src/Testing/MakesGraphQLRequests.php
```diff
@@ -79,9 +79,9 @@ protected function postGraphQL(array $data, array $headers = [])
      * This is used for file uploads conforming to the specification:
      * https://github.com/jaydenseric/graphql-multipart-request-spec
      *
-     * @param  array<string, mixed>|array<int, array<string, mixed>> $operations
-     * @param  array<string, string> $map
-     * @param  array<int, \Illuminate\Http\Testing\File>  $files
+     * @param  array<string, mixed>|array<int, array<string, mixed>>  $operations
+     * @param  array<int, array>  $map
```
spawnia, line 83:
```suggestion
     * @param  array<string, array<int, string>>  $map
```

Integer keys are erased by `json_encode`, which is why all examples use quoted numbers `'0'`. The values are indeed an array (to be precise: a list of strings), good catch.

spawnia:
Looks like PHP is screwing us with inconsistent behaviour:

```
>>> json_encode(['0' => 4])
=> "[4]"
>>> json_encode(['1' => 4])
=> "{"1":4}"
```

jandrodev:
Yes, it seems so. The previous annotation might be more correct (or at least it gave less problems hehe). What do you think?

spawnia:
I noticed that using integer keys also works fine, so I changed the annotation to be more flexible,

spawnia (approved):
Good idea and well done, thank you

jandrodev:
😊 👍

PR #1616, opened 2020-11-19
Add test for HasMany nested mutations with empty list operations

spawnia:


spawnia:
The issue I was experiencing seems like something that could be fixed in Laravel itself: https://github.com/laravel/framework/pull/35293

PR #1624, opened 2020-11-25
Improve Xdebug in Docker

jandrodev:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

I improve the default options a bit so that from IDE like PHPStorm you can work with XDebug by default. You just have to configure the server in the IDE.

No breaking changes.


spawnia (approved):
Cool, thank you

PR #1626, opened 2020-11-27
Add optimized selects

### LINKED ISSUE #1356: Select only necessary columns for optimized queries
I am trying out GraphQL, testing various features in order to include it in some of the web apps I am developing with laravel.  Being curious, I created a Schema and fired up all sorts of queries, followed by poking around in the SQL server to see what was going on.

What I discovered was this:

Suppose I have a Type defined which corresponds to a Model, and the database table has an ID column and 26 more columns (A, B C, ..., Z).

No matter what the Type definition is, which in my case only includes some of the columns ( say, ID and A, B, C), I can see that the SQL server is asked to fetch all 27 columns from the DB, using a SELECT * FROM [Table] statement, even though given the definition of the type at most 4 of them would be returned via the resolver.

This is an unnecessary data transfer between the DB and PHP, in my opinion.

I propose that we construct a more refined SQL query, selecting only the columns that are requested in the GraphQL query. 

I am not sure I am competent enough to try and implement this on my own, but I will give it a shot. In any case, I thought I'd share my thoughts.

DDynamic:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #1356

**Changes**
This MR optimizes queries generated by eloquent by only including the requested fields.

```graphql
type Query {
  posts: [Post] @all
}

type Post {
  id: ID
  likes: Int
  title: String
  uppercaseTitle: String @select(fields: ["title"]) @method(name: "uppercaseTitle")
}
```
```
{
  posts {
    id
    uppercaseTitle
  }
}
```
The above schema definition and query cause `SELECT * FROM posts` to be executed. This MR adds adds in an opt-in configuration option to change this behavior to select only the fields required, `SELECT id, title FROM posts`. It adds a select directive to specify field dependencies, like in the case of uppercaseTitle.

@spawnia what do you think of this implementation idea? I could update the `AllDirecrive` to get the field selection with `$resolveInfo`, but I am not sure how to "look ahead" to field selection to see if there is a select directive present.


spawnia:
First, I suggest to implement the trivial case without `@select`:

```graphql
{
  posts {
    id
    title
  }
}
```

```sql
select id, title from posts;
```

Next, consider the implementation of https://github.com/nuwave/lighthouse/blob/master/src/Schema/Directives/CacheDirective.php, that should give you an idea how you can grab `@select`. You will have to combine the field selection from `$resolveInfo` with the type information from the schema. Tricky, but definitely doable.

spawnia:
Also, see https://github.com/nuwave/lighthouse/pull/1253

src/Select/SelectDirective.php
```diff
@@ -0,0 +1,23 @@
...
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Specify the SQL column dependencies of this field.
+"""
+directive @select(
+  """
+  SQL columns names to pass to the Eloquent query builder
+  """
+  columns: [String!]
```
spawnia, line 19:
```suggestion
  columns: [String!]!
```

This directive has no use without this argument.

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,99 @@
...
+
+        $selectColumns = [];
+
+        foreach ($fieldSelection as $field) {
+            $fieldDefinition = ASTHelper::firstByName($fieldDefinitions, $field);
+
+            if ($fieldDefinition) {
+                $directivesRequiringLocalKey = ['hasOne', 'hasMany', 'count'];
+                $directivesRequiringForeignKey = ['belongsTo', 'belongsToMany', 'morphTo'];
+                $directivesRequiringKeys = array_merge($directivesRequiringLocalKey, $directivesRequiringForeignKey);
```
spawnia, line 48: Since we are trying to optimize performance here: Those should all be `const`.

spawnia (commented):
Nice to see you pick up work on this branch again, this will be a big improvement when it lands. The code looks pretty good so far.

This feature seems like something that can have a lot of weird corner cases, so don't forget to add comprehensive tests.

DDynamic:
Hey, @spawnia, thanks for checking in. I tried to get some work done on it this weekend, but I hit a wall with some edge cases and Eloquent limitations surrounding polymorphic relationships, so I decided to push what I had. The app that utilizes this only employs the all, paginate, and one-to-one and one-to-many relationship directives, so those use cases are working.

I'm not sure how many more hours I can put into this to address all of the edge cases. It would be an amazing edition though, nonetheless.

spawnia:
A partial solution is better than no solution at all, if there are shortcomings we can document them and offer users the choice. Let's just make sure that whatever is implemented actually works.

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,99 @@
...
+                        array_push($selectColumns, $renamedAttribute);
+                    }
+                } else {
+                    // fallback to selecting the field name
+                    array_push($selectColumns, $field);
+                }
+            }
+        }
+
+        return array_unique($selectColumns);
```
igor-kozhevnikov-avg, line 97: I'm not sure, but if I need an accessor that is not in the table columns, we will get a "Column not found" SQL error.

```suggestion
        $allColumns = Schema::getColumnListing($model->getTable());
        return array_intersect($allColumns, array_unique($selectColumns));
```

hassan-index:
@DDynamic 
@spawnia 
Lately, I've been working on a project where we're revaming our system and implementing GraphQL Lighthouse. One issue I've run into is that some of our tables have around 100 to 120 columns, which slows down the process even when I only need the ID and name. This is because I'm using the @find and @paginate directory, but behind the scenes, the MySQL query was selecting all columns from the "registrations" table.

PR #1628, opened 2020-11-30
Add custom attributes to validation

hosmelq:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This PR adds the ability to pass [custom attributes](https://laravel.com/docs/8.x/validation#customizing-the-validation-attributes) to validations.

**Breaking changes**

No.


src/Validation/RulesGatherer.php
```diff
@@ -117,6 +124,12 @@ public function extractRulesAndMessages(ProvidesRules $providesRules, array $arg
 
         $messages = $providesRules->messages();
         $this->messages += $this->wrap($messages, $argumentPath);
+
+        $attributes = $providesRules->attributes();
+        $this->attributes = array_merge(
+            $this->attributes,
+            isset($rules[0]) ? $attributes : $this->wrap($attributes, $argumentPath)
```
spawnia, line 131: What does this do? Extracting a variable of method with a proper name can make this more cleara.

Also: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#multiline-ternary-expressions

src/Validation/Validator.php
```diff
@@ -31,6 +31,16 @@ public function messages(): array
         return [];
     }
 
+    /**
+     * Get custom attributes for validator errors.
```
spawnia, line 35:
```suggestion
     * Specify a custom attribute name to use in your validation message.
```

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+- Add custom attributes to validations https://github.com/nuwave/lighthouse/pull/1628
```
spawnia, line 12: This should be moved to the end of the list of `### Added` features.

spawnia (changes requested):
Very nice job overall. You paid great attention to tests, naming and docs. Thank you

src/Validation/RulesGatherer.php
hosmelq: Changes applied, thank you for your recommendations.

src/Support/Contracts/ProvidesRules.php
```diff
@@ -22,4 +22,11 @@ public function rules(): array;
      * @return array<string, string>
      */
     public function messages(): array;
+
+    /**
+     * Get custom attributes for validator errors.
```
spawnia, line 27:
```suggestion
     * Specify a custom attribute name to use in your validation message.
```

src/Validation/RulesGatherer.php
```diff
@@ -117,6 +124,16 @@ public function extractRulesAndMessages(ProvidesRules $providesRules, array $arg
 
         $messages = $providesRules->messages();
         $this->messages += $this->wrap($messages, $argumentPath);
+
+        $attributes = $providesRules->attributes();
+        $inputToAttributes = isset($rules[0])
+            // We might be passed just the attribute for a single field
+            // in this case, we just add it to the list.
+            ? $attributes
```
spawnia, line 132: I think we might still need to add the path. Can you add a test with `@rules` on an input field?

hosmelq: You are right, in this case, I do not know how to detect if I need to use the `wrap` method or not because if I use the `@rules` directive I ended up with something like this.

```php
[
	'bar.bar' => 'baz',
]
```

spawnia:
I have come to the conclusion we need a clean separation between validation for:
- single arguments
- sets of arguments

It might be useful/necessary to allow directives to function for both, so we can either do:
- two separate interfaces with no overlap in method names
- one interface with flexible enough types for both

My preference goes towards separating the interfaces like this:

```php
/**
 * Provide validation details for a set of arguments.
 *
 * https://laravel.com/docs/validation
 */
interface ArgumentSetValidation
{
    /**
     * Specify validation rules for the arguments.
     *
     * @return array<string, array<int, mixed>>
     */
    public function argumentSetRules(): array;

    /**
     * Specify custom messages for the rules.
     *
     * @return array<string, string>
     */
    public function argumentSetMessages(): array;

    /**
     * Specify custom attribute names to use in the validation message.
     *
     * @return array<string, string>
     */
    public function argumentSetAttributes(): array;
}

/**
 * Provide validation details for a single argument.
 *
 * https://laravel.com/docs/validation
 */
interface ArgumentValidation
{
    /**
     * Specify validation rules for the argument.
     *
     * @return array<int, mixed>
     */
    public function argumentRules(): array;

    /**
     * Specify custom messages for the rules.
     *
     * @return array<string, string>
     */
    public function argumentMessages(): array;

    /**
     * Specify a custom attribute name to use in the validation message.
     */
    public function argumentAttribute(): string;
}
```

spawnia (approved):
Thanks again @hosmelq, I split the interfaces. While the amount of code is increased, the separation is now much clearer.

PR #1630, opened 2020-12-02
Check current guard in can directive

jandrodev:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

In directives like can, when evaluating authorization, it is not taken into account if a custom guard is being used (which is not declared in default), so authorization will fail at this point. This is especially the case for queries or mutations where 2 different guards can be used.

This would detect the guard with which we are authenticating in the case that a different one is required than the one declared in default.

No breaking changes.

### DIFF
--- a/src/Schema/Directives/CanDirective.php
+++ b/src/Schema/Directives/CanDirective.php
@@ -9,12 +9,14 @@ use Illuminate\Contracts\Auth\Access\Gate;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Auth;
 use Nuwave\Lighthouse\Exceptions\AuthorizationException;
 use Nuwave\Lighthouse\Execution\Arguments\ArgumentSet;
 use Nuwave\Lighthouse\Schema\Values\FieldValue;
 use Nuwave\Lighthouse\SoftDeletes\ForceDeleteDirective;
 use Nuwave\Lighthouse\SoftDeletes\RestoreDirective;
 use Nuwave\Lighthouse\SoftDeletes\TrashedDirective;
+use Nuwave\Lighthouse\Support\Authentication;
 use Nuwave\Lighthouse\Support\Contracts\FieldMiddleware;
 use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;
 use Nuwave\Lighthouse\Support\Utils;
@@ -176,7 +178,7 @@ GRAPHQL;
         // should be [modelClassName, additionalArg, additionalArg...]
         array_unshift($arguments, $model);
 
-        if (! $gate->check($ability, $arguments)) {
+        if (! $gate->forUser(Auth::guard(Authentication::getGuard())->user())->check($ability, $arguments)) {
             throw new AuthorizationException(
                 "You are not authorized to access {$this->nodeName()}"
             );
--- a/src/Schema/Directives/CanDirective.php
+++ b/src/Schema/Directives/CanDirective.php
@@ -9,7 +9,6 @@ use Illuminate\Contracts\Auth\Access\Gate;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Facades\Auth;
 use Nuwave\Lighthouse\Exceptions\AuthorizationException;
 use Nuwave\Lighthouse\Execution\Arguments\ArgumentSet;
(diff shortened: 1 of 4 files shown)


src/Schema/Directives/CanDirective.php
```diff
@@ -9,12 +9,14 @@
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Auth;
```
spawnia, line 12: We try to not use facades: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#laravel-feature-usage

src/Schema/Directives/CanDirective.php
```diff
@@ -176,7 +178,7 @@ protected function authorize(Gate $gate, $ability, $model, array $arguments): vo
         // should be [modelClassName, additionalArg, additionalArg...]
         array_unshift($arguments, $model);
 
-        if (! $gate->check($ability, $arguments)) {
+        if (! $gate->forUser(Auth::guard(Authentication::getGuard())->user())->check($ability, $arguments)) {
```
spawnia, line 181: This line is too long and does too much.

src/Support/Authentication.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support;
+
+use Illuminate\Support\Facades\Auth;
+
+class Authentication
+{
+    /**
+     * @return \Illuminate\Config\Repository|\Illuminate\Contracts\Foundation\Application|\Laravel\Lumen\Application|mixed
```
spawnia, line 10: That return type seems way to imprecise.

src/lighthouse.php
```diff
@@ -59,6 +59,18 @@
 
     'guard' => 'api',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Custom Guards
+    |--------------------------------------------------------------------------
+    |
+    | Specify here the guards that you have defined in your application and that are different from guard option.
+    | They will be used to check authentication in directives such as @can.
```
spawnia, line 68: I don't get it, this seems to just seems to just overrule `guard`?

spawnia (changes requested):

src/Schema/Directives/CanDirective.php
jandrodev: Thanks for the advice !

src/Schema/Directives/CanDirective.php
jandrodev: Ok, no problem.

src/lighthouse.php
jandrodev: No, the idea was to indicate in the Lighthouse configuration the custom guards that are defined in the application to be able to check against them in the can directive, but you can also simply take those that are defined in Laravel (or in Lumen).

src/Support/Authentication.php
jandrodev: If you're right. As I see it it should be string | null, but even changing it to int | string it still gives error in phpstan. I don't really understand it very well.

src/Support/Authentication.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * @return string|null
+     */
+    public static function getGuard()
+    {
+        $guards = array_keys(config('auth.guards'));
+
+        if (! empty($guards)) {
+            foreach ($guards as $guard) {
```
spawnia, line 15: That just checks all guards that exist, disregarding if they are even configured to be used. I would not expect this behaviour as a user, that makes it very dangerous.

This also completely disregards the setting in `lighthouse.guard`, that is what I meant my comment here: https://github.com/nuwave/lighthouse/pull/1630#discussion_r534194985

spawnia:
I still do not understand what you are trying to achieve here. Why do you need to try every possible guard your application offers? As I see it, that is just softening security and opening potential attack vectors.

jandrodev:
Sorry, maybe I should have started by explaining the problem I encountered. This comes by using a custom guard other than the default one, and by making use of the @can directive. I give you an example:

Imagine a mutation like this:

```graphql
updateUser (
   id: Int!,
  obsolete: Boolean!
): User
@guard(with: ["web", "custom_guard"])
@can(ability: "view", find: "id")
@update
```
The default guard in lighthouse config is web, and use the AttemptAuthentication middleware. All works fine except the @can directive. Always throw AuthorizationException. The auth user only has the custom_guard.

I have thought that perhaps the problem is due to the fact that the guard is not taken into account in the can directive when checking, hence my proposal. What do you think?

spawnia:
I don't think `@can` has to care about which guard is used.

PR #1633, opened 2020-12-09
Fix documentation of @orderBy

brunobg:
Documentation says that `@orderBy` parameter in the query is `column`, when it's actually `field`. It's here: https://lighthouse-php.com/4.18/api-reference/directives.html#orderby. The last example in the documentation is correct, but the earlier ones are not. I'm guessing the parameter name changed in time. This PR corrects the documentation. I also added an example of how to set the default value for sorting.

No issue was created.

**Changes**

Only updates docs.

**Breaking changes**

No breaking changes.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -1722,14 +1722,22 @@ type Query {
 }
 ```
 
+You can easily set the default value:
+
+```graphql
+type Query {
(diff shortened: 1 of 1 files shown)


spawnia:
This changes in the upcoming v5, which the docs in `master` reflect. Thanks for the effort

brunobg:
Ah, sorry. May I suggest to add a banner to the doc website saying master is reflecting v5, since it contains breaking changes? It will avoid some running around screaming "why, why doesn't it work" for some people ;)

PR #1636, opened 2020-12-11
Add `FieldBuilderDirective` interface and `@whereAuth` directive

### LINKED ISSUE #1632: Custom extension of builder when used in relationship
Allow custom extension of builder when used with relationship.

We have a scenario where we often want to apply a scope where the first argument is the authenticated user, so naturally we just create a new directive called `authScope`. However we then hit the problem of how to apply the scope to the builder inside of a `morphMany` directive. 

We currently solved it the following way. What we did here is create a super class of argument set, so we could override `enhanceBuilder` method.  

```php
class AuthScopeDirective extends BaseDirective implements FieldMiddleware, DefinedDirective
{
    {
        return 'authScope';
    }

    public static function definition(): string
    {
        return /* @lang GraphQL */ <<<'SDL'
        directive @authScope(
          name: String
        ) on FIELD_DEFINITION
        SDL;
    }

    public function handleField(FieldValue $fieldValue, Closure $next): FieldValue
    {
        $scope = $this->directiveArgValue('name');
        $resolver = $fieldValue->getResolver();

        return $next(
            $fieldValue->setResolver(
                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver, $scope) {
                    $resolveInfo->argumentSet = new ArgumentSetExtender($resolveInfo->argumentSet);

                    $resolveInfo->argumentSet->addExtension(
                        fn($query) => $query->{$scope}($context->user())
                    );

                    /** @var \GraphQL\Deferred|\Illuminate\Database\Eloquent\Model $result */
                    $result = $resolver($root, $args, $context, $resolveInfo);

                    return $result;
                }
            )
        );
    }
}

class ArgumentSetExtender extends ArgumentSet
{
    /** @var array<closure> */
    private array $extensions = [];

    public function __construct(ArgumentSet $argumentSet)
    {
        $this->arguments = $argumentSet->arguments;
        $this->undefined = $argumentSet->undefined;
        $this->directives = $argumentSet->directives;
    }

    public function enhanceBuilder($builder, array $scopes, Closure $directiveFilter = null)
    {
        foreach ($this->extensions as $extension) {
            $builder = $extension($builder);
        }

        return parent::enhanceBuilder($builder, $scopes, $directiveFilter);
    }

    public function addExtension(Closure $extension): self
    {
        $this->extensions[] = $extension;
        return $this;
    }
}
```

I see many more ways of supporting this. A more viable way might be to add a new interface in the same form as `ArgBuilderDirective`, but which is not bound to arguments.

olivernybroe:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1632 

This PR is not completely done, I think I would add more test cases to it first and possibly also allow changing the guard on the `whereAuth`.  
The purpose of this PR was mostly to get support for manipulating the builder on fields, not the new directive. However I feel that this directive shows the power of the new interface.

Was hoping to get some feedback about the idea here before I finished the pr completely.

**Changes**
- Adds a new interface for manipulating the builder on fields
- Adds a new directive for filtering a relationship on the currently authenticated user.

### DIFF
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -6,6 +6,7 @@ use Closure;
 use Nuwave\Lighthouse\Schema\Directives\RenameDirective;
 use Nuwave\Lighthouse\Schema\Directives\SpreadDirective;
 use Nuwave\Lighthouse\Support\Contracts\ArgBuilderDirective;
+use Nuwave\Lighthouse\Support\Contracts\FieldBuilderDirective;
 use Nuwave\Lighthouse\Support\Utils;
 
 class ArgumentSet
@@ -149,6 +150,7 @@ class ArgumentSet
     public function enhanceBuilder(object $builder, array $scopes, Closure $directiveFilter = null): object
     {
         self::applyArgBuilderDirectives($this, $builder, $directiveFilter);
+        self::applyFieldBuilderDirectives($this, $builder);
 
         foreach ($scopes as $scope) {
             $builder->{$scope}($this->toArray());
@@ -198,6 +200,24 @@ class ArgumentSet
                 $argument->value
             );
         }
+    }
+
+    /**
+     * Apply the FieldBuilderDirectives onto the builder.
+     *
+     * TODO get rid of the reference passing in here. The issue is that @search makes a new builder instance,
+     * but we must special case that in some way anyhow, as only eq filters can be added on top of search.
+     *
+     * @param  \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet  $argumentSet
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Relations\Relation  $builder
+     */
+    protected static function applyFieldBuilderDirectives(self $argumentSet, object &$builder): void
+    {
+        $argumentSet->directives
+            ->filter(Utils::instanceofMatcher(FieldBuilderDirective::class))
+            ->each(function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder) {
+                $builder = $fieldBuilderDirective->handleFieldBuilder($builder);
+            });
     }
 
     /**
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -215,7 +215,7 @@ class ArgumentSet
     {
         $argumentSet->directives
             ->filter(Utils::instanceofMatcher(FieldBuilderDirective::class))
-            ->each(function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder) {
+            ->each(static function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder): void {
                 $builder = $fieldBuilderDirective->handleFieldBuilder($builder);
             });
     }
--- a/src/Schema/Directives/WhereAuthDirective.php
+++ b/src/Schema/Directives/WhereAuthDirective.php
@@ -40,17 +40,16 @@ GRAPHQL;
 
     public function handleFieldBuilder(object $builder): object
     {
-        $guard = $this->directiveArgValue('guard', config('lighthouse.guard'));
-        $relation = $this->directiveArgValue('relation');
-        $userId = $this
-            ->authFactory
-            ->guard($guard)
-            ->id();
-
-        // @phpstan-ignore-next-line phpstan cannot find this method.
+        // @phpstan-ignore-next-line Mixins are magic
         return $builder->whereHas(
-            $relation,
-            static function ($query) use ($userId): void {
+            $this->directiveArgValue('relation'),
+            static function ($query): void {
+                $guard = $this->directiveArgValue('guard', config('lighthouse.guard'));
+                $userId = $this
+                    ->authFactory
+                    ->guard($guard)
+                    ->id();
+
                 $query->whereKey($userId);
             }
         );
--- a/src/Schema/Directives/WhereAuthDirective.php
+++ b/src/Schema/Directives/WhereAuthDirective.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Nuwave\Lighthouse\Schema\Directives;
+
+use Illuminate\Contracts\Auth\Factory as AuthFactory;
+use Nuwave\Lighthouse\Support\Contracts\FieldBuilderDirective;
+
+class WhereAuthDirective extends BaseDirective implements FieldBuilderDirective
+{
+    /**
+     * @var \Illuminate\Contracts\Auth\Factory
+     */
+    protected $authFactory;
+
+    public function __construct(AuthFactory $authFactory)
+    {
+        $this->authFactory = $authFactory;
+    }
+
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Filter a type to only return instances owned by the current user.
+"""
+directive @whereAuth(
+  """
+  Name of the relationship that links to the user model.
+  """
+  relation: String!
+  """
+  Specify which guard to use, e.g. "api".
+  When not defined, the default from `lighthouse.php` is used.
+  """
+  guard: String
+) on FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    public function handleFieldBuilder(object $builder): object
+    {
+        $guard = $this->directiveArgValue('guard', config('lighthouse.guard'));
+        $relation = $this->directiveArgValue('relation');
+        $userId = $this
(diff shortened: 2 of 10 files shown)


phpstan.neon
```diff
@@ -69,3 +69,6 @@ parameters:
 
   # This really has no type in root resolvers
   - '#Method .* has parameter \$root with no typehint specified\.#'
+
+  # PhpStan cannot handle higher order collections
```
spawnia, line 73: I would rather not use them in that case, given there is a perfectly fine alternative that preserves type safety.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -200,6 +202,27 @@ static function ($value) use (&$builder, $directiveFilter) {
...
+     *
+     * TODO get rid of the reference passing in here. The issue is that @search makes a new builder instance,
+     * but we must special case that in some way anyhow, as only eq filters can be added on top of search.
+     *
+     * @param  \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet  $argumentSet
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Relations\Relation  $builder
+     */
+    protected static function applyFieldBuilderDirectives(self $argumentSet, object &$builder): void
+    {
+        // Get all directives
```
spawnia, line 216:
```suggestion
```

The code is pretty clear without the line-by-line comments 👍

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -200,6 +202,27 @@ static function ($value) use (&$builder, $directiveFilter) {
...
+        $argumentSet->directives
+            // Only give us field builder directives
+            ->filter(Utils::instanceofMatcher(FieldBuilderDirective::class))
+            // And run the builder method.
+            ->each(function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder) {
+                $builder = $fieldBuilderDirective->handleFieldBuilder($builder);
```
spawnia, line 222: Looks like `Model` is missing from the definition of `handleFieldBuilder`.

src/Schema/Directives/WhereAuthDirective.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Filter a type to only return instances owned by the current user.
+"""
+directive @whereAuth(
+  """
+  Specify the relationship to check for auth
```
spawnia, line 29:
```suggestion
  Name of the relationship that links to the user model.
```

src/Schema/Directives/WhereAuthDirective.php
```diff
@@ -0,0 +1,42 @@
...
+  """
+  relation: String = "user"
```
spawnia, line 31: I am on the fence about the default, maybe it should just be a required argument without default.

src/Schema/Directives/WhereAuthDirective.php
```diff
@@ -0,0 +1,42 @@
...
+) on FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    public function handleFieldBuilder($builder)
+    {
+        return $builder->whereHas($this->directiveArgValue('relation', 'user'), function ($query) {
+            $query->whereKey($this->authFactory->guard()->id());
```
spawnia, line 39: Only one method call per line.

src/Support/Contracts/FieldBuilderDirective.php
```diff
@@ -0,0 +1,14 @@
...
+
+interface FieldBuilderDirective extends Directive
+{
+    /**
+     * Add additional constraints to the builder.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Eloquent\Relations\Relation|\Laravel\Scout\Builder  $builder  The builder used to resolve the field.
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Eloquent\Relations\Relation|\Laravel\Scout\Builder  The modified builder.
+     */
+    public function handleFieldBuilder($builder);
```
spawnia, line 13:
```suggestion
    public function handleFieldBuilder(object $builder): object;
```

tests/Integration/Execution/FieldBuilderDirectiveTest.php
```diff
@@ -0,0 +1,44 @@
...
+            ->graphQL(/** @lang GraphQL */ '
+            query {
+                posts {
+                    id
+                }
+            }
+            ');
+
+        $this->assertTrue(
+            $ownedPosts->map->getKey()->diff($response->json('data.posts.*.id'))->isEmpty()
```
spawnia, line 41: Too much happening in this line.

tests/Integration/Execution/FieldBuilderDirectiveTest.php
```diff
@@ -0,0 +1,44 @@
...
+class FieldBuilderDirectiveTest extends DBTestCase
+{
+    protected $schema = /** @lang GraphQL */ '
+    type Post {
+        id: ID!
+        title: String
+    }
+    ';
+
+    public function test(): void
```
spawnia, line 18:
```suggestion
    public function testWhereAuthWithDefault(): void
```

spawnia (changes requested):
I like the concept, seems like a small addition that enables a bunch of use cases.

src/Schema/Directives/WhereAuthDirective.php
olivernybroe: Sure, I agree. I would prob. also in many cases call that relationship owner or something like that.

olivernybroe:
@spawnia Glad you like it 👍 

Will add in the documentation for it later.

src/Schema/Directives/WhereAuthDirective.php
```diff
@@ -0,0 +1,57 @@
...
+        $relation = $this->directiveArgValue('relation');
+        $userId = $this
+            ->authFactory
+            ->guard($guard)
+            ->id();
+
+        // @phpstan-ignore-next-line phpstan cannot find this method.
+        return $builder->whereHas(
+            $relation,
+            function ($query) use ($userId) {
```
spawnia, line 52:
```suggestion
            static function ($query) use ($userId): void {
```

src/Schema/Directives/WhereAuthDirective.php
```diff
@@ -0,0 +1,57 @@
...
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Filter a type to only return instances owned by the current user.
+"""
+directive @whereAuth(
+  """
+  Name of the relationship that links to the user model.
+  """
+  relation: String!
+  """
```
spawnia, line 31:
```suggestion

  """
```

tests/Integration/Execution/FieldBuilderDirectiveTest.php
```diff
@@ -0,0 +1,87 @@
...
+            ');
+
+        $this->assertCount(
+            0,
+            $ownedPosts
+                ->map(function (Post $post) {
+                    return $post->getKey();
+                })
+                ->diff($response->json('data.posts.*.id'))
+        );
```
spawnia, line 48:
```suggestion
        $this->assertSame(
            $ownedPosts->pluck('id'),
            $response->json('data.posts.*.id'),
        );
```

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -200,6 +202,24 @@ static function ($value) use (&$builder, $directiveFilter) {
...
+     * but we must special case that in some way anyhow, as only eq filters can be added on top of search.
+     *
+     * @param  \Nuwave\Lighthouse\Execution\Arguments\ArgumentSet  $argumentSet
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Relations\Relation  $builder
+     */
+    protected static function applyFieldBuilderDirectives(self $argumentSet, object &$builder): void
+    {
+        $argumentSet->directives
+            ->filter(Utils::instanceofMatcher(FieldBuilderDirective::class))
+            ->each(function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder) {
```
spawnia, line 218:
```suggestion
            ->each(static function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder): void {
```

olivernybroe:
@spawnia Alright I think this is ready now. 

Btw. I am also totally fine with you doing changes directly in my branch 😄

spawnia (approved):

PR #1638, opened 2020-12-14
php 8 support for dependency graphql-php-scalars

Nozibulla:
- [ ] Added PHP 8 support for depndent package `mll-lab/graphql-php-scalars`

**Changes**

In PHP 8 Mixed is a reserved word. So the dependent package `mll-lab/graphql-php-scalars` changed their Classname to `MixedScalar` which needed to change in `WhereConditionsServiceProvider.php`

### DIFF
--- a/src/WhereConditions/WhereConditionsServiceProvider.php
+++ b/src/WhereConditions/WhereConditionsServiceProvider.php
@@ -63,9 +63,7 @@ class WhereConditionsServiceProvider extends ServiceProvider
                         )
                     )
                     ->setTypeDefinition(
-                        Parser::scalarTypeDefinition(
-                        /** @lang GraphQL */
-                        '
+                        Parser::scalarTypeDefinition(/** @lang GraphQL */ '
                             scalar Mixed @scalar(class: "MLL\\\GraphQLScalars\\\MixedScalar")
                         ')
                     );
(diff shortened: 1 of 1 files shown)


src/WhereConditions/WhereConditionsServiceProvider.php
```diff
@@ -127,7 +130,9 @@ public static function createHasConditionsInputType(string $name, string $descri
             ->value;
         $operatorDefault = $operator->defaultHasOperator();
 
-        return Parser::inputObjectTypeDefinition(/** @lang GraphQL */ <<<GRAPHQL
+        return Parser::inputObjectTypeDefinition(
```
spawnia, line 133: I prefer the previous formatting

spawnia (changes requested):

spawnia:
Please revert the unnecessary formatting changes.

src/WhereConditions/WhereConditionsServiceProvider.php
Nozibulla: Reverted formatting to the previous one.

spawnia:
Thank you. There were a few more things that also required updating, so I superseded your PR by https://github.com/nuwave/lighthouse/pull/1639

PR #1640, opened 2020-12-15
Add Sanctum cookie endpoint to default CORS paths

divspace:
Minor documentation update to match the default config value that you would see when opening the CORS config file.


spawnia (approved):
Thank you

Issue #1644, opened 2020-12-20
Add @like directive


olivernybroe:
**What problem does this feature proposal attempt to solve?**
Make it easier for adding a new directive for making like queries.

This is for example something that is really nice when wanting to search for a name in all users.

**Which possible solutions should be considered?**

We currently have a `@search` directive, which is really nice. However this is only for Laravel Scout. A `@like` directive would be usable in many other cases.

So I think this directive should be configurable, so you can define where you want a percentage sign. My thought were the following
```gql
users(name: String @like(percentage: BOTH)): [User!]! @all
```
So we make a `ENUM`for percentage, with the following options
```gql
enum PercentageLocation {
    START
    END
    BOTH
    NONE
}
```

We might consider not having `NONE` as an option and making `BOTH` the default.

olivernybroe:
Implementation is quite simple, it's basically copy/pasting `EqDirective` and making a check for where we want the percentage sign. 

Possibly also escaping `%` already present in the values string.

spawnia:
Great suggestion, I like it!

> We might consider not having NONE as an option and making BOTH the default.

Using `NONE` is the same as just using `@eq`, so I agree we should not have that option. `BOTH` seems like a reasonable default.

olivernybroe:
@spawnia I think this could also be a good first issue. Relatively straight forward.

brunobg:
I'm getting this one. PR soon.

olivernybroe:
@brunobg Amazing! If you need any help, just say so 👍

brunobg:
Yes @olivernybroe  thanks! is there any documentation about what writing/running tests? I'm getting a lot of errors here from existing tests.

olivernybroe:
@brunobg Only documentation about it is what is in https://github.com/nuwave/lighthouse

However you can draw a lot of inspiration from the tests here 
https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Execution/ArgBuilderDirectiveTest.php

The test suite should work, are you running it via docker or without? 
When running without you need to change connection information to the database in `phpunit.xml`.

brunobg:
I mean the structure for running the test itself, like creating a database, etc. I'm running without docker. I found some expected env vars at https://github.com/nuwave/lighthouse/blob/507a45268dccbb95674a8b14234858b27b7179b2/tests/DBTestCase.php. I could write a small doc about this on CONTRIBUTING or somewhere like that if you give me some pointers.

[...]

PR #1645, opened 2020-12-21
Add `@like` directive

brunobg:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/1644

**Changes**

Added the @like directive with its test.


brunobg:
Sorry for the many commits, not being able to test locally made it difficult. The tests are passing and I followed the spec on #1644, so I hope this is acceptable.

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+enum LikePercentageLocation {
+    START
+    END
+    BOTH
+}
+
+"""
+Uses a LIKE query.
```
spawnia, line 19:
```suggestion
Use the client given value to add a `LIKE` conditional to a database query.
```

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,74 @@
...
+namespace Nuwave\Lighthouse\Schema\Directives;
+
+use Nuwave\Lighthouse\Support\Contracts\ArgBuilderDirective;
+
+class LikeDirective extends BaseDirective implements ArgBuilderDirective
+{
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+enum LikePercentageLocation {
```
spawnia, line 12: That cannot be part of this definition, it must only contain the directive definition.

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,74 @@
...
+
+  """
+  Specify the positions of the % in the LIKE comparison. The default is BOTH.
+  """
+  percentage: LikePercentageLocation
+) repeatable on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    protected function escapePercentage(string $value, string $char = '\\'): string
```
spawnia, line 36: `$char` does not ever change, so it can just be a constant.

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,74 @@
...
+  """
+  Specify the database column to compare.
+  Only required if database column has a different name than the attribute in your schema.
+  """
+  key: String
+
+  """
+  Specify the positions of the % in the LIKE comparison. The default is BOTH.
+  """
+  percentage: LikePercentageLocation
```
spawnia, line 31: The name `percentage` indicates that the value of this argument would be a numeric value between `0` and `100`.

spawnia (changes requested):
While looking up some details about the `LIKE` operator, i found that `_` can also be used as a wildcard for a single character. That leads me to think we should avoid fixing the wildcards on the server side and just let clients pass in the entire query string.

src/Schema/Directives/LikeDirective.php
brunobg: Fixed by https://github.com/nuwave/lighthouse/pull/1649

brunobg:
> While looking up some details about the LIKE operator, i found that _ can also be used as a wildcard for a single character. That leads me to think we should avoid fixing the wildcards on the server side and just let clients pass in the entire query string.

While I'm not the OP for this issue I think that's not safe. At least that's not something I'd want in my application. Allowing `query%` is okay in terms of performance if the field has an index, but `%query%` isn't and can be a DoS attack for a large table. It's interesting to use `query%` as an autocomplete for strings that start with a given query substring. 

Also, the `escapePercentage()` method deals with %, \ and _. I know there are some more exotic escapes, like `[%]`, but I couldn't find a more thorough escape function.

src/Schema/Directives/LikeDirective.php
brunobg: I have no better idea for a name, it seemed clear enough to me. Any suggestions?

olivernybroe:
@spawnia another point for why not allowing the end user to define where the wildcards go is if you for example want to query all users and filter by username. However a user in your application signed up with the username `my%annoyingUsername`. This means I should now in the frontend escape `%` to actually make the search I wanted, when I search for `my%a`. 

I could however see this being a parameter on the directive, if we should escape wildcards.

src/Schema/Directives/LikeDirective.php
olivernybroe: Yeah, you are right. Bad naming on my part.

What do you say to `wildcard` and `LikeWildcardLocation`

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,75 @@
...
```
spawnia, line 31: We can have maximum flexibility by making `percentageLocation` optional.

```suggestion
  """
  Fixate the positions of the % wildcards in the LIKE comparison.
  If specified, wildcard characters in the client-given input are escaped.
  If not specified, the client can pass wildcards unescaped.
  """
  percentageLocation: LikePercentageLocation
```

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,75 @@
...
+) repeatable on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    protected function escapePercentage(string $value): string
+    {
+        $char = '\\';
```
spawnia, line 38: That's a variable. Given the value never changes, we should just use an actual `const`.

brunobg: fixed

src/Schema/Directives/LikeDirective.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Apply a "WHERE LIKE $value" clause.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     */
+    public function handleBuilder($builder, $value): object
+    {
+        if (!$value) {
```
spawnia, line 55: This bails on `''`, which could be a reasonable value for clients to search for when they can control wildcards.

brunobg: handling only null now.

src/Schema/Directives/LikeDirective.php
StyxUA: How about `SearchMethod`: `(ENDS, STARTS, CONTAINS)` (`%text`, `text%` and `%text%` respectively)?

spawnia: My proposed solution is https://github.com/nuwave/lighthouse/pull/1645#discussion_r549349245

spawnia:
In order to move this issue along, i implemented a template based approach that provides maximum flexibility. Built-in directives should cover most use cases.

brunobg:
Not OP, but my only question is why the directive is repeatable. Other than that it looks great.

spawnia:
> Not OP of the original issue, but my only question is why the directive is repeatable. Other than that it looks great.

It might be useful to use a single argument value and make sure it shows up in multiple columns, or use it multiple times on fields to hardcode values for different columns.

PR #1659, opened 2021-01-10
Force load type when GlobalId\Node is used

k0ka:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**The problem**
1) Query node() for some object but don't mention its type in the query.
2) This type won't be loaded as it isn't needed to resolve the query.
3) `TypeMiddleware` won't be executed.
4) It won't register itself in the `NodeRegistry`.
5) The query will fail with the `[Class] is not a registered node and cannot be resolved.` error.

**Changes**

This PR forces the type to load during node resolving, so that it can register itself.

### DIFF
--- a/src/GlobalId/NodeRegistry.php
+++ b/src/GlobalId/NodeRegistry.php
@@ -76,7 +76,10 @@ class NodeRegistry
     {
         [$decodedType, $decodedId] = $args['id'];
 
-        // Load the type and register it if it hasn't been loaded yet
+        // This check forces Lighthouse to eagerly load the type, which might not have
+        // happened if the client only references it indirectly through an interface.
+        // Loading the type in turn causes the TypeMiddleware to run and thus register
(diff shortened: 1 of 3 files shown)


src/GlobalId/NodeRegistry.php
```diff
@@ -76,6 +76,11 @@ public function resolve($root, array $args, GraphQLContext $context, ResolveInfo
     {
         [$decodedType, $decodedId] = $args['id'];
 
+        // Load the type and register it if it hasn't been loaded yet
```
spawnia, line 79:
```suggestion
        // This check achieves multiple things:
        // - actually load the type, might be necessary when the client only
        //   references the type indirectly through the use of an interface
        // - provide a clear error message when an unknown type is encountered
```

spawnia (approved):
Thanks @k0ka, great job on documenting and testing this fix!

PR #1662, opened 2021-01-13
Allow spec-compliant definition of message on `@rules` and `@rulesForArray`

spawnia:


spawnia:
@SamChristy can you give this a quick review?

PR #1666, opened 2021-01-14
Resolve field middleware in lexical order

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

It is expected that field middleware run in lexical order, as they are defined in the schema. This PR reverses their previous execution order to match that assumption.

**Breaking changes**

In most cases, this makes no difference. If this comes up, the implementation is likely bugged right now and will be fixed by this.

@ipalaus


webard:
What about:
```
extend type Query @guard {
    getProduct(id: ID! @eq): Product
        @find
        @can(ability:"get product")
}
```
?

First should fire @guard to check if user is authenticated, but after this change first fired directive is @can

andrewmclagan:
This is undocumented?

PR #1678, opened 2021-01-21
Predefine default field ordering by using `@orderBy` on fields

olivernybroe:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
Add field support for order by directive.

This PR makes the order by directive works on fields also. 
This is handy for when you need to add an order by which is always applied to a field. For example so users always returns the newest user first.

**Breaking changes**

No BC's.


olivernybroe:
@spawnia Should this be split into it's own directive file then? Or should change so the same directive can be defined multiple times in one directive class?

src/OrderBy/OrderByDirective.php
```diff
@@ -39,6 +40,22 @@ public static function definition(): string
     """
     columnsEnum: String
 ) on ARGUMENT_DEFINITION
+
+"""
+Sort a result list by a column
+"""
+directive @orderBy(
```
spawnia, line 47: That will lead to errors in the generated `schema-directives.graphql`, directives names must be unique. You can just merge the two definitions and clarify correct usage in the description.

tests/Integration/OrderBy/OrderByDirectiveDBTest.php
```diff
@@ -15,6 +15,8 @@ class OrderByDirectiveDBTest extends DBTestCase
             orderByRestricted: _ @orderBy(columns: ["name"])
             orderByRestrictedEnum: _ @orderBy(columnsEnum: "UserColumn")
         ): [User!]! @all
+
+        latestUsers: [User!]! @all @orderBy(column: "created_at" direction: "DESC")
```
spawnia, line 19: I am trying to move away from having a single large schema per test class, it is rather bad for maintenance. Please define it down in the individual test method instead.

src/OrderBy/OrderByDirective.php
```diff
@@ -84,4 +101,15 @@ public function manipulateArgDefinition(
             $argDefinition->type = Parser::typeReference('['.OrderByServiceProvider::DEFAULT_ORDER_BY_CLAUSE.'!]');
         }
     }
+
+    /**
+     * @param \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder $builder
```
spawnia, line 106:
```suggestion
     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
```

spawnia (changes requested):

src/OrderBy/OrderByDirective.php
olivernybroe: Oh, I thought they just had to be unique based on where they were applied.

spawnia (approved):
@olivernybroe bouncing this back to you, feel free to merge if you agree with my changes.

olivernybroe:
@spawnia I agree with the changes! :) 

Don't have permission to merge unfortunately.

PR #1679, opened 2021-01-21
Abstract away dealing with deferred results

spawnia:


olivernybroe (commented):
I like the overall idea for shortening this and don't see anything I am against here. 

One thing I think you should think about is maybe adding in a method on `FieldValue` that basically just allows me to define the closure inside `Resolved::handle`. 

So basically making the `setResolver` the advanced way, but in many cases using the other method is just easier.

olivernybroe (approved):
Yeah, I like this. 
This is cleaner than I expected and still has the same advantages as before 👍 

Gonna refactor all my directives after this! 😆

spawnia:
Yeah, the `FieldMiddleware` API always felt a little bit clunky too me - still does perhaps - but such higher level APIs help.

olivernybroe:
Will prob. Also make it easier for us in the future to make an even higher level of abstraction

PR #1681, opened 2021-01-22
Allow `@eq` directive on fields

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Allow the existing arg builder directives to be used upon fields to provide default conditions.

**Breaking changes**

No.


spawnia:
@olivernybroe what do you think?

olivernybroe (commented):
Nice! 👍 

Looks great on `EqDirective`.

I think you should move the test to another test file, it does not make sense that we test something related to the fields in ArgBuilder directive

PR #1683, opened 2021-01-22
Fix subscriptions renaming variables

### LINKED ISSUE #1682: Variables are renamed when broadcasting subscriptions
**Describe the bug**

When creating a subscription and using variables within the arguments, the variable is not found.

Example operation:

```
subscription {
    documentModified(number: $testVar) {
        id
    }
}
```

Will result in following being broadcast:
```json
{
  "more": true,
  "result": {
    "errors": [
      {
        "message": "Variable \"$testVar\" of required type \"Int!\" was not provided.",
        "extensions": {
          "category": "graphql"
        }
      }
    ],
    "extensions": {
      "validation": [],
      "lighthouse_subscriptions": {
        "version": 1,
        "channels": []
      }
    }
  }
}
```
It seems like Lighthouse renames the variable within the query from `testVar` to `number` (name of the subscription argument). 

Not renaming variable names :-)

**Lighthouse Version**

5.0.2

alberthaff:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1682 

**Changes**

The method expects variables, however we previously used arguments. This caused variables to change names before query execution. This is now fixed.

**Breaking changes**

None

### DIFF
--- a/src/Subscriptions/SubscriptionBroadcaster.php
+++ b/src/Subscriptions/SubscriptionBroadcaster.php
@@ -86,7 +86,7 @@ class SubscriptionBroadcaster implements BroadcastsSubscriptions
                 $data = $this->graphQL->executeQuery(
                     $subscriber->query,
                     $subscriber->context,
-                    $subscriber->args,
+                    $subscriber->variables,
                     $subscriber
                 );
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Resolve field middleware directives in lexical order https://github.com/nuwave/lighthouse/pull/1666
 - Ensure `Carbon\Carbon` is cast to `Illuminate\Support\Carbon` in date scalars https://github.com/nuwave/lighthouse/pull/1672
 - Fix Laravel 5.6 compatibility for `@withCount` and paginated relationship directives https://github.com/nuwave/lighthouse/pull/1528
+- Fix issue where argument names where used instead of variable names in subscription queries https://github.com/nuwave/lighthouse/pull/1683
 
 ## 5.0.2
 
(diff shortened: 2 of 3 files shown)


src/Subscriptions/SubscriptionBroadcaster.php
```diff
@@ -86,7 +86,7 @@ function (Subscriber $subscriber) use ($root): void {
                 $data = $this->graphQL->executeQuery(
                     $subscriber->query,
                     $subscriber->context,
-                    $subscriber->args,
+                    $subscriber->context->request()->input('variables'),
```
spawnia, line 89: This seems like the wrong place to do this, the `Subscriber` is supposed to hold the serialized information that is necessary to re-execute a query. `->args` should just hold the variables?

spawnia (changes requested):

src/Subscriptions/SubscriptionBroadcaster.php
alberthaff: I think the args are used elsewhere, since we also pass in the whole Subscribe-object as an argument on `executeQuery`. What about adding variables into the Subscriber without touching args?

spawnia: You are right, `Subscriber::$args` is used in many places and is useful as is.

We can just use `$resolveInfo->variableValues` and store that in the `Subscriber` object. That is preferrable to the approach you took here, since it avoids ad-hoc reparsing of the request. The client could have sent the variables in a way that is not compatible with `->input()`, see https://github.com/laragraph/utils/blob/master/src/RequestParser.php

alberthaff: I see. Thanks 👍

spawnia:
Nice. Can you add a note to the changelog to document this fix?

alberthaff:
Done

spawnia (approved):
Sublime

PR #1685, opened 2021-01-22
Fix TTL for subscriptions

### LINKED ISSUE #1684: TTL breaks subscriptions
**Describe the bug**

When setting a TTL for subscriptions, following exception is thrown:
```
Redis::set() expects at most 3 parameters, 4 given
```

**Steps to reproduce**

1. Set the subscription TTL
2. Create a subscription

**Lighthouse Version**

5.0.2

alberthaff:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1684 

**Changes**

When creating a subscription, following exception was thrown:

`Redis::set() expects at most 3 parameters, 4 given`

Phpredis seems to [support simply appending the TTL in seconds](https://github.com/phpredis/phpredis#examples-4). 

Reduced arguments from 4 to 3, by removing "EX" from the Redis-command.

**Breaking changes**

### DIFF
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -94,11 +94,11 @@ class RedisStorageManager implements StoresSubscriptions
         }
 
         // Lastly, we store the subscriber as a serialized string...
-        $this->connection->command('set', array_merge([
+        $this->connection->command('set', [
             $this->channelKey($subscriber->channel),
             $this->serialize($subscriber),
-            // ...and set the EX option to set the ttl in one command.
-        ], $this->ttl ? ['EX', $this->ttl] : []));
+            $this->ttl,
+        ]);
     }
 
     public function deleteSubscriber(string $channel): ?Subscriber
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Resolve field middleware directives in lexical order https://github.com/nuwave/lighthouse/pull/1666
 - Ensure `Carbon\Carbon` is cast to `Illuminate\Support\Carbon` in date scalars https://github.com/nuwave/lighthouse/pull/1672
 - Fix Laravel 5.6 compatibility for `@withCount` and paginated relationship directives https://github.com/nuwave/lighthouse/pull/1528
+- Fix issue with TTL breaking subscriptions https://github.com/nuwave/lighthouse/pull/1685
 
 ## 5.0.2
 
--- a/tests/Unit/Subscriptions/Storage/RedisStorageManagerTest.php
+++ b/tests/Unit/Subscriptions/Storage/RedisStorageManagerTest.php
@@ -81,6 +81,7 @@ class RedisStorageManagerTest extends SubscriptionTestCase
                         'channel' => 'private-lighthouse-foo',
                         'topic' => 'some-topic',
                     ]).'}',
+                    null,
                 ]]
             );
 


spawnia:
@thekonz since you are the author of this implementation, can you give a quick review?

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -97,8 +97,8 @@ public function storeSubscriber(Subscriber $subscriber, string $topic): void
         $this->connection->command('set', array_merge([
             $this->channelKey($subscriber->channel),
             $this->serialize($subscriber),
-            // ...and set the EX option to set the ttl in one command.
-        ], $this->ttl ? ['EX', $this->ttl] : []));
+            // ...and set the TTL in one command.
+        ], $this->ttl ? [$this->ttl] : []));
```
spawnia, line 101: My guess is that this should have been `['EX' => $this->ttl]`. Do we have a test case to cover this?

thekonz:
Ah damn, I think I missed that. ;)

I looked at https://redis.io/commands/set where it shows this example:
grafik

I assumed it would just build the command string like `SET key value EX ttl`, but it sends it to Redis::set first, which takes the ttl simply as a third argument.

@alberthaff good catch!

src/Subscriptions/Storage/RedisStorageManager.php
thekonz: No, explanation in the comment. 'ex' => ttl is a common syntax (so good guess^^), but would break the code aswell, because in \Illuminate\Redis\Connections\Connection::command the framework uses array spread to pass parameters and that does not work for associative arrays.

thekonz (commented):
could also be this:
```
        $this->connection->command('set', [
            $this->channelKey($subscriber->channel),
            $this->serialize($subscriber),
            $this->ttl
        ])
```

but works fine either way

thekonz (dismissed):

src/Subscriptions/Storage/RedisStorageManager.php
thekonz: You could do `[ 'key', 'value', ['ex' => $ttl] ]` (double wrapping the 'ex' part) techincally to set more options for the timeout, because Redis::set also can take arguments like this:

```php
// Will set the key, if it doesn't exist, with a ttl of 10 seconds
$redis->set('key', 'value', ['nx', 'ex' => 10]);

// Will set a key, if it does exist, with a ttl of 1000 milliseconds
$redis->set('key', 'value', ['xx', 'px' => 1000]);
```

spawnia:
> could also be this:
> 
> ```
>         $this->connection->command('set', [
>             $this->channelKey($subscriber->channel),
>             $this->serialize($subscriber),
>             $this->ttl
>         ])
> ```
> 
> but works fine either way

This seems much simpler!

src/Subscriptions/Storage/RedisStorageManager.php
thekonz: > Do we have a test case to cover this?

I guess I did not add a ttl when integration testing.

alberthaff:
Changed it to your suggestion @thekonz . Thanks! 👍

spawnia:
What's up with that test failure though?

thekonz:
That's because RedisStorageManagerTest::testStoreSubscriber expects the command to be run without the null at the end. Before the change, there was no third item in the array.
Just add it to the test assertion.

```php
        $redisConnection->expects($this->exactly(2))
            ->method('command')
            ->withConsecutive(
                ['sadd', [
                    'graphql.topic.some-topic',
                    'private-lighthouse-foo',
                ]],
                ['set', [
                    'graphql.subscriber.private-lighthouse-foo',
                    'C:41:"Tests\Utils\Subscriptions\DummySubscriber":57:{'.\Safe\json_encode([
                        'channel' => 'private-lighthouse-foo',
                        'topic' => 'some-topic',
                    ]).'}',
                    null, // <- right here
                ]]
            );
```

spawnia (approved):

alberthaff:
Ah, sorry. Didn't notice until now. Thanks for the fix.

PR #1686, opened 2021-01-23
Update webonyx/graphql-php to ^14.5 and improve PHPStan

spawnia:


src/Schema/AST/ASTHelper.php
```diff
@@ -195,11 +188,17 @@ public static function hasDirective(Node $definitionNode, string $name): bool
...
-     * @param  iterable<\GraphQL\Language\AST\Node> $nodes
+     * @template TNode of \GraphQL\Language\AST\Node
+     * @param  iterable<TNode> $nodes
+     * @return TNode|null
      */
     public static function firstByName($nodes, string $name): ?Node
     {
         foreach ($nodes as $node) {
+            if (! property_exists($node, 'name')) {
+                throw new Exception('Expected a Node with a name property, got: '.get_class($node));
```
olivernybroe, line 199: Should this be another type of exception? 👍 
Just thinking if this should be a custom exception or a RuntimeException?

src/Subscriptions/Storage/CacheStorageManager.php
```diff
@@ -34,16 +36,26 @@ class CacheStorageManager implements StoresSubscriptions
...
     protected $ttl;
 
     public function __construct(CacheManager $cacheManager)
     {
-        $this->cache = $cacheManager->store(
-            config('lighthouse.subscriptions.storage', 'file')
-        );
-        $this->ttl = config('lighthouse.subscriptions.storage_ttl');
+        /** @var \Illuminate\Contracts\Config\Repository $config */
+        $config = app(ConfigRepository::class);
```
olivernybroe, line 46: Can we take this via the constructor instead?

src/Subscriptions/Storage/CacheStorageManager.php
```diff
@@ -34,16 +36,26 @@ class CacheStorageManager implements StoresSubscriptions
...
+
+        $storage = $config->get('lighthouse.subscriptions.storage') ?? 'file';
+        if (! is_string($storage)) {
+            throw new Exception('Config setting lighthouse.subscriptions.storage must be a string or `null`, got: '.\Safe\json_encode($storage));
```
olivernybroe, line 50: Again, I think we should consider creating a custom exception here or at least make it a runtimeException 👍

olivernybroe (commented):
Looks pretty good, have some small notes 👍

src/Schema/AST/ASTHelper.php
spawnia: Definitely should be, I would probably do an overhaul of all the Exceptions in one go. I like to cut through such concerns vertically in order to devise a wholistic and consistent solution for the whole library.

olivernybroe (approved):
Looks fine for me 👍 Once you are green I'll say, go for the merge!

PR #1687, opened 2021-01-23
Allow using the `@builder` directive on fields

olivernybroe:
Add field support for builder directive.

- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Adds field definition to builder directive.

**Breaking changes**

### DIFF
--- a/src/Schema/Directives/BuilderDirective.php
+++ b/src/Schema/Directives/BuilderDirective.php
@@ -3,14 +3,15 @@
 namespace Nuwave\Lighthouse\Schema\Directives;
 
 use Nuwave\Lighthouse\Support\Contracts\ArgBuilderDirective;
+use Nuwave\Lighthouse\Support\Contracts\FieldBuilderDirective;
 
-class BuilderDirective extends BaseDirective implements ArgBuilderDirective
+class BuilderDirective extends BaseDirective implements ArgBuilderDirective, FieldBuilderDirective
 {
     public static function definition(): string
     {
         return /** @lang GraphQL */ <<<'GRAPHQL'
 """
-Use an argument to modify the query builder for a field.
+Manipulate the query builder with a method.
 """
 directive @builder(
   """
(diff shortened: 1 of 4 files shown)


olivernybroe:
@spawnia Not sure if you wanted to make one big pr where we add field definitions to all the directives missing it or if we should do it individually 👍

src/Schema/Directives/BuilderDirective.php
```diff
@@ -35,4 +42,22 @@ public function handleBuilder($builder, $value): object
...
+    /**
+     * Dynamically call a user-defined method to enhance the builder on a field.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     */
+    public function handleFieldBuilder(object $builder): object
+    {
+        $resolver = $this->getResolverFromArgument('method');
+        $value = $this->directiveArgValue('value');
```
spawnia, line 55: `null` might be a valid value to pass, use `directiveHasArg` to check if `value` is defined at all.

spawnia:
One by one is fine with me, thanks!

One other thing: I appreciate the effort and discipline with the conventional commit messages, but I would prefer to stick to the current convention for now. I can see the benefits when paired with tooling that autogenerates changelogs or even releases and strong, written-down conventions that are universally agreed upon, but find them somewhat less readable then a descriptive sentence.

olivernybroe:
@spawnia Alright :) 

I changed it to just a message.

src/Schema/Directives/BuilderDirective.php
```diff
@@ -19,7 +20,13 @@ public static function definition(): string
   If you pass only a class name, the method name defaults to `__invoke`.
   """
   method: String!
-) repeatable on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+
+  """
+  Pass a value to the builder method.
```
spawnia, line 25:
```suggestion
  Pass a value to the method as the second argument after the query builder.
```

spawnia (commented):
Implementation and tests look great. Now we just need docs & changelog.

- The description of the directive could be more general now, since it does no longer use an argument necessarily.
- The directive definition must be copied to the api reference docs, and an example usage there would also be nice.

olivernybroe:
@spawnia Added some docs to it 👍

PR #1690, opened 2021-01-24
Document local development setup with native tools

olivernybroe:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Add some changes to make local development easier without the usage of a docker container.

**Breaking changes**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -300,3 +300,19 @@ Run the reports that are defined in `phpbench.json` via the command line,
 for example:
 
     vendor/bin/phpbench run --report=ast
+
+## Local setup
+It is also possible to setup the test environment without usage of the docker and make files.
+You would need the following:
+
+- PHP (at least 7.2)
+- Composer
+- Mysql 8
+- Redis 6
+
+For setting everything up, you would need to execute the following:
+
+1. run `composer install`.
+2. copy phpunit dist file to a local `cp ./phpunit.xml.dist ./phpunit.xml`.
+3. Change env parameter to your MySql and Redis connection.
+4. run `composer test`.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -300,3 +300,18 @@ Run the reports that are defined in `phpbench.json` via the command line,
...
 
     vendor/bin/phpbench run --report=ast
+
+## Local setup
+It is also possible to setup the test environment without usage of the docker and make files.
+You would need the following:
+
+- PHP (at least 7.2)
+- Composer
+- A Mysql 8 schema
```
spawnia, line 310: Can you clarify?

CONTRIBUTING.md
```diff
@@ -300,3 +300,18 @@ Run the reports that are defined in `phpbench.json` via the command line,
...
+
```
spawnia, line 311: We also have integration tests that require Redis.

spawnia (commented):
Looks like this is pretty easy to support, nicely done!

CONTRIBUTING.md
olivernybroe: I could prob. write it a bit better yeah :)

CONTRIBUTING.md
olivernybroe: Ah right. Haven't looked into those yet.

spawnia: We could also separate them into another testsuite.

olivernybroe: nah, it was pretty easy to fix it 👍

olivernybroe:
@spawnia alright, redis is also added now 👍 

Think this is rdy.

PR #1691, opened 2021-01-24
Handle builders in a typesafe way

spawnia:
Separate the handling of database query builders and the Scout query builder, since they vary wildly in feature set.

This alleviates the burden on `ArgBuilderDirectives` to add special handling for Scout builders, while centralizing the semantics of what a successful Scout search looks like.


spawnia:
@olivernybroe while adding field builder support to `@like`, i ran into type issues. I want to resolve this once and for all.

src/Schema/Directives/WhereDirective.php
```diff
@@ -34,12 +36,13 @@ public static function definition(): string
...
     /**
      * Add any "WHERE" clause to the builder.
-     *
-     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
-     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
      */
     public function handleBuilder($builder, $value): object
     {
+        if ($builder instanceof ScoutBuilder) {
+            throw new Exception("Using {$this->name()} on queries that use a Scout search is not supported.");
```
olivernybroe, line 43: Yeah, we definitely have to refactor this to a custom exception, so this string is not duplicated all the time 👍

src/Schema/Directives/WithRelationDirective.php
```diff
@@ -64,9 +65,13 @@ protected function loadRelation(Model $parent, ResolveInfo $resolveInfo): Deferr
      */
     protected function decorateBuilder(ResolveInfo $resolveInfo): Closure
     {
-        return function (object $query) use ($resolveInfo): void {
+        return function (object $builder) use ($resolveInfo): void {
+            if ($builder instanceof Relation) {
```
olivernybroe, line 69: Unsure what consequences this can have.

olivernybroe (changes requested):
Instead of adding the if statement in all directives, what if we create a new interface with method `handleScoutBuilder`?

olivernybroe:
Really nice. I think it would be awesome to solve this, I also found it quite annoying.

src/Schema/Directives/RelationDirective.php
```diff
@@ -70,6 +71,10 @@ function (Model $parent, array $args, GraphQLContext $context, ResolveInfo $reso
     protected function makeBuilderDecorator(ResolveInfo $resolveInfo): Closure
     {
         return function (object $builder) use ($resolveInfo) {
+            if ($builder instanceof Relation) {
+                $builder = $builder->getQuery();
```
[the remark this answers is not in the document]
spawnia: The tests show none, but theoretically someone might have added a typehint or code depending on a `Relation` being passed? I actually doubt it, and all that the `Relation` did was to proxy calls to `Builder´ anyway

spawnia:
@olivernybroe this is still rough around the edges, but i found a way to handle everything related to Scout separately. This change should clarify the difference between search and other kinds of querying, for directive developers and API consumers alike.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -141,14 +143,19 @@ function ($value) {
...
      * @param  array<string>  $scopes
      * @param  \Closure  $directiveFilter
      *
-     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Relations\Relation
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Laravel\Scout\Builder
      */
     public function enhanceBuilder(object $builder, array $scopes, Closure $directiveFilter = null): object
     {
+        $scoutEnhancer = new ScoutEnhancer($this, $builder);
+        if ($scoutEnhancer->containsSearch()) {
```
olivernybroe, line 155: Naming of this method makes it look like the Scout Enhancer is the one that contains search. 
Also it is not clear that you are checking for if a directive exist. 

Consider naming it something like `isInQuery` (hopefully better name than this). But my point was to show that the naming does not need to say something that specific about the implementation.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -141,14 +143,19 @@ function ($value) {
...
+            return $scoutEnhancer->enhance();
```
olivernybroe, line 156: Maybe name it `EnhanceBuilder`

src/Schema/Directives/EqDirective.php
```diff
@@ -22,16 +24,18 @@ public static function definition(): string
...
     public function handleBuilder($builder, $value): object
     {
         return $builder->where(
-            $this->directiveArgValue('key', $this->nodeName()),
+            $this->directiveArgValue('key') ?? $this->nodeName(),
+            $value
+        );
+    }
+
+    public function handleScoutBuilder(ScoutBuilder $builder, $value)
```
olivernybroe, line 35: Nice, this is soo clean.

src/Scout/ScoutEnhancer.php
```diff
@@ -0,0 +1,149 @@
...
+
+class ScoutEnhancer
+{
+    /**
+     * @var ArgumentSet
+     */
+    protected $argumentSet;
+
+    /**
+     * @var object
```
olivernybroe, line 22: We should be able to change this to the Scout builder type right?

src/SoftDeletes/TrashedDirective.php
```diff
@@ -29,28 +35,62 @@ public static function definition(): string
...
+        $model = $builder->model;
+
+        $this->assertModelUsesSoftDeletes($model);
+
+        if ($value === null) {
+            return $builder;
+        }
+
+        switch ($value) {
+            case 'with':
```
olivernybroe, line 78: Does without not work with scout?

olivernybroe (commented):
In general I think this is a really great path and cleans things up quite a lot. 

As you said it's still a bit rough, but I think the concept is totally right. 

Mostly I think the confusing logic is inside the Scout Enhancer. 

But the directives are soo clean with this approach, love it

src/Scout/ScoutEnhancer.php
spawnia: This one is actually the non-scout builder

src/SoftDeletes/TrashedDirective.php
spawnia: Apparently not, since the builder lacks the method. I see this discovery as a big win of this PR 💪

olivernybroe: Yeah, this is a perfect example of how awesome this change is!

PR #1697, opened 2021-01-26
Add benchmark to measure performance of queries with many results

k0ka:
**Changes**

I've added benchmarks to test GraphQL queries in production-like environment.


spawnia:
@k0ka nice to see you work on that part of Lighthouse! Please notify me when you get CI to pass 👍

k0ka:
I've fixed the style. And also created the following issue https://github.com/nuwave/lighthouse/issues/1699 with explanation what it tests and why.

I can add another benchmark with all middlewares on to test difference between "default" and "optimized" configurations.

spawnia:
I can confirm a signifant difference depending on the usage of field middleware, for the 10k case:
- none: 0.277
- default: 0.533
- default five times: 1.710

Looks like a great starting point to look into further optimizations!

spawnia:
I am merging this as is, just because it is a nice base to do further performance work. We can definitely still explore what we can do about the global field middleware!

Issue #1699, opened 2021-01-26
Move argumentSet/argumentSetFactory to a middleware


k0ka:
**Are you willing to provide a PR for this issue or aid in developing it? **
yes

**What problem does this feature proposal attempt to solve?**
I've pushed https://github.com/nuwave/lighthouse/pull/1697

I'm optimizing my GraphQL API server. The biggest bottlenecks I've found in lighthouse are global field middlewares. Here are test results with them on:

```
\Benchmarks\HugeResponseBench

    benchmark1..............................I9 [μ Mo]/r: 0.082 0.081 (s) [μSD μRSD]/r: 0.002s 2.44%
    benchmark100............................I9 [μ Mo]/r: 0.099 0.100 (s) [μSD μRSD]/r: 0.002s 2.16%
    benchmark10k............................I9 [μ Mo]/r: 0.856 0.853 (s) [μSD μRSD]/r: 0.019s 2.17%
```

and off (they are off in the PR):
```
\Benchmarks\HugeResponseBench

    benchmark1..............................I9 [μ Mo]/r: 0.070 0.069 (s) [μSD μRSD]/r: 0.001s 1.98%
    benchmark100............................I9 [μ Mo]/r: 0.078 0.078 (s) [μSD μRSD]/r: 0.001s 1.63%
    benchmark10k............................I9 [μ Mo]/r: 0.431 0.428 (s) [μSD μRSD]/r: 0.008s 1.85%
```

There is 2x difference for 10k fields and even for the simplest query which gains 15% performance boost. However, it's not a real problem as they can be easily turned off in the config.

The next evident bottleneck I've found is argumentSetFactory - https://github.com/nuwave/lighthouse/blob/master/src/Schema/Factories/FieldFactory.php#L90. Commenting this line increases speed by 20% in 10k scenario:

```
    benchmark1..............................I9 [μ Mo]/r: 0.068 0.068 (s) [μSD μRSD]/r: 0.002s 2.73%
    benchmark100............................I9 [μ Mo]/r: 0.077 0.076 (s) [μSD μRSD]/r: 0.001s 1.70%
    benchmark10k............................I9 [μ Mo]/r: 0.360 0.357 (s) [μSD μRSD]/r: 0.009s 2.43%
```

**Which possible solutions should be considered?**

I suggest rewriting this code as a field middleware that is enabled globally by default but can be disabled.

[...]

PR #1703, opened 2021-01-27
Fix GlobalId query if SanitizeDirective is not listed in field_directives

k0ka:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**The problem**

If `SanitizeDirective` is disabled in `lighthouse.field_directives`, there is no way to use `node` query. `GlobalIdServiceProvider` overwrites it, so the user can't add `@sanitize` to it.

**Changes**

`GlobalIdServiceProvider` checks if `lighthouse.field_directives` has `SanitizeDirective` disabled and if so adds `@sanitize` to the `node` query.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Fixed
 
-- add `@sanitize` to the `node` field definition if `SanitizeDirective` is not listed in the global `field_middleware` list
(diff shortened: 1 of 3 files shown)


spawnia:
This fix is too specific.

Some global middleware is quite essential for other directives to work. The `SanitizeDirective` handles every `ArgSanitizerDirective`, so it should be present whenever one of those is used.

k0ka:
Ok, the solution from https://github.com/nuwave/lighthouse/issues/1699 may be better. I'll try to implement it.

PR #1705, opened 2021-01-27
Clean up field middleware and slightly improve performance

spawnia:


spawnia:
@k0ka this PR provides some small performance optimizations related to field middleware. The basic idea is to do as much work as possible only once per field, not inside of the resolver call.

I am pondering if there is potential to do the argument transformations less often, if we are dealing with nested fields in a list, that step happens more often then it needs to.

src/Schema/Directives/CanDirective.php
```diff
@@ -82,22 +82,22 @@ public static function definition(): string
...
-        return $next(
-            $fieldValue->setResolver(
-                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($previousResolver) {
-                    $gate = $this->gate->forUser($context->user());
-                    $ability = $this->directiveArgValue('ability');
-                    $checkArguments = $this->buildCheckArguments($args);
+        $fieldValue->setResolver(
+            function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($ability, $previousResolver) {
+                $gate = $this->gate->forUser($context->user());
+                $checkArguments = $this->buildCheckArguments($args);
```
k0ka, line 90: `buildCheckArguments` is querying directive args as well. We have to pass `injectArgs`/`args` to it.

Or it may be a better idea to cache arguments in `directiveArgValue`/`directiveHasArgument` functions - so after the first query to AST, they will be saved locally. Or even load all directive args on the first query. The directive code will be simpler and easier to understand, while performance should be almost the same.

spawnia: The difficulty with this particular code is that PHP has no easy way of representing `undefined`, and there is a difference between:

```graphql
@can(args: null) // Should call check($ability, [null])
@can // Should call check($ability, [])
```

The `directiveArgLookup` is not that expensive, so I think we can just leave this as is.

spawnia: Centralized loading of directive args might be a good idea, I thought about implementing a generalized solution that runs validation during the schema compilation.

k0ka: Here is what I meant to do:
https://github.com/nuwave/lighthouse/pull/1707
Args are loaded once and can be swiftly used in any place later.

spawnia: Since #1707 landed, this no longer costs much performance, so I am going to leave it.

PR #1707, opened 2021-01-28
Cache directive arguments in BaseDirective

k0ka:
**Changes**

`BaseDirective` loads all arguments and caches them after the first `directiveHasArgument`/`directiveArgValue` call

**Breaking changes**

no

### DIFF
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -93,12 +93,8 @@ abstract class BaseDirective implements Directive
     {
         $this->directiveArgs = [];
         foreach ($this->directiveNode->arguments as $node) {
-            if (! property_exists($node, 'name')) {
-                throw new Exception('Expected a Node with a name property, got: '.get_class($node));
-            }
-
             if (array_key_exists($node->name->value, $this->directiveArgs)) {
-                continue;
+                throw new DefinitionException("Directive {$this->directiveNode->name->value} has two arguments with the same name {$node->name->value}");
             }
 
             $this->directiveArgs[$node->name->value] = AST::valueFromASTUntyped($node->value);
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -46,9 +46,11 @@ abstract class BaseDirective implements Directive
     /**
      * Cached directive arguments.
      *
-     * @var array<string, mixed>|null
+     * Lazily initialized.
+     *
+     * @var array<string, mixed>
      */
-    protected $directiveArgs = null;
+    protected $directiveArgs;
 
     /**
      * Returns the name of the used directive.
@@ -68,7 +70,6 @@ abstract class BaseDirective implements Directive
     {
         $this->directiveNode = $directiveNode;
         $this->definitionNode = $definitionNode;
-        $this->directiveArgs = null;
 
         return $this;
     }
@@ -109,7 +110,7 @@ abstract class BaseDirective implements Directive
      */
     public function directiveHasArgument(string $name): bool
     {
-        if ($this->directiveArgs == null) {
+        if (! isset($this->directiveArgs)) {
             $this->loadArgValues();
         }
 
@@ -124,7 +125,7 @@ abstract class BaseDirective implements Directive
      */
     protected function directiveArgValue(string $name, $default = null)
     {
-        if ($this->directiveArgs == null) {
+        if (! isset($this->directiveArgs)) {
             $this->loadArgValues();
         }
 
--- a/src/Schema/Directives/BaseDirective.php
+++ b/src/Schema/Directives/BaseDirective.php
@@ -63,7 +63,7 @@ abstract class BaseDirective implements Directive
     /**
      * The hydrate function is called when retrieving a directive from the directive registry.
      *
-     * @param  ScalarTypeDefinitionNode|ObjectTypeDefinitionNode|FieldDefinitionNode|InputValueDefinitionNode|InterfaceTypeDefinitionNode|UnionTypeDefinitionNode|EnumTypeDefinitionNode|EnumValueDefinitionNode|InputObjectTypeDefinitionNode $definitionNode
+     * @param  ScalarTypeDefinitionNode|ObjectTypeDefinitionNode|FieldDefinitionNode|InputValueDefinitionNode|InterfaceTypeDefinitionNode|UnionTypeDefinitionNode|EnumTypeDefinitionNode|EnumValueDefinitionNode|InputObjectTypeDefinitionNode  $definitionNode
      * @return $this
      */
     public function hydrate(DirectiveNode $directiveNode, Node $definitionNode): self
@@ -120,7 +120,7 @@ abstract class BaseDirective implements Directive
     /**
      * Get the value of an argument on the directive.
      *
-     * @param (mixed|null) $default
+     * @param  mixed|null  $default
      * @return mixed|null
      */
     protected function directiveArgValue(string $name, $default = null)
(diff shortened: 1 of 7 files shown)


src/Schema/Directives/BaseDirective.php
```diff
@@ -52,13 +61,14 @@ public function name(): string
     /**
      * The hydrate function is called when retrieving a directive from the directive registry.
      *
-     * @param  ScalarTypeDefinitionNode|ObjectTypeDefinitionNode|FieldDefinitionNode|InputValueDefinitionNode|InterfaceTypeDefinitionNode|UnionTypeDefinitionNode|EnumTypeDefinitionNode|EnumValueDefinitionNode|InputObjectTypeDefinitionNode  $definitionNode
+     * @param ScalarTypeDefinitionNode|ObjectTypeDefinitionNode|FieldDefinitionNode|InputValueDefinitionNode|InterfaceTypeDefinitionNode|UnionTypeDefinitionNode|EnumTypeDefinitionNode|EnumValueDefinitionNode|InputObjectTypeDefinitionNode $definitionNode
```
spawnia, line 64: The double spaces are supposed to be there, Laravel coding style.

src/Schema/Directives/BaseDirective.php
```diff
@@ -52,13 +61,14 @@ public function name(): string
...
      * @return $this
      */
     public function hydrate(DirectiveNode $directiveNode, Node $definitionNode): self
     {
         $this->directiveNode = $directiveNode;
         $this->definitionNode = $definitionNode;
+        $this->directiveArgs = null;
```
spawnia, line 71:
```suggestion
```

src/Schema/Directives/BaseDirective.php
```diff
@@ -41,6 +43,13 @@ abstract class BaseDirective implements Directive
      */
     protected $definitionNode;
 
+    /**
+     * Cached directive arguments.
+     *
+     * @var array<string, mixed>|null
+     */
+    protected $directiveArgs = null;
```
spawnia, line 51:
```suggestion
     * Lazily initialized.
     *
     * @var array<string, mixed>
     */
    protected $directiveArgs;
```

We don't want `null` assigned to this property, it is just uninitialized at first.

src/Schema/Directives/BaseDirective.php
```diff
@@ -76,36 +86,63 @@ public function getResolverFromArgument(string $argumentName): Closure
...
 
     /**
-     * The name of the node the directive is defined upon.
+     * Does the current directive have an argument with the given name?
      */
-    protected function nodeName(): string
+    public function directiveHasArgument(string $name): bool
     {
-        return $this->definitionNode->name->value;
+        if ($this->directiveArgs == null) {
```
spawnia, line 112:
```suggestion
        if (! isset($this->directiveArgs)) {
```

src/Schema/Directives/BaseDirective.php
```diff
@@ -76,36 +86,63 @@ public function getResolverFromArgument(string $argumentName): Closure
...
+        }
+
+        return array_key_exists($name, $this->directiveArgs);
     }
 
     /**
      * Get the value of an argument on the directive.
      *
-     * @param  mixed|null  $default
+     * @param mixed|null $default
```
spawnia, line 122: nope

src/Schema/Directives/BaseDirective.php
```diff
@@ -76,36 +86,63 @@ public function getResolverFromArgument(string $argumentName): Closure
...
      * @return mixed|null
      */
     protected function directiveArgValue(string $name, $default = null)
     {
-        return ASTHelper::directiveArgValue($this->directiveNode, $name, $default);
+        if ($this->directiveArgs == null) {
```
spawnia, line 127:
```suggestion
        if (! isset($this->directiveArgs)) {
```

spawnia (changes requested):

spawnia:
Looks good! A few discussion points:
- calling `directiveArgValue()` outside of the resolver in `FieldMiddleware` is still generally worth it
- this causes all directive arguments to be loaded, even if only one or few were actually needed

My instinct is that overall performance should still be better.

k0ka:
> this causes all directive arguments to be loaded, even if only one or few were actually needed

I think there are only two real options:
1. The directive doesn't use any args
2. The directive does use them and use them all, because if some of args are being unused, the user may just delete them.

src/Schema/Directives/BaseDirective.php
```diff
@@ -76,19 +87,34 @@ public function getResolverFromArgument(string $argumentName): Closure
...
     {
-        return ASTHelper::directiveHasArgument($this->directiveNode, $name);
+        $this->directiveArgs = [];
+        foreach ($this->directiveNode->arguments as $node) {
+            if (! property_exists($node, 'name')) {
+                throw new Exception('Expected a Node with a name property, got: '.get_class($node));
+            }
+
+            if (array_key_exists($node->name->value, $this->directiveArgs)) {
+                continue;
```
k0ka, line 101: I'm not sure about this check.

It ensures that if there is more than one argument with the same name, the first one will be taken (as it does now). It makes no sense to specify two arguments with the same name, but it works now and GraphQL specification doesn't forbid this explicitly (at least I haven't found it). 

I can add a test to check this case.

spawnia: This should throw a `DefinitionException`. A test for this would be great.

src/Schema/Directives/BaseDirective.php
```diff
@@ -76,19 +87,34 @@ public function getResolverFromArgument(string $argumentName): Closure
...
+            }
+
+            $this->directiveArgs[$node->name->value] = AST::valueFromASTUntyped($node->value);
+        }
     }
 
     /**
-     * The name of the node the directive is defined upon.
+     * Does the current directive have an argument with the given name?
```
spawnia, line 109:
```suggestion
     * Does the current directive have an argument with the given name?
     * TODO change to protected in v6
```

src/Schema/Directives/BaseDirective.php
```diff
@@ -99,7 +125,19 @@ protected function nodeName(): string
      */
     protected function directiveArgValue(string $name, $default = null)
     {
-        return ASTHelper::directiveArgValue($this->directiveNode, $name, $default);
+        if (! isset($this->directiveArgs)) {
+            $this->loadArgValues();
+        }
+
+        return array_key_exists($name, $this->directiveArgs) ? $this->directiveArgs[$name] : $default;
```
spawnia, line 133: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#multiline-ternary-expressions

```suggestion
        return array_key_exists($name, $this->directiveArgs)
            ? $this->directiveArgs[$name]
            : $default;
```

src/Testing/TestingServiceProvider.php
```diff
@@ -23,7 +23,7 @@ static function (): string {
 
     public function register(): void
     {
-        $this->app->singleton(MockDirective::class);
+        $this->app->bind(MockDirective::class);
```
spawnia, line 26: We don't need this binding at all then.

spawnia: Hold on, there is actually a difference in a singleton and static properties. The singleton will be reset when entering new test methods, whereas static properties persist.

k0ka: In this case I can revert back the singleton changes and add 
```
unset($this->directiveArgs);
```
to the BaseDirective::hydrate()

`MockDirective` is the only directive, that can be hydrated more than once.

k0ka: Or I can make `MockResolverService`, that will be a singleton and will be used by `MockDirective` and `mockResolver()`

spawnia: I prefer your second suggestion, it keeps the production code simple and lean.

k0ka:
Everything looks fine to me.

`DefinitionException` is thrown if two arguments with the same name exist.

Mocks for `MockDirective` are stored in the new signleton `MockResolverService`

spawnia (approved):
Very nice, very clean!

Issue #1677, opened 2021-01-20
OR conditions behaving like AND when using for conditions with HAS for nested relations


clayzar:
**Describe the bug**

I'm trying to query for products that are in a category or its subcategories. Categories can have one `parent`, and products can `belongToMany` `categories`. When using an array of `OR` conditions to apply to the relationship `categories` for for the `product` type, the query generated uses `AND` instead of `OR`. See the graphql query here and the SQL output further below.

```graphql
query ($where: WhereConditions) {		
  products(where: $where) {
    data {
      id
      title
    }
  }
} 

variables: {
  "where": {
    "HAS": {
      "relation": "categories",
      "condition": {
        "OR": [   // Currently both of these conditions must be met, which is incorrect and never returns products
          {
            "column": "categories.id",
            "value": 375          
          },
          {
            "HAS": {
              "relation": "parent",
              "condition": {
                "column": "id",
                "value": 375
              }
            }
          }
        ]
      }
    }
  }
}
```

If I remove one of the conditions in the `OR` array above, two different sets of products are returned, but I need the `OR` to properly apply these conditions so that I retrieve both sets of products together.

**Expected behavior/Solution**

I'd imagine this would be an update to use one of the Laravel query builder `orWhere..` clauses, but I haven't yet traced the code in lighthouse to see where the exact issue is. 

**Steps to Reproduce**

Run newly added test `Tests\Integration\WhereConditions\WhereHasConditionsDirectiveTest::testWhereHasBelongsToManyOrNestedConditions`

**Output/Logs**

<details><summary>Click to expand</summary>
  
  
The 2nd `AND` below should be an `OR` operator

```sql
SELECT count(*) AS AGGREGATE 
FROM `products` 
WHERE (EXISTS (
	SELECT * FROM `categories` 
	INNER JOIN `category_product` ON `categories`.`id` = `category_product`.`category_id` 
	WHERE `products`.`id` = `category_product`.`product_id` 
	AND (
		`categories`.`id` = 375 
		AND (EXISTS (
			SELECT * FROM `categories` AS `laravel_reserved_0` WHERE `laravel_reserved_0`.`id` = `categories`.`parent_id` AND `categories`.`id` = 375
		))
	)
))
```

</details></br>

**Lighthouse Version**

Lighthouse v5

spawnia:
A PR with a failing test would be great. You can just add to the existing tests for `@whereHas`

spawnia:
[...]

PR #1713, opened 2021-01-30
Test case for nested OR conditions for HAS relation

clayzar:
Added failing test for issue [#1677](https://github.com/nuwave/lighthouse/issues/1677)

**Describe the bug**

I'm trying to query for products that are in a category or its subcategories. Categories can have one `parent`, and products can `belongToMany` `categories`. When using an array of `OR` conditions to apply to the relationship `categories` for for the `product` type, the query generated uses `AND` instead of `OR`. See the graphql query here and the SQL output further below.

```graphql
query ($where: WhereConditions) {		
  products(where: $where) {
    data {
      id
      title
    }
  }
} 

variables: {
  "where": {
    "HAS": {
      "relation": "categories",
      "condition": {
        "OR": [   // Currently both of these conditions must be met, which is incorrect and never returns products
          {
            "column": "categories.id",
            "value": 375          
          },
          {
            "HAS": {
              "relation": "parent",
              "condition": {
                "column": "id",
                "value": 375
              }
            }
          }
        ]
      }
    }
  }
}
```

If I remove one of the conditions in the `OR` array above, two different sets of products are returned, but I need the `OR` to properly apply these conditions so that I retrieve both sets of products together.

I'd imagine this would be an update to use one of the Laravel query builder `orWhere..` clauses, but I haven't yet traced the code in lighthouse to see where the exact issue is. 

**Steps to Reproduce**

Run newly added test `Tests\Integration\WhereConditions\WhereHasConditionsDirectiveTest::testWhereHasBelongsToManyOrNestedConditions`

**Output/Logs**

The 2nd `AND` below should be an `OR` operator

```sql
SELECT count(*) AS AGGREGATE 
FROM `products` 
WHERE (EXISTS (
	SELECT * FROM `categories` 
	INNER JOIN `category_product` ON `categories`.`id` = `category_product`.`category_id` 
	WHERE `products`.`id` = `category_product`.`product_id` 
	AND (
		`categories`.`id` = 375 
		AND (EXISTS (
			SELECT * FROM `categories` AS `laravel_reserved_0` WHERE `laravel_reserved_0`.`id` = `categories`.`parent_id` AND `categories`.`id` = 375
		))
	)
))
```

</details></br>

**Lighthouse Version**

Lighthouse v5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Move Scout related classes into `\Nuwave\Lighthouse\Scout` https://github.com/nuwave/lighthouse/pull/1698
 - `BaseDirective` loads all arguments and caches them after the first `directiveHasArgument`/`directiveArgValue` call https://github.com/nuwave/lighthouse/pull/1707
 
+### Fixed
+
+- Fix nested `OR` conditions in `HAS` relations https://github.com/nuwave/lighthouse/pull/1713
+
 ### Deprecated
 
 - Specify `@guard(with: "api")` should be changed to `@guard(with: ["api"])`https://github.com/nuwave/lighthouse/pull/1705
--- a/src/Support/Utils.php
+++ b/src/Support/Utils.php
@@ -90,11 +90,11 @@ class Utils
      */
     public static function applyEach(\Closure $callback, $valueOrValues)
     {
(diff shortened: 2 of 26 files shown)


spawnia (dismissed):
Thank you, that test was the perfect kickstarter for me to fix this issue. Hopefully once and for all, with this extra check in place.

clayzar:
Thank you for the quick turn-around on this! This is great.

PR #1714, opened 2021-01-31
Add failing test for with directive on query field

olivernybroe:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Hey @spawnia, I found a bug related to the batch reworking changes.
I am not really sure how to fix it, but I added a failing testcase for you 👍


spawnia:
This never worked, i just tried and added your test case to the `4.x` branch.

olivernybroe:
Huh, I guess it just failed silently for me before? Because after updating, we started to get some exceptions

spawnia:
Yeah, i tightened the typehints from `?Model $root` to `Model $root`. When used upon a root field - a scenario that was not tested - `$root` is indeed `null`. The following code just bailed then, making `@with` a no-op.

spawnia:
This usage of `@with` probably isn't optimal, putting it on specific fields of the type corresponding to the Model should lead to better results.

olivernybroe:
Totally agree. This with should be on the specific field.

spawnia:
Alright, then this usage is a feature that should not be supported anyways. I think it is confusing for this directive to function differently, depending if it is used on a nested field or a root field. In one instance, it is loading something on the parent, in the other case for the expected children.

olivernybroe:
Yeah agreed. 

But I guess we need a case in there so it at least doesn't crash? 

Or maybe add a helpful error message?

spawnia:
We can add a validation step through the `manipulateField()` function that throws if `@with` is defined on a root field. Until then, crashing is better than silent failure.

PR #1715, opened 2021-01-31
Use gate response in authorization errors of `@can` directive

5chulzi:
- [X] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
In order to allow field authorization failures (especially in mutation queries) to return a detailed failure message, this PR changes the `@can` directive in a way, that it uses the `$gate->inspect` method (instead of `$gate->check`, which just returns a boolean) to receive an authorization response, which may hold a custom message returned by a policy method. Those Laravel responses are documented [here](https://laravel.com/docs/8.x/authorization#gate-responses).

This also changes the default error message from `You are not authorized to access {$this->nodeName()}` to the default error message of a Laravel auth access response `This action is unauthorized.` (found [here](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Auth/Access/AuthorizationException.php#L27)). I'm pretty sure this default message does not fit the context it is used in, but just wanted to get an opinion about this PR for now.

**Breaking changes**

### DIFF
--- a/src/Schema/Directives/CanDirective.php
+++ b/src/Schema/Directives/CanDirective.php
@@ -15,6 +15,7 @@ use Nuwave\Lighthouse\Schema\Values\FieldValue;
 use Nuwave\Lighthouse\SoftDeletes\ForceDeleteDirective;
 use Nuwave\Lighthouse\SoftDeletes\RestoreDirective;
 use Nuwave\Lighthouse\SoftDeletes\TrashedDirective;
+use Nuwave\Lighthouse\Support\AppVersion;
 use Nuwave\Lighthouse\Support\Contracts\FieldMiddleware;
 use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;
 use Nuwave\Lighthouse\Support\Utils;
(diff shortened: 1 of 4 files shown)


tests/Utils/Policies/UserPolicy.php
```diff
@@ -2,17 +2,28 @@
...
         return $user->name === self::ADMIN;
     }
 
+    public function superAdminOnly(User $user): Response
+    {
+        if ($user->name === self::SUPER_ADMIN) {
+            return Response::allow();
+        }
+
+        return Response::deny('Only super admins allowed');
```
spawnia, line 24: That can go in a constant and be referenced from the test to make the connection clear.

5chulzi: Done.

src/Schema/Directives/CanDirective.php
```diff
@@ -176,10 +176,20 @@ protected function authorize(Gate $gate, $ability, $model, array $arguments): vo
         // should be [modelClassName, additionalArg, additionalArg...]
         array_unshift($arguments, $model);
 
-        if (! $gate->check($ability, $arguments)) {
-            throw new AuthorizationException(
-                "You are not authorized to access {$this->nodeName()}"
-            );
+        if (is_array($ability)) {
```
spawnia, line 179: This can be replaced by `Utils::applyEach()`

spawnia (commented):
That's a great change. Thanks for paying attention to tests, docs and code style :)

lorado:
Interesting... Policy Responses are there from Laravel 6.x, never heard about them yet. 
Nice change!

src/Schema/Directives/CanDirective.php
5chulzi: Done. Makes the code look nicer, thanks for suggestion!

5chulzi:
@spawnia Thanks, I appreciate it!

@lorado You may be right about that. But either way, good point. It should be backward compatible in my opinion, too. Looking into that!

5chulzi:
Seems to be working now :) Is it okay using `AppVersion` to check for compatibility at runtime?

spawnia (approved):

PR #1720, opened 2021-02-03
Request information in manipulate result

MaartenStaa:
- [x] Added or updated tests
- [x] Documented user facing changes *(N/A?)*
- [x] Updated CHANGELOG.md

**Changes**

This adds a query field to the `ManipulateResult` event, so event listeners can know which query to manipulate the result for. We're using [a plugin that sends tracing to Apollo Studio](https://github.com/brightalley/lighthouse-apollo), which for Lighthouse v4 fetches the GraphQL request singleton, but for v5 that no longer exists.

**Breaking changes**

I guess technically it could be considered a breaking change that you now need to pass a second parameter to instantiate a `ManipulateResult`, though I imagine that's not really a typical use case. Let me know if you want me to add this to the breaking changes.

### DIFF
--- a/src/Events/ManipulateResult.php
+++ b/src/Events/ManipulateResult.php
@@ -19,8 +19,19 @@ class ManipulateResult
      */
     public $result;
 
-    public function __construct(ExecutionResult &$result)
+    /**
+     * @var \GraphQL\Language\AST\DocumentNode|string
+     */
+    public $query;
+
+    /**
+     * ManipulateResult constructor.
+     * @param ExecutionResult $result
+     * @param \GraphQL\Language\AST\DocumentNode|string $query
+     */
+    public function __construct(ExecutionResult &$result, $query)
     {
         $this->result = $result;
+        $this->query = $query;
     }
 }
--- a/src/Events/ManipulateResult.php
+++ b/src/Events/ManipulateResult.php
@@ -25,8 +25,6 @@ class ManipulateResult
     public $query;
 
     /**
-     * ManipulateResult constructor.
-     * @param ExecutionResult $result
      * @param \GraphQL\Language\AST\DocumentNode|string $query
      */
     public function __construct(ExecutionResult &$result, $query)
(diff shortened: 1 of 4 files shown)


src/GraphQL.php
```diff
@@ -239,7 +239,7 @@ function (array $errors, callable $formatter): array {
 
         // Allow listeners to manipulate the result after each resolved query
         $this->eventDispatcher->dispatch(
-            new ManipulateResult($result)
+            new ManipulateResult($result, $query)
```
spawnia, line 242: Just the query string without the operation name is ambigious, there might be multiple operations in there.

src/Events/ManipulateResult.php
```diff
@@ -19,8 +19,17 @@ class ManipulateResult
...
-    public function __construct(ExecutionResult &$result)
+    /**
+     * @var \GraphQL\Language\AST\DocumentNode|string
+     */
+    public $query;
+
+    /**
+     * @param \GraphQL\Language\AST\DocumentNode|string $query
+     */
+    public function __construct(ExecutionResult &$result, $query)
```
spawnia, line 30: We should monomorphize that type.

spawnia (changes requested):

spawnia:
We can use https://github.com/webonyx/graphql-php/pull/755 once it lands to grab the `OperationDefinitionNode` for the fitting operation.

spawnia:
I am enhancing the lifecycle events in https://github.com/nuwave/lighthouse/pull/1726, `StartExecution` now holds the executed query. You can grab it off of there.

The reason why I do not want to add the query string to `ManipulateResult` is that this would just create multiple ways of achieving the same thing. I see how it might be convenient for certain users, but a similar argument could then be made for including the query (and the request perhaps) in all other lifecycle events as well.

MaartenStaa:
@spawnia Maybe I missed some details in scrolling through that PR on my phone, but with the new events how would you able to correlate the query from the `StartExecution` to whatever result object you receive in the `ManipulateResult` or `EndExecution` event?

spawnia:
The events are guaranteed to arive in order for every executed operation. Your listener will have to be stateful and keep track of the last seen `StartExecution` event when receiving `ManipulateResult`.

PR #1725, opened 2021-02-05
Fix docs: explicit foreign key for `author_id`

ccfelix:
The "author_id" field is a foreign key with a custom name other than the default. It needs to be explicitly stated.

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**


spawnia:
Thanks

ccfelix:
👍

Issue #911, opened 2019-08-12
Support for Apollo Federation


juusopiikkila:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%240%20Funded-%2300A156.svg" />](https://issuehunt.io/r/nuwave/lighthouse/issues/911)

A while ago Apollo introduced a new strategy for merging GraphQL schemas together in a more sensible way than schema stitching. It would be great to see this supported in Lighthouse too.

https://www.apollographql.com/docs/apollo-server/federation/introduction/
https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/nuwave/lighthouse/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

spawnia:
Are you willing to work on this?

sl0wik:
I'm also searching for a way to implement it. Anyone else working on it?

juusopiikkila:
I could try but I don't think that I have a good enough understanding of the code to make that big implementations yet.

RPSimon:
I'm trying to implement this to. Does anybody found some clear information about this specification. Or examples of other implementations?

matthewhall-ca:
@RPSimon here's a Ruby implementation of it https://github.com/Gusto/apollo-federation-ruby

matthewhall-ca:
We're happy to fund development of this if anyone is interested!

renepardon:
Hey,
if no one is doing this right now, I would read the specs and start to implement this.

spawnia:
@renepardon go for it!

renepardon:
One question in advance: how do you run the tests? They run into timeout even if I do `COMPOSER_PROCESS_TIMEOUT=600 composer tests` within the docker container.
For developing I usually run one test but before I start within a new project I would like to ensure everything works as expected ;) @spawnia

renepardon:
I came across the following idea:

key Directive: https://gist.github.com/renepardon/6751168cc12bc9b32f32c7897169d112
federation Query: https://gist.github.com/renepardon/58b34ba1be427fade9cd2f8020dd1d89
federation.graphql: https://gist.github.com/renepardon/a5a65a050893d56a8a9994b70c475a11
[...]

PR #1728, opened 2021-02-08
Support for Apollo Federation

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/911

**Changes**

Allows Lighthouse to function as a federation capable service in Apollo Federation.

**Breaking changes**

No.


tristanpemble:
hello! I was in the middle of trying to implement Apollo Federation in Lighthouse myself. from what I could tell, there was no good way to extend the webonyx/graphql-php schema printer to handle printing directives due to the use of static methods. I ended up adapting it to write my own. I did not expect to share this, and went a little crazy with generators as an experiment, but it seems to function well. take whatever you need: https://gist.github.com/tristanpemble/f6d427034a33a06392f44ed17fb278dc

tristanpemble:
it would seem the repo linked in the original report has a much simpler [extension of SchemaPrinter](https://github.com/awardis/federation/blob/master/src/Schema/SchemaPrinter.php)

spawnia:
Thank you @tristanpemble, that was some good inspiration :)

tristanpemble:
from what I see most of what is needed for federation is implemented - is there anywhere else I can help (maybe just by testing it out)?

spawnia:
I am not quite sure how to interpret https://www.apollographql.com/docs/federation/federation-spec/#create-stub-types

> A federated service should take the @external fields and types and create them locally so the service can run on its own.

Do we need some kind of mocking / generator?

tristanpemble:
hm. I am definitely not an expert on this, but, I interpret that sentence as leaving the onus on the framework and not on the end-user. I'm not sure how you would automatically mock out types/fields from other federated services if the user did not provide them.

I think the take away from that section could impact this implementation could be here:

> All fields referred to in @key, @requires, and @provides directives need to have corresponding @external fields in the same service.

perhaps you could insert schema validation wrapped in helpful messaging for the library user to tell them they need to define the external type or field in their schema if it is not present.

tristanpemble:
I dug a little deeper on the official JS implementation, I think these validations will help you: https://github.com/apollographql/federation/tree/main/federation-js/src/composition/validate

specifically these are similar to what I was imagining:

- [requiresFieldsMissingExternal.ts](https://github.com/apollographql/federation/blob/main/federation-js/src/composition/validate/postComposition/requiresFieldsMissingExternal.ts)
- [providesFieldsMissingExternal.ts](https://github.com/apollographql/federation/blob/main/federation-js/src/composition/validate/postComposition/providesFieldsMissingExternal.ts)
- [externalUnused.ts](https://github.com/apollographql/federation/blob/main/federation-js/src/composition/validate/postComposition/externalUnused.ts)

I don't fully understand these two:
- [externalMissingOnBase.ts](https://github.com/apollographql/federation/blob/main/federation-js/src/composition/validate/postComposition/externalMissingOnBase.ts)
- [externalTypeMistmatch.ts](https://github.com/apollographql/federation/blob/main/federation-js/src/composition/validate/postComposition/externalTypeMismatch.ts)

anyway, I think digging through their validations will help!

Issue #1724, opened 2021-02-04
Echo subscription driver: Bug, Inconsistency, Missing documentation


ssi-anik:
**Describe the bug**
I was trying to figure out the subscription with the `echo` driver. I found the following things are wrongly done. 

- Package generates a unique **private** channel to listen to as named `private-*`
https://github.com/nuwave/lighthouse/blob/dba839e85631f2262c12d7e5c96b8b25048001f6/src/Subscriptions/Subscriber.php#L170-L173

But when broadcasting, it then broadcasts to a presence channel. which prepends `presensce-` to the channel making `presence-private-lighthouse-*`. So, it should always be handled as a presence channel? https://github.com/nuwave/lighthouse/blob/dba839e85631f2262c12d7e5c96b8b25048001f6/src/Subscriptions/Events/EchoSubscriptionEvent.php#L29-L32

- Is it intentional to keep separate event names?
https://github.com/nuwave/lighthouse/blob/dba839e85631f2262c12d7e5c96b8b25048001f6/src/Subscriptions/Events/EchoSubscriptionEvent.php#L41
https://github.com/nuwave/lighthouse/blob/dba839e85631f2262c12d7e5c96b8b25048001f6/src/Subscriptions/Broadcasters/PusherBroadcaster.php#L14

- Even if user wants to publish the subscription data in `sync`, it gets pushed to the queue.
https://github.com/nuwave/lighthouse/blob/dba839e85631f2262c12d7e5c96b8b25048001f6/src/lighthouse.php#L313
Because, it uses `ShouldBroadcast`, which is pushed to the queue. It should be [`ShouldBroadcastNow`](https://github.com/illuminate/broadcasting/blob/feb1cb1da7e7670d8fa02220e42581ff99a42d78/BroadcastManager.php#L115-L117) to send it in sync manner.
https://github.com/nuwave/lighthouse/blob/dba839e85631f2262c12d7e5c96b8b25048001f6/src/Subscriptions/Events/EchoSubscriptionEvent.php#L8

- To broadcast a message through the `echo` driver, it requires to install `illuminate/broadcasting` package when using Laravel. If using pusher, it's not required. It's not mentioned in the documentation.

**Lighthouse Version**: * v5.1.0

spawnia:
@stayallive @alberthaff what do you think?

ssi-anik:
[...]

PR #1733, opened 2021-02-17
Fix broadcasting in the echo subscription driver

ssi-anik:
- [x] Updated CHANGELOG.md
- [x] Added or updated tests

#1724 

**Changes**
- For `echo` driver-based subscriptions, the event name will be 'lighthouse-subscription'. 
- Echo driver will always broadcast through the `private` channel. 
- If want to broadcast a sync event, it will not be queued anymore. 
- Pusher driver can log entries if defined in the `broadcasting` config file.

**Breaking changes**
- The event name can be a breaking change `lighthouse-subscription`

### DIFF
--- a/src/Subscriptions/BroadcastManager.php
+++ b/src/Subscriptions/BroadcastManager.php
@@ -8,6 +8,7 @@ use Nuwave\Lighthouse\Subscriptions\Broadcasters\LogBroadcaster;
 use Nuwave\Lighthouse\Subscriptions\Broadcasters\PusherBroadcaster;
 use Nuwave\Lighthouse\Subscriptions\Contracts\Broadcaster;
 use Nuwave\Lighthouse\Support\DriverManager;
+use Psr\Log\LoggerInterface;
 use Pusher\Pusher;
 use RuntimeException;
 
@@ -53,6 +54,10 @@ class BroadcastManager extends DriverManager
         $options = Arr::get($driverConfig, 'options', []);
 
         $pusher = new Pusher($appKey, $appSecret, $appId, $options);
+
+        if ($driverConfig['log'] ?? false) {
+            $pusher->setLogger($this->app->make(LoggerInterface::class));
+        }
 
         return new PusherBroadcaster($pusher);
     }
--- a/tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
+++ b/tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
@@ -61,7 +61,7 @@ class AuthorizeRequestsTest extends TestCase
     {
         $response = $this->querySubscription();
 
-        $channel = $response->json('extensions.lighthouse_subscriptions.channels.taskUpdated');
+        $channel = $response->json('extensions.lighthouse_subscriptions.channel');
         $this
             ->postJson('graphql/subscriptions/auth', [
                 'channel_name' => 'anything-before-'.$channel,
--- a/tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
+++ b/tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
@@ -44,7 +44,7 @@ class AuthorizeRequestsTest extends TestCase
     {
         $response = $this->querySubscription();
 
-        $channel = $response->json('extensions.lighthouse_subscriptions.channels.taskUpdated');
+        $channel = $response->json('extensions.lighthouse_subscriptions.channel');
         $this
             ->postJson('graphql/subscriptions/auth', [
                 'channel_name' => 'presence-'.$channel,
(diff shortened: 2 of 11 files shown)


src/Subscriptions/BroadcastManager.php
```diff
@@ -54,6 +55,10 @@ protected function createPusherDriver(array $config): PusherBroadcaster
 
         $pusher = new Pusher($appKey, $appSecret, $appId, $options);
 
+        if ($driverConfig['log'] ?? false) {
```
spawnia, line 58: If this option is not documented, how are users supposed to know about it? Please add a default value and short description in `lighthouse.php`.

tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
```diff
@@ -27,6 +27,23 @@ public function testEchoClientAuthorizesSuccessfully(): void
     {
         $response = $this->querySubscription();
 
+        $channel = $response->json('extensions.lighthouse_subscriptions.channels.taskUpdated');
```
spawnia, line 30:
```suggestion
        $channel = $response->json('extensions.lighthouse_subscriptions.channel');
```

tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
```diff
@@ -40,14 +57,26 @@ public function testEchoClientAuthorizesSuccessfully(): void
             ]);
     }
 
+    public function testEchoClientAuthorizationFailsOtherThanPresenceChannel(): void
+    {
+        $response = $this->querySubscription();
+
+        $channel = $response->json('extensions.lighthouse_subscriptions.channels.taskUpdated');
```
spawnia, line 64:
```suggestion
        $channel = $response->json('extensions.lighthouse_subscriptions.channel');
```

spawnia (changes requested):

src/Subscriptions/BroadcastManager.php
```diff
@@ -54,6 +55,10 @@ protected function createPusherDriver(array $config): PusherBroadcaster
...
+            $pusher->setLogger($this->app->make(LoggerInterface::class));
```
spawnia, line 59: Can you cover this with a test?

src/Subscriptions/BroadcastManager.php
ssi-anik: this log key belongs to the `broadcasting` config file, not the lighthouse config. And if you see Laravel's config, that's also not even in the config and documented. https://github.com/laravel/laravel/blob/ebf2646c347b941e63709f7e69ab79416f6d5124/config/broadcasting.php#L33-L42

I found out while going through their code. So, thought to implement the same.

src/Subscriptions/BroadcastManager.php
ssi-anik: May try. Can't assure.

tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
ssi-anik: I don't understand it. Do you mean to update the variable name because of the purpose? It's been all over the class, should I change each of them as well or just the test I've added?

src/Subscriptions/BroadcastManager.php
ssi-anik: Added test for logger.

tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
spawnia: I changed existing uses https://github.com/nuwave/lighthouse/pull/1735, you can just apply the suggestion,.

src/Subscriptions/BroadcastManager.php
spawnia: I see, thank you.

tests/Integration/Subscriptions/EchoBroadcaster/AuthorizeRequestsTest.php
```diff
@@ -28,6 +28,23 @@ public function testEchoClientAuthorizesSuccessfully(): void
...
+                    'user_id', 'user_info',
+                ],
+            ]);
+    }
+
+    public function testEchoClientAuthorizesPresenceChannelForBackwardCompatibility(): void
+    {
+        $response = $this->querySubscription();
+
+        $channel = $response->json('extensions.lighthouse_subscriptions.channels.taskUpdated');
```
spawnia, line 47:
```suggestion
        $channel = $response->json('extensions.lighthouse_subscriptions.channel');
```

spawnia (approved):
Nicely done, thank you

PR #1739, opened 2021-02-23
Prepend rule arguments that refer to other arguments with the full path

sudo-rm-rf-star:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Produce a test case to show `required_without` rule is not working as expected during validation. 
The `testValidateRequiredWithout` does not pass, but should pass and `testValidateRequiredWithout_Unexpected` does pass but should not pass. A validator should not need to hardcode its name within the rules for the rules to work.

**Breaking changes**
No changes


spawnia:
Is there a way for us to get an intermediate represenation of the parsed validator rules from Laravel? I am looking for some kind of structure where it is easy for us to find rules that contain a path, such as `required_without`, and prepend the full input path.

spawnia:
@sudo-rm-rf-star sending it back to you for re-review, what do you think?

src/Validation/RulesGatherer.php
```diff
@@ -175,20 +191,53 @@ protected function wrap(array $rulesOrMessages, array $path): array
...
+                /**
+                 * @var array{
+                 *   0: string,
+                 *   1: array<int, mixed>,
+                 * } $parsed
+                 */
+                $parsed = ValidationRuleParser::parse($rule);
+
+                $name = $parsed[0];
+                if ($name === 'RequiredWithout') {
```
sudo-rm-rf-star, line 231: Is there a clean way of making this more abstract to work with any validator of this format? 
e.g. `RequiredWith`, `RequiredWithAll`, `RequiredWithoutAll`, `RequiredUnless`, `Same`

sudo-rm-rf-star:
Thanks a lot for your efforts! Just left one comment.

src/Validation/RulesGatherer.php
spawnia: We can treat all rules that have the argument reference as the first argument in the same way. Are there any rules that also reference args but in another position, e.g. `some_rule:3,arg_name`?

If you could dig in and provide a comprehensive list, that would be of great help :)

sudo-rm-rf-star: [This array](https://github.com/illuminate/validation/blob/master/Validator.php#L198-L225) gives us a complete list of validation rules which depend on other fields as parameters.  However, it both lists validation rules which are dependant on scalar values (e.g. `after:date`)  as references to other attributes (e.g. `required_with`)

Combining this information with the [laravel docs on validation](https://laravel.com/docs/8.x/validation), we can extract the following complete list of validation rules where one of the arguments are references to other fields.
```
    /**
     * The validation rules which depend on other attributes as parameters.
     *
     * @var string[]
     */
    protected $dependentRules = [
        'Different',
        'ExcludeIf',
        'ExcludeUnless',
        'ExcludeWithout',
        'Gt',
        'Gte',
        'Lt',
        'Lte',
        'RequiredIf',
        'RequiredUnless',
        'RequiredWith',
        'RequiredWithAll',
        'RequiredWithout',
        'RequiredWithoutAll',
        'Same'
    ];
```

All of these validation rules take `field` or `fields` as their first parameter. `exclude_if` and `exclude_unless` take a value as second argument.

PR #1741, opened 2021-02-24
Support `pusher/pusher-php-server` version 5

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Pusher 5 now throws an exception when triggering the broadcast fails, we catch that exception and report it to the framework so the user can debug.

See changes: https://github.com/pusher/pusher-http-php/blob/master/CHANGELOG.md#500-2021-02-19

As far is I can tell the catch with the new exception has no actual effect when the exception class doesn't exist and it has no negative effects having it there for 5.x, code still works fine with 4.x (it just doesn't enter the "catch" ever).

**Breaking changes**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Add events to cover the lifecycle of a GraphQL request: `EndExecution`, `EndRequest` https://github.com/nuwave/lighthouse/pull/1726
 - Include the client given query, variables and operation name in the `StartExecution` event https://github.com/nuwave/lighthouse/pull/1726
 - Apply `log` option from the `broadcasting` config to the Pusher subscription driver https://github.com/nuwave/lighthouse/pull/1733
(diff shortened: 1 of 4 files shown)


spawnia (dismissed):
Nicely done

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Add events to cover the lifecycle of a GraphQL request: `EndExecution`, `EndRequest` https://github.com/nuwave/lighthouse/pull/1726
 - Include the client given query, variables and operation name in the `StartExecution` event https://github.com/nuwave/lighthouse/pull/1726
 - Apply `log` option from the `broadcasting` config to the Pusher subscription driver https://github.com/nuwave/lighthouse/pull/1733
+- Support for `pusher/pusher-php-server` version 5.0 https://github.com/nuwave/lighthouse/pull/1741
```
spawnia, line 18:
```suggestion
- Support `pusher/pusher-php-server` version 5 https://github.com/nuwave/lighthouse/pull/1741
```

PR #1759, opened 2021-03-10
Fix mget call with empty subscribers

### LINKED ISSUE #1757: ERR wrong number of arguments for 'mget' command
Laravel 8
Php 8
Postgresql 13

I am trying to implement subscription as specified in docs but after setup, when i try to run mutation

```

<?php

namespace App\GraphQL\Mutations;

use App\Models\GroupMessage;
use Nuwave\Lighthouse\Execution\Utils\Subscription;

class SendGroupMessage
{
    /**
     * @param  null  $_
     */
    public function __invoke($_, array $args)
    {
        extract($args);

        $message =  GroupMessage::create([
            'sender_id' => auth()->id(),
            'poll_id' => $poll_id,
            'message' => $message
        ]);

        Subscription::broadcast('groupMessageCreated', $message);

        return $message;
    }
}
```

I am getting this error:-

thekonz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1757 

**Changes**

Fix the mget call when there are no subscribers

### DIFF
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -64,6 +64,9 @@ class RedisStorageManager implements StoresSubscriptions
         // Since we store the individual subscribers with a prefix,
         // but not in the set, we have to add the prefix here.
         $subscriberIds = array_map([$this, 'channelKey'], $subscriberIds);
+        if (empty($subscriberIds)) {
+            return new Collection();
+        }
 
         // Using the mget command, we can retrieve multiple values from redis.
         // This is like using multiple get calls (getSubscriber uses the get command).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Apply error handling and debug settings consistently https://github.com/nuwave/lighthouse/pull/1749
 - Fix typo `comparision` to `comparison` in generated input types for `@whereHas`
+- Fix redis `mget` being called with an empty list of subscriber ids https://github.com/nuwave/lighthouse/pull/1759
 
 ## 5.3.0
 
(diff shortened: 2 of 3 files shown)


spawnia:
@thekonz thanks for digging in and owning up to the code you contributed :) Invited you as a collaborator, you can join our Slack so I can invite you to our collaborators channel, too https://join.slack.com/t/lighthouse-php/shared_invite/zt-4sm280w1-wu21r94f3kLRRtBXRbXVfw

thekonz:
Not sure what is happening with the schema tests, but I could reproduce the mget issue with a test and the early return in the storage manager fixes it.

spawnia:
Btw, fixing the test failure in https://github.com/nuwave/lighthouse/pull/1760

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -64,6 +64,9 @@ public function subscribersByTopic(string $topic): Collection
         // Since we store the individual subscribers with a prefix,
         // but not in the set, we have to add the prefix here.
         $subscriberIds = array_map([$this, 'channelKey'], $subscriberIds);
+        if (empty($subscriberIds)) {
```
spawnia, line 67:
```suggestion
        if (count($subscriberIds) === 0) {
```

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -64,6 +64,9 @@ public function subscribersByTopic(string $topic): Collection
         // Since we store the individual subscribers with a prefix,
         // but not in the set, we have to add the prefix here.
         $subscriberIds = array_map([$this, 'channelKey'], $subscriberIds);
+        if (count($subscriberIds) === 0) {
```
spawnia, line 67: Can we do this return even earlier?

src/Subscriptions/Storage/RedisStorageManager.php
thekonz: Any reason for that?

src/Subscriptions/Storage/RedisStorageManager.php
thekonz: Oh yes we could, true.

src/Subscriptions/Storage/RedisStorageManager.php
spawnia: Yup, empty is just generally too loose.

spawnia (approved):
Nice work, thank you

PR #1766, opened 2021-03-23
Fix documentation of getting an argument inside a validator in UPGRADE.md

cwdt:
**Changes**

Small documentation fix of getting an argument inside a validator in the UPGRADE.md file.


spawnia (approved):
Good find and well done.

PR #1768, opened 2021-03-25
Set and restore the Lighthouse guard name for authenticated subscription iterator

### LINKED ISSUE #1607: "@guard" directive is fired, fails on queued subscription broadcast
**Describe the bug**

The @guard directive requires general login auth to use the GraphQL services within its scope. When a Subscription is guarded with @guard with a Redis broadcaster set up to handle the broadcasting service, the Guard directive is run on the job queue instance which is processing the broadcast event, at which point the context has changed. An application using the "api" auth guard will not fail the @guard test because the authorization header is missing, but there's no reason for @guard to run in this context because it amounts to the application guarding against itself.

If the broadcaster is configured to not queue, the task completes as expected, because it is fulfilled in the main thread.

Guard directive should not be applied to Laravel job queue instances or some other method of disabling on subscription events should be provided.

**Steps to reproduce**

1. Set up subscription with Redis-based broadcaster such as https://github.com/thekonz/lighthouse-redis-broadcaster/pulls
2. Broadcast an event on a subscription with "@guard" directive.
3. Check the event contents and see it will always be "null" value. The auth error is stripped from the data set before packaging for broadcast.

**Output/Logs**

See description

**Lighthouse Version**

Latest

**References**

Past issue regarding auth on subscription: https://github.com/nuwave/lighthouse/issues/1302
PR which provides the user in context without actually performing authentication, thus auth checks still fail. https://github.com/nuwave/lighthouse/pull/1306

stayallive:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #1607.

**Changes**

This fixes subscription with `@guard` (without argument) not having the correct authentication context available.

I do however have a hard time thinking on how to test this to be honest so until I have the time to sit down for that I created this PR to at least have it documented since you can copy the class to your own app to start using these changes.

**Breaking changes**

I don't believe so since I consider this to be a bugfix.

### DIFF
--- a/src/Subscriptions/Iterators/AuthenticatingSyncIterator.php
+++ b/src/Subscriptions/Iterators/AuthenticatingSyncIterator.php
@@ -49,31 +49,33 @@ class AuthenticatingSyncIterator implements SubscriptionIterator
         /** @var \Nuwave\Lighthouse\Subscriptions\SubscriptionGuard $guard */
         $guard = $this->authFactory->guard(SubscriptionGuard::GUARD_NAME);
 
-        $subscribers->each(static function (Subscriber $item) use ($handleSubscriber, $handleError, $guard): void {
-            // If there is an authenticated user set in the context, set that user as the authenticated user
-            $user = $item->context->user();
-            if ($user !== null) {
-                $guard->setUser($user);
-            }
-
-            try {
-                $handleSubscriber($item);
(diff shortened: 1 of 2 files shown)


spawnia:
Thanks @stayallive for the work you did here, I incorporated the changes into https://github.com/nuwave/lighthouse/pull/1769

PR #1777, opened 2021-03-31
Failing test for required_without in array

wimski:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

The validation rule `required_without` doesn't seem to work properly (within arrays). I've added a test which should pass, but fails. This is not so much intended as an actual PR, but more as a demonstration of a bug. If it's not a bug, can someone tell me what I did wrong?

It produces the following error which shouldn't occur.
```json
{
    "errors": [
        {
            "message": "Validation failed for field [createStuff].",
            "extensions": {
                "validation": {
                    "input.items.0.thing.an_id": [
                        "The input.items.0.thing.an_id field is required when items.0.thing.some data is not present."
                    ]
                },
                "category": "validation"
            }
        }
    ]
}
```

### DIFF
--- a/tests/Integration/Validation/RequiredWithoutTest.php
+++ b/tests/Integration/Validation/RequiredWithoutTest.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Tests\Integration\Validation;
+
+use Tests\TestCase;
+
+class RequiredWithoutTest extends TestCase
+{
+    public function testValidatesRequiredWithoutInArray(): void
+    {
+        $this->schema = /** @lang GraphQL */ '
+            type Query {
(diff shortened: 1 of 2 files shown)


spawnia:
Related to https://github.com/nuwave/lighthouse/pull/1739, i will probably pull in the test case and solve both there.

wimski:
> Related to #1739, i will probably pull in the test case and solve both there.

Is this related though? I'm already using the full path in my rules.

spawnia:
> I'm already using the full path in my rules.

Actually not, `input` is not referenced.

wimski:
Ah, `input` is also part of the full path! That wasn't clear to me at all. This solved my issue.

spawnia:
Fixed with https://github.com/nuwave/lighthouse/pull/1739

PR #1779, opened 2021-03-31
fix redis subscription store to pass correct syntax to redis to set ttl

ninetwentyfour:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Using redis with a ttl results in an error:

`ERR syntax error`

when trying to set a subscriber.

The problem is that predis expects the set with ttl as `command('set', [$key, $value, 'EX', $ttl])`
but the code is passing `command('set', [$key, $value, $ttl])`

This correctly sets a ttl on the items in redis.

**Breaking changes**

### DIFF
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -102,7 +102,8 @@ class RedisStorageManager implements StoresSubscriptions
             $this->serialize($subscriber),
         ];
         if ($this->ttl !== null) {
-            $setArguments [] = $this->ttl;
+            $setArguments[] = 'EX';
+            $setArguments[] = $this->ttl;
         }
         $this->connection->command('set', $setArguments);
     }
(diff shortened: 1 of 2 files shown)


thekonz:
related: #1684 

So one client expects 3, the other 4? -.-

spawnia:
Looks like @alberthaff was using `phpredis` and based https://github.com/nuwave/lighthouse/pull/1685 off of that. @ninetwentyfour are you using `predis`?

ninetwentyfour:
@spawnia yeah, sorry I should have included that I am using predis v1.1.6

spawnia:
Can we check which driver is used before making the call and adapt it accordingly?

thekonz:
Yes I think there is a `$connection->client()`.
Either we check for which client it is or we just fire another `$connection->command('expire', ...`.
I would probably choose the latter. What do you think?

spawnia:
I would rather keep the parameterised `set`, because it is an atomic operation and thus safer if the PHP process dies.

thekonz:
I fixed the issue by using the setex command. See #1814

PR #1783, opened 2021-04-04
Fix the tinker command for user factory

saadj55:
Fix the tinker command for user factory

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**


spawnia:
Thank you

PR #1788, opened 2021-04-08
Use selected cache store on clear-cache command

stayallive:
The current method only clears the cache for the default cache store of the application and did not read the (optional) custom cache store set in the Ligthouse config.

- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The current method only clears the cache for the default cache store of the application and did not read the (optional) custom cache store set in the Ligthouse config.

I've used this code as "inspiration":

https://github.com/nuwave/lighthouse/blob/9e37fb38e0430f8594146466b29c6032dbc72650/src/Schema/AST/ASTBuilder.php#L99-L106

**Breaking changes**

None (well, actually clearing the cache now, but that is a good "breaking" change).


spawnia (approved):

stayallive:
I'm an idiot... of course @spawnia haha, thanks for fixing!

spawnia:
All good @stayallive, your code worked fine 👍 It is just that using interface based dependency injection improves static validation and let's me sleep at night.

Issue #1794, opened 2021-04-14
Problem with table name on HAS condition with relationship on same table


akyrey:
**Describe the bug**
I think there's a bug with `WhereConditionsBaseDirective` using `HAS` condition when the relation is between the same model.  
The `prefixConditionWithTableName` method doesn't take in consideration possible renaming of the joined table/model.

For example, I have a `Location` model that can be represented by a tree, having a `parent_id` foreign key column and 2 relationships, _parent_ and _children_, that allow to retrieve related nodes.
My schema is similar to the following:
```graphQL
type Location {
    id: ID!
    parent: Location @belongsTo
    children: [Location!] @hasMany
}

enum LocationColumn {
    ID @enum(value: "id")
}

extend type Query {
    locations(
        where: _ @whereConditions(columnsEnum: "LocationColumn"),
        hasParent: _ @whereHasConditions(columnsEnum: "LocationColumn"),
        hasChildren: _ @whereHasConditions(columnsEnum: "LocationColumn")
    ): [Location!]!
}
```

The models i currently have saved in the database are
```json
[
  {
    "id": 1,
    "parent_id": null,
  },
  {
    "id": 2,
    "parent_id": 1,
  }
]
```

The actual query performed using this code
```graphQL
query locations($hasChildren: QueryLocationsHasChildrenWhereHasConditions) {
  locations(hasChildren: $hasChildren) {
    id
    parent {
      id
    }
    children {
      id
    }
  }
}
```
and these variables
```graphQL
{
  "hasChildren": {
    "column": "ID",
    "value": 2
  }
}
```
is
```sql
select *
from `locations`
where (exists(select *
              from `locations` as `laravel_reserved_0`
              where `locations`.`id` = `laravel_reserved_0`.`parent_id`
                and `locations`.`id` = 8))
limit 1000 offset 0;
```
I have the same problem the other way around, performing the query using the _hasParent_ condition

**Expected behavior/Solution**

I'd expect the query to be like this
```sql
select *
from `locations`
where (exists(select *
              from `locations` as `laravel_reserved_0`
              where `locations`.`id` = `laravel_reserved_0`.`parent_id`
                and `laravel_reserved_0`.`id` = 8))
limit 1000 offset 0;
```
to notice last line of the inner select

**Lighthouse Version**

5.3.0

PR #1795, opened 2021-04-14
Fix problem with table name on HAS condition with relationship on same table

akyrey:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Tests and fix for #1794

Changing the code like this, using directly the builder instead of the model, make it work correctly. Let me know if I'm missing something.

I've removed `nestedRelatedModel` method since this will be handled internally by the builder itself

**Breaking changes**

None


akyrey:
Changed `handleHasCondition` method to use builder itself instead of custom method to retrieve table name

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -178,18 +176,6 @@ protected static function assertValidColumnReference(string $column): void
         }
     }
 
-    protected function nestedRelatedModel(Model $model, string $nestedRelationPath): Model
```
spawnia, line 181: Pretty sure that change is just straight up removing a feature. Looks like it was not tested, but I can see the use case behind it.

akyrey: I'll try to create a test for a nested relationship and let you know. I think that could be the potential break

akyrey: I tried to create the test `testWhereConditionsHasNestedTables` where I traverse multiple nested relationships and it still pass. I'm unsure if it is what you meant though. Checking `WhereConditionsDirectiveTest` class a can see many tests similar to this one, with the only difference in this case the same table is used multiple times.

My understanding is that the logic from `nestedRelatedModel` is already implemented in `whereHas`/`has` Builder method, but I could be completely wrong

spawnia: The test you wrote is not really representative of the potential issue, since no matter how deep you nest `parent.parent....`, it always stays within the same model.

I created https://github.com/nuwave/lighthouse/pull/1807/files, but it seems there are some problems with it. Looks like the implementation is already bugged.

akyrey: Ah I understand now. It's not using the inner most relation table name. I'll check if I'm able to fix this

akyrey: I just found out the problem was related to `posts` relationship in `Category` model. Related class was set to `Category` instead of `Post`.

Changing that both the hold implementation and the new one work as expected (at least the test case you created passes)

spawnia: Great find, thank you.

spawnia (approved):
Nicely done, thanks for working through this with me.

PR #1798, opened 2021-04-14
Fix webonyx.github.io/graphql-php documentation links

asolopovas:
Hi, I have discovered that few 'webonyx.github.io/graphql-php' links are broken, I have fixed them for the master branch and for 5.3


spawnia:
Thanks. It was indeed me who broke them 🙈

Issue #1787, opened 2021-04-07
Suggestion: Exclude "lighthouse_subscriptions" extension when no channel provided


SirLamer:
**What problem does this feature proposal attempt to solve?**

If subscriptions are in use, every API query will always include an "extensions.lighthouse_subscriptions" property, even when the query response is not expecting a subscription channel.

**Which possible solutions should be considered?**

Exclude "lighthouse_subscriptions" from the response when no new channel is provided. If no other extensions are in use then the "extensions" property itself should also be excluded. The purpose is to reduce the size of API responses by a few bytes.

This is potentially a breaking change though a good client implementation should already anticipate this will be excluded without cause to avoid dependency problems.

```js
  getChannelFromResponse(response) {
    return response?.extensions?.lighthouse_subscriptions?.channel ?? null;
  }
```

spawnia:
This is a reasonable micro-optimization. Because of the potential breakage, I marked this for the next major version.

gregpeden:
I'm looking at submitting a PR for this, as well as contributing more in general.

spawnia:
I am fine with hiding this behind a configuration setting in v5, making it the default in v6 and removing the setting in v7. That will allow for a smooth migration.

stayallive:
This was implemented and released in [`v5.4.0`](https://github.com/nuwave/lighthouse/releases/tag/v5.4.0).

Thanks @GregPeden, lovely optimization 👍

PR #1799, opened 2021-04-15
Add config option `subscriptions.exclude_empty`

GregPeden:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Ref: https://github.com/nuwave/lighthouse/issues/1787

**Changes**

Adds config option 'subscriptions.exclude_empty' (default false). When true, and subscriptions are in use, an API response without a channel will exclude the 'lighthouse_subscriptions' extension response as a small performance optimization.

**Breaking changes**

None - default setting 'false' functions the same as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Add `GraphQLContext` to `StartExecution` event
 - Add `connect` and `disconnect` operations in nested mutations for HasMany and MorphMany relations https://github.com/nuwave/lighthouse/pull/1730
 - Add `ValidateSchema` event https://github.com/nuwave/lighthouse/pull/1764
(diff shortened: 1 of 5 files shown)


GregPeden:
BTW I added two tests to cover this setting and when I ran the tests my new tests pass but other already existing tests failed. Ran using using the "php" container within the included Docker dev environment.

GregPeden:
eeeek OOOOPS

Wait I wonder why that test didn't trigger as a fail on my end, hmmmmmmmmm........

spawnia:
@GregPeden no idea. Almost there now: https://github.com/nuwave/lighthouse/pull/1799/checks?check_run_id=2355553554

GregPeden:
Oh "foo" returns 42, not null, that's why. Okay I can fix.

I think that I wrote the test but it ran on the code base without the tests, as though the changes weren't saved or something.

spawnia:
Nicely done, thank you.

PR #1810, opened 2021-04-16
Add `@setCache`, `@deleteCache` and `@cleanInput` directives

### LINKED ISSUE #1772: Invalidate caches created by `@cache`
Yes I want to help you develop a lighthouse.

To delete the cache

```graphql
extend type Query {
    categories: [Category!]! @all @cache(key: "categories") @softDeletes
    categories2: [Category!]! @all @cache @softDeletes # If the key is not filled, use categories2 itself by default categories2
}

extend type Mutation {
    createCategory(input: CategoryInput! @spread): Category! @trim @create @cache(key: "categoryAll", delete: true)
}
```

When creating, editing and deleting (operations performed on the database), by adding a key and deleting it to @cache, we can delete it using cache. And save with the desired name using the key

Rasool-deldar:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #1772

**Changes**

Several features were added to the package
1 - @setCache
2 - @deleteCache
3 - @cleanInput

**Breaking changes**

None.

### DIFF
--- a/src/Schema/Directives/CleanInputDirective.php
+++ b/src/Schema/Directives/CleanInputDirective.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Nuwave\Lighthouse\Schema\Directives;
+
+use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
+use Nuwave\Lighthouse\Execution\Arguments\ArgumentSet;
+use Nuwave\Lighthouse\Schema\Values\FieldValue;
+use Nuwave\Lighthouse\Support\Contracts\ArgDirective;
+use Nuwave\Lighthouse\Support\Contracts\ArgSanitizerDirective;
+use Nuwave\Lighthouse\Support\Contracts\FieldMiddleware;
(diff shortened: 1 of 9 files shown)


spawnia:
I don't see how those features are supposed to be used and cannot be sure they are working. In order for me to consider inclusion, they need to be tested.

stayallive:
For reference, this is related to #1772.

src/Schema/Directives/CleanInputDirective.php
```diff
@@ -0,0 +1,91 @@
...
+    protected function transformArgumentSet(ArgumentSet $argumentSet): ArgumentSet
+    {
+        foreach ($argumentSet->arguments as $argument) {
+            $argument->value = $this->sanitize($argument->value);
+        }
+
+        return $argumentSet;
+    }
+
+    protected function transformLeaf($value)
```
MarkG-IEL, line 76: This set feels arbitrary. Wouldn't built-in PHP functions be better for this?

spawnia:
I don't think those features are well thought out or coherent. I am open to smaller pull requests with more polish.

PR #1814, opened 2021-04-20
Fix subscription redis storage for predis users

thekonz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1779 

**Changes**

Use [setex](https://redis.io/commands/setex) (SETEX is atomic) instead of set with multiple parameters which fails when using predis instead of phpredis (see #1779)


spawnia (approved):
Awesome!

PR #1819, opened 2021-04-21
Add method `assertGraphQLErrorMessage()` to `TestResponse` mixin

spawnia:
```php
$this
    ->graphQL(/** @lang GraphQL */ '
    {
        shouldFail
    }
    ')
    ->assertGraphQLErrorMessage('Expected error message.');
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Added
 
-- Add method `assertGraphQLErrorMessage()` to `TestResponse` mixin
+- Add method `assertGraphQLErrorMessage()` to `TestResponse` mixin https://github.com/nuwave/lighthouse/pull/1819
 
 ## 5.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add method `assertGraphQLErrorMessage()` to `TestResponse` mixin
+
 ## 5.4.0
 
 ### Added
--- a/_ide_helper.php
+++ b/_ide_helper.php
@@ -53,7 +53,18 @@ namespace Illuminate\Testing {
     class TestResponse
     {
         /**
-         * Asserts that the response contains an error from a given category.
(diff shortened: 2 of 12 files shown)


wimski:
@spawnia It's nice to see a native method to check for error messages. I used my own trait for this until now. However the current implementation only works if the response contains exactly one error message. I usually have tests where a lot of errors are returned, but I'm only interested in one. This results in the following failure for example:

```json
Unable to find JSON: 

[{
    "errors": [
        {
            "message": "Field MainInput.integer of required type Int! was not provided."
        }
    ]
}]

within response JSON:

[{
    "errors": [
        {
            "message": "Field MainInput.string of required type String! was not provided.",
            "extensions": {},
            "locations": []
        },
        {
            "message": "Field MainInput.integer of required type Int! was not provided.",
            "extensions": {},
            "locations": []
        }
    ]
}].
```

I see two possible solutions. Which of these has your preference?
1. Refactor the current `assertGraphQLErrorMessage` method to pass when multiple errors are present. This would break the situation where people specifically want to test that only one message is present.
2. Add a new `assertGraphQLErrorMessagesContains` method to check if an error message exists among possible multiple. Which could look something like this:

```php
public function assertGraphQLErrorMessagesContains(): Closure
{
    return function (string $message) {
        static::assertContains(
            $message,
            $this->json('errors.*.message')
        );

        return $this;
    };
}
```

spawnia:
@wimski I prefer option 1, asserting that only one error is present does not seem that useful to me. If needed, that can be done by using something like `assertExactJson()` anyways. Can you create a PR?

wimski:
I would love to create a PR (and more besides this one), but as long as the test suite relies on an actual MySQL database instead of the [standard in-memory sqlite 'testing' database](https://packages.tools/testbench/getting-started/configuration.html#in-memory-sqlite-connection), I can't run the test suite and therefore not reliable work on the code.

spawnia:
Have you considered the `docker-compose` based Setup? It comes with the database included. Also, not all tests require a database.

Background info on why we use MySQL for testing: SQLite lacks some features we need to test for.

wimski:
How did I miss the docker file!? 🙈

spawnia:
https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#setup

Issue #1817, opened 2021-04-21
Add @withSum @withMin, @withMax, @withAvg directive


P-Shakibafar:
**What problem does this feature proposal attempt to solve?**

There is currently a directive @withCount:

`type User {
  activityStatistics: ActivityStatistics! @withCount(relation: "posts")
}`

In the last few months, new methods have been added to Laravel that make it easier to work with models, such as:

`$posts = Post::withSum('comments', 'votes')->get();‍‍`

`‍$posts = Post::withAvg('comments', 'votes')->get();`

`$posts = Post::withMin('comments', 'votes')->get();`

`$posts = Post::withMax('comments', 'votes')->get();`

We can have this help method.

**Which possible solutions should be considered?**

A solution similar to @withCount directive may be used.

spawnia:
Please use the issue template and add a description.

spawnia:
That would be a Laravel 8+ only feature then?

P-Shakibafar:
yes, and laravel 8 lts

spawnia:
Ok. I am accepting merge requests for this. The implementation will probably look very similar to `@withCount`, but should abstract appropriate to keep it DRY. Tests should also be added.

spawnia:
With the newly implemented `@aggregate` directive, I see little use for a non-resolving version. If more people are interested, we might implement something like `@withAggregate`, but I am closing this for now.

PR #1834, opened 2021-04-29
Laravel8 new aggregate functions

P-Shakibafar:
- [X] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

develop withSum aggregate function in laravel 8+ 

Adds a directive to eager load relationship sum.

**Breaking changes**

None

`type Post {
            comments_sum_votes: Int!
                @withSum(relation: "comments", column: "votes")
        }`


P-Shakibafar:
add method withSum #1817

src/Execution/DataLoader/RelationSumLoader.php
```diff
@@ -0,0 +1,85 @@
...
+
+        return $sum;
+    }
+
+    /**
+     * Reload the models to get the `{relation}_sum_{column}` attributes of models set.
+     *
+     * @param  array<string, \Closure> $relations
+     */
+    public static function loadSum(EloquentCollection $parents, array $relations, ?string $column): void
```
spawnia, line 57: This was pretty much just copy-pasted from `loadCount()`, but why did you throw out the deprecation notice? Given the sum aggregate function is Laravel 8 only, we should be able to just use `EloquentCollection::loadSum()`.

src/Schema/Directives/CountDirective.php
```diff
@@ -90,4 +90,9 @@ protected function relationLoader(ResolveInfo $resolveInfo): RelationLoader
             $this->decorateBuilder($resolveInfo)
         );
     }
+
+    protected function relationColumn(): ?string
```
spawnia, line 94: Having to implement this useless method here tells me the abstraction is wrong.

src/Schema/Directives/WithRelationDirective.php
```diff
@@ -45,16 +50,18 @@ function (Model $parent, array $args, GraphQLContext $context, ResolveInfo $reso
     protected function loadRelation(Model $parent, ResolveInfo $resolveInfo): Deferred
     {
         $relationName = $this->relationName();
+        $relationColumn = $this->relationColumn();
 
         // There might be multiple directives on the same field, so we differentiate by relation too
         $uniquePath = $resolveInfo->path;
         $uniquePath [] = $relationName;
+//        $uniquePath [] = $relationColumn;
```
spawnia, line 58: What if there are two `@withSum` for different columns?

src/Schema/Directives/WithSumDirective.php
```diff
@@ -0,0 +1,92 @@
...
+if the field should simply return the relation count.
+"""
+directive @withSum(
+  """
+  Specify the relationship method name in the model class.
+  """
+  relation: String!
+
+  """
+  Specify the relationship column name in the model class.
```
spawnia, line 35: This is just lazily copy-pasted and could be a lot more descriptive.

src/Schema/Directives/WithSumDirective.php
```diff
@@ -0,0 +1,92 @@
...
+         */
+        $relation = $this->directiveArgValue('relation');
+
+        return $relation;
+    }
+
+    protected function relationColumn(): string
+    {
+        /**
+         * We validated the argument during schema manipulation.
```
spawnia, line 74: I can see no such validation, this comment lies.

tests/Integration/Schema/Directives/WithSumDirectiveTest.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    public function testEagerLoadsRelationSum(): void
+    {
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            posts: [Post!] @all
+        }
+
+        type Post {
+            comments_sum_votes: Int!
```
spawnia, line 21: This use case is pretty much precisely what the `@with*` directives are explicitly *not* meant for, see the description of `@withCount`. There should be a `@sum` directive for this.

tests/Integration/Schema/Directives/WithSumDirectiveTest.php
```diff
@@ -0,0 +1,90 @@
...
+                        'comments_sum_votes' => 1,
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->assertSame(2, $queries);
+    }
+
+    public function _testFailsToEagerLoadRelationSumWithoutRelation(): void
```
spawnia, line 65: Why the underscore?

tests/Integration/Schema/Directives/WithSumDirectiveTest.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->withExceptionHandling();
```
spawnia, line 67: Why is this used? We have it nowhere else in the codebase.

tests/database/migrations/2018_02_28_000005_create_testbench_comments_table.php
```diff
@@ -15,6 +15,8 @@ public function up(): void
             $table->unsignedInteger('user_id');
             $table->unsignedInteger('post_id');
 
+            $table->integer('votes')->default(0);
```
spawnia, line 18: Missing from the model PHPDoc.

tests/Integration/Schema/Directives/WithSumDirectiveTest.php
spawnia, line 65: I do not understand the intent of this test. Please clarify.

spawnia (changes requested):
Thanks for giving this a go.

I think the level of abstraction you chose is wrong. Specifying a column name should not leak to other relation loaders that do not have such a requirement.

It looks like there are a bunch more aggregate functions related to `sum` which also require a column to be named, so it seems fitting to abstract over that. We should think about how we can support the other aggregate functions, as you indicated in https://github.com/nuwave/lighthouse/issues/1817. I suggest you try and add support for one more and see where the similarities in implementation are. Perhaps we could make them into a single `@withAggregate` directive?

Consider the difference between `@count` and `@withCount`. The same applies to `@withSum`.

I am marking this as a draft for now, as it needs a lot more work.

src/Execution/DataLoader/RelationMaxLoader.php
```diff
@@ -0,0 +1,85 @@
...
+
+    public function extract(Model $model, string $relationName, string $column)
+    {
+        return self::extractMax($model, $relationName, $column);
+    }
+
+    public static function extractMax(Model $model, string $relationName, string $column): int
+    {
+        /**
+         * This is the name that Eloquent gives to the attribute that contains the min.
```
pyrou, line 36:
```suggestion
         * This is the name that Eloquent gives to the attribute that contains the max.
```

spawnia:
Please address the open conversations so this can move forward.

P-Shakibafar:
https://github.com/nuwave/lighthouse/pull/1834#issuecomment-835480741 

I have completely changed the approach of doing this feature, please check again.

src/Execution/DataLoader/RelationAggregateBatchLoader.php
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * @var string
+     */
+    protected $relationName;
+
+    /**
+     * @var string
+     */
+    protected $relationColumn;
+    /**
```
spawnia, line 33:
```suggestion

    /**
```

src/Execution/DataLoader/RelationAggregateBatchLoader.php
```diff
@@ -0,0 +1,108 @@
...
+     * Check if a loader has been registered.
+     */
+    public function hasRelationAggregateLoader(): bool
+    {
+        return $this->relationAggregateLoader !== null;
+    }
+
+    /**
+     * Register a relation loader.
```
spawnia, line 49:
```suggestion
```

src/Execution/DataLoader/RelationAggregateLoader.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Nuwave\Lighthouse\Execution\DataLoader;
+
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Illuminate\Database\Eloquent\Model;
+
+interface RelationAggregateLoader
+{
+    /**
+     * Load the given relation on the given parent models.
```
spawnia, line 11: Those comments are copy-pasted and not accurate.

src/Execution/DataLoader/RelationAvgLoader.php
```diff
@@ -0,0 +1,85 @@
...
+    protected $decorateBuilder;
+
+    public function __construct(Closure $decorateBuilder)
+    {
+        $this->decorateBuilder = $decorateBuilder;
+    }
+
+    public function load(EloquentCollection $parents, string $relationName, string $column): void
+    {
+        self::loadAvg($parents, [$relationName => $this->decorateBuilder], $column);
```
spawnia, line 25: Why not use `EloquentCollection::loadAvg()`? Same with the other aggregate loaders.

src/Schema/Directives/WithAvgDirective.php
```diff
@@ -0,0 +1,90 @@
...
+if the field should simply return the relation count.
+"""
+directive @withAvg(
+  """
+  Specify the relationship method name in the model class.
+  """
+  relation: String!
+
+  """
+  Specify the relationship column name in the model class.
```
spawnia, line 35: This description is misleading, it is not a column of the model class but rather the relation.

tests/Integration/Schema/Directives/WithAvgDirectiveTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+        type Book {
+            title: String!
+            price: Int!
+        }
+
+        type Author {
+            name: String!
+            books:[Book!]! @belongsToMany
+            books_avg_price: Int!
```
spawnia, line 28: *Sigh*

As I have now repeated multiple times: this usage of the `@withAvg` directive is wrong! If you need to get the actual aggregate value directly as the field result, a simpler version of this directive should be used. It would be called `@avg`.

The `@with` directives are meant to load a relation (or a relation aggregate) as a side effect when a certain field is requested, assuming the field resolver subsequently makes use of that field.

Even if this test is technically working, it is setting a wrong example. What you are doing here would be much simpler with `@avg`, and the test does not demonstrate proper use of `@withAvg`. Consider the test case for `@withCount` and how the assertion for proper loading is made there.

spawnia (changes requested):

src/Schema/Directives/WithAvgDirective.php
```diff
@@ -0,0 +1,90 @@
...
+         */
+        $relation = $this->directiveArgValue('relation');
+
+        return $relation;
+    }
+
+    protected function relationColumn(): string
+    {
+        /**
+         * @var string $column
```
spawnia, line 74: How do you know this is true?

spawnia:
I took the time to go through the PR again, please take the time to address each and every review comment. Many of them apply generally to the different aggregate variants (sum, min, max, avg).

P-Shakibafar:
@spawnia According to you, instead of development @with{sum-min-max-avg} I worked on development @{sum-min-max-avg} please check again. And I think this putty is useful. I have developed @sum.

PR #1845, opened 2021-05-11
Combine eager loads when same

### LINKED ISSUE #1555: Relations get reloaded even when they are already present.
**Describe the bug**
Not 100% sure if this is a bug. But for me it is unexpected behaviour with large performance implications.

RelationFetcher always reloads a relation even when it is already present on the parent model. 

I think if a relation is already present on the parent model it should not have to be reloaded.
What appears to fix the issue is updating line 39 of RelationFetcher.php (in function loadedParentModels) from:

`$modelsOfSameClass->load($relation);`

to

`$modelsOfSameClass->loadMissing($relation);`

This fix appears to be working, but i'm curious as to if there is a reason for load to be used over loadMissing in the first place.

**Steps to reproduce**

1.  Create an Eloquent Model with a relationship.
2. Force relationship to always load with the model.
`protected $with = [ 'contacts'];`
3. Create a type in your graphql schema to match your new model and add your relationship with relationship direcive (eg. @hasOne)
3. Create a query that will return your new type
4. Add code to the boot() of your AppServiceProvider to dump DB events
`        DB::listen(function ($event) {
                dump($event->sql);
                dump($event->bindings);
        });
`
5. Run new query
6. Check the dumped data. You will notice that your relationship is loaded twice from the DB.

**Lighthouse Version**

v4.16.2

### LINKED ISSUE #1665: Combine eager-loads that are the same together

In queries that involve multiple eager-loads on the same model, multiple database calls are made.

```graphql
type Post {
  someFieldInvolvingComments: Int @with(relation: "comments")
  anotherField: String @with(relation: "comments")
}
```

This query would cause the same eager-load on `comments` to be done 3 times.

```graphql
{
  posts {
    comments
    someFieldInvolvingComments
    anotherField
  }
}
```

Lighthouse does not recognize that eager-loading `comments` once suffices when the different fields are called.

On a given model, eager-loads of the same relation should be combined.

There can be some tricky edge-cases, for instance if one of the eager-loads imposes query conditions on the relation:

```graphql
{
  posts {
    comments(last: 20)
    someFieldInvolvingComments
  }
}
```

@ipalaus

cradigan:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

[Resolves #1665](https://github.com/nuwave/lighthouse/issues/1665)
[Resolves #1555](https://github.com/nuwave/lighthouse/issues/1555)

**Changes**
Should batch eager loads more intelligently.

**Breaking changes**
 Shouldn't cause any breaking changes.

### DIFF
--- a/src/Execution/DataLoader/BatchLoaderRegistry.php
+++ b/src/Execution/DataLoader/BatchLoaderRegistry.php
@@ -25,10 +25,10 @@ abstract class BatchLoaderRegistry
     {
         // serves as the unique key for the instance
         $instanceKey = self::instanceKey(
-            $relationName, 
-            $pathToField, 
-            $relationScopes, 
-            !empty($fieldArguments) // we want to batch load per field if any field args exist.
+            $relationName,
+            $pathToField,
+            $relationScopes,
+            ! empty($fieldArguments) // we want to batch load per field if any field args exist.
         );
 
         if (isset(self::$instances[$instanceKey])) {
@@ -50,30 +50,28 @@ abstract class BatchLoaderRegistry
     }
 
     /**
-     * Generate a unique key for the instance
+     * Generate a unique key for the instance.
      *
-     * @param  string  $relationName
      * @param  array<int|string>  $pathToField
      * @param  array<string>  $relationScopes
-     * @param  bool  $batchPerField
      */
     protected static function instanceKey(string $relationName, array $pathToField, array $relationScopes = [], bool $batchPerField = false): string
     {
-        if($batchPerField){
+        if ($batchPerField) {
             // Create separate batch loader per field
             // There might be multiple directives on the same field, so we differentiate by relation too
             return self::keyPartFromPath($pathToField)
-                ."|relation__".$relationName;
+                .'|relation__'.$relationName;
         }else{
 
             // Remove field from path array
             $pathWithoutField = $pathToField;
             array_pop($pathWithoutField);
 
-             // Since we aren't batching per field we need to unique it by scopes as well
+            // Since we aren't batching per field we need to unique it by scopes as well
             return self::keyPartFromPath($pathWithoutField)
-                ."|relation__".$relationName
-                ."|scopes__".implode(',',$relationScopes);
+                .'|relation__'.$relationName
+                .'|scopes__'.implode(',', $relationScopes);
         }
     }
 
--- a/src/Execution/DataLoader/BatchLoaderRegistry.php
+++ b/src/Execution/DataLoader/BatchLoaderRegistry.php
@@ -21,10 +21,15 @@ abstract class BatchLoaderRegistry
      *
      * @throws \Exception
      */
-    public static function instance(string $loaderClass, array $pathToField, array $constructorArgs = []): object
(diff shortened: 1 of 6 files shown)


src/Execution/DataLoader/BatchLoaderRegistry.php
```diff
@@ -21,10 +21,15 @@ abstract class BatchLoaderRegistry
      *
      * @throws \Exception
      */
-    public static function instance(string $loaderClass, array $pathToField, array $constructorArgs = []): object
+    public static function instance(string $loaderClass, string $relationName, array $pathToField, array $relationScopes = [], array $fieldArguments, array $constructorArgs = []): object
```
spawnia, line 24: Everything about relations should probably go into `RelationBatchLoader`. That keeps this class generically useful.

cradigan: You are right, but i'm not quite sure how this should work.  Currently the BatchLoaderRegistry generates the instance key before even creating a RelationBatchLoader instance.  And only long after the instance is created with it's key in the registry do we register the relation info inside the RelationBatchLoader.
`
        /** @var \Nuwave\Lighthouse\Execution\DataLoader\RelationBatchLoader $relationBatchLoader */
        $relationBatchLoader = BatchLoaderRegistry::instance(RelationBatchLoader::class, $uniquePath);

        if (! $relationBatchLoader->hasRelationLoader()) {
            $relationBatchLoader->registerRelationLoader($this->relationLoader($resolveInfo), $relationName);
        }
`

I have a couple ideas though:

1) BatchLoaderRegistry@Instance accepts $constructorArgs for building the BatchLoader.  Instead of registering the relation data after, maybe we could:
          a) add a constructor to RelationBatchLoader and use this instead of the registerRelationLoader method later
          b) create a batchKeySuffix method directly on RelationBatchLoader that allows it to partially handle part of the instanceKey generation itself.
          c) Pass RelationBatchLoader data through BatchLoaderRegistry@instance $constructorArgs
          d) Have BatchLoaderRegistry@Instance generate the RelationBatchLoader instance with $constructorArgs
          e) HaveBatchLoaderRegistry@instance generate the key normally based on the path (with the field).
          f) Check if a batchKeySuffix method exists on the RelationBatchLoader and if it does remove the last part of the path (the field) and replace it with the batchKeySuffix.

2) A more simple solution we could just have withRelationDirective add all the relation data to the path array before it is passed into BatchLoaderRegistry@instance.  The current code is already doing something like this with the relation name (i think maybe to allow the use of multiple @with directives on the same field).

Let me know what you think of either of those solutions.

spawnia:
I gave this a go and found the current abstractions hard to extend, so I did somewhat of a rewrite of all things related to batch loading and relation loading. @cradigan please review https://github.com/nuwave/lighthouse/pull/1849

src/Execution/DataLoader/BatchLoaderRegistry.php
spawnia: Solution 2 sent me on the track I am now on and lead to an even cleaner solution. Thanks.

PR #1848, opened 2021-05-12
Allow clients to order a list of models by an aggregated value of their relations with `@orderBy`

### LINKED ISSUE #1467: Order by nested relation, and aggregation results
I will gladly provide a PR for this issue

When you use OrderBy, you can't order by anything that isn't a direct column, and also you can't order by aggregations like min, max, and so on.

In the OrderBy directive, create support for relation, and aggregation

**Notes**

- Maybe it can somehow relate or use to another issue I made regarding aggregations.

- When using relation, and not putting a column, it will order by the aggregation of the relation.

- If this idea and the other one with the aggregation will be accepted, there would now be 3 things that depend on relations; so I think [this proposal](https://github.com/nuwave/lighthouse/issues/1409#issuecomment-641866097) should also be considered, for easier usage of thosw functionalities.

- Also maybe it should be a different a different directive, so it wouldn't break the original directive, and it would still be available for easier use.

**Request Example**
```graphql{
products(
    orderBy: {
         priorities: [
             {
                 relation: 'purchases'
                 column: 'payed'
                 aggregation: AVG
                 sort: ASC
             }
         ]
    }
){
    name
}
```

**Generated Type**
```graphql{
input OrderByClauses
{
  priorities: [OrderByClause!]
}

input OrderByClause
{
  relation: String

  column: String

  aggregation: AggregationFunction

  sort: SortOrder = DESC
}

enum AggregationFunction
{
  MIN
  
  MAX

  AVG

  SUM

  COUNT

  DISTINCT
}

enum SortOrder
{
  ASC
  
  DESC
}
```

exodusanto:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1467

**Changes**

Add new behavior to `@orderBy` directive.
Now in schema you can declare root type columns and the relations (this will use `withCount` etc.)

```graphql
@orderBy(
  columns: ["name", "created_at"], 
  relations: [
    { relation: "posts" }, # Only COUNT possible
    { relation: "ratings", columns: ["quality", "design", "performance"] } # COUNT and other aggregates dynamically available
  ]
)
``` 

An in your queries

```graphql
authors(
  orderBy: [
    { posts: { aggregate: COUNT }, order: DESC },
    { ratings: { aggregate: AVG, column: DESIGN }}
  ]
)
```


spawnia:
Please see https://github.com/nuwave/lighthouse/pull/1848/files and add tests for the currently uncovered implementation.

spawnia:
@exodusanto thanks for your work on this. It was not polished to a level I felt comfortable merging, so I did all that.

PR #1851, opened 2021-05-17
Make use of the new graphql-php lazy field definition feature

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This introduces 2 changes, which should result in a nice performance boost especially for larger schema's where much types would go unused in the validation for a single query.

Makes use of the lazy field definitions: webonyx/graphql-php#861.

In addition this also removes the need to eager load subscriptions and only resolve them when they are used.

This also bumps `graphql-php` to `14.7`.

**Breaking changes**

None, except apps (especially with large schema's) should see their response times go down 😉 

Invalid schema definitions might go unnoticed during development, because a smaller part of the schema is actually built on each request. Make sure to run `php artisan lighthouse:validate-schema` as part of your CI/CD pipeline.


src/Schema/TypeRegistry.php
```diff
@@ -337,24 +337,27 @@ function () use ($objectDefinition): array {
...
 
                 foreach ($typeDefinition->fields as $fieldDefinition) {
-                    /** @var \Nuwave\Lighthouse\Schema\Factories\FieldFactory $fieldFactory */
-                    $fieldFactory = app(FieldFactory::class);
-                    $fieldValue = new FieldValue($typeValue, $fieldDefinition);
-
-                    $fields[$fieldDefinition->name->value] = $fieldFactory->handle($fieldValue);
+                    $fields[$fieldDefinition->name->value] = static function () use ($typeValue, $fieldDefinition): array {
+                        /** @var \Nuwave\Lighthouse\Schema\Factories\FieldFactory $fieldFactory */
+                        $fieldFactory = app(FieldFactory::class);
```
spawnia, line 355: Perhaps we can micro-optimize by instantiating this only once and cache it in a property of `TypeRegistry`?

stayallive: I tried that before... but you mentioned this broke your app... but this would be another pretty big improvement if that would be possible (the container is not as fast as you would think)!

But because of last time I did not dare to do this (see #961) 😅

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -88,14 +98,24 @@ public function has(string $key): bool
      */
     public function keys(): array
```
spawnia, line 99: This seems unused, should we deprecate it?

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -88,14 +98,24 @@ public function has(string $key): bool
...
      * Get instance of subscription.
      */
     public function subscription(string $key): GraphQLSubscription
     {
+        if (isset($this->subscriptions[$key])) {
+            return $this->subscriptions[$key];
+        }
+
+        if ($this->subscriptionType !== null) {
+            $this->subscriptionType->getField($key);
```
spawnia, line 116:
```suggestion
            // Loading the field has the side effect of triggering a call to 
            // \Nuwave\Lighthouse\Support\Contracts\ProvidesSubscriptionResolver::provideSubscriptionResolver()
            // which is then expected to call static::register().
            $this->subscriptionType->getField($key);
```

Sounds pretty convoluted when writing it down like this, perhaps we can simplify.

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -53,12 +52,18 @@ class SubscriptionRegistry
...
+     */
+    protected $subscriptionType;
+
     public function __construct(ContextSerializer $serializer, StoresSubscriptions $storage, SchemaBuilder $schemaBuilder, ConfigRepository $configRepository)
     {
         $this->serializer = $serializer;
         $this->storage = $storage;
         $this->schemaBuilder = $schemaBuilder;
         $this->configRepository = $configRepository;
+        $this->subscriptionType = $schemaBuilder->schema()->getSubscriptionType();
```
spawnia, line 66: I think this might force building a schema on requests that do not use GraphQL at all.

tests/Unit/Schema/Directives/RenameDirectiveTest.php
```diff
@@ -44,7 +44,7 @@ public function testThrowsAnExceptionIfNoAttributeDefined(): void
 
         $this->schema = /** @lang GraphQL */ '
         type Query {
-            foo: String! @rename
+            fooBar: String! @rename
```
spawnia, line 47: Why?

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -140,11 +156,11 @@ public function subscriptions(Subscriber $subscriber): Collection
                     ->all();
             })
             ->map(function ($subscriptionField): GraphQLSubscription {
-                return Arr::get(
-                    $this->subscriptions,
-                    $subscriptionField,
-                    new NotFoundSubscription
-                );
+                if (! $this->has($subscriptionField)) {
```
spawnia, line 159: Let's invert this negative conditional.

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -88,14 +98,24 @@ public function has(string $key): bool
...
     /**
      * Get instance of subscription.
      */
     public function subscription(string $key): GraphQLSubscription
     {
+        if (isset($this->subscriptions[$key])) {
+            return $this->subscriptions[$key];
+        }
+
+        if ($this->subscriptionType !== null) {
```
spawnia, line 115: I think we can throw if `$this->subscriptionType === null`.

src/Schema/TypeRegistry.php
spawnia: Since then the `FieldFactory` has actually changed to be safe to reuse for multiple fields - perhaps I was subconsciously thinking about making this optimization at some point, but then forgot to go through with it. Let's give it another go.

tests/Unit/Schema/Directives/RenameDirectiveTest.php
stayallive: Because the lazy fields have this as a side effect, when fields are not accessed they also won't throw.

This is not an issue with the validation command, that works perfectly but it is an issue when you are not calling the field you are testing 😄

src/Subscriptions/SubscriptionRegistry.php
stayallive: Agreed, lemme solve this another way!

src/Subscriptions/SubscriptionRegistry.php
stayallive: Yes, I agree, but honestly I'm out of ideas without a huge overhaul where this whole registry is just removed at all... if you have a clean idea I'm all ears!

src/Schema/TypeRegistry.php
stayallive: In that case 💪 let me put that in!

PR #1864, opened 2021-05-27
Allow JSON notation `->` in column references for complex where conditions

### LINKED ISSUE #1863: Query JSON column in whereConditions
**Describe the bug**
If a database column is defined as JSON, it's currently not possible to query using whereConditions, since method `assertValidColumnReference` doesn't allow symbols `->` used as column name

Having a table with a column of type JSON, it should be possible to define such column with JSON notation `->`

**Steps to reproduce**

1. Create a migration to add a table with a JSON column
2. Add related model and define a query to retrieve all models in that table using `@all` directive
3. Define an enum that has a value to query one JSON field (for example `enum ExampleColumn { VALUE @enum(value: "json_column->json_key") }`)
4. Add an argument to that query as `where: _ @whereConditions(columnsEnum: "ExampleColumn")`
5. Perform a query on that table after creating some mock data

**Additional info**
I think it should be possible to fix this changing the regex used with this one:  
`/^(?![0-9.-])([A-Za-z0-9_.-]|->)*$/`  
I'm opening related PR #1864 

**Lighthouse Version**
5.12.1

akyrey:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1863 

Add a test to query inner JSON. This test also required defining a new Scalar type JSON (ref. https://gist.github.com/lorado/0519972d6fcf01f1aa5179d09eb6a315)

**Changes**

Edited regex to check column name validity

**Breaking changes**

No breaking changes


src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -134,7 +134,7 @@ public function manipulateArgDefinition(
         ObjectTypeDefinitionNode &$parentType
     ): void {
         if ($this->hasAllowedColumns()) {
-            $restrictedWhereConditionsName = ASTHelper::qualifiedArgType($argDefinition, $parentField, $parentType).$this->generatedInputSuffix();
+            $restrictedWhereConditionsName = ASTHelper::qualifiedArgType($argDefinition, $parentField, $parentType) . $this->generatedInputSuffix();
```
spawnia, line 137: Please revert unnecessary formatting changes.

src/WhereConditions/WhereConditionsBaseDirective.php
```diff
@@ -168,7 +168,7 @@ protected static function assertValidColumnReference(string $column): void
         // - must not start with a digit, dot or hyphen
         // - must contain only alphanumerics, digits, underscores, dots or hyphens
         // Dots are allowed to reference a column in a table: my_table.my_column.
-        $match = \Safe\preg_match('/^(?![0-9.-])[A-Za-z0-9_.-]*$/', $column);
+        $match = \Safe\preg_match('/^(?![0-9.-])([A-Za-z0-9_.-]|->)*$/', $column);
```
spawnia, line 171: When would this be necessary? Please provide a test case that fails without this change and passes with it.

spawnia (changes requested):

src/WhereConditions/WhereConditionsBaseDirective.php
akyrey: Is there a specific test model that I can modify adding a column or any is good?

src/WhereConditions/WhereConditionsBaseDirective.php
akyrey: Fixed

src/WhereConditions/WhereConditionsBaseDirective.php
spawnia: Any model with a string column should suffice, no?

akyrey: Nope, would need a model with a JSON column in the database

spawnia: Ok, just pick any one and add a nullable column.

akyrey:
Should I update anything in the documentation?

akyrey:
Sorry for multiple pushes for Style CI

spawnia:
> Sorry for multiple pushes for Style CI

StyleCI has an option to download a patch.

spawnia:
I saw you added a custom JSON type. It is not needed for purposes of this PR, but if you want to use one for your projects, I recommend https://github.com/mll-lab/graphql-php-scalars/blob/master/src/JSON.php

PR #1869, opened 2021-06-06
Prevent resolver being called when a cached field contains a falsy value

pyrou:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

When caching a boolean, string or integer value using `@cache` directive, falsy value like `false`, `0` or `""` aren't consider as valid cache. This PR resolves this issue.

**Breaking changes**

No

### DIFF
--- a/src/Schema/Directives/CacheDirective.php
+++ b/src/Schema/Directives/CacheDirective.php
@@ -91,6 +91,29 @@ GRAPHQL;
                 if ($value !== null) {
                     return $value;
                 }
+                // In Laravel cache, null is considered as "non-existant" value. As mentioned in laravel documentation,
+                // https://laravel.com/docs/8.x/cache#checking-for-item-existence
+                // > The `has` method [...] will also return false if the item exists but its value is null.
+                //
+                // If caching `null` value becomes something worthwhile, one possible way to achieve it is to
+                // encapsulate the `$result` at writing time :
+                //
+                //    $storeInCache = static function ($result) use ($cacheKey, $maxAge, $cache): void {
+                //        $value = ['rawValue' => $result];
+                //        $maxAge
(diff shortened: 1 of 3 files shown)


src/Schema/Directives/CacheDirective.php
```diff
@@ -87,7 +87,8 @@ function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
 
                 // We found a matching value in the cache, so we can just return early
                 // without actually running the query
-                if ($value = $cache->get($cacheKey)) {
+                $value = $cache->get($cacheKey);
+                if ($value !== null) {
```
spawnia, line 91: What if the value is `null`?

pyrou: IMO, It is a bad idea to cache raw `null`.  

In Laravel cache, `null` is considered as "non-existant" value. As mentioned in [laravel documentation](https://laravel.com/docs/8.x/cache#checking-for-item-existence).

> The `has` method [...] will also return false if the item exists but its value is null



If caching `null` result from resolver is wanted, we must encapsulate the value. Something like :
```php
$cache->put($cacheKey, ["value" => $result], Carbon::now()->addSeconds($maxAge));
```

And then :

```php
if ($cache->offsetExists($cacheKey)) {
  return $cache->get($cacheKey)["value"];
}
```

But this would include some BC, specially if you migrate lighthouse without clearing cache. So handling `null` (again, _if_ it is something wanted) possibly deserves a dedicated PR.

spawnia: Makes sense. Please add the Link and snippet as a code comment so this knowledge is kept.

pyrou: Done.

spawnia (approved):
Nicely done. Changes like this can be tricky, you took proper care - thank you.

PR #1882, opened 2021-06-27
Fix model name determination when using interfaces

enzonotario:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

When using interfaces (and a `@model` directive attached to it), the `ASTHelper::modelName` function should also handle instances of `InterfaceTypeDefinitionNode`.

**Changes**

Allows `ASTHelper::modelName` function to also handle instances of `InterfaceTypeDefinitionNode` and change `ModelDirective::modelClass`'s `$node` arg type hint to `GraphQL\Language\AST\Node`.

**Breaking changes**

None.


spawnia:
Thank you, nicely done

PR #1894, opened 2021-07-14
Call lighthouse:clear-cache in other commands

spawnia:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Actually clears the cache in `lighthouse:validate-schema` and `lighthouse:print-schema` when using schema cache version 2.

**Breaking changes**

None


thekonz:
:+1: good catch! Did not see that.

PR #1895, opened 2021-07-15
Add directive `@inherits` for reuse of parent type definitions

yondifon:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
Introduced a new Directive `@extends` that makes it easier to inherit from a parent type. 
This copies the merges the properties of the child to the parent while overriding existing properties.


docs/master/api-reference/directives.md
```diff
@@ -1111,6 +1111,47 @@ class PlacedOrder
 }
 ```
 
+## @extends
```
spawnia, line 1114: `@extends` would be the most natural name, but it is already used by Apollo Federation, which Lighthouse supports. How about `@inherits(from: "Foo")`?

docs/master/api-reference/directives.md
```diff
@@ -1111,6 +1111,47 @@ class PlacedOrder
...
+it inherits properties from the parent type. 
+It overides same attributes present in the parent type.
+'""
+"""
+directive @extends(
+    """
+    The parent Type
+    """
+    parent: String!
+) on OBJECT
```
spawnia, line 1129:
```suggestion
) on OBJECT | INPUT_OBJECT | ENUM | INTERFACE | UNION
```

All compound types should work.

spawnia, line 1129: What about multiple inheritance? I don't see why it shouldn't be allowed.

docs/master/api-reference/directives.md
```diff
@@ -1111,6 +1111,47 @@ class PlacedOrder
...
 ```
 
+## @extends
+
+```graphql
+"""
+This allows types to extends other Types
+
+it inherits properties from the parent type. 
+It overides same attributes present in the parent type.
```
spawnia, line 1121: This is not totally clear to me. Which attributes have precedence?

docs/master/api-reference/directives.md
```diff
@@ -1111,6 +1111,47 @@ class PlacedOrder
 }
 ```
 
+## @extends
+
+```graphql
+"""
+This allows types to extends other Types
+
+it inherits properties from the parent type.
```
spawnia, line 1120: What about directives on the parent type? Are those inherited as well?

src/Schema/Directives/ExtendsDirective.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @param   \Nuwave\Lighthouse\Schema\AST\DocumentAST         $documentAST
+     * @param   \GraphQL\Language\AST\TypeDefinitionNode  $typeDefinition
+     *
+     * @return  void
+     */
+    public function manipulateTypeDefinition(DocumentAST &$documentAST, TypeDefinitionNode &$typeDefinition)
+    {
+        $parent_name = $this->getMethodArgumentParts('parent')[0];
```
spawnia, line 39:
```suggestion
        $parent_name = $this->directiveArgValue('parent');
```

src/Schema/Directives/ExtendsDirective.php
```diff
@@ -0,0 +1,62 @@
...
+    """
+    The parent Type
+    """
+    parent: String!
+) on OBJECT
+GRAPHQL;
+    }
+
+    /**
+     * @param   \Nuwave\Lighthouse\Schema\AST\DocumentAST         $documentAST
```
spawnia, line 32: Please consider the code style of this project.

src/Schema/Directives/ExtendsDirective.php
spawnia, line 39: Please consider the naming conventions of this project.

src/Schema/Directives/ExtendsDirective.php
```diff
@@ -0,0 +1,62 @@
...
+     * @param   \GraphQL\Language\AST\TypeDefinitionNode  $typeDefinition
+     *
+     * @return  void
+     */
+    public function manipulateTypeDefinition(DocumentAST &$documentAST, TypeDefinitionNode &$typeDefinition)
+    {
+        $parent_name = $this->getMethodArgumentParts('parent')[0];
+
+        $parent = $documentAST->types[$parent_name];
```
spawnia, line 41: This should check and throw a developer friendly exception if the type does not exist.

src/Schema/Directives/ExtendsDirective.php
```diff
@@ -0,0 +1,62 @@
...
+
+        if ($typeDefinition->kind !== $parent->kind) {
+            throw new DefinitionException(
+                "The {$typeDefinition->name->value} ({$typeDefinition->kind} Found!) kind must the same as {$parent->name->value} ({$parent->kind} Found!)."
```
spawnia, line 45: Do not shout (no `!`).

src/Schema/Directives/ExtendsDirective.php
```diff
@@ -0,0 +1,62 @@
...
+            );
+        }
+
+        foreach (get_object_vars($parent) as $type => $value) {
```
spawnia, line 49: This is quite fragile, let's do the individual properties explicitly.

tests/Integration/Schema/Directives/ExtendsDirectiveTest.php
```diff
@@ -0,0 +1,158 @@
...
+namespace Tests\Integration\Schema\Directives;
+
+use Illuminate\Testing\Fluent\AssertableJson;
+use Tests\TestCase;
+
+class ExtendsDirectiveTest extends TestCase
+{
+    public function testCanExtendTypes()
+    {
+        $this->mockResolver(function ($root, array $args): array {
```
spawnia, line 12: You can examine the result of calling `buildSchema()` directly instead of having to indirectly test it via queries. This feature is strictly schema manipulation, no need to do any actual calls for testing it.

tests/Integration/Schema/Directives/ExtendsDirectiveTest.php
```diff
@@ -0,0 +1,158 @@
...
+            'query TestChildType{
+               childtypeQuery {
+                   attribute_1
+                   attribute_2
+               }
+            }'
+        );
+
+        $response->assertJson(
+            fn (AssertableJson $json) => $json->has(
```
spawnia, line 84: Not available in older Laravel versions.

spawnia (changes requested):

docs/master/api-reference/directives.md
yondifon: That of the child.

docs/master/api-reference/directives.md
yondifon: Yes. They are.

src/Schema/Directives/ExtendsDirective.php
yondifon: Suggestions?

docs/master/api-reference/directives.md
yondifon: True.

I think the default extends adds certain attributes to an already existing type. 
🤔

spawnia: That is not at all what `@extends` does in the context of federation. We don't have to discuss that here though; I am just saying that the name is taken.

docs/master/api-reference/directives.md
spawnia: This might be problematic with many directives, e.g. `@namespace`, `@interface` or `@enum`.

docs/master/api-reference/directives.md
spawnia: The description should be changed to make this completely disambiguous.

docs/master/api-reference/directives.md
yondifon: Sorry meant extend.

enzonotario:
Hi there! I've made a similar directive for a personal project, so let me introduce those changes and if you like it, we can continue working from here.

docs/master/api-reference/directives.md
spawnia: Oh, yeah that is right. Another good reason to choose a different name for this.

src/Schema/Directives/InheritsDirective.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        return /* @lang GraphQL */ <<<'GRAPHQL'
+"""
+This allows types to inherits fields from another type.
+"""
+directive @inherits(
+    """
+    The parent type.
+    """
+    parent: String!
```
spawnia, line 22:
```suggestion
    from: String!
```

src/Schema/Directives/ExtendsDirective.php
spawnia: I am not sure what else I can suggest, what part of my comment was not clear?

spawnia:
In order for this merge request to move forward please address each of my comments.

yondifon (commented):
Hey @enzonotario 

I think this will bring up bugs. 

- Not all types have fields like unions
- Then what about other attributes? I think inherits means inherit more than just fields.
Correct me if I'm wrong.

enzonotario:
Yes you're right. So, we have to define what kind of types it should handle (so far I think ObjectTypes, Inputs and Enums). So something like this should work:
```
        switch ($typeDefinition->kind) {
            case 'EnumTypeDefinition':
                $typeDefinition->values = $parentType->values->merge($typeDefinition->values);
                break;
            case 'ObjectTypeDefinition':
            case 'InputTypeDefinition':
                $typeDefinition->fields = $parentType->fields->merge($typeDefinition->fields);
                break;
        }
```

yondifon:
> Yes you're right. So, we have to define what kind of types it should handle (so far I think ObjectTypes, Inputs and Enums). So something like this should work:
> 
> ```
>         switch ($typeDefinition->kind) {
>             case 'EnumTypeDefinition':
>                 $typeDefinition->values = $parentType->values->merge($typeDefinition->values);
>                 break;
>             case 'ObjectTypeDefinition':
>             case 'InputTypeDefinition':
>                 $typeDefinition->fields = $parentType->fields->merge($typeDefinition->fields);
>                 break;
>         }
> ```



> Yes you're right. So, we have to define what kind of types it should handle (so far I think ObjectTypes, Inputs and Enums). So something like this should work:
> 
> ```
>         switch ($typeDefinition->kind) {
>             case 'EnumTypeDefinition':
>                 $typeDefinition->values = $parentType->values->merge($typeDefinition->values);
>                 break;
>             case 'ObjectTypeDefinition':
>             case 'InputTypeDefinition':
>                 $typeDefinition->fields = $parentType->fields->merge($typeDefinition->fields);
>                 break;
>         }
> ```


Actually this solved the problem

yondifon:
> Yes you're right. So, we have to define what kind of types it should handle (so far I think ObjectTypes, Inputs and Enums). So something like this should work:
> 
> ```
>         switch ($typeDefinition->kind) {
>             case 'EnumTypeDefinition':
>                 $typeDefinition->values = $parentType->values->merge($typeDefinition->values);
>                 break;
>             case 'ObjectTypeDefinition':
>             case 'InputTypeDefinition':
>                 $typeDefinition->fields = $parentType->fields->merge($typeDefinition->fields);
>                 break;
>         }
> ```
Actually, this solved the issue. Since it checks to see if the parent and child are of the same type. 

@spawnia. Please can you review this, I don't think it will be an issue. There's a check that ensures that the two types in question are of the same so obviously have the same attributes. 

[1st Commits](https://github.com/nuwave/lighthouse/pull/1895/commits/c09c9d0f8ebdcce845e06fb69c99535a57ae9356#diff-5055d6975708def4e0cef74c1f1dea58fc4bef5dbfd28d41948380cf07fcbad6R46-R60)

spawnia:
@yondifon I think your approach should in fact work, but it has some problems.

`NodeList::merge()` does not eliminate duplicates. Use `ASTHelper::mergeUniqueNodeLists()` instead. I would like to see a test case where such a duplicate is present and resolved correctly, with the child items having precedence.

Why is there special treatment for the `name` attribute?

yondifon:
> @yondifon I think your approach should in fact work, but it has some problems.
> 
> `NodeList::merge()` does not eliminate duplicates. Use `ASTHelper::mergeUniqueNodeLists()` instead. I would like to see a test case where such a duplicate is present and resolved correctly, with the child items having precedence.
> 
> Why is there special treatment for the `name` attribute?

😄 I don't know why I did that on the name attribute. 

Thanks for the tip. I've restructured the tests to use `buildschema`. it's a lot easy and accurate. 

But I think it's all good now. I think we've resolved all the comments. 
Here's the latest commit that takes care of it. 
(https://github.com/nuwave/lighthouse/pull/1895/commits/53b5cb9af0aa9158971e5690a475634c2ac4076c)

Might need some formatting and minor fixes but it's all good.

spawnia:
Looks much better, definitely on the right track.

> I think we've resolved all the comments.

There are still 3 open comments which are not sufficiently addressed.

yondifon:
> Looks much better, definitely on the right track.
> 
> > I think we've resolved all the comments.
> 
> There are still 3 open comments which are not sufficiently addressed.

They are outdated 

I think it was for documentation. Just check the documentation once more. @enzonotario tweaked it a lil.

spawnia:
> They are outdated

No. I have looked at them. They are still relevant. Please address them.

yondifon:
> > They are outdated
> 
> No. I have looked at them. They are still relevant. Please address them.

Hey @spawnia. Just commit this (https://github.com/nuwave/lighthouse/pull/1895/commits/952a3706523b8af1068af9bec9e3655dd42f0037). 

Check this out. But I think you get the whole idea behind the whole, you can lil tweaks to better document the PR.

docs/master/api-reference/directives.md
spawnia: @yondifon you keep marking this as resolved, yet have not addressed this comment at all.

yondifon: I think you are looking at the wrong commit. It says (outdated) which I have updated. 


Here's the latest that addresses this. (https://github.com/nuwave/lighthouse/pull/1895/commits/952a3706523b8af1068af9bec9e3655dd42f0037)

spawnia: How does this address multiple inheritance? Do you have a test for it? The directive is missing the `repeatable` keyword as of now.

tests/Integration/Schema/Directives/InheritsDirectiveTest.php
```diff
@@ -0,0 +1,139 @@
...
+        ' . self::PLACEHOLDER_QUERY);
+
+        $childType = $schema->getType('ChildType');
+
+        $this->assertNotNull($childType->getField('attribute_1'));
+        $this->assertNotNull($childType->getField('attribute_2'));
+        $this->assertNotNull($childType->getField('attribute_3'));
+    }
+
+    public function testSchemaOrderIsInsignificant(): void
```
spawnia, line 49: This test highlights a problem with the current implementation. Given that the schema can be spread across multiple different files, the order of type definitions might not be predictable in every case. Regardless of definition order, inheritance should always lead to the same result.

There could be other solutions to this, but my instinct is to recursively resolve the inheritance of parent types before merging the attributes into a type.

tests/Integration/Schema/Directives/InheritsDirectiveTest.php
```diff
@@ -0,0 +1,154 @@
...
+                attribute_1: String
+            }
+            
+            type ChildType @inherits(from: "ParentType") {
+                attribute_2: String
+            }
+        ' . self::PLACEHOLDER_QUERY);
+    }
+
+    public function testForbidsInheritanceCycle(): void
```
spawnia, line 129: This test highlights another issue that we need to solve. Depending on the implementation, cyclic inheritance would either lead to nonsensical/incomplete results or endless loops. We need to detect cycles before resolving the inheritance tree.

spawnia (changes requested):

tests/Integration/Schema/Directives/InheritsDirectiveTest.php
yondifon: https://github.com/nuwave/lighthouse/pull/1895/commits/4b752906da64675b5bae7da88677a2c6855c2a5f

Just took care of this.

spawnia: Thanks, but that implementation is not sufficient - see [`fd894da` (#1895)](https://github.com/nuwave/lighthouse/pull/1895/commits/fd894dabd336bc8c90cd7d40da2f5c29d2827efa)

spawnia: https://github.com/webonyx/graphql-php/blob/master/src/Validator/Rules/NoFragmentCycles.php is doing a similar cycle detection and might give you an idea how to properly implement this.

Issue #1784, opened 2021-04-04
Switch test setup to MySQL 8


spawnia:
The newest Ubuntu 20.04 in GitHub actions is now using MySQL 8: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#mysql.

MySQL 8.0 requires a new authentication method that is only supported in PHP 7.4+.defaults:
Once we drop support for older PHP versions, we can use ubuntu-20.04.
Also, default ordering semantics in MySQL 8 seem to have changed, so we will need
to change a bunch of `assertJson()` to `assertSimilarJson()`.

spawnia:
Should we actually do this though? The most popular version is still 5.7 - see https://www.eversql.com/mysql-8-adoption-usage-rate.

PR #1898, opened 2021-07-20
Ensure compatibility with graphql-php 15

spawnia:


olivernybroe:
Just wanted to say that we are actually running with webonyx 15 alpha and this branch of lighthouse and are having no issues with it so far :)

olivernybroe (dismissed):
Lots of changes in this branch 👍 

But my general look through seems like this is fine. 
Also wanna say that we have actually been running this in production for quite a while, no issues for us. 😅

.github/workflows/validate.yml
```diff
@@ -197,13 +140,24 @@ jobs:
         laravel-version:
           - ^8
 
+    services:
+      mysql:
+        image: mysql:5.7
```
olivernybroe, line 145: We should consider at some point upgrading the mysql version (separate task ofc.)

composer.json
```diff
@@ -26,44 +26,45 @@
...
-        "nunomaduro/larastan": "^0.6 || ^0.7 || ^1 || ^2",
-        "orchestra/testbench": "3.6.* || 3.7.* || 3.8.* || 3.9.* || ^4 || ^5 || ^6 || ^7",
-        "phpbench/phpbench": "^1",
-        "phpstan/phpstan": "~0.12 || ^1",
-        "phpstan/phpstan-mockery": "~0.12 || ^1",
-        "phpstan/phpstan-phpunit": "~0.12 || ^1",
-        "phpunit/phpunit": "^7.5 || ^8.4 || ^9",
+        "mockery/mockery": "^1.5",
+        "nesbot/carbon": "^2.62.1",
+        "nunomaduro/larastan": "^0.7 || ^1 || ^2",
```
olivernybroe, line 57: As php 8 is now required, I think we can tighten this dev dependency to a newer version

src/Federation/Types/Any.php
```diff
@@ -14,9 +14,9 @@
  */
 class Any extends ScalarType
 {
-    public $name = '_Any';
+    public string $name = '_Any';
 
-    public $description = /** @lang Markdown */ <<<'DESCRIPTION'
+    public ?string $description = /** @lang Markdown */ <<<'DESCRIPTION'
```
olivernybroe, line 19: we should prob. standardize if we use `?string` or `string|null`

olivernybroe (approved):
A lot of changes, so might have missed some stuff, but looks good to me. 

I have added a few comments for things to think about, but not blockers

src/Schema/TypeRegistry.php
```diff
@@ -68,14 +68,17 @@ class TypeRegistry
     /**
      * Map from type names to resolved types.
      *
-     * @var array<string, \GraphQL\Type\Definition\Type>
+     * May contain `null` if the type is known to not exist, this allows
+     * short-circuiting repeated lookups for the same type.
+     *
+     * @var array<string, (\GraphQL\Type\Definition\Type&\GraphQL\Type\Definition\NamedType)|null>
      */
     protected $types = [];
```
olivernybroe, line 76: how can the type be null? 🤔

tests/Integration/DefaultSchemaTest.php
```diff
@@ -45,7 +45,7 @@ public function testPassesValidation(): void
 
     public function testFindRequiresExactlyOneArgument(): void
     {
-        if (AppVersion::below(8.0)) {
+        if (AppVersion::below(8.59)) {
```
olivernybroe, line 48: Should we just bump min. version to 8.59 instead?

composer.json
spawnia: We only require PHP 7.4, but a bump of larastan/phpstan was possible nonetheless.

src/Schema/TypeRegistry.php
spawnia: As explained above. I changed the description to hopefully be more clear.

.github/workflows/validate.yml
spawnia: It is still the most popular version though: https://www.eversql.com/mysql-8-adoption-usage-rate/

src/Federation/Types/Any.php
spawnia: Yes, when we make the switch to PHP 8.

.github/workflows/validate.yml
spawnia: Also, see https://github.com/nuwave/lighthouse/issues/1784.

src/Federation/Types/Any.php
olivernybroe: ah right, forgot it's php 8 😆

olivernybroe (dismissed):
lgtm

jeremynikolic:
Hi 👋 

I've recently found some issues caused (in our codebase, not the package itself) by the removal of `category => "validation"` in ValidationException by this commit.
Which was set initially by 
```PHP
    public function extensionsContent(): array
    {
        return [
            self::CATEGORY => $this->validator->errors()->messages(),
        ];
    }
```
As we had some frontend tools relying on the checking the category.

Before digging more I wanted to reach out for your input regarding, if :
- this is a cleanup from old features and could safely be restored by custom code in our project
- if restoring it would cause broader issue regarding compatibility with GQL which was the point of this commit ?

spawnia:
See https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md#breaking-removed-error-extension-field-category, it was a non-standard feature. As described, you can emulate it, replace or just omit it.

jeremynikolic:
Thanks a lot @spawnia 🙏

PR #1900, opened 2021-07-26
Allow using date validation rules that reference other arguments

ChrisExP:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The Laravel date validation rules (after, after_or_equal, before,
before_or_equal) allow the user to specify a date or another field from
the request to compare.
This didn't work with Lighthouse because the full path to the field
wasn't added to the rule arguments.
In this commit the argument is checked, if it is a date then nothing
happens, otherwise it is assumed to be another field from the request
and so adds the full argument path.

**Breaking changes**

This shouldn't break any existing projects as it uses the same check that Laravel uses when seeing if the argument is a date or a field reference, so it should behave in the same way.

### DIFF
--- a/src/Validation/RulesGatherer.php
+++ b/src/Validation/RulesGatherer.php
@@ -2,8 +2,8 @@
 
 namespace Nuwave\Lighthouse\Validation;
 
-use Carbon\Carbon;
 use Illuminate\Contracts\Validation\Rule;
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
 use Illuminate\Support\DateFactory;
 use Illuminate\Validation\ValidationRuleParser;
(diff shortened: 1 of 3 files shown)


src/Validation/RulesGatherer.php
```diff
@@ -4,6 +4,7 @@
 
 use Illuminate\Contracts\Validation\Rule;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Date;
```
spawnia, line 7: We do not use facades to stay compatible with Lumen.

spawnia (changes requested):
Well done, if you can replace the facade usage with something else I am happy to merge this.

ChrisExP:
Sorry about that, I have changed it to use the `DateFactory` class instead.

Issue #1901, opened 2021-07-28
Allow custom validation to use field reference arguments


ChrisExP:
**Lighthouse has no way of knowing when a custom validation rule wants to reference another field.**

I have a custom validation rule called `MaxDifference` which ensures that one date field cannot be more or less than a specified interval from another field in the request.  
Because of the way graphQL works the path to the field in the request is ambiguous and cannot be guaranteed at the point the validation rule is defined.  
Lighthouse works around this for the default validation fields by injecting the path when it gathers up the rules that need it, but there is no equivalent logic for custom rules.

**Possible solutions**
- The `rules` method on the `ArgumentSetValidation` and `ArgumentValidation` interfaces could accept an optional argument that provides the path to the fields being validated so the developer could create a custom directive that injects the path into the rule themselves.
> This would allow anyone to use references in any custom rules however it does put the work on the developer and I don't know if this would include any breaking changes.

- Have an interface, something like `ReferencesOtherFields`, which defines a method `updateReferencePath` which the developer can use in the custom rule class. The `RulesGatherer` class can then check for this interface and call the method with the absolute path to the fields, then the rule can store that and use it when validating.
> This would definitely have no breaking changes however it would only work for custom `Rule` classes and not for closures or extending the `Validator` class.

I am happy to create a pull request for this but I wanted to start a discussion if someone has any better ideas?

spawnia:
Your first solution could be doable through an enhanced version of `@rules` and `@rulesForArray`, perhaps some kind of extra syntax for a path on rules? To work with `@validator`, which basically only points to a validator class and does not specify the rules itself, we would have to extend the specific syntax to work there as well.

Your second solution seems like it should work, but requires changing the rules to make them work with Lighthouse.

Perhaps we can combine both approaches and use composition. We could define a custom rule such as `WithReference` and also register it through `Validator::extend()` to have it available in short string form. It would work by wrapping your own custom rule, something like this:

```graphql
@rules(apply: ["with_reference:max_difference,0,foo,5"])
```

Its arguments are:
1. Wrapped rule
2. Index of which argument of the wrapped rule is a reference
3. Arguments of the wrapped rule (any number)

ChrisExP:
I like that. Ok I will make the PR tomorrow.

PR #1903, opened 2021-07-29
Add support for custom rules that use references

ChrisExP:
Laravel allows creating custom validation rules. These could reference
other fields sent in the request. Due to the nature of graphQL when you
generate the schema the input types could be placed in multiple contexts
so the absolute field path cannot be known when generating the rules.
Lighthouse deals with the default validation fields by adding the
argument path to the parameters that need it but it didn't support
custom validation.

- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Pull request for #1901

**Changes**

This change allows the developer to wrap their custom rule in a
`with_reference` rule specifying the rule and the index of the parameter
that should be used as a reference. Alternatively with custom rule
classes the rule can implement the `WithReferenceRule` interface and
manipulate the path with the `setArgumentPath` method.


ChrisExP:
I was able to get the composition working without even creating a rule or an extension because the rule has to go through the rule gatherer however it is defined, I was able to just unwrap it there.

However I couldn't think of a way to use composition for custom rule classes because it is up to the developer how that accesses the other fields. There was no way I could think to get around that, so I fell back to the interface method I suggested.

I am happy to make any changes if you have other suggestions.

I wasn't sure if this was a big enough change to add it into the documentation.

src/Support/Contracts/WithReferenceRule.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
+interface WithReferenceRule
+{
+    public function setArgumentPath(string $argumentPath): void;
```
spawnia, line 7: This should have a docblock explaining what the argument path exactly is and how to use it.

src/Validation/RulesGatherer.php
```diff
@@ -224,6 +225,10 @@ protected function qualifyArgumentReferences(array $rules, array $argumentPath):
         return array_map(
             static function ($rule) use ($argumentPath) {
                 if (is_object($rule)) {
+                    if ($rule instanceof WithReferenceRule) {
+                        $rule->setArgumentPath(implode('.', $argumentPath));
```
spawnia, line 229: We can simply pass the array, access to more structured data is generally better.

src/Validation/RulesGatherer.php
```diff
@@ -238,6 +243,20 @@ static function ($rule) use ($argumentPath) {
                 $name = $parsed[0];
                 $args = $parsed[1];
 
+                if ($name === 'WithReference') {
+                    $indexes = explode('_', $args[1]);
+                    array_splice($args, 1, 1);
+                    foreach ($indexes as $index) {
+                        $index = (int) $index + 1;
```
spawnia, line 250:
```suggestion
                        // Skipping over the first index, which is the name
                        $index = (int) $index + 1;
```

spawnia:
This is a new feature and thus warrants a minor version bump and documentation. How else is anyone supposed to know about this?

ChrisExP:
The tests are failing because in Laravel 8 they added an interface that allows custom rule classes to interact with the validator data.
Before that this wouldn't have been possible with custom rule classes so the test wouldn't apply to versions before 8.
Is there any way to tell it to skip a test and for PHPStan to skip a file for Laravel versions below 8, or should I leave the test out?

spawnia:
You can just test this without the interface. The validation rule for the test does not have to make sense, for our purposes it is probably easiest if it always fails and includes the argument path in the resulting error message.

PR #1917, opened 2021-08-17
Feature/update can directive

### LINKED ISSUE #1916: Add a `field` argument to the @can directive to resolve models by custom key
It's not possible to resolve the model in a query like this where the primary key is the uuid property and not the id. `articleByUUID(uuid: String @eq): Article @can(ability: "view", find: "uuid") @find @softDeletes`

The above code will try to resolve the Article model by the id property and return an error.
`No query results for model [App\\Models\\Article]`

Adding a field argument will allow us to make the query like this and resolve the model as expected.
`articleByUUID(uuid: String @eq): Article @can(ability: "view", find: "uuid", field: "uuid") @find @softDeletes`

A custom directive, but i believe this addition is the correct approach.

SylvesterNielsen:
- [ ] Added or updated tests
- [X] Updated CHANGELOG.md

Resolves #1916

**Changes**
Added a `field` argument to the `@can` directive which allows for models to be resolved by a custom key.
See the issue for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


spawnia:
Let's explore possible solutions in #1916 first.

PR #1919, opened 2021-08-17
Add `query` and `scopes` arguments to the `@can` directive

SylvesterNielsen:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/1916

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -32,11 +32,11 @@
 
     <php>
         <env name="LIGHTHOUSE_TEST_DB_DATABASE" value="test"/>
-        <env name="LIGHTHOUSE_TEST_DB_HOST" value="localhost"/>
+        <env name="LIGHTHOUSE_TEST_DB_HOST" value="mysql"/>
         <env name="LIGHTHOUSE_TEST_DB_PORT" value="3306"/>
         <env name="LIGHTHOUSE_TEST_DB_USERNAME" value="root"/>
         <env name="LIGHTHOUSE_TEST_DB_PASSWORD" value=""/>
-        <env name="LIGHTHOUSE_TEST_REDIS_HOST" value="localhost"/>
+        <env name="LIGHTHOUSE_TEST_REDIS_HOST" value="redis"/>
         <env name="LIGHTHOUSE_TEST_REDIS_PASSWORD" value="null"/>
         <env name="LIGHTHOUSE_TEST_REDIS_PORT" value="6379"/>
         <env name="LIGHTHOUSE_TEST_REDIS_DB" value="0"/>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -32,11 +32,11 @@
 
     <php>
         <env name="LIGHTHOUSE_TEST_DB_DATABASE" value="test"/>
-        <env name="LIGHTHOUSE_TEST_DB_HOST" value="mysql"/>
+        <env name="LIGHTHOUSE_TEST_DB_HOST" value="localhost"/>
         <env name="LIGHTHOUSE_TEST_DB_PORT" value="3306"/>
         <env name="LIGHTHOUSE_TEST_DB_USERNAME" value="root"/>
         <env name="LIGHTHOUSE_TEST_DB_PASSWORD" value=""/>
-        <env name="LIGHTHOUSE_TEST_REDIS_HOST" value="redis"/>
+        <env name="LIGHTHOUSE_TEST_REDIS_HOST" value="localhost"/>
         <env name="LIGHTHOUSE_TEST_REDIS_PASSWORD" value="null"/>
         <env name="LIGHTHOUSE_TEST_REDIS_PORT" value="6379"/>
         <env name="LIGHTHOUSE_TEST_REDIS_DB" value="0"/>
--- a/src/Schema/Directives/CanDirective.php
+++ b/src/Schema/Directives/CanDirective.php
@@ -79,7 +79,6 @@ directive @can(
 
   """
   Apply scopes to the underlying query.
-  The scopes can only be used in combination with the findByArgs argument
   """
   scopes: [String!]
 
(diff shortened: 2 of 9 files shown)


phpunit.xml.dist
```diff
@@ -32,11 +32,11 @@
 
     <php>
         <env name="LIGHTHOUSE_TEST_DB_DATABASE" value="test"/>
-        <env name="LIGHTHOUSE_TEST_DB_HOST" value="mysql"/>
+        <env name="LIGHTHOUSE_TEST_DB_HOST" value="localhost"/>
```
spawnia, line 35: You can override this through your own `phpunit.xml`. The existing config works within the included docker-compose stack.

src/Schema/Directives/CanDirective.php
```diff
@@ -225,4 +260,11 @@ protected function buildCheckArguments(array $args): array
 
         return $checkArguments;
     }
+
+    public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefinitionNode &$fieldDefinition, ObjectTypeDefinitionNode &$parentType)
+    {
+        if ($this->directiveHasArgument('field') && $this->directiveHasArgument('findByArgs')) {
```
spawnia, line 266:
```suggestion
        if ($this->directiveHasArgument('find') && $this->directiveHasArgument('findByArgs')) {
```

src/Schema/Directives/CanDirective.php
```diff
@@ -95,7 +112,25 @@ function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
...
+                    if ($this->directiveArgValue('find') !== null) {
+                        throw new DefinitionException("Can not use the argument 'find' when 'findByArgs' is true");
+                    }
+                    $models = $resolveInfo
+                        ->argumentSet
+                        ->enhanceBuilder(
+                            $this->getModelClass()::query(),
+                            $this->directiveArgValue('scopes', [])
+                        )->get();
+                    if (! $models->count()) {
```
spawnia, line 125: I don't see why this check is necessary. An empty result is not necessarily an error.

src/Schema/Directives/CanDirective.php
```diff
@@ -95,7 +112,25 @@ function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
                 $gate = $this->gate->forUser($context->user());
                 $checkArguments = $this->buildCheckArguments($args);
 
-                foreach ($this->modelsToCheck($resolveInfo->argumentSet, $args) as $model) {
+                if ($this->directiveArgValue('findByArgs') === true) {
+                    if ($this->directiveArgValue('find') !== null) {
```
spawnia, line 116: Already happens during schema validation.

src/Schema/Directives/CanDirective.php
```diff
@@ -95,7 +112,25 @@ function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
...
+                        throw new DefinitionException("Can not use the argument 'find' when 'findByArgs' is true");
+                    }
+                    $models = $resolveInfo
```
spawnia, line 119: Given this is resolving a list of models to check permissions against, this should live in `modelsToCheck()`.

src/Schema/Directives/CanDirective.php
```diff
@@ -66,6 +71,18 @@ public static function definition(): string
...
 
+  """
+  Resolve the model by directive args like @eq, @where etc.
+  You may not use the find argument in combination with findByArgs
+  """
+  findByArgs: Boolean = false
+
+  """
+  Apply scopes to the underlying query.
+  The scopes can only be used in combination with the findByArgs argument
```
spawnia, line 82: Nope, given you already added it in `modelsToCheck()`. It makes sense to enable it for both use cases.

tests/Integration/Schema/Directives/CanDirectiveDBTest.php
```diff
@@ -287,4 +288,239 @@ public function testWorksWithSoftDeletes(): void
...
+    {
+        $user = new User();
+        $user->name = UserPolicy::ADMIN;
+        $this->be($user);
+
+        $user = factory(User::class)->create([
+            'name' => 'foo',
+        ]);
+
+        $this->schema = /** @lang GraphQL */ '
```
spawnia, line 466:
```suggestion
        $this->buildSchema(/** @lang GraphQL */ '
```

Validated at build time, no need to fire a query.

spawnia (changes requested):
Nice work on the comprehensive test suite.

If I could go back, I would rename `find` to something else and use it for this new functionality. Is there something better then `findByArgs`? Some ideas:
- instance
- withModel
- resolve
- query

Laravel does not provide much guidance here. The Laravel default of using just the model class is actually the default in Laravel, see https://laravel.com/docs/8.x/authorization#methods-without-models

phpunit.xml.dist
SylvesterNielsen: Ups. That was not supposed to be commited. Nice catch. 😃

src/Schema/Directives/CanDirective.php
SylvesterNielsen: @spawnia The reason for this check is to keep the same behavior as the current methods. 
The `findOrFail()` will throw this exception.

spawnia: In order to avoid breakage, we should probably leave the existing behaviour as it is. In the case of `find` which only deals with primary keys, it is somewhat sensible to throw. Not finding a model by its primary key is usually indicative of an error.

This new argument does not mandate that the arguments have to look for a primary key, it can be any kind of constraint expressible through args and scopes. Empty results can be perfectly normal.

SylvesterNielsen:
@spawnia I think all is updated except for the naming. I'm not sure if any of the other ideas are better at this time.
Let me know if you find anything else and thank you for the thorough review.

spawnia (approved):
Thanks for your work and the productive back-and-forth. I am confident we have now arrived at a solution that will make `find` obsolete, perhaps I will remove it in v6.

PR #1921, opened 2021-08-20
Include definition when generating directives via `artisan lighthouse:directive`

spawnia:
- [x] Added or updated tests (manually)
- [x] Updated CHANGELOG.md

**Changes**

The required method `definition()` is now included and reasonably populated by default.

**Breaking changes**

Use `choice()` for multi-select of interfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v5.20.0
+
+### Added
+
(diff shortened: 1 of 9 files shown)


olivernybroe (dismissed):
LGTM 👍 

I didn't test it out locally though, but the code looks pretty fine.

I think the only thing I see as missing here would be if we could add an automated test for it. Should be possible by mocking the FileSystem class.

spawnia:
I think we had a go at adding an automated test for this before, it was pretty difficult.

PR #1925, opened 2021-08-23
Support schema without `Query` type when using federation

cappuc:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

When using federation it is possible to have a schema without the `Query` type and only types extensions.
With this change it is possible to omit the `Query` type

**Breaking changes**

none.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Support schema without `Query` type when using federation https://github.com/nuwave/lighthouse/pull/1925
+
 ## v5.22.1
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


spawnia:
This is not that easy, see https://github.com/graphql/graphql-spec/pull/631

If we make this possible, it should not be limited to Federation. We need to make sure that validation and introspection still work.

spawnia:
See https://github.com/webonyx/graphql-php/pull/920

cappuc:
Yes it makes sense to implement it globally but it's more complex because it needs to handle validation and introspection correctly.

With this "federation only" implementation the `Query` type is always defined with at least `_service` and `_entities` queries so the rest of the implementation doesn't need changes.

Anyway I did some tests with apollo gateway and this implementation (only type extension and no root Query) and it seems to works fine

PR #1939, opened 2021-09-19
`@spread`: resolver and array/list support

LastDragon-ru:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

### Motivation

We have some models with a lot of prefixed attributes eg `a_*`, `b_*` etc, in the GraphQL they returned as the objects (`A {a: Int,b: Int}, B {a: Int,b: Int}`). We also want to sort by these properties but with the same structure of types instead of real properties (= client should not know that `A` refers to the prefixed properties `a_*`). The problem is [`@sortBy`](https://github.com/LastDragon-ru/lara-asp/tree/master/packages/graphql#sortby-directive) cannot distinguish relations and prefixed properties. So I thought maybe we can use `@spread`? Unfortunately, `@spread` doesn't support prefixes and arrays.

### Changes

1. Support for arrays (`@spread` will be processed recursive inside arrays)
2. Support for `@spread(resolver: class)` that can change the name of the attribute (eg it can convert it the `snake_case` or `camelCase` or etc)

###  Examples

I our case will be possible:

```graphql
type Query {
  objects(
      order: ObjectsSort @sortBy
    ): [Object!]!
}

type Object {
  id: ID!
}

type ObjectsSort {
   a: ASort @spread(resolver: "App\\GraphQL\\SpreadResolver")
}

type ASort {
  a: string
  b: string
}
```

```php
<?php declare(strict_types = 1);

namespace App\GraphQL\SpreadResolver;

class Spread {
    public function __invoke(string $parent, string $name): string {
        return "{$parent}_{$name}";
    }
}
```

And 🥳 (`@sortBy` will receive `['a_a': 'desc']`)
```graphql
query {
  objects(order: {a: {a: desc}}) {
      id
  }
}
```

Or for example from the docs:
```
type Mutation {
  updatePost(id: ID!, input: PostInput! @spread): Post @update
}

input PostInput {
  title: String!
  image: PostImage @spread(resolver: `CamelCaseResolver`)
}

input PostContent {
  url: String
}
```

```
[
    'id' => 12,
    'title' => 'My awesome title',
    'imageUrl' = 'https://some.site/image.jpg',
]
```

**Breaking changes**

`@spread` will be processed inside List/array.

_PS: `@rename` and `@spread` are very similar now, probably the new interface can be added, eg `ArgSetTransformerDirective`._

### DIFF
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -85,11 +85,12 @@ class ArgumentSet
             if ($value instanceof self) {
                 // Recurse down first, as that resolves the more deeply nested spreads first
                 $value = $value->spread();
-
-                if ($argument->directives->contains(
+                $directive = $argument->directives->first(
                     Utils::instanceofMatcher(SpreadDirective::class)
-                )) {
-                    $argumentSet->arguments += $value->arguments;
+                );
+
+                if ($directive) {
+                    $argumentSet->arguments += $directive->transformArguments($name, $value->arguments);
                     continue;
                 }
             }
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -78,23 +78,34 @@ class ArgumentSet
         $argumentSet->directives = $this->directives;
 
         foreach ($this->arguments as $name => $argument) {
-            $value = $argument->value;
+            // Recursively apply the spread to nested inputs.
+            // We look for further ArgumentSet instances, they
+            // might be contained within an array.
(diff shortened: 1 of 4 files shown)


spawnia:
I fixed the application of `@spread` within lists as of https://github.com/nuwave/lighthouse/releases/tag/v5.23.1, thank you for bringing that up.

If you consider https://github.com/nuwave/lighthouse/pull/1944, it should become apparent that the functionality you desire for merging the parent keys can be implemented as a custom directive. I am not convinced that this use case is common enough to warrant the extra complexity in `@spread`.

PR #1941, opened 2021-09-22
Add an example for querying `morphTo` relations to make it clearer for new comers.

StashWorks:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md
- [X] Updated docs

**Changes**
No change in behavior, just added an example of how to query `morphTo` relations to the docs.

**Breaking changes**
None.


spawnia:
Thank you, I updated the docs with a modified example.

PR #1945, opened 2021-09-30
 Batchloading relations with different database connections #1252

### LINKED ISSUE #1252: Batchloading relations with different database connections
### The Scenario
- I have 3 models
- Model A, B, C
- A Exists on database A
- B and C exist on database B
- C is a pivot relationship/table that exists on database B as well.
- I am Using schema stitching
- I am trying to use the relationship `@belongsToMany(type: 'paginator')` - **This is where things break**

I implemented a database connection where all A Records Exists and called it connection2
B and C exist on the default database connection using a test controller I can verify that the connection and data exist and is queryable. Using a custom resolver also works.

---
#### The implementation (That doesnt work)

I created a graphQL file that basically defines the ff
```graphql
extend type A {
 bRelationship: [B!]! @belongsToMany(type: "paginator", relation: "bRelationship")
}
```
When trying to query this I get this error
```
SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation \"domain_names\" does not exist\nLINE 1: select \"organizations\".*, (select count(*) from \"domain_name...\n                                                        ^ (SQL: select \"organizations\".*, (select count(*) from \"domain_names\" inner join \"domain_name_organization\" on \"domain_names\".\"id\" = \"domain_name_organization\".\"domain_name_id\" where \"organizations\".\"id\" = \"domain_name_organization\".\"organization_id\" and \"domain_names\".\"deleted_at\" is null) as \"domain_names_count\" from \"organizations\" where \"organizations\".\"id\" in (1))
 ```

- Assume that "domain_names" is A.
- Assume that "organizations" is B
- Assume that domain_name_oragnization is C

---

#### The implementation (***That works***)

```graphQL
extend type A {
B: [B!]! @paginate(resolver: "App\\GraphQL\\Queries\\A@resolveB")
}
```

And the contents of that resolver look something like this:

```php
 $values = $rootValue->domainNames()->paginate();
 return $values;
```

This resolves the query and returns a propper graphql response.

---

####

I think there is a problem when trying to resolve the "paginator" type on the `@belongsToMany` relationship (or quite possibly all relationships, I havent tested) that is preventing it from being resolved.

I think that writing custom resolvers for each of these implementations defeats the purpose of having the custom default directives in the first place.

**Environment**

Lighthouse Version: 4.10
Laravel Version: 6

dolfje:
Make sure that we eager load relations when they are from another connection. So that we don't try to create an inner subquery that is impossible

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**

### DIFF
--- a/src/Schema/Directives/RelationDirective.php
+++ b/src/Schema/Directives/RelationDirective.php
@@ -35,8 +35,11 @@ abstract class RelationDirective extends BaseDirective implements FieldResolver
                 /** @var \Illuminate\Database\Eloquent\Relations\Relation $relation */
                 $relation = $parent->{$relationName}();
          
-                $sameConnection = $relation->getParent()->getConnectionName() == $relation->getRelated()->getConnectionName();
-                if (config('lighthouse.batchload_relations') && $sameConnection) {
+                if (
+                    config('lighthouse.batchload_relations')
+                    // Batch loading joins across both models, thus only works if they are on the same connection
+                    && $relation->getParent()->getConnectionName() === $relation->getRelated()->getConnectionName()
+                ) {
                     /** @var \Nuwave\Lighthouse\Execution\BatchLoader\RelationBatchLoader $relationBatchLoader */
                     $relationBatchLoader = BatchLoaderRegistry::instance(
                         $this->qualifyPath($args, $resolveInfo),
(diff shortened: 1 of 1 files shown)


spawnia:
Good find, thank you! How can we test if this works, and ensure it does not break in the future?

https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Execution/DataLoader/RelationBatchLoaderTest.php has a test that can determine if batchloading happens by the number of executed queries, that can be a starting point.

What if one model explicitly names the default connection, whereas the other simply has `null`? In this case, both will actually resolve to the same connection, right?

spawnia:
Incorporated your changes into https://github.com/nuwave/lighthouse/pull/1979, thanks. I just don't merge PRs without tests.

PR #1948, opened 2021-10-07
Change simple models loader to use `loadMissing` instead of `load`

olivernybroe:
The reason for this change is that `loadMissing` does not delete already loaded relationships,
it simple adds the new relationships to the already loaded relationships. Whereas `load`
completely reloads the relationships

- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**


olivernybroe:
Yep, seems this will cause issues for relationships which would be loaded in again later, but with a scope instead. 

I guess the solution then is that we need to group them.

PR #1955, opened 2021-10-14
Manipulate interfaces

thekonz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Currently, interfaces cannot use field manipulation directives. This PR aims to remove that restriction.
I added a test to demonstrate the use case.

Downside of the current implementation is that the $parentType of FieldManipulator::manipulateFieldDefinition cannot typehint both object and interface type. 
In php 8.0 we could use a union type. Alternatively, we need to have a combining type for object and interface type in webonyx/graphql-php (they are very similar anyways, so we could even have one extend from the other).

Any thoughts?


spawnia:
Given this changes interface signatures, this will be a breaking change.

> we need to have a combining type for object and interface type in webonyx/graphql-php

We are going to have a common superclass `TypeWithFields` in the upcoming `webonyx/graphql-php` 15. I am continuously testing compatiblity in https://github.com/nuwave/lighthouse/pull/1898.

I haven't committed to starting work on Lighthouse v6 just yet, I would like to go through the open pull requests and issues first and see if there is something that is easy or important to get into v5. As soon as that is done, we can start breaking stuff.

spawnia:

Issue #1958, opened 2021-10-18
Custom exception in middleware does not handle with the lighthouse error handler


ehsanrasta:
**Describe the bug**
I create a middleware "**CheckAdminEnabled2FA**" and put it in middleware array in config file:

```
       'middleware' => [
            \Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::class,
            \Nuwave\Lighthouse\Support\Http\Middleware\AcceptJson::class,

            \Nuwave\Lighthouse\Support\Http\Middleware\AttemptAuthentication::class,
            \App\Http\Middleware\CheckAdminEnabled2FA::class,
        ],
```

CheckAdminEnabled2FA:
```
namespace App\Http\Middleware;

use App\Exceptions\AdminTwoFactorAuthenticationNotEnabledException;
use Closure;
use Illuminate\Http\Request;
use PragmaRX\Google2FALaravel\Support\Authenticator;

class CheckAdminEnabled2FA
{
    /**
     * Handle an incoming request.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle(Request $request, Closure $next)
    {
        $authenticator = app(Authenticator::class)->boot($request);

        if ($authenticator->isActivated()) {
            return $next($request);
        }

        throw new AdminTwoFactorAuthenticationNotEnabledException('Enable 2FA for this account before using service');
    }
}
 
```
In middleware I throw an exception "**AdminTwoFactorAuthenticationNotEnabledException**" and this exception impliments RendersErrorsExtensions 

```
namespace App\Exceptions;

use Exception;
use Nuwave\Lighthouse\Exceptions\RendersErrorsExtensions;

class AdminTwoFactorAuthenticationNotEnabledException extends Exception implements  RendersErrorsExtensions
{
    public const MESSAGE = 'test.';
    public const CATEGORY = 'test';

    public function __construct(string $message)
    {
        parent::__construct($message);
    }

    public function getCategory(): string
    {
        return self::CATEGORY;
    }

    public function isClientSafe(): bool
    {
        return true;
    }

    /**
     * @return array<string, array<string>>
     */
    public function extensionsContent(): array
    {
        return ['key' => 'value'];
    }

    public static function fromLaravel(Exception $laravelException): self
    {
        return new static($laravelException->getMessage(), $laravelException->getCode());
    }

}

```
[...]

PR #1959, opened 2021-10-19
Render `GraphQL\Error\ClientAware` exceptions outside of GraphQL context with extensions content

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/1958


spawnia:
@ehsanrasta can you try if this solution works for you? I think it should be generalized to work in every Laravel application by default.

PR #1968, opened 2021-10-27
added header to graphql

emtudo:
- [X] Added

Added headers to query/mutation on tests


spawnia:
Implemented with proper tests in https://github.com/nuwave/lighthouse/pull/1968

PR #1954, opened 2021-10-14
Run tests on PHP 8.1

spawnia:

PR #1986, opened 2021-11-28
Test PHP 8.1 fixes

stayallive:
Expand on #1954 to move forward with 8.1 support even when the dependencies of `bensampo/laravel-enum` does not support it yet.


stayallive:
I overlooked like 9000 things, back to the drawing board 😎

PR #1992, opened 2021-12-01
Add directive `@void` to unify the definition of fields with no return value

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

See https://github.com/graphql/graphql-spec/issues/906

**Changes**

- Add directive `@void` to unify the definition of fields with no return value

**Breaking changes**

None, since the service provider is optional. When added by default, it will reserve the directive name `@void` and the type name `Unit`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v5.28.0
+
+### Added
+
+- Add directive `@void` to unify the definition of fields with no return value
+
 ## v5.27.1
 
 ### Changed
--- a/docs/5/performance/schema-caching.md
+++ b/docs/5/performance/schema-caching.md
@@ -14,7 +14,7 @@ using the [cache](../api-reference/commands.md#cache) artisan command:
 
 The structure of the serialized schema can change between Lighthouse releases.
 In order to prevent errors, use cache version 2 and a deployment method that
-atomically updates both the cache file and the dependencies, e.g. K8s. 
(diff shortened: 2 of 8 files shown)


stayallive (approved):
I like it, on a technical basis!

I just don't care for the choice for `Unit`, it seems like a random name that easily clashes with userland (I now learned that this is a [type theory](https://en.wikipedia.org/wiki/Unit_type) thing and makes sense but only if you know the sauce). Why not use `enum Void { VOID }` or `enum Empty { EMPTY }` or `enum Accepted { ACCEPTED }` (close to a HTTP 202 response) or even `enum Null { NULL }` or something a little closer to what it is actually doing?

Regardless, this seems like a nice addition!

pyrou:
This PR becomes a bit old.
But for what it worth, here is what I use following in my project :
```graphql
scalar Void @scalar(class: "App\\GraphQL\\Scalars\\VoidType")
```

With following dummy implementation (a class simply named "Void" isn't allowed in PHP) :
```php
class VoidType extends ScalarType
{
    public function serialize($value): mixed {
        return null;
    }

    public function parseValue($value): mixed {
        return null;
    }

    public function parseLiteral(Node $valueNode, ?array $variables = null): mixed {
        return null;
    }
}
```

By making Void a type rather than a directive, it becomes much more clear when defining the schema :
```graphql
        type Query {
            foo: Void
        }
```

The approach proposed with `@void` directive ends with very unnatural definition in schema, like the one seen in unit test :
```graphql
        type Query {
            foo: Int @void
        }
```

Is it void or Int ?

PR #1994, opened 2021-12-02
Fix undefined array key with missing cache config

stayallive:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

We have this config:

```php
    'cache' => [
        /*
         * Setting to true enables schema caching.
         */
        'enable'  => env('LIGHTHOUSE_CACHE_ENABLE', !env('APP_DEBUG', false)),

        /*
         * Allowed values:
         * - 1: uses the store, key and ttl config values to store the schema as a string in the given cache store.
         * - 2: uses the path config value to store the schema in a PHP file allowing OPcache to pick it up.
         */
        'version' => env('LIGHTHOUSE_CACHE_VERSION', 2),

        /*
         * File path to store the lighthouse schema.
         * Only relevant if version is set to 2.
         */
        'path'    => env('LIGHTHOUSE_CACHE_PATH', base_path('bootstrap/cache/lighthouse-schema.php')),
    ],
```

It's missing some keys for the v1 cache strategy which currently breaks even when using version 2.

Refactored it a bit to prevent that.

**Breaking changes**

None.

### DIFF
--- a/src/Schema/AST/ASTCache.php
+++ b/src/Schema/AST/ASTCache.php
@@ -59,12 +59,12 @@ class ASTCache
         $this->enable = $cacheConfig['enable'];
 
         $version = $cacheConfig['version'] ?? 1;
-        
+
         switch ($version) {
             case 1:
                 $this->store = $cacheConfig['store'] ?? null;
-                $this->key   = $cacheConfig['key'];
-                $this->ttl   = $cacheConfig['ttl'];
+                $this->key = $cacheConfig['key'];
+                $this->ttl = $cacheConfig['ttl'];
                 break;
             case 2:
                 $this->path = $cacheConfig['path'] ?? base_path('bootstrap/cache/lighthouse-schema.php');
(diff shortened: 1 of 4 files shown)


olivernybroe (approved):
LGTM

src/Schema/AST/ASTCache.php
```diff
@@ -59,18 +59,21 @@ public function __construct(ConfigRepository $config)
...
-        if (! in_array($version, [1, 2])) {
-            throw new UnknownCacheVersionException($version);
-        }
-        $this->version = $version;
 
-        // Version 1
-        $this->store = $cacheConfig['store'] ?? null;
-        $this->key = $cacheConfig['key'];
-        $this->ttl = $cacheConfig['ttl'];
+        switch ($version) {
```
olivernybroe, line 63: if we think this is gonna be more and more of a thing, with new caching versions etc. we could consider making a driver based system, so all of the configs needed depending on which version is in it's own class

stayallive: Yeah this might even make more sense in the future!

spawnia (commented):
nicely done

PR #1997, opened 2021-12-04
Fix undefined array key with interfaces and two renamed models

LavaDesu:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

In `TypeRegistry->typeResolverFallback`, use `array_pop` instead of `[0]`, as `array_intersect` might return an array with a key that isn't 0.
This can happen with two renamed models that implement an interface, and a result with the second model is returned (which has key `1`)

**Breaking changes**

None.

### DIFF
--- a/src/Schema/TypeRegistry.php
+++ b/src/Schema/TypeRegistry.php
@@ -489,7 +489,7 @@ EOL
                         );
                     }
 
-                    return $this->get($actuallyPossibleTypes[0]);
+                    return $this->get(array_pop($actuallyPossibleTypes));
                 }
 
                 return $this->get(class_basename($root));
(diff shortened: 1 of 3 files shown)


src/Schema/TypeRegistry.php
```diff
@@ -489,7 +489,7 @@ protected function typeResolverFallback(array $possibleTypes): Closure
                         );
                     }
 
-                    return $this->get($actuallyPossibleTypes[0]);
+                    return $this->get(array_pop($actuallyPossibleTypes));
```
spawnia, line 492:
```suggestion
                    return $this->get(end($actuallyPossibleTypes));
```

spawnia:
Thanks, released with https://github.com/nuwave/lighthouse/releases/tag/v5.27.3

PR #2004, opened 2021-12-10
Detect usage of deprecated elements in queries

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

**Breaking changes**


spawnia:
Should use `App::terminating()` for error reporting

stayallive:
Am I understanding correctly that this is just for testing basically? I am unsure what the use of this rule is outside of the testing harnass but maybe I'm missing something 😄

spawnia:
@stayallive I just added some docs. Let me know what you think

stayallive:
Oooooooh. Yes. This is very cool! And really handy 💪

Issue #2007, opened 2021-12-13
Call format() on null in Tracing when using subscriptions and a queue


MaartenStaa:
We just started experimenting with subscriptions, and are running into an error in our testing environment, which is running background jobs using Laravel Horizon:

```Call to a member function format() on null```

This seems to be caused by the `StartRequest` event not being triggered (which makes sense), but the `Tracing` class assuming that it will.

**Expected behavior/Solution**

The `Tracing` class should use a value other than the request start time to set the `startTime` in the tracing extension.

**Steps to reproduce**

1. Trigger a subscription
2. The queue driver should not be set to `sync`
3. Run either `php artisan queue:work` or `php artisan horizon` (not sure if it matters)

**Output/Logs**

<details><summary>Stack trace</summary>

```
test.ERROR: Call to a member function format() on null {"exception":"[object] (Error(code: 0): Call to a member function format() on null at vendor/nuwave/lighthouse/src/Tracing/Tracing.php:82)
[stacktrace]
#0 vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php(424): Nuwave\\Lighthouse\\Tracing\\Tracing->handleBuildExtensionsResponse()
#1 vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php(249): Illuminate\\Events\\Dispatcher->Illuminate\\Events\\{closure}()
#2 vendor/nuwave/lighthouse/src/GraphQL.php(218): Illuminate\\Events\\Dispatcher->dispatch()
#3 app/GraphQL/GraphQL.php(33): Nuwave\\Lighthouse\\GraphQL->executeQuery()
#4 vendor/nuwave/lighthouse/src/Subscriptions/SubscriptionBroadcaster.php(87): App\\GraphQL\\GraphQL->executeQuery()
#5 vendor/nuwave/lighthouse/src/Subscriptions/Iterators/AuthenticatingSyncIterator.php(54): Nuwave\\Lighthouse\\Subscriptions\\SubscriptionBroadcaster->Nuwave\\Lighthouse\\Subscriptions\\{closure}()
#6 vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php(242): Nuwave\\Lighthouse\\Subscriptions\\Iterators\\AuthenticatingSyncIterator::Nuwave\\Lighthouse\\Subscriptions\\Iterators\\{closure}()
[...]

PR #2009, opened 2021-12-15
Use `StartExecution` event to determine per-operation Tracing timestamps

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/2007

**Changes**

Use `StartExection` over `StartRequest` to mark the beginning of an operation in tracing. This has two advantages:
- accurate timing when using batched requests
- works in a non-http context

**Breaking changes**

No, but the timing values will differ.

### DIFF
--- a/src/Tracing/TracingServiceProvider.php
+++ b/src/Tracing/TracingServiceProvider.php
@@ -2,13 +2,14 @@
 
 namespace Nuwave\Lighthouse\Tracing;
 
+use GraphQL\Language\Parser;
 use Illuminate\Contracts\Events\Dispatcher as EventsDispatcher;
 use Illuminate\Support\ServiceProvider;
 use Nuwave\Lighthouse\Events\BuildExtensionsResponse;
 use Nuwave\Lighthouse\Events\ManipulateAST;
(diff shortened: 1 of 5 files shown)


src/Tracing/TracingServiceProvider.php
```diff
@@ -28,12 +29,12 @@ static function (): string {
...
-            Tracing::class . '@handleManipulateAST'
-        );
-
-        $eventsDispatcher->listen(
-            StartRequest::class,
-            Tracing::class . '@handleStartRequest'
+            static function (ManipulateAST $manipulateAST): void {
+                ASTHelper::attachDirectiveToObjectTypeFields(
+                    $manipulateAST->documentAST,
+                    Parser::constDirective('@tracing')
```
stayallive, line 35: An optimization could be to maybe put the `Parser::constDirective('@tracing')` outside the closure so it only has to parse it once instead of for every object type in the schema. I'm not a 100% sure it matters (especially with schema caching not) but still.

spawnia: done https://github.com/nuwave/lighthouse/commit/b09492276e175fad34e33d60f172055704ec7f2b

PR #2013, opened 2021-12-26
CSRF protection middleware

### LINKED ISSUE #1582: CSRF protection for Lighthouse
**Are you willing to provide a PR for this issue or aid in developing it?**
------
Yes

------
CSRF is a widely known type of attacks that occurs when a malicious web site causes a user's web browser to perform an unwanted action on a trusted site when the user is authenticated.

The rule of the thumb is that modern frameworks include and enable CSRF protection by default.

Right now the default Lighthouse installation allows CSRF. For example, the attacker may insert the following simple code in a blog comment:

```html
<img src="/graphql?query={EvilMutation{}}" />
```

When the victim opens this comment, it will cause the browser to make a request to the graphql endpoint with all authentication cookies. And execute the EvilMutation.

Though clean Lighthouse installation doesn't create this vulnerability by itself (the user has to write some code that enables this vulnerability), I suppose it should encourage applying right practices and warn users about possible caveats.

------

The simplest solution I've found is to enforce the "Content-type" request header to be "application/json" and reject any requests that have an invalid or missing header.

Here is why it would work. The attacker may forge "Content-type" only with POST form or XHR:
1) the POST form content type is limited to only three formats: "application/x-www-form-urlencoded", "multipart/form-data" and "text/plain".
2) if the attacker can execute an arbitrary XHR from your site, it means they can execute any javascript, so the site is already vulnerable to XSS.

So by enforcing the "Content-type" request header to be "application/json", we ensure, that request is either legitimate or the  attacker already has gained full control over the client. 

How we can achieve it:
1) removing the AcceptJson middleware from the default lighthouse config. Add a warning, that enabling this middleware may cause your application to be vulnerable to CSRF attacks.
2) accepting only POST requests
3) adding the "EnforceJsonContentType" middleware, which rejects any request that doesn't have a "Content-type" request header set to "application/json".
4) adding an explanation to the documentation why these middlewares should be changed very cautiously or with understanding of the possible caveats.

Some more features may be added, these are the most important ones.

Further reading
------
https://stackoverflow.com/questions/11008469/are-json-web-services-vulnerable-to-csrf-attacks
https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html

k0ka:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1582

**Changes**

Added the `EnforceJsonContentType` as we discussed in the issue above.

**Breaking changes**

None. The middleware is not enabled by default.


spawnia (changes requested):
I think a broader set of content types should be valid, see https://github.com/laragraph/utils/blob/master/src/RequestParser.php
- `application/graphql`
- `application/graphql+json`

k0ka:
I've rewritten the middleware to make it more permissive and general. Now it allows any request that is not `GET` or `form POST`. And also renamed it to make it more clear.

spawnia (approved):

spawnia:
Thanks, nicely done

Issue #651, opened 2019-03-06
Automatic persisted queries


spawnia:
Let's up the feature parity with Apollo and implement APQ:

https://blog.apollographql.com/automatic-persisted-queries-and-cdn-caching-with-apollo-server-2-0-bf42b3a313de

**Challenges**

How do we modify the incoming request?

Is APQ interoperable with other advanced features:
- Batching
- Defer
- Tracing
- File Uploads

How about mutations, are those allowed?

spawnia:
@chrissm79 i take it you have been working on this in a project of yours.

Do you think it has a place in Lighthouse core?

spawnia:
Supported as of https://github.com/nuwave/lighthouse/releases/tag/v5.36.0

PR #2017, opened 2021-12-30
Cache queries

k0ka:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Query caching has been added: the result of `GraphQL\Language\Parser::parse()` will be serialized and stored in the Laravel cache. It can be disabled in the config.

I tested it on my production queries and it shows about 5-10% performance boost. I use Relay with a lot of fragments. 

I use sha256 hash of query as a key. So we may develop it further towards implementing [Apollo Automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/). While Relay currently uses md5 hashes for their [Persisted queries](https://relay.dev/docs/next/guides/persisted-queries/), I guess they will [change it](https://github.com/facebook/relay/pull/3666) soon to a better hashing algorithm. 

**Breaking changes**

`GraphQL::executeQuery()` is deprecated and should be removed in the next version. `parseAndExecuteQuery()` or `executeParsedQuery()` should be used instead. This PR has a simple `executeQuery()` stub that executes one of above methods depending on the `query` type.


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Added the query caching
```
```suggestion
- Added query caching
```

UPGRADE.md
```diff
@@ -126,6 +126,11 @@ the new behaviour before upgrading by setting `unbox_bensampo_enum_enum_instance
 public function scopeByType(Builder $builder, AOrB $aOrB): Builder
 ```
 
+### Nuwave\Lighthouse\GraphQL::executeQuery() is deprecated
+
+Use `parseAndExecuteQuery()` for executing a string query or `executeParsedQuery()` for 
+executing already parsed `DocumentNode`
```
spawnia, line 132:
```suggestion
executing already parsed `DocumentNode`.
```

docs/master/performance/query-caching.md
```diff
@@ -0,0 +1,8 @@
+# Query caching
+
+In order to speed up the GraphQL query parsing, its results can be stored in the Laravel cache.
+
+Query caching is enabled in non-local environments by default. You can define cache store and cache duration,
+see `config/lighthouse.php`.
+
+The GraphQL schema doesn't affect query parsing, so you have to flush query cache only on the package update.
```
spawnia, line 8:
```suggestion
Make sure to flush the query cache when you deploy an upgraded version of the dependency `webonyx/graphql-php`:

    some-one-liner that clears the cache?
```

src/GraphQL.php
```diff
@@ -101,7 +103,7 @@ public function __construct(
     /**
      * Run one ore more GraphQL operations against the schema.
      *
-     * @param  \GraphQL\Server\OperationParams|array<int, \GraphQL\Server\OperationParams>  $operationOrOperations
```
spawnia, line 104: please revert to the previous formatting in `@param`

src/lighthouse.php
```diff
@@ -126,6 +126,37 @@
...
+    |
+    | Caches user queries so we won't have to parse string queries on each request
+    |
+    */
+
+    'query_cache' => [
+        /*
+         * Setting to true enables query caching.
+         */
+        'enable' => env('LIGHTHOUSE_QUERY_CACHE_ENABLE', 'local' !== env('APP_ENV')),
```
spawnia, line 142: why not local?

src/lighthouse.php
```diff
@@ -126,6 +126,37 @@
...
+        'store' => env('LIGHTHOUSE_QUERY_CACHE_STORE', null),
+
+        /*
+         * Duration in seconds the query should remain cached, null means forever.
+         */
+        'ttl' => env(
+            'LIGHTHOUSE_QUERY_CACHE_TTL',
+            \Nuwave\Lighthouse\Support\AppVersion::atLeast(5.8)
+            ? null
+            : 365 * 24 * 60 // For Laravel < 5.8 the exact value must be specified and it is counted in minutes
```
spawnia, line 156:
```suggestion
                ? null
                : 365 * 24 * 60 // For Laravel < 5.8 the exact value must be specified and it is counted in minutes
```

spawnia:
I need to think about the deprecation of `executeQuery()` and the naming of the new methods.

Perhaps we should extract a method `parse(string $query): DocumentNode`?

docs/master/performance/query-caching.md
k0ka: I can add another command like `lighthouse:query:cache:clear` but it only clears the cache store. So I guess it's easier to just use the laravel default command.

src/lighthouse.php
k0ka: I just copied it from schema caching. It works on any env and all tests are passed with enabled query cache.

k0ka:
> Perhaps we should extract a method `parse(string $query): DocumentNode`?

This method might also throw a `SyntaxError` which must be caught and wrapped in the `ExecutionResult`. So it's better to leave `parseAndExecuteQuery()` and it is not too big and complex to split it.

src/lighthouse.php
spawnia:
```suggestion
        'enable' => env('LIGHTHOUSE_QUERY_CACHE_ENABLE', true),
```

docs/master/performance/query-caching.md
spawnia: That Laravel command clears all caches?

spawnia:
I could imagine the memoized `parse()` method being useful of its own.

docs/master/performance/query-caching.md
k0ka: it fully clears the default store. If the user specified non-default store for query cache they should add an argument to the command
`php artisan cache:clear --store=query_cache_store`

k0ka:
I extracted the `GraphQL::parse()` method

src/GraphQL.php
```diff
@@ -185,15 +188,77 @@ public function executeQuery(
...
+            $parsedQuery = $this->parse($query);
+        } catch (SyntaxError $syntaxError) {
+            return new ExecutionResult(null, [$syntaxError]);
         }
 
+        return $this->executeParsedQuery($parsedQuery, $context, $variables, $rootValue, $operationName);
+    }
+
+    /**
+     * Parses string query to DocumentNode. Caches queries if required.
```
spawnia, line 221:
```suggestion
     * Parse the given query string into a DocumentNode.
     *
     * Caches the parsed result if the query cache is enabled in the configuration.
```

src/GraphQL.php
```diff
@@ -185,15 +188,77 @@ public function executeQuery(
...
+        if (! $cacheConfig['enable']) {
+            return Parser::parse($query);
+        }
+
+        /** @var \Illuminate\Contracts\Cache\Factory $cacheFactory */
+        $cacheFactory = app(CacheFactory::class);
+        $store = $cacheFactory->store($cacheConfig['store']);
+
+        return $store->remember(
+            'lighthouse:query:' . hash('sha256', $query),
```
spawnia, line 238: We don't use global helpers such as `hash()`, they are usually not compatible with Lumen.

src/GraphQL.php
```diff
@@ -185,15 +188,77 @@ public function executeQuery(
...
+            $cacheConfig['ttl'],
+            function () use ($query) {
```
spawnia, line 240:
```suggestion
            static function () use ($query): DocumentNode {
```

src/lighthouse.php
```diff
@@ -126,6 +126,37 @@
         'tags' => false,
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Query Cache
+    |--------------------------------------------------------------------------
+    |
+    | Caches user queries so we won't have to parse string queries on each request
```
spawnia, line 134:
```suggestion
    | Caches the result of parsing incoming query strings to boost performance on subsequent requests.
```

spawnia (commented):
Like the extracted method, just a few more nits regarding the docs.

Considering [Automated persisted queries](https://github.com/nuwave/lighthouse/issues/651), would you foresee further changes to the proposed API?

src/GraphQL.php
k0ka: `hash()` is a global php function - https://www.php.net/manual/en/function.hash.php

k0ka:
> Considering [Automated persisted queries](https://github.com/nuwave/lighthouse/issues/651), would you foresee further changes to the proposed API?

I don't thinks this API will have any changes. We will have to add another method like `getParsedQueryByHash(string $sha256hash): DocumentNode`. It should get the corresponding cache value and return apollo compatible error if this value is not present in the cache.

spawnia:
The API of the `GraphQL` class is one thing that might have to change, but I am also concerned about the configuration format. We have some questionable structure in the config file (`schema.register`...) already. I would like to implement both features at once to make sure they fit together nicely, The config for APQ would probably overlap with `query_cache` quite a bit.

I can look into this myself, but am also quite happy if you push it further.

src/GraphQL.php
spawnia: You are right, thanks!

k0ka:
I added the APQ support. `webonyx/graphql-php` already supports it so it was quite easy. 

As for config schema: I guess the better cache format will be like this:
```php
'cache' => [
    'schema' => [
          'version' => env('LIGHTHOUSE_CACHE_VERSION', 1),

          // other parameters that are currently in the cache.*
    ],

    'query' => [
          // parameters from query_cache
    ],

    'validate' => [
         // future parameters for https://github.com/nuwave/lighthouse/issues/2018
    ],
]
```

But it will break backward compatibility. So I guess it's easier to leave as is and refactor it in the next major version.

spawnia (approved):
Awesome job!

spawnia:
@k0ka invited you as a collaborator. If you would like to [join our Slack](https://join.slack.com/t/lighthouse-php/shared_invite/zt-4sm280w1-wu21r94f3kLRRtBXRbXVfw), I can add you to our collaborator chat as well.

PR #2020, opened 2022-01-05
Fix ASTCache with `LIGHTHOUSE_CACHE_VERSION=1` in env

pyrou:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR cast explicitly the cache version as an integer in `ASTCache`, resolving a "Path cannot be empty" error in `ASTCache::put` method when using `LIGHTHOUSE_CACHE_VERSION=1` in `.env` file.

Why this issue wasn't seen before ?

1. Most of user using cache version 1 just use the provided configuration file without any particular `env` override. In this situation, `config('lighthouse.cache.version') === 1`, because of the default value used in config file.
2. In the meantime, users using `LIGHTHOUSE_CACHE_VERSION=2` ends up with a `config('lighthouse.cache.version') === "2"`, which pass fine the switch/case statement in `ASTCache::__construct`, and also isn't an issue, for `if (1 === $this->version)` comparison, because so far 1 != "2".

So when is it an issue ?

This issue is easily reproductible if you set `LIGHTHOUSE_CACHE_VERSION=1` in your `.env` file. In this situation `config('lighthouse.cache.version') === "1"`, (as a string not an int) which still pass fine the switch/case statement, but not any other ASTCache method relying on a `===` comparison. 

Also, In my case, I used `LIGHTHOUSE_CACHE_VERSION=2` in `.env`, but for my Unit test, and I chosen to override this environnement variable with a `<server name="LIGHTHOUSE_CACHE_VERSION" value="1" />` in my `phpunit.xml`. This time also, I ended up with a `config('lighthouse.cache.version') === "1"`

**Breaking changes**

This is bugfix and code wasn't working in the described condition.

### DIFF
--- a/src/Schema/AST/ASTCache.php
+++ b/src/Schema/AST/ASTCache.php
@@ -58,7 +58,7 @@ class ASTCache
 
         $this->enable = $cacheConfig['enable'];
 
-        $version = $cacheConfig['version'] ?? 1;
+        $version = (int) ($cacheConfig['version'] ?? 1);
 
         switch ($version) {
             case 1:
--- a/src/Schema/AST/ASTCache.php
+++ b/src/Schema/AST/ASTCache.php
@@ -58,7 +58,7 @@ class ASTCache
 
         $this->enable = $cacheConfig['enable'];
 
-        $version = (int) ($cacheConfig['version'] ?? 1);
+        $version = $cacheConfig['version'] ?? 1;
 
         switch ($version) {
             case 1:
@@ -73,7 +73,7 @@ class ASTCache
                 throw new UnknownCacheVersionException($version);
         }
 
-        $this->version = $version;
+        $this->version = (int) $version;
     }
 
     public function isEnabled(): bool


spawnia:
I think the code should work as is, given that `switch` uses loose comparison: https://3v4l.org/7J4rk

So I think this change actually does not fix anything, but is actively harmful because it casts nonsensical values to `0` and gives a worse error message.

pyrou:
Of course, `switch` uses loose comparison, this is exactly why there is currently no issue when using `LIGHTHOUSE_CACHE_VERSION=2` in .env.

"2" goes well in `case 2` (as expected) 
and "2" don't pass the `if ($this->version === 1)` (as expected) 


Also without any LIGHTHOUSE_CACHE_VERSION at all in .env (or in phpunit.xml)

1 goes well in `case 1` (as expected) 
and 1 pass the `if ($this->version === 1)` (as expected) 


Problem only exists when you have `LIGHTHOUSE_CACHE_VERSION=1` in .env (or in phpunit.xml)
This time, (so in lighthouse 5.33) you have a string instead of an int :

"1" still goes well in `case 1` (as expected) 
BUT don't pass anymore the strict comparison `if ($this->version === 1)`

Hope it's a bit more clear now ?

pyrou:
Speaking about potential breaking change that this PR could involve, I don't see any possible value for config('cache.version') that could prior fall in `case 1` or `case 2` and that would now fall in default producing a UnknownCacheVersionException due to the int cast.

spawnia:
I don't foresee breakage, but a worse error message. Let's just do the loose comparison and then memorize the proper int.

spawnia:
Thanks, released with https://github.com/nuwave/lighthouse/releases/tag/v5.33.1

PR #2023, opened 2022-01-07
Update Github action to run on latest linux and lowest composer dep

olivernybroe:
- [x] Added or updated tests
- [ ] ~Documented user facing changes~
- [x] Updated CHANGELOG.md

**Changes**

- Running on latest linux
- Renaming of the actions for better visibility
- Removing unused action steps
- Added services for mysql and redis
- Added runs for lowest and highest composer dependencies

**Dependency changes**
- Added min laravel 5.6 version
- Added min laravel 5.8 version


olivernybroe:
Hey @spawnia, would like your feedback on the changes to github actions here. 

With the changes to run on both lowest and highest possible dependencies, it actually shows us some issues in our dependencies.

spawnia:
Good find! I am fine with requiring certain minor/bugfix versions of dependencies in a minor release of Lighthouse. If you can make the minimal required bumps that would be great.

olivernybroe:
@spawnia we are having an issue with nesbot and laravel 5.6. 

L5.6 is limited to nesbot v1.26, however `createFromIsoFormat` was first added in v1.37 and we use that method in `DateTimeUtc`. 

So either we need to move `DateTimeUtc` into a separate package, or just drop L5.6 🤔

spawnia:
We can just skip the test for `DateTimeUtc` in Laravel 5.6, we do something similar in a bunch of places.

tests/database/factories/ContractorFactory.php
```diff
@@ -7,6 +7,6 @@
 
 $factory->define(Contractor::class, function (Faker $faker): array {
     return [
-        'position' => $faker->jobTitle,
+        'position' => $faker->word,
```
olivernybroe, line 10: Changed because of method not being available in old laravel versions

spawnia: Fine.

spawnia:
Thanks @olivernybroe for your work on this. I included parts of this in https://github.com/nuwave/lighthouse/pull/1898, but do not plan to make potentially breaking dependency changes in this version.

olivernybroe:
awesome, glad you could use some of it.

Issue #2027, opened 2022-01-13
From v.5.40: "Call to undefined function Safe\substr()" when run `php artisan lighthouse:ide-helper`


jBernavaPrah:
Hi,

Starting from v5.40 I receive the error when I try to run `php artisan lighthouse:ide-helper`. 
This is because the function `Safe/substr`  (on file `src/Schema/Types/LaravelEnumType.php`) is not anymore present on the package `thecodingmachine/safe@v2`. 

Using specifically v1 on my composer is fixing the error. 

Happy to help further if needed.
JBP

spawnia:
In order for us to constinually ensure we are compatible with v2, we need https://github.com/thecodingmachine/phpstan-safe-rule/pull/28 to be merged.

I installed the fork locally and fixed the errors - including the one you reported - and will release a fixed Lighthouse version soon. Please leave this issue open until for now though, I plan to update the fix with a continuous solution.

spawnia:
Fixed with https://github.com/nuwave/lighthouse/releases/tag/v5.35.2

olivernybroe:
It looks like this issue has been fixed, if my assumption is wrong I'll reopen 👍

PR #2028, opened 2022-01-13
Ensure compatibility with thecodingmachine/safe:2

spawnia:
https://github.com/nuwave/lighthouse/issues/2027

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "illuminate/support": "5.6.* || 5.7.* || 5.8.* || ^6 || ^7 || ^8",
         "illuminate/validation": "5.6.* || 5.7.* || 5.8.* || ^6 || ^7 || ^8",
         "laragraph/utils": "^1",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -39,7 +39,7 @@
         "illuminate/support": "5.6.* || 5.7.* || 5.8.* || ^6 || ^7 || ^8",
         "illuminate/validation": "5.6.* || 5.7.* || 5.8.* || ^6 || ^7 || ^8",
         "laragraph/utils": "^1",
-        "thecodingmachine/safe": "^1 || ^2",
```
brunogaspar, line 42: Hello @spawnia 

Was it intentional to remove the version 2 constraint?

spawnia: Oops, that was wrong. Good catch.

brunogaspar: Fantastic!! Thank you very much for the speedy release!

PR #2047, opened 2022-02-02
Improve query caching docs

thekonz:
**Changes**

Just some documentation changes that came up over the last weeks using query caching.


spawnia:
Thank you, nicely done

PR #2051, opened 2022-02-04
Add field `hasMorePages` for pagination type `SIMPLE`

thekonz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

As you can see in here, the `simplePaginate` method actually supports `hasMorePages` by just querying an additional line. So I added it (back) in :)

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Builder.php#L833-L836
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Pagination/Paginator.php#L70-L72

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## Added
+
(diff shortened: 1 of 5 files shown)


spawnia:
Thanks, nicely done

PR #2070, opened 2022-02-23
Normalize rules input to ensure Laravel 9 support

andershagbard:
I recently updated to Laravel 9.0 in my application, and these kind of inputs started failing.

`validatesAttributes@validateSize` does not receive correct `$parameters`. Its just receives an empty array.


spawnia:
Without looking further into this, a fix for this will probably need to happen in https://github.com/nuwave/lighthouse/blob/master/src/Validation/RulesGatherer.php

andershagbard:
I will probably not look into this myself in the near future. If anyone if up for it, go for it.

spawnia:
Thanks, released with https://github.com/nuwave/lighthouse/releases/tag/v5.42.1

PR #2084, opened 2022-03-11
Handle invalid subscriber cache entries from redis

majchrosoft:
Resolves https://github.com/nuwave/lighthouse/issues/2085

### DIFF
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -74,14 +74,17 @@ class RedisStorageManager implements StoresSubscriptions
 
         return (new Collection($subscribers))
             ->map(function (?string $subscriber): ?Subscriber {
-                if (!is_string($subscriber)) {
+                // Some entries may be expired
+                if ($subscriber === null) {
                     return null;
                 }
-                $deserializedSubscriber = unserialize($subscriber);
-                if($deserializedSubscriber === false){
+
+                // Other entries may contain invalid values
+                try {
+                    return unserialize($subscriber);
+                } catch (\ErrorException $e) {
                     return null;
                 }
-                return $deserializedSubscriber;
             })
             ->filter();
     }
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -136,7 +136,7 @@ class RedisStorageManager implements StoresSubscriptions
         $subscriber = $this->connection->command('get', [$channelKey]);
 
         return is_string($subscriber)
-            ? unserialize($subscriber)
+            ? @unserialize($subscriber)
             : null;
     }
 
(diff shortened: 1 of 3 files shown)


spawnia:
Please add a failing test case first.

majchrosoft:
@spawnia i've extended test case when unserializa returns false

majchrosoft:
@spawnia sorry for all this commits, but my child was interrupting me and i've done some mistakes ...

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -74,9 +74,11 @@ public function subscribersByTopic(string $topic): Collection
 
         return (new Collection($subscribers))
             ->map(function (?string $subscriber): ?Subscriber {
-                return is_string($subscriber)
-                    ? unserialize($subscriber)
-                    : null;
+                return
+                    is_string($subscriber)
+                    && @unserialize($subscriber) !== false
```
spawnia, line 79: We do not use error suppression through `@`. Also, calling `unserialize` twice is wildly inefficient.

spawnia (changes requested):

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -74,9 +74,13 @@ public function subscribersByTopic(string $topic): Collection
 
         return (new Collection($subscribers))
             ->map(function (?string $subscriber): ?Subscriber {
-                return is_string($subscriber)
-                    ? unserialize($subscriber)
-                    : null;
+                $deserializedSubscriber = unserialize($subscriber);
```
spawnia, line 77: Now this is potentially calling `unserialize` with `null`.

spawnia (changes requested):

src/Subscriptions/Storage/RedisStorageManager.php
majchrosoft: sure i known it,
just wanted to fast 

right now it's better

spawnia:
`unserialize()` fails with an exception - see https://github.com/nuwave/lighthouse/runs/5534931661?check_suite_focus=true#step:11:41

majchrosoft:
> `unserialize()` fails with an exception - see https://github.com/nuwave/lighthouse/runs/5534931661?check_suite_focus=true#step:11:41

@spawnia I have had exactly the same issue when i was testing locally.
I've already have tried with
```

                $deserializedSubscriber = (function () use ($subscriber) {
                    try {
                        return unserialize($subscriber);
                    } catch (\Throwable $unserializeFailed) {
                        return false;
                    }
                })();
```
but still without success.
The only solution was to suppress with `@`

majchrosoft:
@spawnia 
here i found conversation  about that topic
https://stackoverflow.com/questions/12684871/how-to-catch-unserialize-exception

spawnia:
No idea what setup you were using to run the tests, but it worked fine with the docker setup.

majchrosoft:
> No idea what setup you were using to run the tests, but it worked fine with the docker setup.

You are right. 

I've made mistake and edited another copy of class (the one in my project - which overrides in container the default one).
I 've run test once again without knowing that i did not changed nothing.

Thanks for fast responses

PR #2092, opened 2022-03-22
Add @drop directive for arguments and input fields, removing the argu…

eNzyOfficial:
…ment from the argument set

- [*] Added or updated tests
- [*] Documented user facing changes
- [ ] Updated CHANGELOG.md

**Changes**

Add the ability to use the `@drop` directive. This will remove the argument/input_field from the argument set.

**Breaking changes**

No breaking changes.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -928,6 +928,31 @@ type Query {
 Deprecated elements are not included in introspection queries by default,
 but they can still be queried by clients.
 
+## @drop
+
+```graphql
+"""
+Drop and remove an argument or input field from being passed to the resolver.
+"""
+directive @drop on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+```
+
+This can often be useful when wanting to confirm an input but don't want
+to use it as a column when creating a row in the database.
+
+```graphql
+type User {
+  email: String!
+}
+
+type Mutation {
+    createUser(
+        email: String!
+        email_confirmation: String! @drop
+    ): User @create
+}
+```
+
 ## @field
 
 ```graphql
--- a/src/Schema/Directives/DropArgsDirective.php
+++ b/src/Schema/Directives/DropArgsDirective.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Nuwave\Lighthouse\Schema\Directives;
+
+use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
+use Nuwave\Lighthouse\Execution\Arguments\ArgumentSet;
+use Nuwave\Lighthouse\Schema\Values\FieldValue;
+use Nuwave\Lighthouse\Support\Contracts\Directive;
+use Nuwave\Lighthouse\Support\Contracts\FieldMiddleware;
+use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;
+use Nuwave\Lighthouse\Support\Utils;
+
+class DropArgsDirective extends BaseDirective implements FieldMiddleware
+{
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Apply the @drop directives on the incoming arguments.
+"""
+directive @dropArgs on FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    /**
+     * @param FieldValue $fieldValue
+     * @param Closure $next
+     * @return mixed
+     */
+    public function handleField(FieldValue $fieldValue, Closure $next)
+    {
+        $resolver = $fieldValue->getResolver();
+
+        return $next(
+            $fieldValue->setResolver(
+                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver) {
+                    $argumentSet = $resolveInfo->argumentSet;
+                    $this->drop($argumentSet);
+
+                    return $resolver(
+                        $root,
+                        $argumentSet->toArray(),
+                        $context,
+                        $resolveInfo
+                    );
+                }
+            )
+        );
+    }
+
+    /**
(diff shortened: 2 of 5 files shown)


eNzyOfficial:
After speaking with @spawnia, this is my PR for the `@drop` directive. If something needs adding/changing, please let me know and I'll make said ammendments.

src/Schema/Directives/DropDirective.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Nuwave\Lighthouse\Schema\Directives;
+
+use Nuwave\Lighthouse\Schema\Values\FieldValue;
+use Nuwave\Lighthouse\Support\Contracts\FieldResolver;
+
+class DropDirective extends BaseDirective implements FieldResolver
```
spawnia, line 8:
```suggestion
class DropDirective extends BaseDirective
```

src/Schema/Directives/DropDirective.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Change the internally used name of a field or argument.
```
spawnia, line 14: Out of date

src/Schema/Directives/DropDirective.php
```diff
@@ -0,0 +1,34 @@
...
+This does not change the schema from a client perspective.
+"""
+directive @drop on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    /**
+     * @param FieldValue $fieldValue
+     * @return FieldValue
+     */
```
spawnia, line 25: Superfluous

src/Schema/Directives/DropArgsDirective.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * @param ArgumentSet $old
+     * @return void
+     */
+    protected function drop(ArgumentSet &$old): void
+    {
+        foreach ($old->arguments as $name => $argument) {
+            // Recursively apply the renaming to nested inputs.
```
spawnia, line 59: Leftover from copy-paste

tests/Unit/Schema/Directives/DropDirectiveTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace Tests\Unit\Schema\Directives;
+
+use Tests\TestCase;
+
+class DropDirectiveTest extends TestCase
```
spawnia, line 7:
```suggestion
final class DropDirectiveTest extends TestCase
```

spawnia (changes requested):
Good work overall, nice tests and docs. Needs some cleanup, but I am happy to merge this once done.

src/Schema/Directives/DropArgsDirective.php
eNzyOfficial: Not sure I understand this one, would you mind ellaborating?

Is it the `Utils::applyEach(` section that needs removing as we no longer need to recursively loop arguments, because dropping the root argument should be sufficient?

spawnia: There is no renaming going on, the comment is clearly just copy-pasted - and no longer accurate here.

Please go through the PR and make sure every line of code actually makes sense.

eNzyOfficial: Right, understood.

Would you mind checking my latest commit, I've rearranged the code so the recursive check is only called if the "root" argument isnt an instance of DropDirective.

src/Schema/Directives/DropDirective.php
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public function resolveField(FieldValue $fieldValue): FieldValue
+    {
+        return $fieldValue->setResolver(
+            function ($rootValue) use ($fieldValue) {
+                return data_get($rootValue, $fieldValue->getFieldName());
+            }
+        );
+    }
```
spawnia, line 26: This has no use

docs/master/api-reference/directives.md
```diff
@@ -928,6 +928,31 @@ type Query {
 Deprecated elements are not included in introspection queries by default,
 but they can still be queried by clients.
 
+## @drop
+
+```graphql
+"""
+Drop and remove an argument or input field from being passed to the resolver.
```
spawnia, line 935:
```suggestion
Ignore the user given value, don't pass it to the resolver.
```

src/Schema/Directives/DropArgsDirective.php
```diff
@@ -0,0 +1,80 @@
...
+                    );
+                }
+            )
+        );
+    }
+
+    /**
+     * @param ArgumentSet $old
+     * @return void
+     */
```
spawnia, line 55: Redundant

src/Schema/Directives/DropArgsDirective.php
```diff
@@ -0,0 +1,80 @@
...
+    protected function drop(ArgumentSet &$old): void
```
spawnia, line 56:
```suggestion
    protected function drop(ArgumentSet &$argumentSet): void
```

`$old` would make sense if we were constructing `$new`, but we mutate instead. You can change this accordingly in the rename directive

spawnia (changes requested):

src/Schema/Directives/DropArgsDirective.php
spawnia: Good catch, have not even thought of that extra efficiency gain.

eNzyOfficial:
I've made the ammendents suggested.

spawnia (approved):
Thanks!

Issue #2078, opened 2022-03-04
@can directive can't detect model from Type when relation @hasmany on field is paginator. 


webard:
**Describe the bug**

@can directive can't detect model from Type when relation @hasMany on field  is paginator. I have schema:

```gql
type Product @model(class: "App\\Product\\Product") {
    id: ID
    name: String
    reviews: [ProductReview!] @hasMany(type: PAGINATOR) @can(ability: "list reviews")
}

type ProductReview @model(class: "App\\Product\\Reviews") {
    id: ID!
    author: ID!
}

query {
    products: [Product] @paginate
}
```

As you see, I define @model in type, so @can directive and other directives should read proper model from Type. Everything works fine with @can on fields, until field is not paginated relation. In example above, error is:

```
Failed to find class ReviewsPaginator in namespaces [App\\Models] for directive @can.
```

but when we delete `type: PAGINATOR` from @hasMany, everything works fine and @can directive can detect proper model.
I know I can declare `model` argument in @can directive, but using @model on type simplifies life, so detecting paginated relation should works.

**Output/Logs**

<details><summary>Click to expand</summary>

```
Failed to find class ReviewsPaginator in namespaces [App\\Models] for directive @can.
```

</details></br>

**Lighthouse Version**

5.42.0, Laravel 9.3.0

spawnia:
Can you add a failing test case to https://github.com/nuwave/lighthouse/blob/master/tests/Integration/Schema/Directives/HasManyDirectiveTest.php? I think we can then fix this by adding `@model` to the generated `type ReviewsPaginator`.

webard:
Ok it will be ready soon.

PR #2093, opened 2022-03-22
Annotate generated schema with correct model in paginated relations

webard:
- [x] Added or updated tests

Added tests to #2078.

**Changes**

**Breaking changes**

No breaking changes


spawnia:
Looks like the test successfully failed 😉

If you want to accelerate a fix: we need to adapt https://github.com/nuwave/lighthouse/blob/master/src/Pagination/PaginationManipulator.php by adding `@model` to the generated types.

src/Pagination/PaginationManipulator.php
```diff
@@ -236,16 +238,14 @@ protected static function countArgument(?int $defaultCount = null, ?int $maxCoun
...
         }
         $description .= "\"\n";
 
-        $definition = 'first: Int'
-            . (
-                $defaultCount
-                ? ' = ' . $defaultCount
-                : '!'
-            );
+        $definition = 'first: Int!';
```
olivernybroe, line 245: Interface field argument Account.files(first:) expects type Int but Company.files(first:) is type Int!.

PR #2104, opened 2022-03-28
Revert breaking schema change in generate pagination types

spawnia:
Reverts the breakage mentioned in https://github.com/nuwave/lighthouse/pull/2093/files/a5e76b202966e988b30760131a50a8fb2fe4989f#diff-76d791c7e3599495909ebc9bff7dc2d6cf89f30d8e33f9bf68d4eb91f935c9ca

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


olivernybroe (approved):
Nice 👍 

Would love if we could just have ignored it, but yeah, I guess we have to honor BC's.

V6 is going to be awesome!

PR #2114, opened 2022-04-09
throw http error 400 on wrong json in request

k0ka:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Sending malformed json request to the graphql endpoint results is throwing `JsonException` by `Laragraph`. It is left uncaught and is shown to user as http error 500.

This error is not right as it means something wrong with the server while something wrong happens with the client.

This PR catches that exception (and another one that might be thrown by `Laragraph`) and wraps it into `BadRequestHttpException` so the client will receive http error 400.

### DIFF
--- a/src/Support/Http/Controllers/GraphQLController.php
+++ b/src/Support/Http/Controllers/GraphQLController.php
@@ -2,6 +2,7 @@
 
 namespace Nuwave\Lighthouse\Support\Http\Controllers;
 
+use GraphQL\Server\RequestError;
 use Illuminate\Contracts\Events\Dispatcher as EventsDispatcher;
 use Illuminate\Http\Request;
 use Laragraph\Utils\RequestParser;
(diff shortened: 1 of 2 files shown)


src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -26,7 +29,12 @@ public function __invoke(
             new StartRequest($request)
         );
 
-        $operationOrOperations = $requestParser->parseRequest($request);
+        try {
+            $operationOrOperations = $requestParser->parseRequest($request);
+        } catch (JsonException|RequestError $e) {
+            throw new BadRequestHttpException('Error parsing GraphQL request', $e);
```
spawnia, line 35: This erases valuable information

k0ka: Why? I pass the original exception to `BadRequestHttpException` so it is logged too.

spawnia: The error message shown to the user will be just that of the parent.

k0ka: Uhm, currently it is showing just "error 500" for the production site and full exception details (with nested exceptions) for the dev. 

Anyways, should I add this PR to the `laragraph/utils`?

spawnia: Yeah, I would appreciate a PR there.

spawnia:
Superseded by https://github.com/laragraph/utils/pull/10

PR #2120, opened 2022-04-16
Fix assert for builder argument on @all & @paginate directives

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Recently an assert was added to `@all` & `@paginate` that was incomplete, it was missing the `Relation` type which should also be allowed.

Unrelated, this also fixes the use of `mkdir --parents` that doesn't work under macOS, changed to `mkdir -p` to work on both macOS and Linux OSs.

**Breaking changes**

None.

### DIFF
--- a/src/Pagination/PaginateDirective.php
+++ b/src/Pagination/PaginateDirective.php
@@ -7,7 +7,7 @@ use GraphQL\Language\AST\ObjectTypeDefinitionNode;
 use GraphQL\Type\Definition\ResolveInfo;
 use Illuminate\Contracts\Pagination\Paginator;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
-use Illuminate\Database\Eloquent\Relations\Relation as EloquentRelation;
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Builder as QueryBuilder;
 use Nuwave\Lighthouse\Schema\AST\DocumentAST;
 use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
@@ -112,7 +112,7 @@ GRAPHQL;
 
                 $query = $builderResolver($root, $args, $context, $resolveInfo);
                 assert(
-                    $query instanceof QueryBuilder || $query instanceof EloquentBuilder || $query instanceof EloquentRelation,
+                    $query instanceof QueryBuilder || $query instanceof EloquentBuilder || $query instanceof Relation,
                     "The method referenced by the builder argument of the @{$this->name()} directive on {$this->nodeName()} must return a Builder or Relation."
                 );
             } else {
--- a/src/Pagination/PaginateDirective.php
+++ b/src/Pagination/PaginateDirective.php
@@ -7,6 +7,7 @@ use GraphQL\Language\AST\ObjectTypeDefinitionNode;
 use GraphQL\Type\Definition\ResolveInfo;
 use Illuminate\Contracts\Pagination\Paginator;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Relations\Relation as EloquentRelation;
 use Illuminate\Database\Query\Builder as QueryBuilder;
 use Nuwave\Lighthouse\Schema\AST\DocumentAST;
 use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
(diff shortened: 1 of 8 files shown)


src/Pagination/PaginateDirective.php
```diff
@@ -7,6 +7,7 @@
 use GraphQL\Type\Definition\ResolveInfo;
 use Illuminate\Contracts\Pagination\Paginator;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Relations\Relation as EloquentRelation;
```
spawnia, line 10: We can leave this as `Relation` - in contrast to `Builder` there is no ambiguity.

tests/Integration/Pagination/PaginateDirectiveDBTest.php
```diff
@@ -88,6 +89,51 @@ public function testSpecifyCustomBuilder(): void
...
+            id: ID!
+            posts: [Post!]! @paginate(builder: "' . $this->qualifyTestResolver('builderForRelation') . '")
+        }
+
+        type Query {
+            user(id: ID!): User @find
+        }
+        ';
+
+        // The custom builder is supposed to change the sort order
```
spawnia, line 111: I know this was not the purpose of adding the test, but it would still be great if the result would prove that the builder works.

Perhaps just add 2 posts, query for 2 but let the builder exclude one of them by some condition.

tests/Unit/Schema/Source/SchemaStitcherTest.php
```diff
@@ -17,7 +17,7 @@ public function setUp(): void
         parent::setUp();
 
         // @phpstan-ignore-next-line using the Safe variant crashes PHPStan
-        exec('mkdir --parents ' . self::SCHEMA_PATH);
+        exec('mkdir -p ' . self::SCHEMA_PATH);
```
spawnia, line 20: Please add a comment if you would like this change to stick, otherwise I am keen to refactor short CLI arguments into their long form when I see them 😉

tests/Integration/Schema/Directives/AllDirectiveTest.php
```diff
@@ -200,8 +201,54 @@ public function testSpecifyCustomBuilder(): void
...
+            id: ID!
+            posts: [Post!]! @all(builder: "' . $this->qualifyTestResolver('builderForRelation') . '")
+        }
+
+        type Query {
+            user(id: ID!): User @find
+        }
+        ';
+
+        // The custom builder is supposed to change the sort order
```
spawnia, line 223: Same as the other test, it would be good if this test were to prove that the builder was actually used.

spawnia (changes requested):

tests/Integration/Pagination/PaginateDirectiveDBTest.php
stayallive: Ah chips, also missed this comment I copied and pasted, oeps.

I've fixed it to do what the comment said (reversing the order).

tests/Integration/Pagination/PaginateDirectiveDBTest.php
```diff
@@ -88,6 +89,57 @@ public function testSpecifyCustomBuilder(): void
         ]);
     }
 
+    public function testSpecifyCustomBuilderForRelation(): void
+    {
+        $user = factory(User::class)->create();
+        $post = factory(Post::class, 2)->create([
+            'user_id' => $user->id,
+        ]);
```
spawnia, line 97:
```suggestion
        $user = factory(User::class)->create();
        assert($user instanceof User);

        $posts = factory(Post::class, 2)->make();
        $user->posts()->saveMany($posts);
```

spawnia:
Any idea what is wrong with PHPStan? The error is also present in `master` and other branches, but seems nonsensical:

```
Internal error: Internal error: Call to undefined function Safe\array_combine() in file /workdir/src/Schema/Directives/ModifyModelExistenceDirective.php
     Run PHPStan with -v option and post the stack trace to:
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md
```

Neither this class nor any other code in Lighthouse uses any variant of `array_combine()`.

spawnia:
I can reproduce the error locally after upgrading my dependency. As long as it is still working for you, can you share the versions of PHPStan and Safe from your `composer.lock`?

stayallive:
```
PHP 8.1.4 (cli) (built: Apr 12 2022 18:36:19) (NTS)
Composer version 2.3.4 2022-04-07 21:16:35

thecodingmachine/phpstan-safe-rule v1.2.0    A PHPStan rule to detect safety issues. Must be used in conjunction with thecodingmachine/safe
thecodingmachine/safe              v2.1.3    PHP core functions that throw exceptions instead of returning FALSE on error
phpstan/phpstan                    1.5.6     PHPStan - PHP Static Analysis Tool
```

spawnia:
Thanks @stayallive, the issue was https://github.com/thecodingmachine/safe/issues/344 and should be fixed now.

stayallive:
Ah sweet, cool!

PR #2121, opened 2022-04-18
Add `@upload` directive

### LINKED ISSUE #2094: Directive to upload a file and return storage path

While it is possible to pass a file via input using `Upload` scalar, it would be nice if there was a directive that would upload the file to a storage disk, drop the file from the input and replace it with the `path`.

For example, if I have a `posts` table with a thumbnail url, I would pass the thumbnail image file in an input payload along with other details about my blog post (title, body, tags, etc), uploading the file to storage and storing the url as a column in the database.

Possibly something similar to `image: Upload @upload(attribute:"image_url", disk:"public", path:"uploads/")`.

eNzyOfficial:
…ument, uploaded to disk and replaced with the uploaded file path

- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #2094

**Changes**

Added `@upload` directive. This will take in an UploadedFile, store it to a disk and return the resulting filepath.

If `disk` is omitted, it defaults to config value `filesystems.default`
If `path` is ommited, it defaults to `/`


eNzyOfficial:
Please review the tests, I used existing tests as examples when creating my own, so I could have missed something.

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    public function diskArgValue(): string
+    {
+        return $this->directiveArgValue('disk', config('filesystems.default'));
+    }
+
+    public function pathArgValue(): string
+    {
+        return rtrim(
+            $this->directiveArgValue('path', ''),
```
spawnia, line 67:
```suggestion
            $this->directiveArgValue('path') ?? '/',
```

As documented

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,71 @@
...
+     * @throws Exception
+     */
+    public function diskArgValue(): string
+    {
+        return $this->directiveArgValue('disk', config('filesystems.default'));
+    }
+
+    public function pathArgValue(): string
+    {
+        return rtrim(
```
spawnia, line 66: Why?

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,71 @@
...
+"""
+directive @upload(
+  """
+  The storage disk to be used, defaults to config value `filesystems.default`.
+  """
+  disk: String
+  """
+  The path where the file should be stored, defaults to `/`.
+  """
+  path: String
```
spawnia, line 29:
```suggestion
  The path where the file should be stored.
  """
  path: String! = "/"
```

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,218 @@
...
+use Exception;
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+use Tests\TestCase;
+
+final class UploadDirectiveTest extends TestCase
+{
+    private const DISK_NAME = 'uploadDisk';
+
+    protected $operations = [
```
spawnia, line 14: Use a local variable instead, I would rather duplicate some test code but have it more readable. DRY is not a big concern in testing.

src/Schema/Directives/UploadDirective.php
eNzyOfficial: You're right, originally I was using `Storage::disk()->...` concating the path and filename together. I ended up using `storeAs` which takes in the path and filename separately. I'll remove the `rtrim`

eNzyOfficial:
Requested changes have been made.

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,71 @@
...
+        $filename = $argumentValue->hashName();
+
+        $filepathInStorage = $argumentValue->storeAs(
+            $this->pathArgValue(),
+            $filename,
+            $this->diskArgValue()
+        );
+
+        if ($filepathInStorage === false) {
+            throw new Exception("Unable to upload `{$this->name()}` file to `{$this->pathArgValue()}` via disk `{$this->diskArgValue()}`");
```
spawnia, line 56: Can we use a more appropriate exception class - maybe something from the framework? Alternatively, this would be a `RuntimeException`.

Also `$this->name()` is the name of the directive, which does not really help much here. Let's include the name of the argument (`$this->nodeName()`).

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public function diskArgValue(): string
+    {
+        return $this->directiveArgValue('disk') ?? config('filesystems.default');
+    }
+
+    public function pathArgValue(): string
+    {
+        return $this->directiveArgValue('path', '');
```
spawnia, line 69:
```suggestion
        return $this->directiveArgValue('path') ?? '/';
```

If this does not work, we need to change the description about the default value.

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,299 @@
...
+{
+    public function testUploadArgumentWithDefaultParameters(): void
+    {
+        config()->offsetSet('filesystems.default', 'uploadDisk');
+
+        $filePath = null;
+
+        $this->mockResolver(static function ($root, array $args) use (&$filePath): ?string {
+            $filePath = $args['file'];
+            return $args['file'];
```
spawnia, line 21:
```suggestion
            return $filePath = $args['file'];
```

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        $file = UploadedFile::fake()->create('test.pdf', 500);
+
+        $this->multipartGraphQL(
+            [
+                'query' => /** @lang GraphQL */ '
+                mutation ($file: Upload!) {
+                    file: upload(file: $file)
+                }
+            ',
```
spawnia, line 45:
```suggestion
                    mutation ($file: Upload!) {
                        file: upload(file: $file)
                    }
                ',
```

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        $this->mockResolver(static function ($root, array $args) use (&$filePath): ?string {
+            $filePath = $args['file'];
+            return $args['file'];
+        });
+
+        Storage::fake('uploadDisk');
+
+        $this->schema = /** @lang GraphQL */
+            '
```
spawnia, line 27:
```suggestion
        $this->schema = /** @lang GraphQL */ '
```

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,299 @@
...
+                mutation ($file: Upload!) {
+                    file: upload(file: $file)
+                }
+            ',
+                'variables' => [
+                    'file' => null,
+                ],
+            ],
+            ['0' => ['variables.file']],
+            ['0' => $file]
```
spawnia, line 270: Static analysis fails here - perhaps we should use `Illuminate\Http\Testing\File` for this test? Alternatively, we could loosen the type definition of `$this->multipartGraphQL()` to also accept `Illuminate\Http\UploadedFile` or a common interface.

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,299 @@
...
+                'file' => $filePath,
+            ],
+        ]);
+
+        Storage::disk('uploadDisk')->assertExists($file->hashName());
+    }
+
+    public function testThrowsAnExceptionIfAttributeIsNotUploadedFile(): void
+    {
+        $this->expectException(Exception::class);
```
spawnia, line 282: Let's use `$this->expectExceptionObject()` and actually test that the error message comes out right.

spawnia (changes requested):

eNzyOfficial:
Static analysis is now passing and requested changes have been made.

eNzyOfficial:
Static analysis is now passing, requested changes have been made.

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,73 @@
...
+        $filename = $argumentValue->hashName();
+
+        $filepathInStorage = $argumentValue->storeAs(
+            $this->pathArgValue(),
+            $filename,
+            $this->diskArgValue()
+        );
+
+        if ($filepathInStorage === false) {
+            throw new CannotWriteFileException("Unable to upload `{$this->name()}` file to `{$this->pathArgValue()}` via disk `{$this->diskArgValue()}`");
```
spawnia, line 58:
```suggestion
            throw new CannotWriteFileException("Unable to upload `{$this->nodeName()}` file to `{$this->pathArgValue()}` via disk `{$this->diskArgValue()}`");
```

src/Schema/Directives/UploadDirective.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function diskArgValue(): string
+    {
+        return $this->directiveArgValue('disk') ?? config('filesystems.default');
+    }
+
+    public function pathArgValue(): string
+    {
+        return $this->directiveArgValue('path', '/');
```
spawnia, line 71:
```suggestion
        return $this->directiveArgValue('path') ?? '/';
```

I suggested this change before, and I meant it.

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,300 @@
...
+                ],
+            ],
+            ['0' => ['variables.file']],
+            ['0' => $file]
+        );
+    }
+
+    public function testThrowsAnExceptionIfStoringFails(): void
+    {
+        $this->expectExceptionObject(new CannotWriteFileException("Unable to upload `upload` file to `/` via disk `uploadDisk`"));
```
spawnia, line 232: When exactly is this thrown? Please move this expectation as closely as possible to where the exception is triggered.

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,300 @@
...
+            ],
+            ['0' => ['variables.file']],
+            ['0' => $file]
+        )->assertJson([
+            'data' => [
+                'file' => $filePath,
+            ],
+        ]);
+
+        Storage::disk('uploadDisk')->assertExists($file->hashName());
```
spawnia, line 278: I assume this is dead code?

tests/Unit/Schema/Directives/UploadDirectiveTest.php
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    public function testThrowsAnExceptionIfAttributeIsNotUploadedFile(): void
+    {
+        $this->expectExceptionObject(new InvalidArgumentException("Expected UploadedFile from `baz`"));
```
spawnia, line 283: Same as the other test, this should be as close as possible to the source of error.

spawnia (changes requested):

eNzyOfficial:
Sorry it's taken so long to get around to this. I've made the changes requested above.

eNzyOfficial:
I've modified the directive to include the `public` argument, this allows the ability to set if the visibility should be public/private.

spawnia (approved):
Thanks, looks good now.

Issue #1459, opened 2020-06-26
Convert empty strings to null in multipart-form request 


lorado:
**Describe the bug**

Dunno if it is a bug, but this was for me very unexpected, and it may be so to other people...

The situation: I have an input with a field of type `Float`. In my frontend I send by mistake an empty string. While I was sending a JSON request, everything worked fine. But when I add a file to the request, request becomes a `multipart-form` request, and then I got an error, that empty string can't be a float number, what is sure true. But I was really confused, because without file it works without a problem.

After long debugging I found out, that when request is sent as json, input gets sanitized by middleware `\Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull`, that was there by default, after laravel installation. As the middleware name says, empty strings becomes null. So my empty string variable becomes value null (actually what I was expected from my frontend), and FloatType convertion happens without a problem (because the value is `null`)   

For a multipart-form request, middlewares are executed on whole GraphQL operation is encoded JSON, because decoding happens way later in `Nuwave\Lighthouse\Execution\MultipartFormRequest`. As result sent empty string stay an empty string, because after JSON decoding of `operations` values are not going through the request pipeline defined  in `App\Http\Kernel`. When FloatType then tries to parse float value, it throws an error, because an empty string is not a float number. 

For me that was very unexpected.

**Expected behavior/Solution**

So for me to solve the problem I actually have to disable that `\Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull` middleware, because I don't need it at all. 

Anyway, I thought there might be other people, that could face same issue. The question is - should we mention it an docs, e.g. in installation as a note? Like "be aware of..." hint/advice?

spawnia:
Just recently i also had an issue related to that middleware and disabled it.

I think usually it is included in Laravel's global middleware and used for the entire application. Can it be disabled only for Lighthouse?

lorado:
Well as option we could provide our own `ConvertEmptyStringsToNull` implementation, that extends laravels one, and in `handle()` method we could check the `$request->pathinfo` equals `/graphql`. If so - simply return `$next($request)` so we ignore modifying of request, otherwise return `parent::handle($request, $next)`. In the installation guide we could write a note, that if user still need this middleware, he should remove laravels one, and include ours, as ours ignores `graphql` route.

spawnia:
[...]

PR #2130, opened 2022-04-25
Omit non-nullable arguments in `@convertEmptyStringsToNull`

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

https://github.com/nuwave/lighthouse/issues/1459#issuecomment-1106598749

@LastDragon-ru

**Changes**

Omit non-nullable arguments in `@convertEmptyStringsToNull`.

**Breaking changes**

Behavioural fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v5.48.1
+
+### Fixed
+
+- Omit non-nullable arguments in `@convertEmptyStringsToNull` https://github.com/nuwave/lighthouse/pull/2130
+
 ## v5.48.0
 
 ### Changed
--- a/src/Execution/Arguments/Argument.php
+++ b/src/Execution/Arguments/Argument.php
@@ -49,6 +49,23 @@ class Argument
         return static::toPlainRecursive($this->value);
     }
 
+    public function namedType(): ?NamedType
+    {
+        return static::namedTypeRecursive($this->type);
+    }
+
+    /**
+     * @param \Nuwave\Lighthouse\Execution\Arguments\ListType|\Nuwave\Lighthouse\Execution\Arguments\NamedType|null $type
+     */
+    protected static function namedTypeRecursive($type): ?NamedType
+    {
+        if ($type instanceof ListType) {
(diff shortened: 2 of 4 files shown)


oscar-ol:
I'm using this directive in global field middleware, but with this fix, the value of a mutation input field is not converting to null anymore.
Setting the directive to the field directly, it works.

spawnia:
> I'm using this directive in global field middleware, but with this fix, the value of a mutation input field is not converting to null anymore. Setting the directive to the field directly, it works.

The input field is non-nullable, correct?

oscar-ol:
The input field is a String, it can be nullable. 

updateUser(
        id: ID!
        email: String
    ): User! 
    
updateUser(id: 1, email: "") is not converted to..  email: null

instead, it works:

updateUser(
        id: ID!
        email: String @convertEmptyStringsToNull
    ): User! 


it doesn't work either:

updateUser(
        id: ID!
        email: String 
    ): User! @convertEmptyStringsToNull

spawnia:
@oscar-ol this use case is covered in https://github.com/nuwave/lighthouse/blob/388fa7d9db0bf603cf066bae836713c9b363ed1b/tests/Integration/Schema/Directives/ConvertEmptyStringsToNullDirectiveTest.php#L55-L98. If you can add a PR with a failing test case I can look into it, otherwise I am going to assume the error is with your setup.

skaesdorf:
Same for me. Since this condition simply checks for `ScalarType::STRING` it will skip sanitization no matter if the namedType is nullable or not.

```
if (null !== $namedType && ScalarType::STRING === $namedType->name) {
   continue;
}
```

Cheers

skaesdorf:
> @oscar-ol this use case is covered in
> 
> https://github.com/nuwave/lighthouse/blob/388fa7d9db0bf603cf066bae836713c9b363ed1b/tests/Integration/Schema/Directives/ConvertEmptyStringsToNullDirectiveTest.php#L55-L98
> 
> . If you can add a PR with a failing test case I can look into it, otherwise I am going to assume the error is with your setup.

I found the issue with your test case. When calling `assertJson` you dont pass the `strict` parameter which is required since otherwise PHPUnit will compare the values using `==` instad of `===`. Hence `null` and `""` are equal in the test.

Cheers

Issue #2126, opened 2022-04-21
Set HTTP Cache-Control header for the response


mostafa-rz:
**What problem does this feature proposal attempt to solve?**

The Cache-Control header for any response from Lighthouse is no-cache, private. This header can be used by CDN to set correct caches values for the response. Most of the CDNs have the option to ignore this header, so it is possible to use it as it is, but then it needs to set the invalidation rules manually in the CDN. It is possible to calculate the invalidation rules from the schema and send it by the response header.
[Apollo](https://www.apollographql.com/docs/apollo-server/performance/caching#calculating-cache-behavior) has some rules to calculate the final cache header values. The rules are based on the `@cacheControl` directive. We have the same information in the `@cache` directive.

- The response's maxAge equals the lowest maxAge among all fields. If that value is 0, the entire result is not cached.
- The response's scope is PRIVATE if any field's scope is PRIVATE.

**Which possible solutions should be considered?**

1. Add env variables to enable this feature and some default values for max-age and max-stale.
2. Calculate the final cache header values for the request based on the rules.
4. Pass the final cache values to the Nuwave\Lighthouse\Support\Http\Controllers: GraphQLController
5. Add the headers to the response by createResponse($result)->setCache()

spawnia:
Seems like we can pretty much copy what Apollo does here, their approach seems to be battle-tested and proven.

@stayallive what do you think?

spawnia:
Is `@cacheControl` a prerequisite for this feature request, or can they be implemented independently?

stayallive:
IMO they can be totally separated, maybe even better because the plumbing to get the cache control header back to the controller can and should be re-used. So whichever is implemented first makes the second trivial.

Maybe it's even a good idea to start with `@cacheControl` since it's the cleanest (no config, no magic) and implement `@cache` auto-`@cacheControl` afterwards?

mostafa-rouzbahani:
@stayallive Thanks for sharing your idea.
@spawnia I would like to contribute to implementing the issue. Is anyone working on this?

spawnia:
@mostafa-rouzbahani Not that I know of. Feel free to go ahead with it. Starting with just `@cacheControl` seems good.

mostafa-rouzbahani:
@stayallive To make it clear what I am going to do:

1. Add the `@cacheControl` directive.
2. The `@cacheControl` has nothing to do with server-side caching(at least for now).
[...]

PR #2132, opened 2022-04-28
Add `@cacheControl` directive

mostafa-rouzbahani:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

[#2126](https://github.com/nuwave/lighthouse/issues/2126)

**Changes**

Add `@cacheControl` directive

**Breaking changes**

No breaking changes.


src/CacheControl/CacheControl.php
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            $headerOptions['max_age'] = $maxAge;
+        }
+
+        // Set scope value.
+        if ('public' === $this->calculateScope()) {
+            $headerOptions['public'] = true;
+        } else {
+            $headerOptions['private'] = true;
+        }
```
spawnia, line 56:
```suggestion
        $headerOptions[$this->calculateScope()] = true;
```

src/CacheControl/CacheControlDirective.php
```diff
@@ -0,0 +1,69 @@
...
+    public static function definition(): string
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Set the HTTP Cache-Control headers of the response.
+"""
+directive @cacheControl(
+  """
+  The maximum amount of time the field's cached value is valid, in seconds. 
+  The default value is 0(no-cache), but you can set a different default from config.
```
spawnia, line 31: > but you can set a different default from config

Which config?

src/CacheControl/CacheControlDirective.php
```diff
@@ -0,0 +1,69 @@
...
+    """
+    PRIVATE
+}
+GRAPHQL;
+    }
+
+    public function handleField(FieldValue $fieldValue, Closure $next)
+    {
+        $maxAge = $this->directiveArgValue('maxAge') ?? 0;
+        $scope = $this->directiveArgValue('scope') ?? 'public';
```
spawnia, line 62: The directive docs says the default is `PUBLIC` - which one is it?

src/CacheControl/CacheControlServiceProvider.php
```diff
@@ -0,0 +1,49 @@
...
+class CacheControlServiceProvider extends ServiceProvider
+{
+    public function register()
+    {
+        $this->app->singleton(CacheControl::class);
+    }
+
+    /**
+     * Bootstrap any application services.
+     */
```
spawnia, line 20:
```suggestion
```

src/CacheControl/CacheControlServiceProvider.php
```diff
@@ -0,0 +1,49 @@
...
+                            "The HTTP Cache-Control header set to public."
+                            PUBLIC @enum(value: "public")
+
+                            "The HTTP Cache-Control header set to private."
+                            PRIVATE @enum(value: "private")
+                        }
+                    ')
+                );
+            }
+        );
```
spawnia, line 47: Not necessary, since this directive is only used in the schema

src/Support/Http/Controllers/GraphQLController.php
```diff
@@ -32,11 +34,14 @@ public function __invoke(
         $result = $graphQL->executeOperationOrOperations($operationOrOperations, $context);
 
         $response = $createsResponse->createResponse($result);
+        $response = $response->setCache($cacheControl->makeHeaderOptions());
 
         $eventsDispatcher->dispatch(
             new EndRequest($response)
         );
 
+        app()->forgetInstance(CacheControl::class);
```
spawnia, line 43: Can we do this in a listener of `EndRequest`? That would mean that users who exclude `CacheControlServiceProvider` don't pay for it.

tests/Integration/CacheControl/CacheControlDirectiveTest.php
```diff
@@ -0,0 +1,178 @@
...
+        {
+            user {
+                name
+            }
+        }
+        ')->assertHeader('Cache-Control', $expectedHeaderString);
+    }
+
+    /**
+     * @return string[][]
```
spawnia, line 72:
```suggestion
     * @return array<string, array{string, string}>
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add `@cacheControl` directive
```
```suggestion
- Add `@cacheControl` directive https://github.com/nuwave/lighthouse/pull/2132
```

tests/TestCase.php
```diff
@@ -79,6 +80,7 @@ protected function getPackageProviders($app): array
             PaginationServiceProvider::class,
             SoftDeletesServiceProvider::class,
             ValidationServiceProvider::class,
+            CacheControlServiceProvider::class,
```
spawnia, line 83: alphabetical

src/CacheControl/CacheControl.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace Nuwave\Lighthouse\CacheControl;
+
+final class CacheControl
```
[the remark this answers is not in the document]
mostafa-rouzbahani: when i run `make it`, all protected changes to private.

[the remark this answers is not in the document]
spawnia: Probably due to `final`.

src/CacheControl/CacheControlDirective.php
mostafa-rouzbahani: I don't get this. In doc, it is PUBLIC (enum) and here it is public (enum value).

spawnia: There is no automagical conversion that lowercases enum literals from GraphQL. Unless specified otherwise or converted explicitly, the internal value of parsing an enum value is its string representation.

src/CacheControl/CacheControlDirective.php
```diff
@@ -0,0 +1,69 @@
...
+directive @cacheControl(
+  """
+  The maximum amount of time the field's cached value is valid, in seconds. 
+  The default value is 0(no-cache), but you can set a different default from config.
+  """
+  maxAge: Int = 0
+
+  """
+  If PRIVATE, the field's value is specific to a single user. 
+  The default value is PUBLIC.
```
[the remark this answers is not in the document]
spawnia: This is redundant, the definition already defines the default value.

src/CacheControl/CacheControl.php
```diff
@@ -0,0 +1,61 @@
...
+    protected $scopeList = [];
+
+    public function addToMaxAgeList(int $maxAge): void
+    {
+        $this->maxAgeList[] = $maxAge;
+    }
+
+    public function addToScopeList(string $scope): void
+    {
+        $this->scopeList[] = strtolower($scope);
```
spawnia, line 28:
```suggestion
        $this->scopeList[] = $scope;
```

Let's just always use `PUBLIC` and `PRIVATE`.

src/CacheControl/CacheControlServiceProvider.php
```diff
@@ -0,0 +1,34 @@
...
+            static function (): string {
+                return __NAMESPACE__;
+            }
+        );
+
+        $dispatcher->listen(
+            EndRequest::class,
+            function (EndRequest $request) use ($cacheControl): void {
+                $request->response->setCache($cacheControl->makeHeaderOptions());
+                app()->forgetInstance(CacheControl::class);
```
spawnia, line 30: Will a new singleton be instantiated for new requests? This might happen when using `Octane`.

spawnia, line 30:
```suggestion
                $this->app->forgetInstance(CacheControl::class);
```

tests/Integration/CacheControl/CacheControlDirectiveTest.php
```diff
@@ -0,0 +1,178 @@
...
+        assert($team instanceof Team);
+
+        $users = factory(User::class, 3)->make();
+        $team->users()->saveMany($users);
+
+        $this->graphQL($query)->assertHeader('Cache-Control', $expectedHeaderString);
+    }
+
+    /**
+     * @return string[][]
```
spawnia, line 126:
```suggestion
     * @return array<string, array{string, string}>
```

spawnia (changes requested):

spawnia:
Docs need:
- a section in https://github.com/nuwave/lighthouse/blob/master/docs/master/performance/caching.md that describes basic usage, probably referring to Apollo docs for a deeper explanation
- directive definition in https://github.com/nuwave/lighthouse/blob/master/docs/master/api-reference/directives.md

Thanks for implementing this, you are doing a great job. The tests look well structured and the implementation is done very well.

src/CacheControl/CacheControlServiceProvider.php
mostafa-rouzbahani: With data providers in tests, I needed to remove the instance otherwise tests failed. After the previous review changes, tests passed even without forgetInstance. I don't know what is happened, so pls decide to do forgetInstance or not.

spawnia: My concern was that after this call, the class may no longer be instantiated as a singleton. I tested the behaviour of the container though, and it does behave the way we need it to:

```
>>> class Foo {}
>>> app()->singleton(Foo::class)
=> null
>>> app(Foo::class)
=> Foo {#4859}
>>> app(Foo::class)
=> Foo {#4859}
>>> app()->forgetInstance(Foo::class)
=> null
>>> app(Foo::class)
=> Foo {#4857}
>>> app(Foo::class)
=> Foo {#4857}
```

We should actually keep this, it is crucial for deployments that have long running server processes.

spawnia (changes requested):

PR #2134, opened 2022-04-29
Improve cache control directive docs

spawnia:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -528,30 +528,30 @@ Influences the HTTP `Cache-Control` headers of the response.
 """
 directive @cacheControl(
   """
-  The maximum amount of time the field's cached value is valid, in seconds. 
+  The maximum amount of time the field's cached value is valid, in seconds.
   0 means the field is not cacheable.
   """
   maxAge: Int! = 0
 
   """
-  Is the value specific to a single user? 
+  Is the value specific to a single user?
   """
-  scope: CacheControlScope! = PUBLIC  
+  scope: CacheControlScope! = PUBLIC
 ) on FIELD_DEFINITION | OBJECT | INTERFACE | UNION
 
 """
 Options for the `scope` argument of `@cacheControl`.
 """
 enum CacheControlScope {
-    """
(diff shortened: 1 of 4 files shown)


spawnia:
We need to revamp the implementation strategy to ensure that root-fields are not cached by default. Instead of implementing `@cacheControl` as `FieldMiddleware´, we could
- listen to the `StartExecution` event
- traverse the AST of the incoming query
- visit every field using `Visitor`
- consider the field's `@cacheControl` settings or apply defaults according to https://www.apollographql.com/docs/apollo-server/performance/caching/#default-maxage

mostafa-rz:
> 



> We need to revamp the implementation strategy to ensure that root-fields are not cached by default. Instead of implementing `@cacheControl` as `FieldMiddleware´, we could
> 
> * listen to the `StartExecution` event
> * traverse the AST of the incoming query
> * visit every field using `Visitor`
> * consider the field's `@cacheControl` settings or apply defaults according to https://www.apollographql.com/docs/apollo-server/performance/caching/#default-maxage
> 
> @mostafa-rz are you willing to take this on?

yes, i will go with it.

spawnia:
Alright, thanks. Please use this branch as the basis for your implementation, I will revert the addition of the previous PR to `master` for now.

mostafa-rz:
> We need to revamp the implementation strategy to ensure that root-fields are not cached by default. Instead of implementing `@cacheControl` as `FieldMiddleware`, we could
> 
> * listen to the `StartExecution` event
> * traverse the AST of the incoming query
> * visit every field using `Visitor`
> * consider the field's `@cacheControl` settings or apply defaults according to https://www.apollographql.com/docs/apollo-server/performance/caching/#default-maxage
> 
> @mostafa-rz are you willing to take this on?

I need some help.

1. The `StartExecution` event is just a parsed incoming request. No directive information inside it thus  `ASTHelper::hasDirective` returns nothing.
2. Is any method or helper exist, which I can give the field and get back the directive information? 
3. What is this `Visitor`?

PR #2136, opened 2022-05-03
Add `@cacheControl` directive

mostafa-rz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/2126

**Changes**

Allows setting the `Cache-Control` header through the usage of the `@cacheControl` directive.

**Breaking changes**

Nope.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -772,8 +772,10 @@ type Mutation {
   createPost(input: CreatePostInput!): Post! @convertEmptyStringsToNull
 }
 ```
+
 Non-nullable arguments will *not* be converted when this directive is used on a field,
 but will be converted when it is used directly on the argument.
+
 ```graphql
 type Mutation {
   createPost(
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -528,30 +528,30 @@ Influences the HTTP `Cache-Control` headers of the response.
 """
 directive @cacheControl(
   """
-  The maximum amount of time the field's cached value is valid, in seconds. 
+  The maximum amount of time the field's cached value is valid, in seconds.
   0 means the field is not cacheable.
   """
   maxAge: Int! = 0
 
   """
-  Is the value specific to a single user? 
+  Is the value specific to a single user?
   """
-  scope: CacheControlScope! = PUBLIC  
+  scope: CacheControlScope! = PUBLIC
 ) on FIELD_DEFINITION | OBJECT | INTERFACE | UNION
 
 """
 Options for the `scope` argument of `@cacheControl`.
 """
 enum CacheControlScope {
-    """
-    The value is the same for each user.
-    """
-    PUBLIC
+  """
+  The value is the same for each user.
+  """
+  PUBLIC
 
-    """
-    The value is specific to a single user.
-    """
-    PRIVATE
+  """
+  The value is specific to a single user.
+  """
+  PRIVATE
 }
 ```
 
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -524,19 +524,19 @@ You can find usage examples of this directive in [the caching docs](../performan
 
 ```graphql
 """
-Set the HTTP Cache-Control headers of the response.
+Influences the HTTP `Cache-Control` headers of the response.
 """
 directive @cacheControl(
-    """
-    The maximum amount of time the field's cached value is valid, in seconds.
-    The default value is 0(no-cache).
-    """
-    maxAge: Int = 0
+  """
+  The maximum amount of time the field's cached value is valid, in seconds. 
+  0 means the field is not cacheable.
+  """
+  maxAge: Int! = 0
 
-    """
-    If PRIVATE, the field's value is specific to a single user.
-    """
-    scope: CacheControlScope = PUBLIC
+  """
+  Is the value specific to a single user? 
+  """
+  scope: CacheControlScope! = PUBLIC  
 ) on FIELD_DEFINITION | OBJECT | INTERFACE | UNION
 
 """
@@ -544,18 +544,18 @@ Options for the `scope` argument of `@cacheControl`.
 """
 enum CacheControlScope {
     """
-    The HTTP Cache-Control header set to public.
+    The value is the same for each user.
     """
     PUBLIC
 
     """
-    The HTTP Cache-Control header set to private.
+    The value is specific to a single user.
     """
     PRIVATE
 }
 ```
 
-You can find usage examples of this directive in [the caching docs](../performance/caching.md#http-cache-control-header).
+Find usage examples of this directive in [the caching docs](../performance/caching.md#http-cache-control-header).
 
 ## @cacheKey
 
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -769,7 +769,17 @@ Usage on a field applies the conversion recursively to all inputs.
 
 ```graphql
 type Mutation {
-  createPost(input: CreatePostInput!): Post! @trim
+  createPost(input: CreatePostInput!): Post! @convertEmptyStringsToNull
+}
+```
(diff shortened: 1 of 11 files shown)


spawnia:
Please target `master` with your merge request.

spawnia:
Please merge/rebase against `master` to resolve the conflicts and have a clean diff.

mostafa-rz:
> Please merge/rebase against `master` to resolve the conflicts and have a clean diff.

I already did it according to this [link](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request).

mostafa-rz:
> You did change the target branch, but there are a lot of conflicts:
> 
> image
> 
> This is due to me reverting the original PR.
I merged it and all the tests are passed.

spawnia:
> I merged it and all the tests are passed.

They were all failing due to syntax errors, which I now fixed.

spawnia:
There are still test failures. Please mark as ready when all tests are passing.

docs/master/api-reference/directives.md
```diff
@@ -732,19 +769,7 @@ Usage on a field applies the conversion recursively to all inputs.
...
-Non-nullable arguments will *not* be converted when this directive is used on a field,
-but will be converted when it is used directly on the argument.
-
-```graphql
-type Mutation {
-  createPost(
-      willBeConvertedBecauseExplicitlyMarked: String! @convertEmptyStringsToNull
-      willNotBeConvertedToMaintainInvariants: String!
-  ): Post! @convertEmptyStringsToNull
+  createPost(input: CreatePostInput!): Post! @trim
```
spawnia, line 772: This should stay as-is.

docs/master/performance/caching.md
```diff
@@ -101,3 +101,68 @@ type GithubProfile {
...
+# Cache-Control header: no-cache, PRIVATE
+{
+  me {
+    # 5, PRIVATE
+    tasks {
+      # 50, PUBLIC
+      id # 10, PUBLIC
+      description # no-cache, PUBLIC
+    }
+  }
```
spawnia, line 152:
```suggestion
  me { # 5, PRIVATE
    tasks { # 50, PUBLIC
      id # 10, PUBLIC
      description # no-cache, PUBLIC
    }
  }
```

Otherwise it is unclear which field the comment refers to

src/CacheControl/CacheControlDirective.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @var \Nuwave\Lighthouse\CacheControl\CacheControl
+     */
+    protected $cacheControl;
+
+    public function __construct(CacheControl $cacheControl)
+    {
+        $this->cacheControl = $cacheControl;
+    }
+
```
spawnia, line 18:
```suggestion
```

src/CacheControl/CacheControlServiceProvider.php
```diff
@@ -0,0 +1,98 @@
...
+            static function (): string {
+                return __NAMESPACE__;
+            }
+        );
+
+        $dispatcher->listen(
+            StartExecution::class,
+            function (StartExecution $StartExecution) use ($cacheControl) {
+                $typeInfo = new TypeInfo($StartExecution->schema);
+                Visitor::visit($StartExecution->query, Visitor::visitWithTypeInfo($typeInfo, [
```
spawnia, line 39:
```suggestion
            function (StartExecution $startExecution) use ($cacheControl) {
                $typeInfo = new TypeInfo($startExecution->schema);
                Visitor::visit($startExecution->query, Visitor::visitWithTypeInfo($typeInfo, [
```

src/CacheControl/CacheControlServiceProvider.php
```diff
@@ -0,0 +1,98 @@
...
+                ]));
+            }
+        );
+
+        $dispatcher->listen(
+            EndRequest::class,
+            function (EndRequest $request) use ($cacheControl): void {
+                $maxAge = $cacheControl->calculateMaxAge();
+                if (0 != $maxAge) {
+                    $request->response->setMaxAge($maxAge);
```
spawnia, line 88: We can extract both `$response` and `$headers` to avoid repeated property access.

spawnia (changes requested):

tests/Integration/CacheControl/CacheControlDirectiveTest.php
```diff
@@ -0,0 +1,224 @@
...
+        }
+        ';
+
+        $this->graphQL(/** @lang GraphQL */ '
+        {
+            user {
+                name
+            }
+        }
+        ')->assertHeader('Cache-Control', 'no-cache, private');
```
spawnia, line 37: I am not sure if this is the desired default behaviour, the Apollo docs say https://www.apollographql.com/docs/apollo-server/performance/caching/#calculating-cache-behavior

> The response's scope is PRIVATE if any field's scope is PRIVATE.

That would mean that fields without directives default to `PUBLIC` - that seems to be in line with the default in `@cacheControl`.

In any case, the behaviour is currently inconsistent, as illustrated by the failing test `testRootScalar` I added. Querying the root field `value` by itself defaults the cache to `private`, but adding another field makes it `public` - that does not make sense.

spawnia (changes requested):

tests/Integration/CacheControl/CacheControlDirectiveTest.php
mostafa-rz: You are mixing two things here:

1. The app default behavior was `no-cache, private` without any `@cacheControl`, this probably come from symfony or laravel. I code it in this way if there is no `@cacheControl` don't change anything.
2. When u add any `@cacheControl` then everything is calculated based on this directive, and the default for this directive is public unless u mention the private, thus the root field `value` is also `public` here.

Apollo didn't mention anything about the root field default scope.

mostafa-rz: We can also do one of these:
- Make the default applications response public. 
- Make the route fields default scope public and mention it in doc.

spawnia: I think we should make an informed decision about the default scope for root fields. Is it a potential security risk to expose arbitrary fields with public scope? I would guess so, since a public response could be cached by proxy servers. If it holds private data, it could be accidentally shared with other users. I think they should default to private. Their default `max-age` is 0 anyway, so they won't be cached unless marked with `@cacheControl`.

Since Lighthouse currently follows the application default and this feature would override that, we should make it entirely optional. Let's do that by omitting the service provider from `composer.json` and let users add it explicitly.

mostafa-rz: I made root fields default scope private anyway. Now if someone wants to use this, he must define `@cacheControl` for the root fields.

spawnia: > I made root fields default scope private anyway.

I understand this was your intention, but it does not behave that way - as illustrated by https://github.com/nuwave/lighthouse/pull/2136#discussion_r866679871

spawnia:
@mostafa-rz please review the changes I made in [`f7a94af` (#2136)](https://github.com/nuwave/lighthouse/pull/2136/commits/f7a94afb5ebeb4c3bef11da2b2e766b159feed4f)

mostafa-rz:
> @mostafa-rz please review the changes I made in [`f7a94af` (#2136)](https://github.com/nuwave/lighthouse/pull/2136/commits/f7a94afb5ebeb4c3bef11da2b2e766b159feed4f)

@spawnia review done and made some changes [`ba2fe8f`](https://github.com/nuwave/lighthouse/pull/2136/commits/ba2fe8ff94d2bfc8df057f05c819ffe099998e56)

spawnia (approved):

stayallive (approved):
This looks awesome 💪 

Not sure how I imagined the implementation but this is pretty clean.

PR #2138, opened 2022-05-06
Add `columns` and `distinct` args to `@count` directive

crojasaragonez:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

- Add `columns` and `distinct` args to `@count` directive https://github.com/nuwave/lighthouse/pull/2138

**Breaking changes**

- Nope

### DIFF
--- a/src/Schema/Directives/CountDirective.php
+++ b/src/Schema/Directives/CountDirective.php
@@ -47,10 +47,9 @@ directive @count(
   field: String
 
   """
-  distinct count.
-  exclude duplicated rows
+  Should exclude duplicated rows?
   """
-  distinct: Boolean
+  distinct: Boolean! = false
 ) on FIELD_DEFINITION
 GRAPHQL;
     }
--- a/src/Schema/Directives/CountDirective.php
+++ b/src/Schema/Directives/CountDirective.php
@@ -39,6 +39,18 @@ directive @count(
   Apply scopes to the underlying query.
   """
   scopes: [String!]
+
+  """
+  The field to count.
+  counts only a specific field
+  """
+  field: String
+
+  """
+  distinct count.
+  exclude duplicated rows
+  """
+  distinct: Boolean
 ) on FIELD_DEFINITION
 GRAPHQL;
     }
@@ -52,6 +64,15 @@ GRAPHQL;
                     ->namespaceModelClass($modelArg)::query();
 
                 $this->makeBuilderDecorator($resolveInfo)($query);
+
+                if ($this->directiveArgValue('distinct') ?? false) {
+                    $query->distinct();
+                }
+
+                $field = $this->directiveArgValue('field');
+                if ($field) {
+                    return $query->count($field);
+                }
 
                 return $query->count();
             });
--- a/src/Schema/Directives/CountDirective.php
+++ b/src/Schema/Directives/CountDirective.php
@@ -68,9 +68,9 @@ GRAPHQL;
                     $query->distinct();
                 }
 
-                $field = $this->directiveArgValue('field');
-                if ($field) {
-                    return $query->count($field);
+                $columns = $this->directiveArgValue('columns');
+                if ($columns) {
+                    return $query->count(...$columns);
                 }
 
                 return $query->count();
--- a/src/Schema/Directives/CountDirective.php
+++ b/src/Schema/Directives/CountDirective.php
@@ -41,10 +41,10 @@ directive @count(
   scopes: [String!]
 
   """
-  The field to count.
-  counts only a specific field
+  Count only rows where the given columns are non-null.
+  `*` counts every row.
   """
-  field: String
+  columns: [String!]! = ["*"]
 
   """
   Should exclude duplicated rows?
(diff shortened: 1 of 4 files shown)


tests/Integration/Schema/Directives/CountDirectiveDBTest.php
```diff
@@ -420,4 +420,44 @@ public function testResolveRelationItemsAndCount(): void
...
+    }
+
+    public function testCountModelWithFieldAndDistinct(): void
+    {
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            tasks: Int @count(model: "Task", distinct: true, field: "difficulty")
+        }
+        ';
+        factory(Task::class, 3)->create(['difficulty' => 1]);
```
spawnia, line 451: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#test-data-setup

spawnia (changes requested):

src/Schema/Directives/CountDirective.php
```diff
@@ -39,6 +39,18 @@ public static function definition(): string
   Apply scopes to the underlying query.
   """
   scopes: [String!]
+
+  """
+  The field to count.
+  counts only a specific field
+  """
+  field: String
```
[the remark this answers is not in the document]
crojasaragonez: Sure, let me put together a few scenarios:
let's say we have 100 Projects with different states (the state column allows null values):

- Completed
- Draft
- Cancelled
- In Progress
- On Hold

Let's assume 5 of them have a null state.

| What we want  | Eloquent | lighthouse directive | Result |
| ------------- | ------------- | ------------- | ------------- |
| Count every row  | Project::count()  | @count(model: "Project") | 100 |
| Count every row excluding where state is null  | Project::count('state')  | @count(model: "Project", field: "state", distinct: true) | 95|
| Count every different state  | Project::distinct()->count('state')  | @count(model: "Project", field: "state", distinct: true) | 5|

src/Schema/Directives/CountDirective.php
```diff
@@ -39,6 +39,18 @@ public static function definition(): string
...
+
+  """
+  distinct count.
+  exclude duplicated rows
+  """
+  distinct: Boolean
```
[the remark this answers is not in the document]
crojasaragonez: The second argument **field** below will take the column name, this is only to exclude duplicate rows when counting records.

tests/Integration/Schema/Directives/CountDirectiveDBTest.php
crojasaragonez: I need to create three tasks with the same difficulty in order to test the **field** and **distinct** combination, please let me know if there is a better way. 🙏🏼

spawnia:
```suggestion
        foreach (factory(Task::class, 3)->make() as $task) {
            $task->difficulty = 1;
            $task->save();
        }
```

A bit more verbose, but typesafe and able to be refactored.

src/Schema/Directives/CountDirective.php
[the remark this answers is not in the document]
spawnia: Got it, makes sense to have this as a `Boolean` then. The definition can be cleaned up though, since passing `null` here does not make sense and there is a clear default:

```suggestion
  """
  Should exclude duplicated rows?
  """
  distinct: Boolean! = false
```

src/Schema/Directives/CountDirective.php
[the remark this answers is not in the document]
spawnia: Great example. A test case for the effect of passing only `field` without `distinct` would be nice, asserting that it is indeed different from not passing `field`.

When looking at the signature of `count()`, I see that it can take more then one column. I don't know if it is a common scenario, but we can easily accomodate for it.

```suggestion
  """
  Count only rows where the given columns are non-null.
  `*` counts every row.
  """
  columns: [String!]! = ['*']
```

tests/Integration/Schema/Directives/CountDirectiveDBTest.php
```diff
@@ -420,4 +420,67 @@ public function testResolveRelationItemsAndCount(): void
             ],
         ]);
     }
+
+    public function testCountModelWithSingleColumn(): void
```
spawnia, line 424: I don't see how this test shows the effectiveness of `columns`. Perhaps one of the tasks should have `difficulty = null`?

tests/Integration/Schema/Directives/CountDirectiveDBTest.php
```diff
@@ -420,4 +420,67 @@ public function testResolveRelationItemsAndCount(): void
...
+            tasks
+        }
+        ')->assertExactJson([
+            'data' => [
+                'tasks' => 3,
+            ],
+        ]);
+    }
+
+    public function testCountModelWithMultipleColumns(): void
```
spawnia, line 444: Same issue as with `testCountModelWithSingleColumn`. The test should show that `columns` makes a meaningful difference in results.

spawnia (changes requested):
The implementation and docs look fine, but I am not confident that it actually works as intended without seeing the proper tests.

In addition to the problem with the current tests for `columns`, we are also missing a test that shows the effect of only passing `distinct`. `testCountModelWithFieldAndDistinct` looks good and actually shows a meaningful effect of the directives.

spawnia:
Thanks for your work on this.

PR #2142, opened 2022-05-16
Fix `@convertEmptyStringsToNull` on fields and matrices

spawnia:
#2130 (comment)

- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Fix `@convertEmptyStringsToNull` on fields and matrices.

**Breaking changes**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


madrussa:
> Breaking changes -> none.

This has breaking change implications as the expected behaviour to return the empty value for input objects no longer adheres to this behaviour. We have been caught by this specific change during our upgrade. It's worth calling this out so that your team is aware that going from only mapping initial to mapping recursively can cause issues.

spawnia:
I see your point, but still consider this a bugfix as it restores the indentend behaviour of this directive that was already present in a prior `5.x` release.

PR #2149, opened 2022-05-25
Include field that produced a `RateLimitException` in its message

### LINKED ISSUE #2145: Get `RateLimitException`s origin
We are currently building a new GraphQL-Query which we want to throttle by `@throttle`. We are pretty unsure which rate limit we should use (The query might get executed quite often when the user is mistyping its input). So we want to start with a fairly restrictive rate limit and adjust it along the way. Therefore we wanted to add a new `ErrorHandler` to log the `RateLimitException` to be notified about how often users run into the rate limit. Unfortunately, when we look into the logs we can not determine which query was the cause for such an entry in our log files. So, we have to guess if the rate limit is from someone trying to abuse our API or someone hitting our new query.

It would be great if the `RateLimitException` could be extended to provide the information about the path the exception was thrown at. The exception could get this info by a newly introduced constructor parameter or a setter function. Then we could access this property in our `Errorhandler` to enhance our log message with that info.

If this is desired, we would be happy to bring up a PR.

christian98:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #2145 

**Changes**

As discussed in #2145, this PR adds the query path to the `RateLimitException` message.

**Breaking changes**

Only the already mentioned message of the exception changes. 😉

### DIFF
--- a/tests/Integration/Schema/Directives/ThrottleDirectiveTest.php
+++ b/tests/Integration/Schema/Directives/ThrottleDirectiveTest.php
@@ -97,7 +97,7 @@ final class ThrottleDirectiveTest extends TestCase
         {
             foo
         }
-        ')->assertGraphQLErrorMessage(RateLimitException::MESSAGE);
+        ')->assertGraphQLErrorMessage(RateLimitException::buildErrorMessage('foo'));
     }
 
     public function testInlineLimiter(): void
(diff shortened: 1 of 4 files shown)


spawnia:
Please rebase atop the latest changes in `master`.

spawnia (changes requested):

tests/Integration/Schema/Directives/ThrottleDirectiveTest.php
```diff
@@ -122,6 +122,44 @@ public function testInlineLimiter(): void
...
+            ],
+        ]);
+
+        $this->graphQL(/** @lang GraphQL */ '
+        {
+            foo {
+                bar
+            }
+        }
+        ')->assertGraphQLErrorMessage(RateLimitException::buildErrorMessage('foo.bar'));
```
spawnia, line 163:
```suggestion
```

This test is redundant after switching to type + field.

Issue #2158, opened 2022-06-08
`@can` is reading all records of the table


loureirorg:
When used with a relationship directive, the `@can` directive is loading all the records from the table.

For example, the definition:
```graphql
type Query {
    me: User @auth @guard
}

type User {
    id: ID!
    cats: [Cat]! @hasMany @can(ability: "readCat", query: true)
}
```

And the request:
```graphql
query {
  me {
    id
    cats { name }
  }
}
```

The `readCat` policy is being called for each Cat in the database. If the `cats` table has 1 million records, the `readCat` policy will be called 1M times, even if the user has no cats.

This is not just a performance issue, it also makes `@can` unusable to prevent unauthorized access in nested queries.

Considering the `readCat` definition:
```php
Gate::define('readCat', function (User $user, Cat $cat) {
    return $user->id === $cat->user_id;
});
```

The request always returns `unauthorized`, because some of the 1M cats don't belong to the authenticated user - and `readCat` is called for all 1M cats.

One can argue that there's no need to check for authorization if you are using the Viewer pattern, or something like that, but here's a real-world use-case for this scenario:
https://stackoverflow.com/questions/60435387/laravel-lighthouse-graphql-limit-list-of-models-to-the-ones-user-can-access

**Expected behavior/Solution**
The policy should only be called for the returned records. If the user has only two cats (ids 1 and 2), the policy should only be called for these two records.

Also, for performance reasons, `@can` shouldn't be loading the whole table.

**Steps to reproduce**

1. Create a Laravel+Lighthouse+Sanctum app;
2. Create a relationship for the User model (ex. `php artisan make:model Cat --migration`);
3. Create at least two User records (ex. `User::factory()->count(2)->create()`);
4. Generate many relationships for the second User (ex. `Cat::factory()->count(100)->create(['user_id' => 2])`)
5. Use these definitions:
```graphql
type Query {
    me: User @auth @guard
}

type User {
    id: ID!
    cats: [Cat]! @can(ability: "readCat", query: true)
}

type Cat {
    id: ID!
    name: String!
}
```

```php
// AuthServiceProvider.php
Gate::define('readCat', function (User $user, Cat $cat) {
    return $user->id === $cat->user_id;
});
```

6. On a GraphQL client (GraphQL Playground, Insomnia, Postman, etc)
7. Sign in as User ID 1
8. Run this request:
```graphql
query FetchUser {
  me {
    id
    cats { name }
  }
}
```

[...]

Issue #1989, opened 2021-11-29
Deduplicate database queries made through `@can`


baravak:
**What problem does this feature proposal attempt to solve?**

Checking model-specific policies through `@can` causes duplicate database queries:
1. for the policy check
2. for resolving the result

Given the following field definitions:

```graphql
gift(id: ID! @eq): Gift! @can(ability: "view", query: true) @find
gift(id: ID! @eq): Gift! @can(ability: "view", find: "id") @find
```

Lighthouse will produce two identical database queries:

```sql
select * from `gifts` where `id` = ?
select * from `gifts` where `id` = ?
```

**Expected behavior/Solution**

It would be ideal if Lighthouse could detect scenarios where the queries will be identical and reuse the result of the first query, avoiding a redundant query.

spawnia:
I reformulated the issue into a focused feature request.

I can only see deduplication happening reliably for the variant that uses `@can(query: true)`. The variant with `@can(find: "id")` might not work in all scenarios, since the field might have other arguments which are currently not considered in the policy check query.

This is a difficult problem to solve due to the implementation strategy of `@can` and `@find`. Both are directives which are decoupled from each other and run in separate phases.

PR #2159, opened 2022-06-09
Allow checking policies against resolved models in `@can`

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/2158

Resolves https://github.com/nuwave/lighthouse/issues/1989

**Changes**

Adds the option `resolved` to `@can` as an alternative to `find` and `query`.
Instead of fetching models to check against before calling the resolver, it will instead use the result of
resolving the field and check against that.

**Breaking changes**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v5.57.0
+
+### Added
+
+- Allow checking policies against resolved models in `@can` https://github.com/nuwave/lighthouse/pull/2159
+
 ## v5.56.0
 
 ### Added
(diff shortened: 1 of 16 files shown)


loureirorg:
It also throws this exception when used without pagination:
`Argument #1 ($model) must be of type Illuminate\Database\Eloquent\Model, GraphQL\Deferred given`

Long log error:
```
Nuwave\Lighthouse\Auth\CanDirective::Nuwave\Lighthouse\Auth\{closure}(): Argument #1 ($model) must be of type Illuminate\Database\Eloquent\Model, GraphQL\Deferred given, called in lighthouse-can-issue/vendor/nuwave/lighthouse/src/Support/Utils.php on line 99
```

loureirorg:
It's working perfectly now. Thanks, @spawnia!

For reference, this is what I used (no `query: true` as suggested):
```graphql
cats: [Cat]! @hasMany @can(ability: "readCat", resolved: true)
```

esistgut:
I can confirm this appears to be now working as expected.

PR #2170, opened 2022-06-20
Optimize paginator query when no page info is requested

stayallive:
- [x] Added or updated tests
- [x] ~Documented user facing changes~
- [x] Updated CHANGELOG.md

**Changes**

When doing a paginated query we use the `LengthAwarePaginator` from Laravel. This makes total sense but we can optimize in case the query doesn't request the paginator info, then we can fallback to a simple paginator saving a `count(*)` query which is always nice :)

This is especially useful when you want to do something like getting the first object from a paginated list to see if it exists or to show a data but don't want yet another attribute, instead you can do something like this without the `count(*)` query penalty:

```graphql
# Pseudo schema
type User {
    auditLogs: [AuditLogEntry!]! @paginate
}

type Query {
    viewer: User @auth
}


# Query
{
    viewer {
        auditLogs(first: 1, page: 1) {
            data {
                id
                createdAt
            }
        }
    }
}
```

**Breaking changes**

I think none... maybe API changes I made that are breaking?

### DIFF
--- a/src/Pagination/ConnectionField.php
+++ b/src/Pagination/ConnectionField.php
@@ -41,7 +41,6 @@ class ConnectionField
     /**
      * Resolve edges for connection.
      *
-     * @param  \Illuminate\Pagination\AbstractPaginator  $paginator
      * @param  array<string, mixed>  $args
      */
     public function edgeResolver(AbstractPaginator $paginator, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Collection
--- a/src/Pagination/ConnectionField.php
+++ b/src/Pagination/ConnectionField.php
@@ -4,6 +4,7 @@ namespace Nuwave\Lighthouse\Pagination;
 
 use GraphQL\Type\Definition\ResolveInfo;
 use Illuminate\Contracts\Pagination\LengthAwarePaginator;
+use Illuminate\Pagination\AbstractPaginator;
 use Illuminate\Support\Collection;
 use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;
 
@@ -40,10 +41,10 @@ class ConnectionField
     /**
      * Resolve edges for connection.
      *
-     * @param  \Illuminate\Pagination\LengthAwarePaginator<mixed>  $paginator
+     * @param  \Illuminate\Pagination\AbstractPaginator  $paginator
      * @param  array<string, mixed>  $args
      */
-    public function edgeResolver(LengthAwarePaginator $paginator, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Collection
+    public function edgeResolver(AbstractPaginator $paginator, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Collection
     {
         // We know those types because we manipulated them during PaginationManipulator
         /** @var \GraphQL\Type\Definition\NonNull $nonNullList */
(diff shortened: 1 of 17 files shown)


src/Pagination/PaginateDirective.php
```diff
@@ -126,18 +126,33 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
             $this->directiveArgValue('type') ?? PaginationType::PAGINATOR
         );
+
+        if (null === $resolveInfo || $type->isSimple()) {
+            return $type;
+        }
+
+        $pageInfoFieldName = $type->isPaginator()
+            ? 'paginatorInfo'
+            : 'pageInfo';
```
spawnia, line 148: Those should use constants that are also used in code generation to make the relation clear.

src/Pagination/ConnectionField.php
```diff
@@ -40,10 +41,9 @@ public function pageInfoResolver(LengthAwarePaginator $paginator): array
     /**
      * Resolve edges for connection.
      *
-     * @param  \Illuminate\Pagination\LengthAwarePaginator<mixed>  $paginator
```
spawnia, line 43: It is no longer generic?

src/Pagination/PaginateDirective.php
```diff
@@ -126,18 +126,33 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
-            return PaginationArgs::extractArgs($args, $this->paginationType(), $this->paginateMaxCount())
+            return PaginationArgs::extractArgs($args, $this->paginationType($resolveInfo), $this->paginateMaxCount())
                 ->applyToBuilder($query);
         });
 
         return $fieldValue;
     }
 
-    protected function paginationType(): PaginationType
+    protected function paginationType(?ResolveInfo $resolveInfo = null): PaginationType
```
spawnia, line 136: When is `ResolveInfo` not given?

tests/Integration/Pagination/PaginateDirectiveDBTest.php
```diff
@@ -424,6 +424,96 @@ public function testQueriesPaginationWithNoData(): void
...
+        type User {
+            id: ID!
+        }
+
+        type Query {
+            users: [User!]! @paginate
+        }
+        ';
+
+        DB::enableQueryLog();
```
spawnia, line 442: We typically use another method to count database queries, something with registering a query listener. Let's stay consistent - perhaps abstract?

spawnia (changes requested):
I like the direction of this change.

tests/Integration/Pagination/PaginateDirectiveDBTest.php
stayallive: I copied this from another test in the exact same file...

But I'll browse around then.

src/Pagination/PaginateDirective.php
stayallive: When we build the schema, there we also need the type to correctly add the types but we obviously don't have resolve info nor do we want it there.

tests/Integration/Pagination/PaginateDirectiveDBTest.php
spawnia: Got it. If you could unify the approaches that would be great.

stayallive: I have checked around but was unable to find anywhere we are query counting.

Maybe I'm blind but the `Event::fake` route also wasn't working for me so took a trait that add some nice syntax sugar for future query counting needs :)

PR #2176, opened 2022-06-27
Pass the path array to `CacheKeyAndTags::key()`

mostafa-rz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Pass the resolveInfo->path array to the `key` method of `CacheKeyAndTags` class.

**Breaking changes**

interface signautre change.


src/Cache/CacheKeyAndTags.php
```diff
@@ -11,8 +11,9 @@ interface CacheKeyAndTags
      *
      * @param  int|string|null $id
      * @param  array<string, mixed> $args
+     * @param  array<int, string> $path
      */
-    public function key(?Authenticatable $user, bool $isPrivate, string $parentName, $id, string $fieldName, array $args): string;
+    public function key(?Authenticatable $user, bool $isPrivate, string $parentName, $id, string $fieldName, array $args, array $path): string;
```
spawnia, line 16:
```suggestion
    public function key(?Authenticatable $user, bool $isPrivate, array $args, ResolveInfo $resolveInfo): string;
```

That way we can avoid future breakage and simply have all relevant context available.

mostafa-rz:
@spawnia Would you mind if also changing the signature of the parentTag and fieldTag method? Add resolveInfo and remove parentName and fieldName.

src/Cache/CacheKeyAndTags.php
```diff
@@ -12,19 +13,19 @@ interface CacheKeyAndTags
...
-    public function key(?Authenticatable $user, bool $isPrivate, string $parentName, $id, string $fieldName, array $args): string;
+    public function key(?Authenticatable $user, bool $isPrivate, $id, array $args, ResolveInfo $resolveInfo): string;
 
     /**
      * Generate a tag for the parent.
      *
      * @param  int|string|null $id
      */
-    public function parentTag(string $parentName, $id): string;
+    public function parentTag($id, ResolveInfo $resolveInfo, ?string $parentName = null): string;
```
spawnia, line 23: Why keep `$parentName`? It is never passed.

spawnia (commented):
How about we make the arguments the same for all 3 methods? That way, the implementation really can do whatever it wants.

```php
($rootID, array $args, GraphQLContext $context, ResolveInfo $resolveInfo, bool $isPrivate)

src/Cache/CacheKeyAndTags.php
spawnia: I see, maybe passing `ResolveInfo` is not such a great idea after all, since it pretty much forces them to be called only in resolvers. I can see use cases where keys or tags are generated somewhere else, so we might just go back to your original implementation of just adding `array $path` to `key()`.

Sorry for the back and forth, but I would rather go the extra mile when designing interfaces than to break public APIs often.

mostafa-rz: Absolutely no problem. When it is a breaking change it is better to check all the possibilities.

mostafa-rz:
@spawnia Is there any timetable for version 6 release?

spawnia:
> @spawnia Is there any timetable for version 6 release?

No, but I reckon it will be a few months. If you need this soon, you can fork or overwrite the involved directives.

spawnia:

spawnia:
Missing Changelog. Also, why do you want to make this change at all, what is the use case?

mostafa-rz:
I apply local scopes to fields based on the path. The parent and field name is not enough to differentiate between the cache keys.

PR #2139, opened 2022-05-09
Allow ScoutBuilder in @paginate directive

arrisar:
An issue was not raised, as it seemed simpler to fix the issue.

Introduce in [this commit](https://github.com/nuwave/lighthouse/commit/6a9b5ac971edeef9a5b142d73902a5d4197b04c4), directive types were more strictly asserted. Causing the ScoutBuilder to be an invalid type in some directives.

While this PR only fixes this directive, it is likely there are others.

**Changes**

This PR _very_ simply allows the ScoutBuilder as a valid Builder type in the `@paginate` directive.

**Breaking changes**

None.

PR #2183, opened 2022-07-20
Allow returning `Laravel\Scout\Builder` from custom `builder` in `@all` and `@paginate`

pyrou:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #2139

**Changes**

Allows again the `Laravel\Scout\Builder`, aka ScoutBuilder, as a valid Builder type in `@all` and `@paginate` directives. Support was partial until lighthouse 5.46.

Current implementation of enhanceBuilder produce either a ScoutBuilder (and return it immediately) OR will eventually try to apply scope etc.. (which is not supported on `Laravel\Scout\Builder` instance).

~So, if eventually a custom builder param (from `@paginate` directive) produce a `Laravel\Scout\Builder` instance, I think we can safely skip the enhanceBuilder in that case.~

When a ScoutBuilder is returned in a custom builder for `@all` or `@paginate` directive, ScoutEnhancer is now called. Eventual directives, implementing `ScoutBuilderDirective` interface such as `@eq` or `@trashed`, are now applied to the custom ScoutBuilder instance.

**Breaking changes**

No

### DIFF
--- a/src/Pagination/PaginateDirective.php
+++ b/src/Pagination/PaginateDirective.php
@@ -120,14 +120,12 @@ GRAPHQL;
                 $query = $this->getModelClass()::query();
             }
 
-            if (! $query instanceof ScoutBuilder) {
-                $query = $resolveInfo
-                    ->argumentSet
-                    ->enhanceBuilder(
-                        $query,
-                        $this->directiveArgValue('scopes') ?? []
-                    );
-            }
+            $query = $resolveInfo
+                ->argumentSet
+                ->enhanceBuilder(
+                    $query,
+                    $this->directiveArgValue('scopes') ?? []
+                );
 
             return PaginationArgs::extractArgs($args, $this->optimalPaginationType($resolveInfo), $this->paginateMaxCount())
                 ->applyToBuilder($query);
--- a/src/Pagination/PaginateDirective.php
+++ b/src/Pagination/PaginateDirective.php
@@ -120,12 +120,14 @@ GRAPHQL;
                 $query = $this->getModelClass()::query();
             }
 
-            $query = $resolveInfo
-                ->argumentSet
-                ->enhanceBuilder(
-                    $query,
-                    $this->directiveArgValue('scopes') ?? []
-                );
+            if (!$query instanceof ScoutBuilder) {
+                $query = $resolveInfo
+                    ->argumentSet
+                    ->enhanceBuilder(
+                        $query,
+                        $this->directiveArgValue('scopes') ?? []
+                    );
+            }
 
             return PaginationArgs::extractArgs($args, $this->optimalPaginationType($resolveInfo), $this->paginateMaxCount())
                 ->applyToBuilder($query);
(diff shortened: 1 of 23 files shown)


pyrou:
maybe `@all` directive deserve same update?

src/Pagination/PaginateDirective.php
```diff
@@ -112,19 +113,21 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
                 $query = $this->getModelClass()::query();
             }
 
-            $query = $resolveInfo
-                ->argumentSet
-                ->enhanceBuilder(
-                    $query,
-                    $this->directiveArgValue('scopes') ?? []
-                );
+            if (! $query instanceof ScoutBuilder) {
```
spawnia, line 123: This would skip possibly important steps in https://github.com/pyrou/lighthouse/blob/6a810c0b00aa917b6ceb45c52d4080ba4268a5a9/src/Scout/ScoutEnhancer.php#L99-L106

pyrou: Would be ok if I made change so Nuwave\Lighthouse\Scout\ScoutEnhancer accept also ScoutBuilder instances (see https://github.com/pyrou/lighthouse/commit/233dd7827c3328056e222bc0598d5c2d591e6978) ?

(PaginateDirective and AllDirective still has to be modified of course)

src/Scout/ScoutEnhancer.php
```diff
@@ -145,4 +112,42 @@ function ($value) {
...
+         * We know this argument has this directive, because that is how we found it.
+         *
+         * @var \Nuwave\Lighthouse\Scout\SearchDirective $searchDirective
+         */
+        $searchDirective = $searchArgument
+            ->directives
+            ->first(Utils::instanceofMatcher(SearchDirective::class));
+
+        $searchDirective->search($scoutBuilder);
+    }
```
pyrou, line 152: Obviously I forgot : 

```suggestion
        return $scoutBuilder;
    }
```

src/Pagination/PaginateDirective.php
spawnia: Yeah, this is basically what I suggested in https://github.com/nuwave/lighthouse/pull/2139#issuecomment-1123738101.

spawnia:
The solution looks pretty good! To make sure it really works, could you add a test combining `@search` and `@all/@paginate` with `builder`?

pyrou:
@spawnia Not exactly sure about your Changelog, this PR doesn't change anything about `@search` directive. However, it now apply "SearchDirective" (I speak here about the interface) so directive like `@eq` are now applied on custom builder

tests/Integration/Schema/Directives/AllDirectiveTest.php
```diff
@@ -249,6 +253,126 @@ public function testSpecifyCustomBuilderForRelation(): void
...
+            ],
+        ]);
+    }
+
+    public function testSpecifyCustomBuilderForScoutBuilderWithScoutDirective(): void
+    {
+        $this->setUpScoutEngine();
+
+        /** @var \Mockery\MockInterface&\Laravel\Scout\Builder $builder */
+        $builder = $this->partialMock(\Laravel\Scout\Builder::class);
```
spawnia, line 313: Not supported in older PHPUnit/Laravel versions. Probably can be replaced by a check similar to `SearchDirectiveTest::testSearchWithEq()`.

spawnia (approved):
Nice work, thank you very much

PR #2187, opened 2022-07-29
Handle missing models in `@can` with option `resolved`

leinelissen:
This patch addresses an issue where a non-existing instance is queried while a @can directive exists on the field that is resolved. This will result in a type error since PHP expects the first argument of the closure in question to be a Model. The following error is generated

```
Nuwave\\Lighthouse\\Auth\\CanDirective::Nuwave\\Lighthouse\\Auth\\{closure}(): Argument #1 ($model) must be of type Illuminate\\Database\\Eloquent\\Model, null given, called in /***/vendor/nuwave/lighthouse/src/Support/Utils.php on line 117"
```

- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

The fix is to modify the type of the argument into a nullable type. The
operator is available since PHP 7.1, which means this should be a
backwards-compatible fix.

**Breaking changes**

No breaking changes expected.

### DIFF
--- a/src/Auth/CanDirective.php
+++ b/src/Auth/CanDirective.php
@@ -135,7 +135,7 @@ GRAPHQL;
                             ? $modelLike->items()
                             : $modelLike;
 
-                        Utils::applyEach(function (Model $model) use ($gate, $ability, $checkArguments): void {
+                        Utils::applyEach(function (?Model $model) use ($gate, $ability, $checkArguments): void {
                             $this->authorize($gate, $ability, $model, $checkArguments);
                         }, $modelOrModels);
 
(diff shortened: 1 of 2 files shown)


spawnia:
Please add a test case that reproduces the issue.

leinelissen:
Hey @spawnia. I've added a test case that successfully reproduces the error when the patch is not present (`Model $model` => `?Model $model`). It appears the error appears due the combination of the `@can` and `@find` directives. I hope this is what you're looking for, and if not, I'm happy to help out.

leinelissen:
Hey @spawnia, I see the workflow you approved threw a bunch of errors. I've resolved the ones that popped up, so I'm curious to see the outcome if you try it again. Thanks for the help.

spawnia:
Thanks, released the fix with https://github.com/nuwave/lighthouse/releases/tag/v5.57.1

leinelissen:
Hey, thanks for the quick turnaround on this one. Happy to see it released!

PR #2190, opened 2022-08-04
Fix cursor pagination always returning the first page when no fields from the `pageInfo` are requested

### LINKED ISSUE #2189: `CONNECTION` pagination broken without `pageInfo`
**Describe the bug**

CONNECTION pagination offset doesn't work as expected if no pageInfo is requested. Attribute `after` looks not being used, and page 1 is returned when asking for page 2

I added 2 similar unit tests is [this commit](https://github.com/pyrou/lighthouse/commit/c1609c4ecfe1a736b6a1dd328d3c273707e9a8bb), one with pageInfo, one without. Second test is failing, items for page one is returned.

**Lighthouse Version**

This issue introduced with PR #2170 (Lighthouse 5.55)

stayallive:
- [x] Added or updated tests
- [x] ~Documented user facing changes~
- [x] Updated CHANGELOG.md

Fixes #2189

**Changes**

We calculate an optimal pagination strategy when the paginator info is not requested. But since cursor based pagination needs some special handling to correctly calculate the correct page to retrieve the `PaginationArgs::extractArgs` needs to get the original pagination type instead of the optimal type. We can change the type to the optimal one when applying it to the builder keeping the benefits but fixing the bug :)

Thanks @pyrou for finding and reporting the bug and creating the failing tests!

**Breaking changes**

None.

### DIFF
--- a/tests/Integration/Pagination/PaginateDirectiveDBTest.php
+++ b/tests/Integration/Pagination/PaginateDirectiveDBTest.php
@@ -610,12 +610,10 @@ GRAPHQL;
         }
         ';
 
-        $cursor = Cursor::encode(2);
-
-        $this->assertQueryCountMatches(2, function () use ($users, $cursor): void {
-            $this->graphQL(/** @lang GraphQL */ "
-            {
-                users(first: 2, after: \"{$cursor}\") {
+        $this->assertQueryCountMatches(2, function () use ($users): void {
+            $this->graphQL(/** @lang GraphQL */ '
+            query ($after: String!) {
+                users(first: 2, after: $after) {
                     pageInfo {
                       hasNextPage
                       hasPreviousPage
@@ -633,7 +631,9 @@ GRAPHQL;
                     }
                 }
             }
-            ")->assertJson([
+            ', [
+                'after' => Cursor::encode(2),
+            ])->assertJson([
                 'data' => [
                     'users' => [
                         'pageInfo' => [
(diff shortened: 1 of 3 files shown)


tests/Integration/Pagination/PaginateDirectiveDBTest.php
```diff
@@ -595,6 +596,112 @@ public function testQueriesConnectionWithoutPageInfo(): void
...
+            users: [User!]! @paginate(type: CONNECTION)
+        }
+        ';
+
+        $cursor = Cursor::encode(2);
+
+        $this->assertQueryCountMatches(2, function () use ($users, $cursor): void {
+            $this->graphQL(/** @lang GraphQL */ "
+            {
+                users(first: 2, after: \"{$cursor}\") {
```
spawnia, line 618: Should use variables

tests/Integration/Pagination/PaginateDirectiveDBTest.php
```diff
@@ -595,6 +596,112 @@ public function testQueriesConnectionWithoutPageInfo(): void
...
+            users: [User!]! @paginate(type: CONNECTION)
+        }
+        ';
+
+        $cursor = Cursor::encode(2);
+
+        $this->assertQueryCountMatches(1, function () use ($users, $cursor): void {
+            $this->graphQL(/** @lang GraphQL */ "
+            {
+                users(first: 2, after: \"{$cursor}\") {
```
spawnia, line 681: Also should use variables

spawnia (changes requested):
Nicely done, just needs a stylistic fix.

stayallive:
You are right, I was being lazy here... fixed!

PR #2193, opened 2022-08-04
Allow unlimited pagination from directive

pyrou:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

In configuration file `lighthouse.pagination` the key `defaultCount` and `maxCount` are used as a default value when not specified in schema.   Also, specific behaviors [are documented](https://github.com/nuwave/lighthouse/blob/master/src/lighthouse.php#L210) in  for these key for the `null` values. 

```php
    'pagination' => [
        /*
         * Allow clients to query paginated lists without specifying the amount of items.
         * Setting this to `null` means clients have to explicitly ask for the count.
         */
        'default_count' => null,

        /*
         * Limit the maximum amount of items that clients can request from paginated lists.
         * Setting this to `null` means the count is unrestricted.
         */
        'max_count' => null,
    ],
```
  
  While it is authorized currently to override these default value in schema, by setting value higher or lower than the configuration value; It is not currently supported to explicitly set `null` in schema exceptionally, for a specified field to unrestrict pagination (maxCount: null) OR to make clients have to explicitly ask for the count (defaultCount: null)
  
    This PR solves this little inconsistency between schema and configuration behavior.
  
**Breaking changes**

Even if this might be very unusual, people using explicitly `maxCount: null` or `defaultCount: null` in their schema, are  currently falling back to configuration value if any.     With this PR, `maxCount: null` will unrestrict pagination limitation, while `defaultCount: null` will require clients to explicitly ask for the count. 

People not specifying maxCount or defaultCount in schema and relying on configuration fallback remains backward compatible.

### DIFF
--- a/docs/5/api-reference/directives.md
+++ b/docs/5/api-reference/directives.md
@@ -249,12 +249,14 @@ directive @belongsToMany(
   """
   Allow clients to query paginated lists without specifying the amount of items.
   Overrules the `pagination.default_count` setting from `lighthouse.php`.
+  Setting this to `null` means clients have to explicitly ask for the count.
   """
   defaultCount: Int
 
   """
   Limit the maximum amount of items that clients can request from paginated lists.
   Overrules the `pagination.max_count` setting from `lighthouse.php`.
+  Setting this to `null` means the count is unrestricted.
   """
   maxCount: Int
 
@@ -1496,12 +1498,14 @@ directive @hasMany(
   """
   Allow clients to query paginated lists without specifying the amount of items.
   Overrules the `pagination.default_count` setting from `lighthouse.php`.
+  Setting this to `null` means clients have to explicitly ask for the count.
   """
   defaultCount: Int
 
   """
   Limit the maximum amount of items that clients can request from paginated lists.
   Overrules the `pagination.max_count` setting from `lighthouse.php`.
+  Setting this to `null` means the count is unrestricted.
   """
   maxCount: Int
 
@@ -1586,12 +1590,14 @@ directive @hasManyThrough(
   """
   Allow clients to query paginated lists without specifying the amount of items.
   Overrules the `pagination.default_count` setting from `lighthouse.php`.
+  Setting this to `null` means clients have to explicitly ask for the count.
   """
   defaultCount: Int
 
   """
   Limit the maximum amount of items that clients can request from paginated lists.
   Overrules the `pagination.max_count` setting from `lighthouse.php`.
+  Setting this to `null` means the count is unrestricted.
   """
   maxCount: Int
 
@@ -1967,12 +1973,14 @@ directive @morphMany(
   """
   Allow clients to query paginated lists without specifying the amount of items.
   Overrules the `pagination.default_count` setting from `lighthouse.php`.
+  Setting this to `null` means clients have to explicitly ask for the count.
   """
   defaultCount: Int
 
   """
   Limit the maximum amount of items that clients can request from paginated lists.
   Overrules the `pagination.max_count` setting from `lighthouse.php`.
+  Setting this to `null` means the count is unrestricted.
   """
   maxCount: Int
 
@@ -2103,12 +2111,14 @@ directive @morphToMany(
   """
   Allow clients to query paginated lists without specifying the amount of items.
   Overrules the `pagination.default_count` setting from `lighthouse.php`.
+  Setting this to `null` means clients have to explicitly ask for the count.
   """
   defaultCount: Int
 
   """
   Limit the maximum amount of items that clients can request from paginated lists.
   Overrules the `pagination.max_count` setting from `lighthouse.php`.
+  Setting this to `null` means the count is unrestricted.
   """
   maxCount: Int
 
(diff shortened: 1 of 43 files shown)


pyrou:
@spawnia due to breaking change involved, should this PR target lighthouse 6 ?

docs/5/api-reference/directives.md
```diff
@@ -249,12 +249,14 @@ directive @belongsToMany(
   """
   Allow clients to query paginated lists without specifying the amount of items.
   Overrules the `pagination.default_count` setting from `lighthouse.php`.
+  Setting this to `null` means clients have to explicitly ask for the count.
```
pyrou, line 252:
```suggestion
  When a value is defined in `pagination.default_count` settings, setting this value to `null` in directive means clients have to explicitly ask for the count for this specific pagination.
```

src/Pagination/PaginateDirective.php
```diff
@@ -164,13 +166,11 @@ protected function paginationType(): PaginationType
 
     protected function defaultCount(): ?int
     {
-        return $this->directiveArgValue('defaultCount')
-            ?? config('lighthouse.pagination.default_count');
+        return $this->directiveArgValue('defaultCount', config('lighthouse.pagination.default_count'));
```
spawnia, line 169: I must admit that I changed practically all calls to `directiveArgValue` to use `??` for defaults some time ago. Now, I recognize this was the wrong thing to do, as it disallows explicitly set `null`.

I documented the method, can you change other calls back to providing `$default`?

docs/5/api-reference/directives.md
spawnia: I don't think this longer explanation provides much value, the precedence is explained by the line above.

src/Pagination/PaginateDirective.php
pyrou: I can't see any other directiveArgValue ?? usage where a null value would be preferable than the actual ?? fallback

```graphql
@guard(with: null)
@auth(guard: null)
@whereAuth(guard: null)
@cache(private: null)
@globalId(decode: null)
@orderBy(direction: null)
@like/@eq/@where/... (key:null)
@where (operator:null, clause:null)
...
```

None of them make sense to me. I mean, I don't expect particular behavior compared to not setting attribute at all.

---

BUT, as it looks like keeping null value (like here for maxCount, and defaultCount) is really uncommon, possibly we could change implementation of directiveArgValue to return fallback when null value are provided in schema. And add a new method hasDirectiveArg($name) to handle maxCount and defaultCount

```php
    return $this->hasDirectiveArg('defaultCount')
        ? $this->directiveArgValue('defaultCount')
        : config('lighthouse.pagination.default_count'));
```

And so change all others `directiveArgValue($name) ?? $default` to `directiveArgValue($name, $default)`

spawnia: > And so change all others directiveArgValue($name) ?? $default to directiveArgValue($name, $default)

Given you checked the other use cases, it might not be necessary, but I also see no downside to doing just that. I would always expect explicit `null` to be preserved over any defaults that might exist.

spawnia: > And add a new method hasDirectiveArg($name) to handle maxCount and defaultCount

`directiveHasArgument()` exists and is used in some places - sometimes it makes sense - but the `$default` argument handles the case you described just fine.

pyrou: > it might not be necessary, but I also see no downside to doing just that
Doing this may introduce some new breaking change.

Let's take example of clause argument on where directive.

```php
        $clause = $this->directiveArgValue('clause') ?? 'where';
        return $builder->{$clause}(...)
```

Let's then suppose the following weird notation in graphQL schema :

```graphql
@where(clause: null)
```

This notation is pretty weird, but it "works" today, clause is set to "where", thanks to ?? 'where'.

If we change all other directiveArgValue($name) ?? $default to directiveArgValue($name, $default), the notation above will end up with a PHP Error: Method name must be a string.

spawnia: I am fine with breaking nonsensical uses of explicit `null`. In the future `webonyx/graphql-php` will gain the ability to validate the correctness of schema directive usage statically.

pyrou: I re-read carefully [GraphQL specification #sec-Type-System.Non-Null](http://spec.graphql.org/June2018/#sec-Type-System.Non-Null). I think I was wrong with this pull request.

> Inputs (such as field arguments), are always optional by default. However a non‐null input type is required. In addition to not accepting the value null, it also does not accept omission. For the sake of simplicity nullable types are always optional and non‐null types are always required.

it is not 100% clear if an omitted attribute is conceptually equivalent or not to attribute: null. But maybe doing this differentiation is not correct. If conceptually `maxCount: null` is equivalent of no maxCount parameter at all (which fallback to settings) this PR should NOT change the current behavior.

This being said. We are speaking here about the server schema, not the exposed GraphQL server. Following strictly "by the book" the GraphQL spec for server schema might not be a requirement.

spawnia: > it is not 100% clear if an omitted attribute is conceptually equivalent or not to attribute: null

PR #2199, opened 2022-09-02
Use a cache key prefix.

Saad89:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
Use a cache key prefix to generate new cache with every deployment

**Breaking changes**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 You can find and compare releases at the [GitHub release page](https://github.com/nuwave/lighthouse/releases).
 
 ## Unreleased
+Use a cache key prefix to generate new cache with every deployment
 
 ## v5.58.2
 
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -223,7 +223,7 @@ class GraphQL
         $store = $cacheFactory->store($cacheConfig['store']);
 
         return $store->remember(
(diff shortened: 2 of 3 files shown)


kingtigg:
imo this is a great change.

spawnia:
I don't see why that would be necessary, given the cached value is keyed by its hash value. Please explain if there actual is a practical issue you are solving with this, not just a theoretical concern.

Saad89:
> I don't see why that would be necessary, given the cached value is keyed by its hash value. Please explain if there actually is a practical issue you are solving with this, not just a theoretical concern.

- The idea behind this is to have isolated cache for query every deployment.
- An example case a query that have a resolver and cached , Then within next release the query changed with a new the resolver.

spawnia:
I think you do not fully understand what the query cache is, and you do not understand the determistic nature of a hash function. No matter what changes on the server, the query cache stores only documents sent by clients. If the query changes, it will have a different hash.

k0ka:
> An example case a query that have a resolver and cached , Then within next release the query changed with a new the resolver.

Query cache doesn't cache any resolvers. It only caches how a string query (that came from a client) is parsed to an object tree.

Saad89:
> I think you do not fully understand what the query cache is, and you do not understand the determistic nature of a hash function. No matter what changes on the server, the query cache stores only documents sent by clients. If the query changes, it will have a different hash.

Yes, i checked again and it saves only client query at the end.

PR #2202, opened 2022-09-07
Open EnsureXHR middleware for extension

Deyjandi:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR makes the ```handle``` method of the ```EnsureXHR``` class to inherit ```FORM_CONTENT_TYPES``` const (when EnsureXHR class is extended).

Use case:
By extending the ```EnsureXHR``` class, one can exclude any content-type (like multipart/form-data) without the need to override the middleware's handle method. 

<details>
  <summary>See example code</summary>

```
...

class EnsureXHR extends \Nuwave\Lighthouse\Support\Http\Middleware\EnsureXHR
{
    public const FORM_CONTENT_TYPES = [
        'application/x-www-form-urlencoded',
        'text/plain',
    ];

    public function handle(Request $request, Closure $next)
    {
        return parent::handle($request, $next);
    }
}

```
</details>

**Breaking changes**

### DIFF
--- a/src/Support/Http/Middleware/EnsureXHR.php
+++ b/src/Support/Http/Middleware/EnsureXHR.php
@@ -48,7 +48,7 @@ class EnsureXHR
             throw new BadRequestHttpException('Content-Type header must be set');
         }
 
-        if (Str::startsWith($contentType, self::FORM_CONTENT_TYPES)) {
+        if (Str::startsWith($contentType, static::FORM_CONTENT_TYPES)) {
             throw new BadRequestHttpException("Content-Type $contentType is forbidden");
         }
 


spawnia:
I don't think this middleware does what it is supposed to do when changing the content types, but I believe in extensibility and freedom - even if that means offering possible footguns. Godspeed.

k0ka:
Uhm, really. @Deyjandi what do you want to change in the `FORM_CONTENT_TYPES`? They are not just random, they are chosen to ensure that request is not forged by any external site. Changing them without deep understanding of http security will cause your application to be vulnerable to CSRF attacks.

Deyjandi:
> Uhm, really. @Deyjandi what do you want to change in the `FORM_CONTENT_TYPES`? They are not just random, they are chosen to ensure that request is not forged by any external site. Changing them without deep understanding of http security will cause your application to be vulnerable to CSRF attacks.

I want to whitelist the multipart/form-data requests because I can't find other solution for uploading files without using a 3rd party service. My solution is to allow multipart/form-data requests (while having sanctum's CSRF protection).

The [documentation](https://lighthouse-php.com/master/digging-deeper/file-uploads.html#handling-file-uploads) suggests using multipart/form-data.

Is there any other alternative solution for uploading files without using multipart/form-requests?

I also agree w/ @spawnia's point. Maybe this PR was a mistake.

k0ka:
Here is how I handle it:

1. copy `EnsureXHR` to your project
2. change the last lines (that uses `FORM_CONTENT_TYPES`) to:

```php
        if (!Str::startsWith($contentType, self::FORM_CONTENT_TYPES)) {
            return $next($request);
        }

        if (Str::startsWith($contentType, 'multipart/form-data')
            && $request->header('X-My-GraphQL') === '1'
        ) {
            return $next($request);
        }

        throw new BadRequestHttpException("Content-Type $contentType is forbidden");
```
3. The client must set this `X-My-GraphQL` header for upload request. 

But of course it's better to use [Laravel internal tools](https://laravel.com/docs/9.x/csrf) or [Laravel Sanctum](https://laravel.com/docs/9.x/sanctum)

spawnia:
Side note: adding the `X` prefix to custom headers is no longer recommended - see https://datatracker.ietf.org/doc/html/rfc6648

k0ka:
We might name the header like `Lighthouse-GraphQL-CSRF` and update the middleware/documentation

spawnia:
Let's continue the discussion in https://github.com/nuwave/lighthouse/issues/2206.

PR #2205, opened 2022-09-08
Allow configuring `security.disable_introspection` through the env variable `LIGHTHOUSE_SECURITY_DISABLE_INTROSPECTION`

Deyjandi:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Allow configuring the disable_introspection setting via the .env file.

I find this feature useful because i like using introspection on local environments only.
On testing/production I disable introspection, so having the ability to configure introspection from the  ```.env``` file might be useful.

**Breaking changes**
none

### DIFF
--- a/src/lighthouse.php
+++ b/src/lighthouse.php
@@ -193,7 +193,9 @@ return [
     'security' => [
         'max_query_complexity' => \GraphQL\Validator\Rules\QueryComplexity::DISABLED,
         'max_query_depth' => \GraphQL\Validator\Rules\QueryDepth::DISABLED,
-        'disable_introspection' => \GraphQL\Validator\Rules\DisableIntrospection::DISABLED,
+        'disable_introspection' => (bool) env('LIGHTHOUSE_DISABLE_INTROSPECTION', false)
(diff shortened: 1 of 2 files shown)


src/lighthouse.php
```diff
@@ -193,7 +193,9 @@
     'security' => [
         'max_query_complexity' => \GraphQL\Validator\Rules\QueryComplexity::DISABLED,
         'max_query_depth' => \GraphQL\Validator\Rules\QueryDepth::DISABLED,
-        'disable_introspection' => \GraphQL\Validator\Rules\DisableIntrospection::DISABLED,
+        'disable_introspection' => (bool) env('LIGHTHOUSE_DISABLE_INTROSPECTION', false)
```
spawnia, line 196:
```suggestion
        'disable_introspection' => (bool) env('LIGHTHOUSE_SECURITY_DISABLE_INTROSPECTION', false)
```

spawnia (changes requested):
Please mention this in `CHANGELOG.md`

PR #2220, opened 2022-10-27
Improve handling `ModelNotFoundException`

bastiaandewaele:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md (but the numbering still needs to be fixed)

This links to a conversion at [Slack](https://lighthouse-php.slack.com/archives/CB28A070S/p1666092971554689).

**Changes**

This pull request removes the try catch for a model not found.

https://github.com/nuwave/lighthouse/blob/f46b1f4556d29cd7db486c66feef8ce9bebac9ee/src/Auth/CanDirective.php#L201-L212

I have removed the try catch, which means `ModelNotFoundException` thrown from a custom mutation will now also be captured with the new error handler.

While before a custom mutation:

```php
class UpdateArticleContent {
  pubic function __invoke(...) {
    $article = Article::query()->findOrFail();
  }
}
```

Would result in a error response like below:

```json
{
  "errors": [
    {
      "debugMessage": "No query results for model [App\\Article] 500",
      "extensions": {
        "category": "internal"
      }
    }
  ]
}
```

And when debug is disabled, the response was non descriptive and handled as an internal server error:

```json
{
  "errors": [
    {
      "extensions": {
        "category": "internal"
      }
    }
  ],
  "data": {
    "article": null
  }
}
```

With the new changes:

```json
{
  "errors": [
    {
      "message": "No query results for model [App\\Article] -1",
      "extensions": {
        "category": "not_found"
      }
    }
  ],
  "data": {
    "article": null
  }
}
```

Also the category has been changed from "graphql" into "not_found" to make it easier, to detect if content doesn't exist.

**Breaking changes**

Not sure ...

### DIFF
--- a/src/Execution/ModelNotFoundErrorHandler.php
+++ b/src/Execution/ModelNotFoundErrorHandler.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Nuwave\Lighthouse\Execution;
+
+use Closure;
+use GraphQL\Error\Error;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+
+/**
+ * Wrap model not found exceptions.
+ */
+class ModelNotFoundErrorHandler implements ErrorHandler
+{
+    public function __invoke(?Error $error, Closure $next): ?array
+    {
+        if (null === $error) {
+            return $next(null);
+        }
+
+        $previous = $error->getPrevious();
+
+        if ($previous instanceof ModelNotFoundException) {
+            return $next(new Error($previous->getMessage()));
+        }
+
+        return $next($error);
+    }
+}
--- a/src/Execution/ModelNotFoundErrorHandler.php
+++ b/src/Execution/ModelNotFoundErrorHandler.php
@@ -20,7 +20,7 @@ class ModelNotFoundErrorHandler implements ErrorHandler
         $previous = $error->getPrevious();
 
         if ($previous instanceof ModelNotFoundException) {
-            return $next(new Error($previous->getMessage()));
+            return $next(new Error($previous->getMessage(), null, null, [], null, null, ['category' => 'not_found']));
         }
 
         return $next($error);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,6 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
-## v5.6x.0
-
 ### Fixed
 
 - Fix handle model not found thrown inside a custom mutation.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v5.6x.0
+
+### Fixed
+
+- Fix handle model not found thrown inside a custom mutation.
+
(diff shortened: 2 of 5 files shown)


src/Execution/ModelNotFoundErrorHandler.php
```diff
@@ -0,0 +1,28 @@
...
+    public function __invoke(?Error $error, Closure $next): ?array
+    {
+        if (null === $error) {
+            return $next(null);
+        }
+
+        $previous = $error->getPrevious();
+
+        if ($previous instanceof ModelNotFoundException) {
+            return $next(new Error($previous->getMessage(), null, null, [], null, null, ['category' => 'not_found']));
```
spawnia, line 23:
```suggestion
            return $next(new Error($previous->getMessage(), null, null, [], null, $previous));
```

1. Include `$previous`
2. `category` will be removed in the next major version anyways

src/Execution/ModelNotFoundErrorHandler.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Nuwave\Lighthouse\Execution;
+
+use Closure;
+use GraphQL\Error\Error;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+
+/**
+ * Wrap model not found exceptions.
+ */
+class ModelNotFoundErrorHandler implements ErrorHandler
```
spawnia, line 12: I am not convinced that any `ModelNotFoundException` ever thrown is generally safe to show to users. What if it is not a model the user even queried, but rather just something happening in the background that the user is not supposed to know about? This handler captures anything that happens inside any query or mutation.

spawnia (changes requested):

src/Execution/ModelNotFoundErrorHandler.php
bastiaandewaele: One minor question about removing `category` in next version, will there be a replacement to check for different types of errors?

I use for example the `category` **validation** to detect validation errors and show them differently than other errors.

spawnia: https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md#breaking-removed-error-extension-field-category

I am not going to re-add it in Lighthouse.

bastiaandewaele: No problem I understand, I'll remove it from my implementation :)

src/Execution/ModelNotFoundErrorHandler.php
spawnia: > just removing the try catch inside the canDirective

How would that be better then the previous behaviour? I am wondering if we should even throw if no models are found at all - would it be so bad to continue with an empty result?

Recently, there has been a proposal to change the behaviour of `@can` in regards to not found models - see https://github.com/nuwave/lighthouse/issues/2225. Maybe we can work out an optimal solution there.

6XGate:
I suspect this should really target the `RecordsNotFoundException` exception for more general handling of missing data.

spawnia:
Closing as I am suspicious this is an improvement, see https://github.com/nuwave/lighthouse/pull/2220#discussion_r1007741313.

PR #2232, opened 2022-11-17
Support returning data in a `Paginator` from option `resolver` in `@paginator` directive

Stevemoretz:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md

There is no user facing changes, this doesn't impact backward compatibility but adds a feature which I honestly thought it was included before I tried, then I saw it wasn't.

We can decide on adding an example to doc for it or not, but I didn't think it was a good idea to do it myself.

**Changes**

No changes...
You can now return an instance of LengthAwarePaginator from the builder of your paginate directive which would bypass all the builders, (I use this myself to paginate a simple array). eg:

```php
return new LengthAwarePaginator([
            [
                "ID" => "test",
                "display_name" => "ok",
                "arr" => [
                    "s" => 's'
                ]
            ],
            [
                "ID" => "test2",
                "display_name" => "ok",
                "arr" => [
                    "s" => 's'
                ]
            ],
        ],2, 15);
```

**Breaking changes**

No breaking changes.

### DIFF
--- a/docs/5/api-reference/directives.md
+++ b/docs/5/api-reference/directives.md
@@ -2770,6 +2770,8 @@ class Blog
 
 You can provide your own function to return paginated data instead of using eloquent.
 
+This is mutually exclusive with builder and model and other options such as scopes will have no impact when it is used.
+
 ```graphql
 type Query {
   posts: [Post!]!
--- a/docs/5/api-reference/directives.md
+++ b/docs/5/api-reference/directives.md
@@ -21,13 +21,13 @@ directive @aggregate(
 
   """
   The relationship with the column to aggregate.
-  Mutually exclusive with the `model` argument.
+  Mutually exclusive with `model`.
   """
   relation: String
 
   """
   The model with the column to aggregate.
-  Mutually exclusive with the `relation` argument.
+  Mutually exclusive with `relation`.
   """
   model: String
 
@@ -821,13 +821,13 @@ Returns the count of a given relationship or model.
 directive @count(
   """
   The relationship to count.
-  Mutually exclusive with the `model` argument.
+  Mutually exclusive with `model`.
   """
   relation: String
 
   """
   The model to count.
-  Mutually exclusive with the `relation` argument.
+  Mutually exclusive with `relation`.
   """
   model: String
 
@@ -2255,7 +2255,7 @@ directive @node(
   Consists of two parts: a class name and a method name, seperated by an `@` symbol.
   If you pass only a class name, the method name defaults to `__invoke`.
 
-  Mutually exclusive with the `model` argument.
+  Mutually exclusive with `model`.
   """
   resolver: String
 
@@ -2263,7 +2263,7 @@ directive @node(
   Specify the class name of the model to use.
   This is only needed when the default model detection does not work.
 
-  Mutually exclusive with the `model` argument.
+  Mutually exclusive with `resolver`.
   """
   model: String
 ) on OBJECT
@@ -2334,15 +2334,15 @@ directive @orderBy(
   """
   Restrict the allowed column names to a well-defined list.
   This improves introspection capabilities and security.
-  Mutually exclusive with the `columnsEnum` argument.
+  Mutually exclusive with `columnsEnum`.
   Only used when the directive is added on an argument.
   """
   columns: [String!]
 
   """
   Use an existing enumeration type to restrict the allowed columns to a predefined list.
-  This allowes you to re-use the same enum for multiple fields.
-  Mutually exclusive with the `columns` argument.
+  This allows you to re-use the same enum for multiple fields.
+  Mutually exclusive with `columns`.
   Only used when the directive is added on an argument.
   """
   columnsEnum: String
@@ -2386,21 +2386,21 @@ Options for the `relations` argument on `@orderBy`.
 """
 input OrderByRelation {
   """
-  TODO: description
+  Name of the relation.
   """
   relation: String!
 
   """
   Restrict the allowed column names to a well-defined list.
   This improves introspection capabilities and security.
-  Mutually exclusive with the `columnsEnum` argument.
+  Mutually exclusive with `columnsEnum`.
   """
   columns: [String!]
 
   """
   Use an existing enumeration type to restrict the allowed columns to a predefined list.
-  This allowes you to re-use the same enum for multiple fields.
-  Mutually exclusive with the `columns` argument.
+  This allows you to re-use the same enum for multiple fields.
+  Mutually exclusive with `columns`.
   """
   columnsEnum: String
 }
@@ -2433,7 +2433,7 @@ directive @paginate(
   builder: String
 
   """
-  Reference a function that resolves the field by directly returning data in a `LengthAwarePaginator` instance.
+  Reference a function that resolves the field by directly returning data in a Paginator instance.
   Mutually exclusive with `builder` and `model`. Not compatible with `scopes` and builder arguments such as `@eq`.
   Consists of two parts: a class name and a method name, seperated by an `@` symbol.
   If you pass only a class name, the method name defaults to `__invoke`.
@@ -2775,8 +2775,7 @@ This is mutually exclusive with `builder` and `model`. Not compatible with `scop
 
 ```graphql
 type Query {
-  posts: [Post!]!
-    @paginate(resolver: "App\\Resolver\\Posts@userPosts")
+  posts: [Post!]! @paginate(resolver: "App\\Resolver\\Posts@userPosts")
 }
 ```
 
@@ -2793,10 +2792,12 @@ class Posts {
         //...apply your logic
         return new LengthAwarePaginator([
             [
-                "ID" => 1,
+                'id' => 1,
+                'title' => 'Flying teacup found in solar orbit',
             ],
             [
-                "ID" => 2,
+                'id' => 2,
+                'title' => 'What actually is the difference between cookies and biscuits?',
             ],
         ], 2, 15);
     }
--- a/docs/5/api-reference/directives.md
+++ b/docs/5/api-reference/directives.md
@@ -2433,6 +2433,13 @@ directive @paginate(
   builder: String
 
   """
+  Reference a function to customize the resolving the paginator by returning a LengthAwarePaginator instance from it.
+  Consists of two parts: a class name and a method name, seperated by an `@` symbol.
+  If you pass only a class name, the method name defaults to `__invoke`.
+  """
+  resolver: String
+
+  """
   Apply scopes to the underlying query.
   """
   scopes: [String!]
@@ -2757,6 +2764,39 @@ class Blog
             ->groupBy(...);
     }
 }
+```
+
+### Custom resolver
+
+You can provide your own function to return paginated data instead of using eloquent.
+
+```graphql
+type Query {
+  posts: [Post!]!
+    @paginate(resolver: "App\\Resolver\\Posts@userPosts")
+}
+```
+
+A custom resolver function may look like the following.
+
+```php
+namespace App\Resolver;
+
+use Illuminate\Pagination\LengthAwarePaginator;
+
+class Posts {
+    public function userPosts($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Builder
+    {
+        //...apply your logic
+        return new LengthAwarePaginator([
+            [
+                "ID" => 1,
+            ],
+            [
+                "ID" => 2,
+            ],
+        ], 2, 15);
+    }
 ```
 
 ## @rename
--- a/docs/5/api-reference/directives.md
+++ b/docs/5/api-reference/directives.md
@@ -2769,25 +2769,32 @@ class Blog
 
 ### Custom resolver
 
-You can provide your own function that resolves the field by directly returning data in a `LengthAwarePaginator` instance.
+You can provide your own function that resolves the field by directly returning data in a `\Illuminate\Contracts\Pagination\Paginator` instance.
 
 This is mutually exclusive with `builder` and `model`. Not compatible with `scopes` and builder arguments such as `@eq`.
 
 ```graphql
 type Query {
-  posts: [Post!]! @paginate(resolver: "App\\Resolver\\Posts@userPosts")
+  posts: [Post!]! @paginate(resolver: "App\\GraphQL\\Queries\\Posts")
 }
 ```
 
-A custom resolver function may look like the following.
+A custom resolver function may look like the following:
 
 ```php
-namespace App\Resolver;
+namespace App\GraphQL\Queries;
 
 use Illuminate\Pagination\LengthAwarePaginator;
 
-class Posts {
-    public function userPosts($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Builder
+final class Posts
+{
+    /**
+     * @param  null  $root Always null, since this field has no parent.
+     * @param  array{}  $args The field arguments passed by the client.
+     * @param  \Nuwave\Lighthouse\Support\Contracts\GraphQLContext  $context Shared between all fields.
+     * @param  \GraphQL\Type\Definition\ResolveInfo  $resolveInfo Metadata for advanced query resolution.
+     */
+    public function __invoke($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): LengthAwarePaginator
     {
         //...apply your logic
         return new LengthAwarePaginator([
@@ -2801,6 +2808,7 @@ class Posts {
             ],
         ], 2, 15);
     }
+}
 ```
 
 ## @rename
--- a/docs/5/api-reference/directives.md
+++ b/docs/5/api-reference/directives.md
@@ -2433,7 +2433,8 @@ directive @paginate(
   builder: String
 
   """
-  Reference a function to customize the resolving the paginator by returning a LengthAwarePaginator instance from it.
+  Reference a function that resolves the field by directly returning data in a `LengthAwarePaginator` instance.
+  Mutually exclusive with `builder` and `model`. Not compatible with `scopes` and builder arguments such as `@eq`.
   Consists of two parts: a class name and a method name, seperated by an `@` symbol.
   If you pass only a class name, the method name defaults to `__invoke`.
   """
@@ -2768,9 +2769,9 @@ class Blog
 
 ### Custom resolver
 
-You can provide your own function to return paginated data instead of using eloquent.
+You can provide your own function that resolves the field by directly returning data in a `LengthAwarePaginator` instance.
 
-This is mutually exclusive with builder and model and other options such as scopes will have no impact when it is used.
+This is mutually exclusive with `builder` and `model`. Not compatible with `scopes` and builder arguments such as `@eq`.
 
 ```graphql
 type Query {
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -153,11 +153,8 @@ class ArgumentSet
     {
         $argumentSet->directives
             ->filter(Utils::instanceofMatcher(FieldBuilderDirective::class))
-            ->each(static function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder, $argumentSet): void {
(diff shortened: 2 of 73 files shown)


spawnia:
I am not against an option to return a paginator directly, I see how the schema transformation of `@paginate` is still useful. However, using `builder` for this seems wrong to me, I would rather see this as a new option - probably just called `resolver`. Then, the directive definition and docs can clearly describe that this option is mutually exclusive with `builder` and `model`, and that other options such as `scopes` will have no impact when it is used.

Stevemoretz:
> I am not against an option to return a paginator directly, I see how the schema transformation of `@paginate` is still useful. However, using `builder` for this seems wrong to me, I would rather see this as a new option - probably just called `resolver`. Then, the directive definition and docs can clearly describe that this option is mutually exclusive with `builder` and `model`, and that other options such as `scopes` will have no impact when it is used.

Hello, thanks for responding to this PR. I'm new to lighthouse that's why this took sometime to do, the idea of using resolver instead of builder was great, I also had an idea to set the resolver as a default if App\Query\QueryName was present but needed to ask you wether you think that's a good idea or not? And if yes, a little guidance would be nice I don't know how to get the QueryName in App\Query\QueryName, and I'm pretty sure you got a helper method written for it.

Stevemoretz:
Was going to try to implement nested support for it but amazingly it already got supported with what I wrote yesterday!

Just sharing a simple example:

```graphql
type Query {
    sum: D @paginate(resolver: "App\\GraphQL\\Queries\\Sum")
}

type D {
    ID: String
    display_name: String
    arr: S @paginate(resolver: "App\\GraphQL\\Queries\\Sum")
}

type S{
    s: String
    nested: T @paginate(resolver: "App\\GraphQL\\Queries\\Sum")
}

type T{
    s: String
}
```

./app/GraphQL/Queries/Sum.php

```php
<?php

use Illuminate\Pagination\LengthAwarePaginator;
use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;

final class Sum
{
    /**
     * @param  null  $_
     * @param  array{}  $args
     */
    public function __invoke($_, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
    {
        if($_){
            if($_["arr"] && count($resolveInfo->path) === 4){
                return $_["arr"];
            }

            if( $_["nested"] && count($resolveInfo->path) === 7){
                return $_["nested"];
            }
        }
        return new LengthAwarePaginator([
            [
                "ID" => "test",
                "display_name" => "ok",
                "arr" => new LengthAwarePaginator([
                    ["s" => 's', "nested" => new LengthAwarePaginator([
                        ["s" => 's2']
                    ],1,10)]
                ],1,10)
            ],
            [
                "ID" => "test2",
                "display_name" => "ok",
                "arr" => new LengthAwarePaginator([
                    ["s" => 's',"nested" => new LengthAwarePaginator([
                        ["s" => 's2']
                    ],1,10)]
                ],1,10)
            ],
        ],15, 15);
    }
}
```

----

Query:

```graphql
{
  sum(first: 50) {
    data {
      ID
      arr {
        paginatorInfo{
          count
        }
        data{
          s,
          nested{
            paginatorInfo{
              count
              currentPage
            }
            data{
              s
            }
          }
        }
      }
    }
  }
}
```

result:

```json
{
  "data": {
    "sum": {
      "data": [
        {
          "ID": "test",
          "arr": {
            "paginatorInfo": {
              "count": 1
            },
            "data": [
              {
                "s": "s",
                "nested": {
                  "paginatorInfo": {
                    "count": 1,
                    "currentPage": 1
                  },
                  "data": [
                    {
                      "s": "s2"
                    }
                  ]
                }
              }
            ]
          }
        },
        {
          "ID": "test2",
          "arr": {
            "paginatorInfo": {
              "count": 1
            },
            "data": [
              {
                "s": "s",
                "nested": {
                  "paginatorInfo": {
                    "count": 1,
                    "currentPage": 1
                  },
                  "data": [
                    {
                      "s": "s2"
                    }
                  ]
                }
              }
            ]
          }
        }
      ],
      "ID": 1
    }
  }
}
```

docs/5/api-reference/directives.md
```diff
@@ -2432,6 +2432,13 @@ directive @paginate(
   """
   builder: String
 
+  """
```
spawnia, line 2435: Update the `master` docs instead, see https://github.com/nuwave/lighthouse/blob/master/docs/.github/README.md#documenting-new-features

src/Pagination/PaginateDirective.php
```diff
@@ -43,6 +44,14 @@ public static function definition(): string
   """
   builder: String
 
+  """
+  Reference a function to customize the resolving the paginator by returning directly a LengthAwarePaginator instance.
+  This is mutually exclusive with builder and model and other options such as scopes will have no impact when it is used.
```
spawnia, line 49:
```suggestion
  Reference a function that resolves the field by directly returning data in a `LengthAwarePaginator` instance.
  Mutually exclusive with `builder` and `model`. Not compatible with  `scopes` and builder arguments such as `@eq`.
```

src/Pagination/PaginateDirective.php
```diff
@@ -89,7 +98,9 @@ public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefini
     {
         $paginationManipulator = new PaginationManipulator($documentAST);
 
-        if ($this->directiveHasArgument('builder')) {
+        if ($this->directiveHasArgument('resolver')) {
+            $this->getResolverFromArgument('resolver');
+        } elseif ($this->directiveHasArgument('builder')) {
```
spawnia, line 103: This is fine, but let's also validate that only one of `model`, `builder` and `resolver` is used, else throw a `DefinitionException` indicating that they are mutually exclusive.

src/Pagination/PaginateDirective.php
```diff
@@ -43,6 +44,14 @@ public static function definition(): string
   """
   builder: String
```
spawnia, line 45: Let's also amend this and `model` with a hint that they are mutually exclusive with each other.

Stevemoretz:
Thanks for accepting the PR phpStan should be fine now.

src/Pagination/PaginateDirective.php
```diff
@@ -110,10 +121,24 @@ public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefini
     public function resolveField(FieldValue $fieldValue): FieldValue
     {
         $fieldValue->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Paginator {
+            if ($this->directiveHasArgument('resolver')) {
+                $builderResolver = $this->getResolverFromArgument('resolver');
+
+                $query = $builderResolver($root, $args, $context, $resolveInfo);
+
+                assert(
+                    $query instanceof LengthAwarePaginator,
```
spawnia, line 130: Actually, the return value of this method only says `Illuminate\Contracts\Pagination\Paginator` - I think it would not work with that though, given `PaginatorField` and `ConnectionField` expect a `LengthAwarePaginator`. So I think we should probably change the return value?

src/Pagination/PaginateDirective.php
```diff
@@ -110,10 +121,24 @@ public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefini
     public function resolveField(FieldValue $fieldValue): FieldValue
     {
         $fieldValue->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Paginator {
+            if ($this->directiveHasArgument('resolver')) {
+                $builderResolver = $this->getResolverFromArgument('resolver');
+
+                $query = $builderResolver($root, $args, $context, $resolveInfo);
```
spawnia, line 127:
```suggestion
                $paginator = $this->getResolverFromArgument('resolver')($root, $args, $context, $resolveInfo);
```

spawnia: The variable names you used are misleading.

tests/Unit/Pagination/PaginateDirectiveTest.php
```diff
@@ -696,4 +697,62 @@ public function testDisallowFirstNull(): void
...
+        $builder = addslashes(static::class).'@returnPaginatedDataInsteadOfBuilder';
+        $this->schema = $this->buildSchema(/* @lang GraphQL */ "
+            type Query {
+                users: [User] @paginate(resolver: \"{$builder}\")
+            }
+
+            type User {
+                id: ID
+            }
+            ");
```
spawnia, line 725:
```suggestion
        type Query {
            users: [User] @paginate(resolver: \"{$builder}\")
        }

        type User {
            id: ID
        }
        ");
```

tests/Unit/Pagination/PaginateDirectiveTest.php
```diff
@@ -696,4 +697,62 @@ public function testDisallowFirstNull(): void
...
+    }
+
+    public function testThrowsIfResolverIsNotPresent(): void
+    {
+        $this->expectException(DefinitionException::class);
+        $this->expectExceptionMessage('Failed to find class NonexistingClass in namespaces [] for directive @paginate.');
+
+        $this->buildSchema(/** @lang GraphQL */ '
+        type Query {
+            users: [Query!] @paginate(resolver: "NonexistingClass@notFound")
```
spawnia, line 754:
```suggestion
            users: [Query!]! @paginate(resolver: "NonexistingClass@notFound")
```

spawnia (changes requested):

spawnia:
We have to keep `resolver` explicit for now, defaulting to classes in `App\GraphQL\Queries` would be a breaking change.

docs/5/api-reference/directives.md
Stevemoretz: Working on this after changing the master, wanted to revert v5 and run `yarn release` which I ran after yarn and yarn start, but it gives:

```
error Command "release" not found.
```

I checked package.json and there was no scripts for release indeed:

```json
{
    "license": "MIT",
    "dependencies": {
        "@vuepress/plugin-back-to-top": "1.8.2",
        "@vuepress/plugin-medium-zoom": "1.8.2",
        "fs-extra": "9.1.0",
        "vuepress": "1.8.2",
        "vuepress-theme-default-prefers-color-scheme": "1.0.9"
    },
    "devDependencies": {
        "prettier": "2.2.1"
    },
    "scripts": {
        "start": "vuepress dev .",
        "build": "vuepress build .",
        "prettify": "prettier --write --tab-width=2 **/**/*.md **/*.js .vuepress/*.js"
    }
}
```

Stevemoretz: > Update the `master` docs instead, see https://github.com/nuwave/lighthouse/blob/master/docs/.github/README.md#documenting-new-features

Working on this after changing the master, wanted to revert v5 and run `yarn release` which I ran after yarn and yarn start, but it gives:

```
error Command "release" not found.
```

I checked package.json and there was no scripts for release indeed:

```json
{
    "license": "MIT",
    "dependencies": {
        "@vuepress/plugin-back-to-top": "1.8.2",
        "@vuepress/plugin-medium-zoom": "1.8.2",
        "fs-extra": "9.1.0",
        "vuepress": "1.8.2",
        "vuepress-theme-default-prefers-color-scheme": "1.0.9"
    },
    "devDependencies": {
        "prettier": "2.2.1"
    },
    "scripts": {
        "start": "vuepress dev .",
        "build": "vuepress build .",
        "prettify": "prettier --write --tab-width=2 **/**/*.md **/*.js .vuepress/*.js"
    }
}
```

spawnia: Sorry, it is `make release` now. I just fixed the docs.

Stevemoretz: > Sorry, it is `make release` now. I just fixed the docs.

Thank you but is there still any alternatives for not using docker and releasing it?
I'm on MacOS and don't have docker. (Docker is terrible on MacOS...)

spawnia: `make release` requires `make`, not Docker. If you do not have that, just look into the makefile and see what it does - you will find it to be quite trivial 😉

Stevemoretz: > `make release` requires `make`, not Docker. If you do not have that, just look into the makefile and see what it does - you will find it to be quite trivial 😉

Sorry got confused by :

```
If you use Docker you can start up the environment (including docs) by running:

make setup
make node
```

Thought I need these first to run the release, ran it without these and it worked thanks!

src/Pagination/PaginateDirective.php
Stevemoretz: I went through this now seems like `Illuminate\Pagination\LengthAwarePaginator` implements `Illuminate\Contracts\Pagination\LengthAwarePaginator` which extends `Illuminate\Contracts\Pagination\Paginator` so `Illuminate\Pagination\LengthAwarePaginator` is actually of typeof `Illuminate\Contracts\Pagination\Paginator` so I guess it made more sense if I had written:

```php
                assert(
                    $query instanceof Paginator,
                    "The method referenced by the resolver argument of the @{$this->name()} directive on {$this->nodeName()} must return a Paginator."
                );
```

instead, that means the return type doesn't need to be changed tried with `Illuminate\Pagination\Paginator` instead of `Illuminate\Pagination\LengthAwarePaginator` which also worked.

src/Pagination/PaginateDirective.php
Stevemoretz: Copy paste error sorry :) 
Changed to $resolverResolver

src/Pagination/PaginateDirective.php
Stevemoretz: For now I committed it the way I described above if you see anything wrong with it please let me know.

src/Pagination/PaginateDirective.php
Stevemoretz: Done but I'm not sure if you like the message and the fact it's always mentioning model and builder I could make it show it conditionally but didn't want to make the code dirty! 
Also added tests for it.

spawnia: I have had this issue a couple of time, please review my draft of how we can solve it once and for all: https://github.com/nuwave/lighthouse/pull/2233

src/Pagination/PaginateDirective.php
spawnia: We should define the type to be as wide as possible, but as narrow as needed. The downstream resolvers require `Illuminate\Contracts\Pagination\LengthAwarePaginator`, so that is what we should expect.

Stevemoretz: > We should define the type to be as wide as possible, but as narrow as needed. The downstream resolvers require `Illuminate\Contracts\Pagination\LengthAwarePaginator`, so that is what we should expect.

Wasn't aware that, that was used internally so does this look ok?

```php
        $fieldValue->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): LengthAwarePaginator {
            if ($this->directiveHasArgument('resolver')) {
                $paginator = $this->getResolverFromArgument('resolver')($root, $args, $context, $resolveInfo);

                assert(
                    $paginator instanceof LengthAwarePaginator,
                    "The method referenced by the resolver argument of the @{$this->name()} directive on {$this->nodeName()} must return a LengthAwarePaginator."
                );

                return $paginator;
            }
```

src/Pagination/PaginateDirective.php
```diff
@@ -89,7 +98,19 @@ public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefini
...
-        if ($this->directiveHasArgument('builder')) {
+        $directiveHasBuilderArgument = $this->directiveHasArgument('builder');
+        $directiveHasResolverArgument = $this->directiveHasArgument('resolver');
+        $directiveHasModelArgument = $this->directiveHasArgument('model');
+
+        if ($directiveHasResolverArgument && ($directiveHasModelArgument || $directiveHasBuilderArgument)) {
+            throw new DefinitionException(
+                "Argument 'resolver' is mutually exclusive with 'builder' and 'model'."
+            );
+        }
```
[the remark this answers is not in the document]
Stevemoretz: > They are all exclusive with each other, only one may be given. Let's validate this with #2233 once it is finished.

That pull request looks very interesting! Please let me know when it's finished.

> For docs, consider the hints for mutually exclusive arguments in `CanDirective`.

Just to be clear are you talking about this?

```
  """
  Check the policy against the model instances returned by the field resolver.
  Only use this if the field does not mutate data, it is run before checking.

  Mutually exclusive with `query` and `find`.
  """
  resolved: Boolean! = false
```

Which I should use as a base to alter the hint I wrote for `@pagniate`

```
  """
  Reference a function that resolves the field by directly returning data in a `LengthAwarePaginator` instance.
  Mutually exclusive with `builder` and `model`. Not compatible with `scopes` and builder arguments such as `@eq`.
  Consists of two parts: a class name and a method name, seperated by an `@` symbol.
  If you pass only a class name, the method name defaults to `__invoke`.
  """
  resolver: String
```

src/Pagination/PaginateDirective.php
```diff
@@ -110,10 +131,22 @@ public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefini
     public function resolveField(FieldValue $fieldValue): FieldValue
     {
         $fieldValue->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Paginator {
+            if ($this->directiveHasArgument('resolver')) {
+                $paginator = $this->getResolverFromArgument('resolver')($root, $args, $context, $resolveInfo);
+
+                assert(
+                    $paginator instanceof Paginator,
+                    "The method referenced by the resolver argument of the @{$this->name()} directive on {$this->nodeName()} must return a Paginator."
```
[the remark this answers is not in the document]
Stevemoretz: Turned out making the return type LengthAwarePaginator is not a great idea, hen using the same directive without resolver using eloquent I get this error:

```
"debugMessage": "Nuwave\\Lighthouse\\Pagination\\PaginateDirective::Nuwave\\Lighthouse\\Pagination\\{closure}(): Return value must be of type Illuminate\\Contracts\\Pagination\\LengthAwarePaginator, Illuminate\\Pagination\\Paginator returned",
```

src/Pagination/PaginateDirective.php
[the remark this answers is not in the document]
Stevemoretz: Got the tests working locally and LengthAwarePaginator actually fails a few tests, AbstractPaginator passes all tests.

Stevemoretz:
@spawnia I haven't pushed my latest commit that uses `AbstractPaginator` (was waiting for your approval on it), please let me know if you want me to push it since I saw the tests failed but sound like for a different reason (conflict) but anyways... if they didn't fail there they would have later

src/Pagination/PaginateDirective.php
Stevemoretz: @spawnia Used your PR now, everything looks great, thanks! Please check the latest commit.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -153,8 +153,11 @@ protected static function applyFieldBuilderDirectives(self $argumentSet, object
     {
         $argumentSet->directives
             ->filter(Utils::instanceofMatcher(FieldBuilderDirective::class))
-            ->each(static function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder): void {
-                $builder = $fieldBuilderDirective->handleFieldBuilder($builder);
+            ->each(static function (FieldBuilderDirective $fieldBuilderDirective) use (&$builder, $argumentSet): void {
+                // @phpstan-ignore-next-line
+                $fieldBuilderDirective->argsForHandleFieldBuilder = $argumentSet->toArray();
+                // @phpstan-ignore-next-line
+                $builder = $fieldBuilderDirective->handleFieldBuilder($builder, $argumentSet->toArray());
```
spawnia, line 160: Those changes seem unrelated, please clean up.

Stevemoretz: Oops messed how that got there from another branch :)

Stevemoretz:
Changes requested
1 review requesting changes by reviewers with write access. 

@spawnia I can't find that change request, could you mention me there?

PR #2234, opened 2022-11-19
Add args parameter support for @builder field

Stevemoretz:
- [X] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Let's start with an example:

```graphql
posts(search: String, ok : String) : [Post]! @builder(method: "App\\Models\\Post@scoutScopeOk") @paginate
```
So right now we get only one argument in scoutScopeOk function which is the builder, we can add a value to the @builder as well.

```graphql
posts(search: String, ok : String) : [Post]! @builder(method: "App\\Models\\Post@scoutScopeOk",value: "valueHere") @paginate
```

So now we get two arguments in scoutScopeOk function first which is the builder, second is the "valueHere" string it's not so practical, and can be improved in my opinion.

Two ideas to throw out there:

1. What if there was a method to get current args in any resolver i.e. save args, (even context and all 4 params can be saved) and then be accessed, that will fix the issue here as well but since I'm new to this I have no idea how easy or hard that is to implement and I didn't go for it in this PR because I don't know a lot about it, but some global data would be great anyways! Also I would love a function to check for instance the current request is a graphql request or not and stuff like that...
2. Just add args to builder method, this is what this PR does.

This PR targets 100% backward compatibility (I have ran the tests locally), it sounds a little dirty at some parts which we can discuss and fix it specially the parts that I have disabled PHPstan (on purpose so you would see them better), a suggestion is to clean it up together a little and add this to current latest version (v5), and a cleaner implementation (with breaking changes) can be introduced for v6 basically we need to add one more parameter to `handleFieldBuilder` and whatever implemented FieldBuilderDirective (From inside library or custom user directives) should also add that to their method parameters and that's pretty much it, since that would be a breaking change we can pass more data there too maybe context and other stuff if they are useful too?

Another thing for this version, currently I use `$argumentSet->toArray()` for passing the arguments if `$argumentSet` itself does provide more helpful utilities such as getting parent and child params too maybe we should pass that itself instead? (I'm totally new to lighthouse.)

Some other ways to do this:

- do this logic in another interface which has one more parameter and use that one instead, when it's needed which would have no breaking changes even for v6
- apply this logic specifically and only for BuilderDirective only, again no breaking changes and a lot cleaner.

**Breaking changes**

No, but I explained above about this in more depth.


Stevemoretz:
I feel like also adding a `$resolveInfo` can help too, for instance you can optimize the query even further. What do you think?

spawnia:
If we change this, we should just pass all 4 resolver arguments through, that way we will not have to break anything again. It seems to me like doing this properly in v5 is a lot of hassle - I plan to make the cut to start on v6 soon anyway, thus opening the door to breaking changes.

Stevemoretz:
> If we change this, we should just pass all 4 resolver arguments through, that way we will not have to break anything again. It seems to me like doing this properly in v5 is a lot of hassle - I plan to make the cut to start on v6 soon anyway, thus opening the door to breaking changes.

Thanks for reviewing this PR, ok so let me start with another implementation for v6 to start it, let me first verify the main change which is in this interface's signature with you, is this good enough?

```php
<?php

namespace Nuwave\Lighthouse\Support\Contracts;

interface FieldBuilderDirective extends Directive
{
    /**
     * Add additional constraints to the builder.
     *
     * @param object $builder the builder used to resolve the field
     * @param $root
     * @param array $args
     * @param GraphQLContext $context
     * @param ResolveInfo $resolveInfo
     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder the modified builder
     */
    public function handleFieldBuilder(object $builder, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): object;
}
```

spawnia:
Looks about right. We use FQCN and two spaces after `@param` though.

Stevemoretz:
> Looks about right. We use FQCN in PHPDocs and two spaces after `@param` though.

Thanks for responding and the hints.

spawnia:
Passing `ArgumentSet` and `ResolveInfo` is redundant, we can simplify that.

PHPDocs that type as `mixed` need comments.

Stevemoretz:
> Passing `ArgumentSet` and `ResolveInfo` is redundant, we can simplify that.
> 
> PHPDocs that type as `mixed` need comments.

Thanks for the tip on php-docs fixed my problem with php-cs-fixer.

Removed the extra `ArgumentSets` and updated the list on : https://github.com/nuwave/lighthouse/pull/2234#issuecomment-1322284787

Stevemoretz:
@spawnia Didn't get notification about the latest changes you pushed to main, just saw it while checking the repo.
Fixed the small conflict with new changed.

src/Auth/CanDirective.php
```diff
@@ -156,18 +155,23 @@ function ($modelLike) use ($gate, $ability, $checkArguments) {
 
     /**
      * @param  array<string, mixed>  $args
+     * @param  array<string, mixed> $args
```
spawnia, line 158: duplicate

src/Auth/CanDirective.php
```diff
@@ -189,7 +193,7 @@ protected function modelsToCheck(ArgumentSet $argumentSet, array $args): iterabl
                 $queryBuilder->withTrashed();
             }
 
-            $directivesContainsRestore = $argumentSet->directives->contains(
+            $directivesContainsRestore = $resolveInfo->argumentSet->directives->contains(
```
spawnia, line 196: `$resolveInfo->argumentSet->directives` is repeated

src/Support/Contracts/FieldBuilderDirective.php
```diff
@@ -2,14 +2,18 @@
...
      * Add additional constraints to the builder.
      *
      * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder  the builder used to resolve the field
+     * @param  array<string, mixed>  $args
+     * @param  \Nuwave\Lighthouse\Support\Contracts\GraphQLContext  $context
      *
      * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder the modified builder
      */
-    public function handleFieldBuilder(object $builder): object;
+    public function handleFieldBuilder(object $builder, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): object;
```
spawnia, line 18: This is a breaking change and should be listed as such in both CHANGELOG.md and UPGRADE.md

src/Support/Contracts/FieldBuilderDirective.php
```diff
@@ -2,14 +2,18 @@
...
+use GraphQL\Type\Definition\ResolveInfo;
+
 interface FieldBuilderDirective extends Directive
 {
     /**
      * Add additional constraints to the builder.
      *
      * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder  the builder used to resolve the field
+     * @param  array<string, mixed>  $args
+     * @param  \Nuwave\Lighthouse\Support\Contracts\GraphQLContext  $context
```
spawnia, line 14: redundant, only add PHPDocs when they add information

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -79,12 +147,13 @@ public function testCallsCustomBuilderMethodOnFieldWithoutValue(): void
     }
 
     /**
-     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @param  \Illuminate\Database\Query\Builder|Builder  $builder
```
spawnia, line 150: Always use FQCN in PHPDoc

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -79,12 +147,13 @@ public function testCallsCustomBuilderMethodOnFieldWithoutValue(): void
...
+     * @param  int  $value
```
spawnia, line 151: Why not a real type?

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -79,12 +147,13 @@ public function testCallsCustomBuilderMethodOnFieldWithoutValue(): void
...
      *
-     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
+     * @return \Illuminate\Database\Query\Builder|Builder
      */
-    public static function limit(object $builder, int $value = 2): object
+    public static function limit(object $builder, $value): object
     {
-        return $builder->limit($value);
+        return $builder->limit($value ?: 2);
```
spawnia, line 157: Why not a real default?

spawnia (changes requested):

src/Auth/CanDirective.php
Stevemoretz: Made it them both into one variable with the name of `$argumentSetDirectives` if you have a better name in mind please let me know.

src/Support/Contracts/FieldBuilderDirective.php
Stevemoretz: I couldn't get changelog.md to get automatically generated, I tried a couple of tools like: 

But they give me the diff result like:

```diff
4a5,6
> use GraphQL\Type\Definition\ResolveInfo;
>
10a13,14
>      * @param  array<string, mixed>  $args
>      * @param  \Nuwave\Lighthouse\Support\Contracts\GraphQLContext  $context
14c18
<     public function handleFieldBuilder(object $builder): object;
---
>     public function handleFieldBuilder(object $builder, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): object;
```

That's not the format you have used which have + and -, would you mind telling me how to make one like yours?

src/Support/Contracts/FieldBuilderDirective.php
Stevemoretz: Not as redundant as they seem, they fix phpstan errors would you mind giving me some short descriptions on `$args` and `$context` so I can add it and make it more useful for users too?

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
Stevemoretz: I wasn't quite sure `?int $value` is supported on all the php versions you support, added it now thanks.

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
Stevemoretz: Well some arguments are added after it so a default would be useless in `BuilderDirective.php` we have the following now:

```php
    public function handleFieldBuilder(object $builder, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): object
    {
        $resolver = $this->resolver();

        if ($this->directiveHasArgument('value')) {
            return $resolver(
                $builder,
                $this->directiveArgValue('value'),
                $root,
                $args,
                $context,
                $resolveInfo
            );
        }

        return $resolver($builder, null, $root, $args, $context, $resolveInfo);
    }
```

So null is the default, unless we decide to move the `$value` at the end but that means the users need to always define 6 variables in their custom methods.

src/Support/Contracts/FieldBuilderDirective.php
spawnia: It is only necessary for `$args`. Docs are at https://lighthouse-php.com/master/api-reference/resolvers.html#resolver-function-signature

src/Support/Contracts/FieldBuilderDirective.php
spawnia: I hand-write it - it is quite easy to do, just annotate the codeblock as `diff` and put + or - in the beginning of lines.

CHANGELOG.md
```diff
@@ -9,6 +9,33 @@ You can find and compare releases at the [GitHub release page](https://github.co
...
++ use GraphQL\Type\Definition\ResolveInfo;
+
+interface FieldBuilderDirective extends Directive
+{
+    /**
+     * Add additional constraints to the builder.
+     *
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder  the builder used to resolve the field
++     * @param  array<string, mixed>  $args
++     * @param  \Nuwave\Lighthouse\Support\Contracts\GraphQLContext  $context
```
spawnia, line 30: Again, this line specifically does not add anything over the native type.

CHANGELOG.md
```diff
@@ -9,6 +9,33 @@ You can find and compare releases at the [GitHub release page](https://github.co
...
+### Changed
+
+- New signature for `handleFieldBuilder` on `FieldBuilderDirective` interface
+
+```diff
+<?php
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
++ use GraphQL\Type\Definition\ResolveInfo;
```
spawnia, line 21: Missing import for `GraphQLContext`

CHANGELOG.md
```diff
@@ -9,6 +9,33 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+
+- New signature for `handleFieldBuilder` on `FieldBuilderDirective` interface
+
+```diff
+<?php
+
```
spawnia, line 18:
```suggestion
```

CHANGELOG.md
```diff
@@ -9,6 +9,33 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+
+- New signature for `handleFieldBuilder` on `FieldBuilderDirective` interface
+
+```diff
```
spawnia, line 16: Let's just put this in UPGRADE.md, I prefer to limit the changelog to single lines

UPGRADE.md
```diff
@@ -9,6 +9,33 @@ Compare your `lighthouse.php` against the latest [default configuration](src/lig
 
 ## v5 to v6
 
+### New signature for `handleFieldBuilder` on `FieldBuilderDirective` interface
+
+Lighthouse previously, did not pass any data about the resolved node to `handleFieldBuilder` but now it does, this change gives you more control over the field builder.
```
spawnia, line 14:
```suggestion
Lighthouse now passes the typical 4 resolver arguments to `handleFieldBuilder`.
```

CHANGELOG.md
```diff
@@ -9,6 +9,33 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+
+- New signature for `handleFieldBuilder` on `FieldBuilderDirective` interface
```
```suggestion
- Pass resolver arguments to `FieldBuilderDirective::handleFieldBuilder()`
```

spawnia (changes requested):

src/Support/Contracts/FieldBuilderDirective.php
Stevemoretz: @spawnia Thanks for this really helped with the description!

src/Support/Contracts/FieldBuilderDirective.php
Stevemoretz: @spawnia Oh thank you for teaching me about this I thought you auto-generate it somehow!

CHANGELOG.md
Stevemoretz: They're both in the same namespace though, is it necessary to add another use statement for it?

Stevemoretz:
@spawnia Updated docs on master for this

docs/master/api-reference/directives.md
```diff
@@ -486,13 +486,17 @@ type Query {
...
+use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;
 
 class MyClass
 {
-    public function minimumHighscore(Builder $builder, int $minimumHighscore): Builder
+    public function minimumHighscore(Builder $builder, ?int $minimumHighscore, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): Builder
     {
-        return $builder->whereHas('game', static function (Builder $builder) use ($minimumHighscore): void {
-            $builder->where('score', '>', $minimumHighscore);
+        return $builder->when($minimumHighscore !== null, static function(Builder $builder) use ($minimumHighscore): void {
```
spawnia, line 496: Why complicate this by making `minimumHighscore` nullable?

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -148,13 +151,14 @@ static function ($value) use (&$builder, $directiveFilter) {
      * but we must special case that in some way anyhow, as only eq filters can be added on top of search.
      *
      * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder  $builder
+     * @param  array<string, mixed>  $args
      */
-    protected static function applyFieldBuilderDirectives(self $argumentSet, object &$builder): void
+    protected static function applyFieldBuilderDirectives(ResolveInfo $resolveInfo, object &$builder, $root, array $args, GraphQLContext $context): void
```
spawnia, line 156:
```suggestion
    protected static function applyFieldBuilderDirectives(object &$builder, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): void
```

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -83,8 +151,8 @@ public function testCallsCustomBuilderMethodOnFieldWithoutValue(): void
      *
      * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder
      */
-    public static function limit(object $builder, int $value = 2): object
+    public static function limit(object $builder, ?int $value): object
     {
-        return $builder->limit($value);
+        return $builder->limit($value ?: 2);
```
spawnia, line 156:
```suggestion
        return $builder->limit($value ?? 2);
```

CHANGELOG.md
spawnia: I changed the example to be user-facing - it concerns them when they implement custom directives.

docs/master/api-reference/directives.md
Stevemoretz: @spawnia For that specific example where the user is passing `value: 1000` in schema there is no need for it being nullable you are right, I took it a little more general so if the user doesn't pass a value it wouldn't make:

```php
$builder->whereHas('game', static function (Builder $builder) use ($minimumHighscore): void {
    $builder->where('score', '>', $minimumHighscore);
});
```
to get to be :

```php
$builder->whereHas('game', static function (Builder $builder) use ($minimumHighscore): void {
    $builder->where('score', '>', null);
});
```
Just wanted to make it a little more complete no other reasons, we could revert it or add another example to it for instance:

```graphql
type Query {
    users(
        minimumHighscore: Int @builder(method: "App\MyClass@minimumHighscore")
    ): [User!]! @all
    highrankedUsers: [User!]! @all @builder(method: "App\MyClass@minimumHighscore")
}
```

so it would make sense.

Stevemoretz:
@spawnia Sorry :) didn't get notifications about the conflict with your latest commits again, fixed now!
Github should really add a notification for the conflict.

Small conflict resolved, tests ran, all good now 👍

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -74,20 +76,21 @@ public function has(string $key): bool
      * @phpstan-param  TBuilder  $builder
      *
      * @param  array<string>  $scopes
+     * @param  array<string, mixed> $args
```
spawnia, line 79:
```suggestion
     * @param  array<string, mixed>  $args
```

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -32,6 +35,71 @@ public function testCallsCustomBuilderMethod(): void
...
+        $mock->shouldReceive('limit')->once()->withArgs(function (Builder $builder, $value, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) {
+            $this->assertIsNumeric($value);
+            $this->assertEquals(1, $value);
+            $this->assertEquals([
+                'arg1' => 'Hello',
+                'arg2' => 'World',
+            ], $args);
+
+            return true;
+        })->andReturn(User::query());
```
spawnia, line 51:
```suggestion
        $this->app->instance(__CLASS__, $mock);
        $mock->shouldReceive('limit')
            ->once()
            ->withArgs(function (Builder $builder, $value, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) {
                $this->assertSame(1, $value);
                $this->assertSame([
                    'arg1' => 'Hello',
                    'arg2' => 'World',
                ], $args);

                return true;
            })
            ->andReturn(User::query());
```

Stevemoretz:
@spawnia thanks for the suggestions no idea why it (github) didn't run tests after I committed them, so I ran the tests locally and it worked.

docs/master/api-reference/directives.md
spawnia: I figured out the value could be `null` even in the current example - when the client explicitly passed `minimumHighscore: null`. I modified the example.

spawnia:
Thanks @Stevemoretz, this should be good to go when I make the cut towards v6.

Stevemoretz:
> Thanks @Stevemoretz, this should be good to go when I make the cut towards v6.

You're welcome! Thanks for working on it with me 👍

spawnia:

src/Auth/CanDirective.php
```diff
@@ -198,9 +202,13 @@ protected function modelsToCheck(ArgumentSet $argumentSet, array $args): iterabl
             }
 
             try {
-                $enhancedBuilder = $argumentSet->enhanceBuilder(
+                $enhancedBuilder = $resolveInfo->argumentSet->enhanceBuilder(
```
spawnia, line 205:
```suggestion
                $enhancedBuilder = $resolveInfo->enhanceBuilder(
```

src/Execution/ResolveInfo.php
```diff
@@ -42,13 +43,21 @@ public function __construct(
...
      * @phpstan-return TBuilder
      */
-    public function enhanceBuilder(object $builder, array $scopes, \Closure $directiveFilter = null): object
+    public function enhanceBuilder(object $builder, array $scopes, $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo, \Closure $directiveFilter = null): object
     {
-        return $this->argumentSet->enhanceBuilder($builder, $scopes, $directiveFilter);
+        return $this->argumentSet->enhanceBuilder(
+            $builder,
+            $scopes,
+            $directiveFilter,
```
spawnia, line 57: Missing root, order is messed up.

src/Schema/Directives/AggregateDirective.php
```diff
@@ -96,6 +96,21 @@ enum AggregateFunction {
 
     public function resolveField(FieldValue $fieldValue): FieldValue
     {
+        $modelArg = $this->directiveArgValue('model');
+        if (is_string($modelArg)) {
+            $fieldValue->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($modelArg) {
+                /** @var EloquentBuilder $query */
```
spawnia, line 102: use `assert`

src/Schema/Directives/AggregateDirective.php
```diff
@@ -96,6 +96,21 @@ enum AggregateFunction {
...
+                $query = $this
+                    ->namespaceModelClass($modelArg)::query();
```
spawnia, line 104:
```suggestion
                $query = $this->namespaceModelClass($modelArg)
                    ::query();
```

spawnia (changes requested):

phpunit.xml.dist
```diff
@@ -19,19 +17,19 @@
...
             <directory suffix=".php">src</directory>
-        </include>
-    </coverage>
+        </whitelist>
+    </filter>
 
     <php>
         <env name="LIGHTHOUSE_TEST_DB_DATABASE" value="test"/>
-        <env name="LIGHTHOUSE_TEST_DB_HOST" value="mysql"/>
+        <env name="LIGHTHOUSE_TEST_DB_HOST" value="127.0.0.1"/>
```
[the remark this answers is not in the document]
Stevemoretz: My fault that's how I can test it locally didn't mean to push it though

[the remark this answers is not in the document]
spawnia: As mentioned in the link, you can add `phpunit.xml`.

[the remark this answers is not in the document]
Stevemoretz: Thanks for that I was having a hard time working with it :)

src/Schema/Directives/AggregateDirective.php
spawnia: https://www.php.net/manual/de/function.assert.php over `@var`

Stevemoretz: Thanks done and pushed

spawnia, line 102:
```suggestion
```

Stevemoretz: Wouldn't that remove autocomplete of IDE too?

PR #2235, opened 2022-11-21
Support optimized select for top-level query

Lyrisbee:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

I made some updates from https://github.com/nuwave/lighthouse/pull/1626.

Tests passed directives

1. `@all`
2. `@find`
3. `@paginate`

The custom builder and scout builder are not optimized.

```graphql
type Query {
  users: [User!]! @all(builder: "CustomUsersBuilder") // custom builder
  posts: [Post!]! @paginate(builder: "CustomPostsBuilder") // custom builder
  postsSearch(search: String @search): [Post!]! @paginate // scout builder
}
```

Tests passed relation types
1. hasOne
2. hasMany
3. morphOne
4. morphMany
5. belongsTo
6. belongsToMany

Only the root query will be optimized. The nested queries are not optimized. 

```graphql
type User {
  name: String!
  posts: [Post!]! @hasMany
}

type Post {
  id: ID!
  title: String!
}

type Query {
  users: [User!]! @all
}
```

```graphql
{
  users {
    name
    posts {
      id
      title
    }
  }
}
```

```sql
# query logs
# optimized
select `id`, `name` from `users`
# the nested query is not optimized
select * from `posts` where `posts`.`user_id` in (1, 2) and `posts`.`deleted_at` is null
```

**Changes**
Only query the selection fields.

**Breaking changes**
* Custom attribute  needs to use `@select`
```php
public function getCompanyNameAttribute(): string
{
    return $this->company->name;
}
```

```graphql
type User {
    id: ID!
    companyName: String! @select(columns: ["company_id"])
}
```
* The result may change when using optimized select. For predictable results, you should use `@orderBy`
https://dba.stackexchange.com/a/6053

**Note**
This PR was only tested in our own project, and there may be some edge cases that are not considered.


src/ClientDirectives/ClientDirective.php
```diff
@@ -73,7 +73,7 @@ protected function definition(): Directive
 
         $definition = $schema->getDirective($this->name);
         if (! $definition instanceof Directive) {
-            throw new DefinitionException("Missing a schema definition for the client directive $this->name");
+            throw new DefinitionException("Missing a schema definition for the client directive {$this->name}");
```
spawnia, line 76: Please rebase atop the latest changes from `master` to slim the diff.

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,137 @@
...
+                        if (in_array($directiveType, self::DirectivesRequiringLocalKey)) {
+                            $relationName = ASTHelper::directiveArgValue($directive, 'relation', $field);
+
+                            if (method_exists($model, $relationName)) {
+                                if (AppVersion::below(5.7)) {
+                                    $relation = $model->{$relationName}();
+                                    $rc = new \ReflectionClass($model->{$relationName}());
+                                    $localKey = $rc->getProperty('localKey');
+                                    $localKey->setAccessible(true);
+                                    array_push($selectColumns, $localKey->getValue($relation));
```
spawnia, line 85: Use `Utils::accessProtected()`

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,137 @@
...
+                                }
+                            }
+                        }
+
+                        if (in_array($directiveType, self::DirectivesRequiringForeignKey)) {
+                            $relationName = ASTHelper::directiveArgValue($directive, 'relation', $field);
+
+                            if (method_exists($model, $relationName)) {
+                                if (AppVersion::below(5.8)) {
+                                    array_push($selectColumns, $model->{$relationName}()->getForeignKey());
```
spawnia, line 97: `array_push($selectColumns, ` is the same between both branches

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,137 @@
...
+                        continue 2;
+                    }
+                }
+
+                if (ASTHelper::hasDirective($fieldDefinition, 'select')) {
+                    // append selected columns in select directive to seletion
+                    $directive = ASTHelper::directiveDefinition($fieldDefinition, 'select');
+
+                    if ($directive) {
+                        $selectFields = ASTHelper::directiveArgValue($directive, 'columns') ?? [];
```
spawnia, line 113:
```suggestion
                        $selectFields = ASTHelper::directiveArgValue($directive, 'columns', []);
```

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,137 @@
...
+                        array_push($selectColumns, $renamedAttribute);
+                    }
+                } else {
+                    // fallback to selecting the field name
+                    array_push($selectColumns, $field);
+                }
+            }
+        }
+
+        $selectColumns = array_filter($selectColumns, function ($column) use ($model) {
```
spawnia, line 131: Types?

src/Schema/Directives/AllDirective.php
```diff
@@ -56,13 +58,48 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
+
+                    $selectColumns = SelectHelper::getSelectColumns(
+                        $this->definitionNode,
+                        $fieldSelection,
+                        $this->getModelClass()
+                    );
+
+                    if (empty($selectColumns)) {
+                        return $builder->get();
+                    }
```
spawnia, line 80: Shouldn't this throw?

src/Schema/Directives/AllDirective.php
```diff
@@ -56,13 +58,48 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
+
+                    $query = $builder instanceof EloquentBuilder ? $builder->getQuery() : $builder;
```
spawnia, line 82: Duplicate check for `$builder instanceof EloquentBuilder`

src/Schema/Directives/AllDirective.php
```diff
@@ -56,13 +58,48 @@ public function resolveField(FieldValue $fieldValue): FieldValue
...
+
+                        $expressions = array_filter($query->columns, function ($column) {
+                            return $column instanceof Expression;
+                        });
+
+                        $builder = $builder->select(array_unique(array_merge($selectColumns, $expressions)));
+
+                        foreach ($bindings as $type => $binding) {
+                            $builder = $builder->addBinding($binding, $type);
+                        }
```
spawnia, line 95: What does this do?

src/Schema/Directives/SelectDirective.php
```diff
@@ -0,0 +1,21 @@
...
+    {
+        return /** @lang GraphQL */ <<<'GRAPHQL'
+"""
+Specify the SQL column dependencies of this field.
+"""
+directive @select(
+  """
+  SQL columns names to pass to the Eloquent query builder
+  """
+  columns: [String!]
```
spawnia, line 17:
```suggestion
  """
  SQL column names to include in the `SELECT` part of the query.
  """
  columns: [String!]!
```

src/lighthouse.php
```diff
@@ -355,6 +355,19 @@
...
     'batchload_relations' => true,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Optimized Selects
+    |--------------------------------------------------------------------------
+    |
+    | If set to true, Eloquent will only select the columns necessary to resolve
+    | a query. You must use the select directive to resolve advanced field dependencies
+    | on other columns.
```
spawnia, line 365:
```suggestion
    | If set to true, Eloquent will only select the columns necessary to resolve a query.
    | Use the @select directive to specify column dependencies of compound fields.
```

tests/AssertsQueryCounts.php
```diff
@@ -15,7 +15,7 @@ trait AssertsQueryCounts
 {
     protected function countQueries(?int &$count): void
     {
-        DB::listen(function () use (&$count): void {
+        DB::listen(function ($query) use (&$count): void {
```
spawnia, line 18: Why?

tests/Integration/Schema/Directives/FindDirectiveTest.php
```diff
@@ -196,4 +196,48 @@ public function testReturnsCustomAttributes(): void
             ],
         ]);
     }
+
+    public function testReturnsOptimizedSelect(): void
+    {
+        $company = factory(Company::class)->create();
+        $user = factory(User::class)->create([
+            'company_id' => $company->id,
+        ]);
```
spawnia, line 205:
```suggestion
        $company = factory(Company::class)->create();

        $user = factory(User::class)->make();
        assert($user instanceof User);
        $user->company()->associate($company);
        $user->save();
```

tests/Integration/Schema/Directives/HasManyDirectiveTest.php
```diff
@@ -183,29 +183,20 @@ public function testQueryHasManyWithConditionInDifferentAliases(): void
...
-                        ],
-                    ],
-                ],
             ]);
+
+        $response->assertJsonCount(2, 'data.users');
+
+        $firstTaskData = empty($response->json('data.users.0.firstTasks'))
+            ? $response->json('data.users.1.firstTasks.0')
+            : $response->json('data.users.0.firstTasks.0');
```
spawnia, line 192: Why? The test should be deterministic.

tests/Integration/Schema/Directives/LimitDirectiveTest.php
```diff
@@ -183,28 +183,19 @@ public function testLimitsWithCache(): void
...
-                    ],
-                ],
-            ]);
+            ');
+
+        $response->assertJsonCount(2, 'data.user');
+
+        $user1Tasks = 1 === intval($response->json('data.user.0.id'))
+            ? $response->json('data.user.0.tasks.0')
+            : $response->json('data.user.1.tasks.0');
```
spawnia, line 192: The test should be deterministic

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
```diff
@@ -107,18 +107,17 @@ public function testOperatorIn(): void
...
-                ],
-            ],
-        ]);
+        ');
+
+        $response->assertJsonCount(2, 'data.users');
+
+        $users = $response->json('data.users');
+
+        $ids = collect($users)->pluck('id')->all();
```
spawnia, line 116: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#laravel-feature-usage

spawnia (changes requested):
Great work on this, the tests look comprehensive and the overall direction looks right.

src/Schema/Directives/AllDirective.php
Lyrisbee: We only want to replace the `*` with `id`. The sub-query and bindings should be preserved.

From `testOrderByRelationCount()`
```graphql
{
  users(
     orderBy: [
       {
          tasks: { aggregate: COUNT }
          order: ASC
       }
     ]
  ) {
    id
  }
}
```

query dump:
```php
columns: array:2 [
    0 => "users.*"
    1 => Illuminate\Database\Query\Expression^ {#7717
      #value: "(select count(*) from `tasks` where `users`.`id` = `tasks`.`user_id` and `tasks`.`deleted_at` is null and `name` != ?) as `tasks_count`"
    }
  ],
bindings: array:9 [
    "select" => array:1 [
      0 => "cleaning"
    ]
    "from" => []
    "join" => []
    "where" => []
    "groupBy" => []
    "having" => []
    "order" => []
    "union" => []
    "unionOrder" => []
  ]
```

src/Pagination/PaginateDirective.php
```diff
@@ -129,6 +131,26 @@ public function resolveField(FieldValue $fieldValue): FieldValue
                     $this->directiveArgValue('scopes', [])
                 );
 
+            if (config('lighthouse.optimized_selects')) {
+                if (($query instanceof QueryBuilder || $query instanceof EloquentBuilder) && ! $this->directiveHasArgument('builder')) {
+                    $fieldSelection = $resolveInfo->getFieldSelection(4);
```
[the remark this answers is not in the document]
Lyrisbee: Fixed. Depth 2 is enough that can get `$fieldSelection['edges']['node']` fields.

src/Schema/Directives/AllDirective.php
spawnia: I do not understand why you need to first get and then re-set the bindings. As far as I can tell, `->select()` only clears `columns` and `bindings['select']`.

Lyrisbee: Do you mean it only needs to re-set the `bindings['select']` but not the whole `bindings`?

spawnia: I think these 3 lines and `$bindings = $query->getRawBindings();` are completely unnecessary.

Lyrisbee: The above schema is the case. Before `->select()`, the raw query is below. The subquery has a select binding `cleaning` that `name != ?` needs. If we don't re-set the bindings, it will throw `Illuminate\Database\QueryException: SQLSTATE[HY093]: Invalid parameter number`. Because the select bindings were clear by `->select()`
```sql
SELECT 
    `users.*`,
    (SELECT 
            COUNT(*)
        FROM
            `tasks`
        WHERE
            `users`.`id` = `tasks`.`user_id`
                AND `tasks`.`deleted_at` IS NULL
                AND `name` != ?
	) AS `tasks_count`
FROM
    `users`
ORDER BY `tasks_count` ASC
```

src/Schema/Directives/AllDirective.php
Lyrisbee: In some cases,`$selectColumns` will be empty, and it shouldn’t throw exceptions because this is an expected behavior.

```graphql
type User {
    posts: [Post!]! #@hasMany
}

type Post {
    id: ID!
}

type Query {
    users: [User!]! @all
}
```
```graphql
{
    users {
        posts {
            id
        }
    }
}
```
Use the above schema as an example. The `posts: [Post!]!` will not be optimized due to empty `$selectColumns`. In the current implementation, the `@hasMany` directive is a must-have directive to let the optimized select work normally.

The optimized select is an improvement feature, even if this function is not working normally due to insufficient directives, it shouldn’t break original queries.

spawnia: In this example, the query would have to select the users `id` so that `posts` can be resolved without `@hasMany`, right?

In that case, I believe this check is insufficient in order to not break the query. If the client queries an additional field from `User`, `$selectColumns` would no longer be empty and result in `SELECT name FROM users`, thus breaking the resolution of `posts`.

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @reference https://github.com/nuwave/lighthouse/pull/1626
+     */
+    public static function getSelectColumns(Node $definitionNode, array $fieldSelection, string $modelName): array
+    {
+        $returnTypeName = ASTHelper::getUnderlyingTypeName($definitionNode);
+
+        $astBuilder = Container::getInstance()->make(ASTBuilder::class);
+
+        assert($astBuilder instanceof ASTBuilder);
```
spawnia, line 55:
```suggestion
        $astBuilder = Container::getInstance()->make(ASTBuilder::class);
        assert($astBuilder instanceof ASTBuilder);
```

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,141 @@
...
+
+        $documentAST = $astBuilder->documentAST();
+
+        assert($documentAST instanceof DocumentAST);
```
spawnia, line 59:
```suggestion
        $documentAST = $astBuilder->documentAST();
        assert($documentAST instanceof DocumentAST);
```

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,141 @@
...
+
+        if (Str::contains($returnTypeName, ['SimplePaginator', 'Paginator'])) {
+            $returnTypeName = str_replace(['SimplePaginator', 'Paginator'], '', $returnTypeName);
+        }
+
+        $type = $documentAST->types[$returnTypeName];
+
+        $fieldDefinitions = $type->fields;
+
+        assert($fieldDefinitions instanceof NodeList);
```
spawnia, line 69:
```suggestion
        $fieldDefinitions = $type->fields;
        assert($fieldDefinitions instanceof NodeList);
```

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,141 @@
...
+
+        $model = new $modelName();
+
+        assert($model instanceof Model);
```
spawnia, line 73:
```suggestion
        $model = new $modelName();
        assert($model instanceof Model);
```

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,141 @@
...
+                    // fallback to selecting the field name
+                    $selectColumns[] = $field;
+                }
+            }
+        }
+
+        /** @var array<int, string> $selectColumns */
+        $selectColumns = array_filter($selectColumns, function ($column) use ($model): bool {
+            return ! $model->hasGetMutator($column) && ! method_exists($model, $column);
+        });
```
spawnia, line 137:
```suggestion
        $selectColumns = array_filter($selectColumns, function (string $column) use ($model): bool {
            return ! $model->hasGetMutator($column) && ! method_exists($model, $column);
        });
```

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,141 @@
...
+                } else {
+                    // fallback to selecting the field name
+                    $selectColumns[] = $field;
+                }
+            }
+        }
+
+        /** @var array<int, string> $selectColumns */
+        $selectColumns = array_filter($selectColumns, function ($column) use ($model): bool {
+            return ! $model->hasGetMutator($column) && ! method_exists($model, $column);
```
spawnia, line 136: I don't think this condition is sufficient to determine if a field is a column.

```graphql
type Foo {
  foo: String @method(name: "bar") # falsely assumed to be a column due to not recognizing the directive
  bar: String # falsely assumed not to be a column due to the presence of method bar()
  baz: String # falsely assumed not to be a column due to the presence of a get mutator
}
```

```php
final class Foo extends Model {
    public function bar(): string { ... }
    public function getBazAttribute(): string {
        return $this->attributes['baz'];
    }
}
```

Given the existence of custom field resolver directives, I don't see how we could reliably determine if a field is a column without explicit configuration.

src/Schema/Directives/AllDirective.php
spawnia: > Do you mean it only needs to re-set the bindings['select'] but not the whole bindings?

That sounds reasonable then. Saves unnecessary work and is more explicit about what is happening and why.

src/lighthouse.php
```diff
@@ -355,6 +355,18 @@
...
+    |--------------------------------------------------------------------------
+    | Optimized Selects
+    |--------------------------------------------------------------------------
+    |
+    | If set to true, Eloquent will only select the columns necessary to resolve a query.
+    | Use the @select directive to specify column dependencies of compound fields.
+    |
+    */
+
+    'optimized_selects' => true,
```
spawnia, line 368: We should consider starting this setting with `false` and wait for reports of forgotten edge cases to come in.

tests/Integration/Schema/Directives/FindDirectiveTest.php
```diff
@@ -196,4 +196,53 @@ public function testReturnsCustomAttributes(): void
...
+    {
+        $company = factory(Company::class)->create();
+
+        $user = factory(User::class)->make();
+
+        assert($user instanceof User);
+
+        $user->company()->associate($company);
+
+        $user->save();
```
spawnia, line 210:
```suggestion
        $user = factory(User::class)->make();
        assert($user instanceof User);
        $user->company()->associate($company);
        $user->save();
```

tests/Integration/Schema/Directives/HasManyDirectiveTest.php
```diff
@@ -146,7 +146,7 @@ public function testQueryHasManyWithConditionInDifferentAliases(): void
         }
 
         type Query {
-            users: [User!]! @all
+            users: [User!]! @all @orderBy(column: "id")
```
spawnia, line 149: Why did you add this?

Lyrisbee: Simple queries like `select * from users` may be returned in the order they are stored on disk, in primary key order, or others. When we select specific columns, it is more unpredictable without using `orderBy`.

https://dba.stackexchange.com/a/6053

After some commits, I removed it and made the primary key as the default order.

src/Select/SelectHelper.php
Lyrisbee: Good example. I removed it and only checked if the method existed for the relation. And by default, select the primary key to avoid no column selected if the method exists and can't find any select column. Then, we can use `@select` if any other columns need to be chosen.

bepsvpt:
@spawnia

Is there anything that needs to change for this pull request to be merged?

src/Select/SelectHelper.php
```diff
@@ -0,0 +1,160 @@
...
+                    } elseif ($relation instanceof BelongsTo) {
+                        $selectColumns[] = self::getForeignKey($relation);
+                    } elseif ($relation instanceof HasOneOrMany) {
+                        $selectColumns[] = self::getLocalKey($relation);
+                    } else {
+                        $selectColumns[] = $model->getKeyName();
+                    }
+                } else {
+                    // fallback to selecting the field name
+                    $selectColumns[] = $field;
```
spawnia, line 133: This just goes wrong for the very simple case of there being a custom getter for the field - or any custom directive that we can not possibly know about. Consider the following:

```graphql
type MyModel {
  foo: Int @someCustomDirectiveThatWeDoNotKnowAbout
  bar: ID @field(resolver: "SomeCustomClass@andAMethodWeCanNotPossiblyLookInto")
}
```

I don't think any amount of magic can help us here, any approach of trying to determine columns without explicit configuration is just fundamentally flawed.

src/lighthouse.php
```diff
@@ -355,6 +355,18 @@
 
     'batchload_relations' => true,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Optimized Selects
+    |--------------------------------------------------------------------------
+    |
+    | If set to true, Eloquent will only select the columns necessary to resolve a query.
+    | Use the @select directive to specify column dependencies of compound fields.
```
spawnia, line 364: Due to the problems I outlined in `SelectHelper`, I think we need to change the approach to this entirely. I believe we need to require that every field in a model that needs to select any columns has an `@select` configuration. I know that this is cumbersome, but perhaps we can allow iterative adoption by marking some models as being optimizable and only applying the optimization to those.

```graphql
type Foo @select {
  id: ID @select(columns: ["id"])
}
```

`@select` on the model type signifies it can be considered for optimization.

spawnia (changes requested):

spawnia:
Thank you for your continued efforts on this. Unfortunately this pull request needs significant rework for the reasons I outlined in my review.

spawnia:
You can add the following entry to `CHANGELOG.md`:

```
### Added

- Allow optimizing `SELECT` https://github.com/nuwave/lighthouse/pull/2235 by @Lyrisbee, @bepsvpt
```

HelloAlexPan:
Thanks @spawnia, could we add a @storipress tag as well — we'll get back to this one sometime next month as we're working on some other features right now. We expect this to be done EO March

spawnia:
> Thanks @spawnia, could we add a @storipress tag as well — we'll get back to this one sometime next month as we're working on some other features right now. We expect this to be done EO March

Sure, that sounds fine to me.

PR #2253, opened 2022-12-12
Update Adding Directives in extending-lighthouse.md

daniel-de-wit:
The link towards the Integration test example is no longer valid. PR of refactoring to use RegisterDirectiveNamespaces started here https://github.com/nuwave/lighthouse/pull/950.

Took me a while to figure out how to add my custom directive namespace for a package. This might help others :)

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

none.

**Breaking changes**

none.

### DIFF
--- a/docs/5/digging-deeper/extending-lighthouse.md
+++ b/docs/5/digging-deeper/extending-lighthouse.md
@@ -13,11 +13,55 @@ You can find a complete list of all dispatched events [in the events API referen
 
 ## Adding Directives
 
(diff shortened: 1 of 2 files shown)


spawnia:
Thanks, but proposing to use the config file does not really apply for this part of the documentation. It is meant to target package developers - so they really should be using the event!

Also, please modify `docs/master/` - see https://github.com/nuwave/lighthouse/blob/master/docs/.github/README.md

Issue #2254, opened 2022-12-12
`$argumentSet` is a dynamic property on the `ResolveInfo` class which is now a deprecated feature in PHP 8.2


ChrisExP:
The `GraphQL\Type\Definition\ResolveInfo` class is modified and a dynamic property `$argumentSet` is added in a number of places within Lighthouse:
```
Nuwave\Lighthouse\Schema\Factories\FieldFactory::113
Nuwave\Lighthouse\Schema\Directives\ArgTraversalDirective::23
Nuwave\Lighthouse\Schema\Directives\ConvertEmptyStringsToNullDirective::44
Nuwave\Lighthouse\Schema\Directives\SpreadDirective::30
Nuwave\Lighthouse\Schema\Directives\TrimDirective::50
```

This behaviour is [deprecated in PHP 8.2](https://php.watch/versions/8.2/dynamic-properties-deprecated) and throws errors in static analysis tools.

I'm happy to create a pull request to fix this issue, but I'm not sure the best way to keep backwards compatibility or make upgrading as painless as possible, as the `$argumentSet` property is used a lot.

stevebauman:
Hey @ChrisExP,

Either adding `#[AllowDynamicProperties]` onto the `ResolveInfo` class or adding the property `public $argumentSet` to it should be fine. I don't see any issues with that breaking backwards compatibility. 👍

ChrisExP:
But that class isn't part of Lighthouse. It's in the webonyx graphql-php package. I could go and create a pull request to add that on their repository but I'm not sure if they would go for that.

stevebauman:
Gotcha. I don't see why they wouldn't. They'll have to do one or the other to fix the PHP 8.2 deprecation error. Some members of this repository appear to be members of that repository (@spawnia may be -- correct me if I'm wrong).

ChrisExP:
As I understand it `$argumentSet` is a workaround used by Lighthouse to make the arguments available to different parts of the code so they can be modified and used at different points in the request. It is never used in the webonyx repository. So it's not a bug for webonyx as they didn't intend `ResolveInfo` to be used that way.

stevebauman:
I understand. The only other way to prevent any BC break is to extend the `ResolveInfo` class with our own implementation and add one of above mentioned patches. That is -- if webonyx GraphQL allows the `ResolveInfo` to be overridden (I do not know if this is the case).

Any other change is a BC break.

ChrisExP:
[...]

PR #2256, opened 2022-12-13
Create a new `ResolveInfo` class to extend with `$argumentSet` in response to PHP 8.2 deprecating dynamic properties

ChrisExP:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

[Issue](https://github.com/nuwave/lighthouse/issues/2254)

**Changes**

Allows Lighthouse to confirm to new restrictions in PHP 8.2 where dynamic properties are deprecated.  
The `ResolveInfo` class was used as a carrier for `$argumentSet` without it being explicitly defined.

This pull request creates a class that extends `ResolveInfo` and defines the necessary.

**Breaking changes**

This should not affect any existing use cases.


src/Schema/ResolveInfo.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Nuwave\Lighthouse\Schema;
```
spawnia, line 3:
```suggestion
namespace Nuwave\Lighthouse\Execution;
```

src/Pagination/PaginateDirective.php
```diff
@@ -152,7 +152,6 @@ public function resolveField(FieldValue $fieldValue): FieldValue
             }
 
             $query = $resolveInfo
-                ->argumentSet
                 ->enhanceBuilder(
```
spawnia, line 155: We can format this section nicer now:

```
 $query = $resolveInfo->enhanceBuilder(
    $query,
);
```

Same with other places like this

src/Schema/Factories/FieldFactory.php
```diff
@@ -109,10 +110,22 @@ public function handle(FieldValue $fieldValue): array
...
+                $resolveInfo->fieldDefinition,
+                $resolveInfo->fieldNodes,
+                $resolveInfo->parentType,
+                $resolveInfo->path,
+                $resolveInfo->schema,
+                $resolveInfo->fragments,
+                $resolveInfo->rootValue,
+                $resolveInfo->operation,
+                $resolveInfo->variableValues,
+                $argumentSet
```
spawnia, line 125:
```suggestion
            $wrappedResolveInfo = new ResolveInfo(
                $resolveInfo->fieldDefinition,
                $resolveInfo->fieldNodes,
                $resolveInfo->parentType,
                $resolveInfo->path,
                $resolveInfo->schema,
                $resolveInfo->fragments,
                $resolveInfo->rootValue,
                $resolveInfo->operation,
                $resolveInfo->variableValues,
                $this->argumentSetFactory->fromResolveInfo($args, $resolveInfo)
```

spawnia (changes requested):
Nicely done, just some minor tweaks and this should be good to go.

PR #2259, opened 2022-12-16
Do not cache schema when runing tests

Deyjandi:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Lighthouse config file disables schema caching **only** if `APP_ENV` is set to `local`

Out of the box, `APP_ENV`  is set to `testing` when running PHPUnit tests (as defined in phpunit.xml).

This means that when running tests using `php artisan test` or `sail test` the schema gets cached.
Any further changes to schema require the file ./bootstrap/cache/lighthouse-schema.php to be deleted for the tests to pickup the schema changes, which I don't think is the intended behavior.

This PR fixes that behavior by disabling cache for both `local` and `testing` application environments.

**Breaking changes**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Do not cache schema when running tests https://github.com/nuwave/lighthouse/pull/2259
+
 ## v5.69.1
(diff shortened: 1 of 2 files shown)


spawnia:
Use `RefreshesSchemaCache` in your tests, see https://lighthouse-php.com/master/testing/phpunit.html#setup.

PR #2276, opened 2023-01-16
CanDirective uses the context from GuardDirective

### LINKED ISSUE #2260: `@can` throws 403 on non-default guard
When using `@guard` with any other guard than lighthouse config's default `@can` always returns 403

`@can` should use the guard specified in `@guard`

Inside `CanDirective.php` the code

```php
$this->gate->forUser($context->user());
```
`$context->user()` returns null if using other guard, given `Nuwave\Lighthouse\Auth\AuthServiceProvider.php` `guard()` function only tries to use lighthouse or auth config file default guard.

Lighthouse Version: v5.69.1

MissaelAnda:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #2260

**Changes**
Removed the recreation of `Gate` in `CanDirective` given it already has the correct context.
`@guard` must be used before `@can` so the context is correct.


tests/Unit/Tests/Unit/Auth/GuardDirectiveTest.php
```diff
@@ -132,4 +134,62 @@ public function testGuardHappensBeforeOtherDirectivesIfAddedFromType(): void
             ')
             ->assertGraphQLErrorCategory(AuthenticationException::CATEGORY);
     }
+
+    public function testMultiGuardWithAuthorization(): void
+    {
+        /** @var \Illuminate\Contracts\Config\Repository $config */
```
spawnia, line 140: Use `assert`

tests/Unit/Tests/Unit/Auth/GuardDirectiveTest.php
```diff
@@ -132,4 +134,62 @@ public function testGuardHappensBeforeOtherDirectivesIfAddedFromType(): void
...
+                'driver' => 'session',
+                'provider' => 'teams',
+            ],
+        ]));
+
+        $team = new Team();
+        $team->id = 1;
+        $team->name = 'Test';
+
+        $this->app['auth']->guard('team')->setUser($team);
```
spawnia, line 161: Use `make()` + `assert`

tests/Utils/Policies/TeamPolicy.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Tests\Utils\Policies;
+
+use Illuminate\Auth\Access\Response;
+use Tests\Utils\Models\Team;
+
+final class TeamPolicy
+{
+    public function onlyTeams($user): bool
```
spawnia, line 10: type of `$user`?

src/Auth/CanDirective.php
```diff
@@ -124,19 +124,18 @@ public function handleField(FieldValue $fieldValue, \Closure $next): FieldValue
         $resolved = $this->directiveArgValue('resolved');
 
         $fieldValue->setResolver(function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($previousResolver, $ability, $resolved) {
-            $gate = $this->gate->forUser($context->user());
```
spawnia, line 127: I am unsure if removing this will have side effects. Let's investigate when and why it was added.

MissaelAnda: That line was added in PR #1043 and it was not addressed why.
But I think it wouldn't have side effects given `Gate` uses the config's default guard as `Nuwave\Lighthouse\Schema\Context`, the difference is `Context` resolves the user before `@guard` and `Gate` resolves the user until it tries to authorize.
Another solution could be that `Context` resolves the user until `user()` is called, so if it's called after `@guard` the default guard changes and doesn't return null.

tests/Utils/Policies/TeamPolicy.php
MissaelAnda: I intended the authorization to assert that the logged user type is `Team` and not null or other guard.

src/Schema/Context.php
```diff
@@ -16,17 +16,9 @@ class Context implements GraphQLContext
      */
     public $request;
 
-    /**
-     * An instance of the currently authenticated user.
-     *
-     * @var \Illuminate\Contracts\Auth\Authenticatable|null
-     */
-    public $user;
```
spawnia, line 24: This is a breaking change, please mention it in `CHANGELOG.md`

spawnia: Oh I also think this will be problematic with `@inject`.

MissaelAnda:
So I've been checking this and there are two possible solutions:
- Make `Context` resolve the user when `user()` is called (leaving `$this->user` attribute available) but it breaks some tests that access directly to the `$context->user`.
- Let `Gate` resolve the user by itself, wich already works due to `@guard` using `$guard->shouldUse()`, the only problem is that `Gate` would ignore the `lighthouse.guard` configuration, which I don't think is not a big deal because you are already using `@guard`.

what do you think @spawnia ?

tests/Unit/Tests/Unit/Auth/GuardDirectiveTest.php
spawnia:
```suggestion
```

unnecessary with assert

tests/Unit/Tests/Unit/Auth/GuardDirectiveTest.php
```diff
@@ -132,4 +135,66 @@ public function testGuardHappensBeforeOtherDirectivesIfAddedFromType(): void
...
+                'driver' => 'session',
+                'provider' => 'teams',
+            ],
+        ]));
+
+        $team = new Team();
+        $team->id = 1;
+        $team->name = 'Test';
+
+        /** @var \Illuminate\Contracts\Auth\Factory */
```
spawnia, line 163:
```suggestion
```

tests/Unit/Tests/Unit/Auth/GuardDirectiveTest.php
```diff
@@ -132,4 +135,66 @@ public function testGuardHappensBeforeOtherDirectivesIfAddedFromType(): void
...
+        $auth = $this->app->make('auth');
```
spawnia, line 164:
```suggestion
        $auth = $this->app->make(AuthFactory::class);
```

tests/Unit/Tests/Unit/Auth/GuardDirectiveTest.php
```diff
@@ -132,4 +135,66 @@ public function testGuardHappensBeforeOtherDirectivesIfAddedFromType(): void
...
+        type Query {
+            team: Team!
+                @guard(with: ["team"])
+                @can(ability: "onlyTeams", model: "Tests\\\\Utils\\\\Models\\\\Team")
+                @mock
+        }
+        ';
+
+        $this
+            ->graphQL(/** @lang GraphQL */ '
```
spawnia, line 184:
```suggestion
        $this->graphQL(/** @lang GraphQL */ '
```

spawnia:
We can also make `$this->user` a magic attribute - using `__get()` - and resolve it lazily that way.

MissaelAnda:
> We can also make `$this->user` a magic attribute - using `__get()` - and resolve it lazily that way.

I tried this approach locally and it works, the problem is still it breaks some tests that change directly the attribute `$user` of `Context`, should I commit it like that or go with the default `Gate` approach?

spawnia:
> it breaks some tests that change directly the attribute $user

It looks to me like there is exactly `testSetsAndResetsGuardContextAfterEachIteration`. Perhaps it can be modified to use a different implementation of the context interface that allows setting `$user`, I think it is not a common use case.

MissaelAnda:
As you mentioned there are 3 failing tests now, but I'm not sure how those are supposed to be modified.

src/Schema/Context.php
```diff
@@ -36,11 +30,23 @@ public function __construct(Request $request)
...
+    public function __get(string $name): mixed
+    {
+        return method_exists($this, $name)
+            ? $this->{$name}()
+            : $this->{$name};
+    }
+
+    public function __isset(string $name): bool
+    {
+        return method_exists($this, $name) || property_exists($this, $name);
```
spawnia, line 50:
```suggestion
        return $this->__get($name) !== null;
```

And make `__get()` not throw on undefined keys. `isset()` is supposed to return `false` if the value is `null`.

pyrou: I dug deeper on failing SubscriptionTest::testWithGuard().

Turns out resolving user 'on-demand' can't work properly as expected. (Even if you comment the logout part in the test). Using magical `__get()`, and this attempt to mitigate data_get and @inject directive issue with an `__isset()` is probably not the way to go.

I'll give some more try with a $context->setUser().

pyrou:
Implementation with `setUser()` looks ok with existing and new unit test.

Let me know what you think about it. 
UPGRADE.md will be updated then.

Failing UT on Laravel 10 configuration is unrelated to this PR, I re-ran manually a PHP 8.2 / Laravel 10 configuration on another recent PR to confirm this : https://github.com/nuwave/lighthouse/actions/runs/4113307323/jobs/7171623286 

It fails too now, due to recent changes in Laravel 10. But this deserve a separated PR for fix.

src/Auth/GuardDirective.php
```diff
@@ -92,6 +95,7 @@ protected function authenticate(array $guards): void
      * Handle an unauthenticated user.
      *
      * @param  array<string|null>  $guards
+     * @throws AuthenticationException
```
spawnia, line 98:
```suggestion
     * @return never
```

Should also make the phpstan config change obsolete

src/Auth/GuardDirective.php
```diff
@@ -74,14 +75,16 @@ public function handleField(FieldValue $fieldValue, \Closure $next): FieldValue
...
      * @throws \Illuminate\Auth\AuthenticationException
      */
-    protected function authenticate(array $guards): void
+    protected function authenticate(array $guards): Authenticatable
     {
         foreach ($guards as $guard) {
-            if ($this->auth->guard($guard)->check()) {
+            $user = $this->auth->guard($guard)->user();
+
+            if (!is_null($user)) {
```
spawnia, line 83:
```suggestion
            if ($user) {
```

phpstan.neon
```diff
@@ -35,6 +35,10 @@ parameters:
   # As long as we support multiple Laravel versions at once, there will be some dead spots
   reportUnmatchedIgnoredErrors: false
 
+  earlyTerminatingMethodCalls:
+    Nuwave\Lighthouse\Auth\GuardDirective:
+      - unauthenticated
```
spawnia, line 40: should not be necessary with `@return never`

src/Schema/Context.php
```diff
@@ -39,6 +35,11 @@ public function user(): ?Authenticatable
         return $this->user;
     }
 
+    public function setUser(?Authenticatable $user): void
+    {
+        $this->user = $user;
+    }
```
spawnia, line 41: That seems like a much, much simpler solution

spawnia (commented):
I like the direction of this, just needs a few tweaks

src/Auth/GuardDirective.php
pyrou: I kept both

PR #2280, opened 2023-01-25
Add wildcard to inject directive

### LINKED ISSUE #1206: Add wildcard path for @inject directive
Considering this example schema

`
type User {
    id: ID!
    name: String!
}

type Post {
    id: ID!
    title: String!
    body: String!
    user: User! @belongsTo
}

type Comment {
    id: ID!
    title: String!
    body: String!
    user: User! @belongsTo
}

input UpdateUserInput {
    comments: UpdateCommentsHasMany
}

input UpdateCommentsHasMany {
    create: [CreateCommentInput!]
}

input CreateCommentInput {
    post_id: ID!
    title: String!
    body: String!
}

type Mutation @guard {
    updateUser(input: UpdateUserInput!): User 
        @update
        @inject(context: "user.id", name: "comments.create.*.user_id")
}
`

It would be awesome if Lighthouse could inject data in the argument set using wildcard in the dot notation, this would cover the nested mutations in the `HasMany` case, while ensuring that a comment will be created for the authenticated user or a mutation that uses a list of input types.

Of course, today, I can create something like:

`
input CreateCommentInput {
    post_id: ID!
    title: String!
    body: String!
}

type Mutation @guard {
    createComment(input: CreateCommentInput!): Comment!
        @create
        @inject(context: "user.id", name: "customer_id")
}
`

But, well, this way I'm not using the power of nested mutations.

Maybe I'm wrong, but I think the change needed is just use `data_fill` instead of `Illuminate\Support\Arr::add` in the `Nuwave\Lighthouse\Schema\Directives\InjectDirective` class and rewrite the `addValue` method of `Nuwave\Lighthouse\Execution\Arguments\ArgumentSet` class to behave like `data_fill` as actually it is behaving like `Illuminate\Support\Arr::add`

Or maybe create a new directive and method.

I'm willing to help with a PR when as soon as there is time available.

Lighthouse Version: 4.9
Laravel Version: 6.12.0

andershagbard:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1206

~I need some help finishing this. It throws a weird error, probably due to how I made the test. If you run `dd($this->toArray)` on line 73 in [`ArgumentSet`](https://github.com/andershagbard/lighthouse/blob/inject-wildcard/src/Execution/Arguments/ArgumentSet.php), you can see it injects the context with the correct value.~

**Changes**

You can now use wildcard in the inject directive.

**Breaking changes**

None I think?

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -2060,6 +2060,8 @@ directive @inject(
   The target name of the argument into which the value is injected.
   You can use dot notation to set the value at arbitrary depth
   within the incoming argument.
+  Use an asterisk `*` as a path segment to inject the value into every
+  element of a list found at that position.
   """
   name: String!
 ) repeatable on FIELD_DEFINITION
@@ -2083,6 +2085,30 @@ type Mutation {
   createTask(input: CreateTaskInput!): Task!
     @inject(context: "user.id", name: "input.user_id")
     @create
+}
+```
+
+If you need to inject a value into every element of a list, such as a nested `create` mutation
+for a `HasMany` relation, use an asterisk `*` as a wildcard path segment.
+
+```graphql
+type Mutation {
+  updateUser(input: UpdateUserInput!): User
+    @update
+    @inject(context: "user.id", name: "input.tasks.create.*.user_id")
+}
+
+input UpdateUserInput {
+  id: ID!
+  tasks: UpdateTasksHasManyInput
+}
+
+input UpdateTasksHasManyInput {
+  create: [CreateTaskInput!]
+}
+
+input CreateTaskInput {
+  title: String!
 }
 ```
 
--- a/src/Execution/Arguments/ArgumentSet.php
+++ b/src/Execution/Arguments/ArgumentSet.php
@@ -3,6 +3,7 @@
 namespace Nuwave\Lighthouse\Execution\Arguments;
 
 use Illuminate\Support\Collection;
+use Nuwave\Lighthouse\Exceptions\DefinitionException;
 
 class ArgumentSet
 {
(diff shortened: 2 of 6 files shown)


tests/Integration/Schema/Directives/InjectDirectiveTest.php
```diff
@@ -56,4 +56,61 @@ public function testCreateFromInputObjectWithDeepInjection(): void
...
+            name: String!
+            user: User @belongsTo
+        }
+
+        type User {
+            id: ID
+        }
+
+        type Mutation {
+            createTasks(input: [CreateTaskInput!]! @spread): Task @create @inject(context: "user.id", name: "input.*.user_id")
```
spawnia, line 77: This is not a valid usage of `@create`, it can only create a single task.

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -61,19 +61,38 @@ public function has(string $key): bool
...
+     * @param  mixed  $value any value to inject
+     * @param  ArgumentSet  $argumentSet
+     */
+    private static function applyValue(array $keys, $value, ArgumentSet $argumentSet): void
+    {
         while (count($keys) > 1) {
             $key = array_shift($keys);
 
+            if ($key === '*') {
+                foreach ($argumentSet as $argument) {
```
andershagbard, line 89: Not sure if we should throw an error with it's not an array here?

src/Schema/Directives/InjectDirective.php
```diff
@@ -28,6 +28,7 @@ public static function definition(): string
   The target name of the argument into which the value is injected.
   You can use dot notation to set the value at arbitrary depth
   within the incoming argument.
+  Asterisks may be used to indicate wildcards.
```
spawnia, line 31: This does not really explain what it actually does.

```suggestion
  Use an asterisk `*` as a path segment where the value of the argument is a list to be traversed.
```

Also, the definition needs to be copied to the docs - just look for `@inject` in `/docs`. I think we should add an example there.

src/Execution/Arguments/ArgumentSet.php
andershagbard: I've updated it now. Not sure if i used the correct exception.

docs/master/api-reference/directives.md
```diff
@@ -1756,6 +1757,16 @@ type Mutation {
 }
 ```
 
+If you have an array you need to inject a value into, you can use an asterisk `*`. 
+
+```graphql
+type Mutation {
+  createTask(input: CreateTaskInput!): Task
+    @create
+    @inject(context: "user.id", name: "input.*.user_id")
```
spawnia, line 1766: This example does not work.

src/Execution/Arguments/ArgumentSet.php
spawnia: Use `DefinitionException`, in this case the error must be solved by the developer.

docs/master/api-reference/directives.md
andershagbard: I dont know if the docs has an existing example we can use, but I've updated it to a `updateUser` mutation.

docs/master/api-reference/directives.md
```diff
@@ -1756,6 +1757,16 @@ type Mutation {
...
 ```
 
+If you have an array you need to inject a value into, you can use an asterisk `*`. 
+
+```graphql
+type Mutation {
+  updateUser(input: UpdateUser!): Task
+    @update
+    @inject(context: "user.id", name: "input.tasks.create.*.user_id")
+}
```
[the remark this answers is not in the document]
andershagbard: Makes sense. Perhaps we should check whether the fields are mandatory, and if not, just `break` instead of throwing an error?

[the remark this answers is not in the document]
andershagbard: I guess the same "issue" could exists with just dot notation. But it would make sense to discard it if an array does not exists at the asterisk segment.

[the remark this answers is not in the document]
spawnia: With just dot notation, we force the value at the defined position - and create intermediary inputs if necessary. I guess this could not be desired in every case, but it is how it currently works. At least it is well defined.

In the case of an array, we have the problem that it is less clear what to insert. What number of arguments at the desired position should be added? 1 or another number? I would argue it should always be 0 - that being equivalent to injecting nothing at all.

andershagbard:
I've taken a slightly different approach, just to make it easy. I've implemented `ArrayAccess` and let `data_set` handle the logic.

I've also removed all empty arrays from the values. It caused a `array to string conversion error` later in the code. In this test: `testWillRejectValuesNotPlacedAtArrayWithWildcardInjection`

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -102,4 +102,34 @@ public function argumentsWithUndefined(): array
...
+
+        return $argumentSet->arguments[$offset];
+    }
+
+    public function offsetSet(mixed $offset, mixed $value): void
+    {
+        $argumentSet = $this;
+
+        $argument = new Argument();
+        $argument->value = $value ?: new self();
```
spawnia, line 125: This means `false`, `0` or `''` would not be possible to set

spawnia (changes requested):

src/Execution/Arguments/ArgumentSet.php
```diff
@@ -61,38 +61,38 @@ public function has(string $key): bool
...
+        data_set($argumentSet, $path, $value);
+        self::removeEmptyArrays($argumentSet);
 
         return $this;
     }
 
+    /**
+     * Remove all arguments which is an empty array.
+     */
+    private static function removeEmptyArrays(ArgumentSet $argumentSet): void
```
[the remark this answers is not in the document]
andershagbard: I think this was a symptom of `data_set` setting object properties. Will remove.

andershagbard:
Not sure how to solve the unit tests.

src/Execution/Arguments/Argument.php
```diff
@@ -85,4 +85,42 @@ protected static function toPlainRecursive($value)
 
         return $value;
     }
+
+    public function offsetExists(mixed $offset): bool
+    {
+        $argument = $this;
```
spawnia, line 91: I do not get why you keep aliasing `$this` - why not just it directly?

src/Execution/Arguments/Argument.php
```diff
@@ -85,4 +85,42 @@ protected static function toPlainRecursive($value)
...
+
+        unset($argument[$offset]);
+    }
+
+    public function getIterator(): \ArrayIterator
+    {
+        $value = $this->value;
+
+        return new \ArrayIterator($value);
```
spawnia, line 124:
```suggestion
        return new \ArrayIterator($this->value);
```

spawnia (commented):
Have you considered adding to `ArgumentSetTest`? It should be pretty straightforward to test the many possible edge cases there and examine exactly what the result will be.

PR #2281, opened 2023-01-30
Feat: Add support for global id in entity resolver

olivernybroe:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds decoding of the global id when using the model entity resolver

**Breaking changes**

This will break everyone who used the global id directive and expecting to be able to use database id's when using `_entities`. I would however argue that it is almost a bug :)

### DIFF
--- a/src/Federation/EntityResolverProvider.php
+++ b/src/Federation/EntityResolverProvider.php
@@ -15,9 +15,12 @@ use Illuminate\Support\Collection;
 use Nuwave\Lighthouse\Exceptions\DefinitionException;
 use Nuwave\Lighthouse\Exceptions\FederationException;
 use Nuwave\Lighthouse\Federation\Directives\KeyDirective;
+use Nuwave\Lighthouse\GlobalId\GlobalIdDirective;
+use Nuwave\Lighthouse\Schema\AST\ASTHelper;
 use Nuwave\Lighthouse\Schema\DirectiveLocator;
 use Nuwave\Lighthouse\Schema\Directives\ModelDirective;
 use Nuwave\Lighthouse\Schema\SchemaBuilder;
+use Nuwave\Lighthouse\Support\Contracts\GlobalId;
 use Nuwave\Lighthouse\Support\Utils;
 
 /**
@@ -60,16 +63,23 @@ class EntityResolverProvider
      */
     protected $resolvers;
 
+    /**
+     * @var \Nuwave\Lighthouse\Support\Contracts\GlobalId
+     */
+    protected $globalId;
+
     public function __construct(
         SchemaBuilder $schemaBuilder,
         DirectiveLocator $directiveLocator,
         ConfigRepository $configRepository,
-        Container $container
+        Container $container,
+        GlobalId $globalId,
     ) {
         $this->schema = $schemaBuilder->schema();
         $this->directiveLocator = $directiveLocator;
         $this->configRepository = $configRepository;
         $this->container = $container;
+        $this->globalId = $globalId;
     }
 
     public static function missingResolver(string $typename): string
@@ -179,11 +189,11 @@ class EntityResolverProvider
 
         $keyFieldsSelections = $this->keyFieldsSelections($definition);
 
-        return function (array $representation) use ($keyFieldsSelections, $modelClass): ?Model {
+        return function (array $representation) use ($keyFieldsSelections, $modelClass, $definition): ?Model {
             $builder = $modelClass::query();
             assert($builder instanceof Builder);
 
-            $this->constrainKeys($builder, $keyFieldsSelections, $representation);
+            $this->constrainKeys($builder, $keyFieldsSelections, $representation, $definition);
 
             $results = $builder->get();
             if ($results->count() > 1) {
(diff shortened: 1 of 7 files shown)


src/Federation/EntityResolverProvider.php
```diff
@@ -60,16 +63,23 @@ class EntityResolverProvider
      */
     protected $resolvers;
 
+    /**
+     * @var GlobalId
```
spawnia, line 67: fqcn

src/Federation/EntityResolverProvider.php
```diff
@@ -198,12 +208,13 @@ static function (string $classCandidate): bool {
      * @param  \Illuminate\Support\Collection<\GraphQL\Language\AST\SelectionSetNode>  $keyFieldsSelections
      * @param  array<string, mixed>  $representation
      */
-    protected function constrainKeys(EloquentBuilder $builder, Collection $keyFieldsSelections, array $representation): void
+    protected function constrainKeys(EloquentBuilder $builder, Collection $keyFieldsSelections, array $representation, ObjectTypeDefinitionNode $definition): void
```
spawnia, line 211: Could `$definition` also be an `InterfaceTypeDefinitionNode`?

src/Federation/EntityResolverProvider.php
```diff
@@ -252,15 +263,38 @@ protected function applySatisfiedSelection(EloquentBuilder $builder, SelectionSe
 
             $subSelection = $field->selectionSet;
             if (null === $subSelection) {
-                $builder->where($fieldName, $value);
+                if ($this->hasFieldWithDirective($definition, $fieldName, 'globalId')) {
+                    $builder->where($fieldName, $this->globalId->decodeID($value));
+                } else {
+                    $builder->where($fieldName, $value);
+                }
```
spawnia, line 270:
```suggestion
                $builder->where(
                    $fieldName,
                    $this->hasFieldWithDirective($definition, $fieldName, 'globalId')
                        ? $this->globalId->decodeID($value)
                        : $value
                );
```

src/Federation/EntityResolverProvider.php
```diff
@@ -252,15 +263,38 @@ protected function applySatisfiedSelection(EloquentBuilder $builder, SelectionSe
...
 
-            $this->applySatisfiedSelection($builder, $subSelection, $representation);
+            $this->applySatisfiedSelection($builder, $subSelection, $representation, $definition);
         }
     }
 
+    private function hasFieldWithDirective(ObjectTypeDefinitionNode $definition, string $fieldName, string $directiveName): bool
+    {
+        foreach ($definition->fields as $definitionField) {
+            /** @var FieldDefinitionNode $definitionField */
```
spawnia, line 282: use `assert()` or omit entirely

src/Federation/EntityResolverProvider.php
```diff
@@ -252,15 +263,38 @@ protected function applySatisfiedSelection(EloquentBuilder $builder, SelectionSe
...
+            if ($definitionField->name->value == $fieldName) {
```
spawnia, line 283:
```suggestion
            if ($definitionField->name->value === $fieldName) {
```

src/Federation/EntityResolverProvider.php
```diff
@@ -252,15 +263,38 @@ protected function applySatisfiedSelection(EloquentBuilder $builder, SelectionSe
...
+                foreach ($definitionField->directives as $directive) {
+                    /** @var DirectiveNode $directive */
+                    if ('globalId' === $directive->name->value) {
```
spawnia, line 286: 2nd usage - let's define a constant `NAME` in `GlobalIdDirective`

tests/Integration/Federation/FederationEntitiesModelTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+        $user = factory(User::class)->create();
+
+        $userRepresentation = [
+            '__typename' => 'User',
+            'id' => (string) $user->getKey(),
+        ];
+
+        $this->graphQL(/** @lang GraphQL */ '
+        query ($_representations: [_Any!]!) {
```
spawnia, line 36:
```suggestion
        query ($representations: [_Any!]!) {
```

src/Federation/EntityResolverProvider.php
olivernybroe: It is coming from the method `\Nuwave\Lighthouse\Federation\EntityResolverProvider::typeDefinition` which says it is instance of `ObjectTypeDefinitionNode`, so I guess not 🤔

src/Federation/EntityResolverProvider.php
```diff
@@ -252,13 +262,35 @@ protected function applySatisfiedSelection(EloquentBuilder $builder, SelectionSe
...
+                    if ($directive->name->value === $directiveName) {
+                        return true;
+                    }
+                }
+
+                return false;
+            }
         }
+
+        return false;
```
spawnia, line 293:
```suggestion
        $field = ASTHelper::firstByName($definition->fields, $fieldname);
        if (! $field) {
          return false;
        }

         return ASTHelper::hasDirective($field, $directiveName);
```

olivernybroe: Ah nice, couldn't find the helper, was searching for the wrong thing 😆 

Thanks!

spawnia:
Proposed changelog under `### Add`:

```
- Decode fields with directive `@globalId` in federation model entity resolver
```

spawnia (approved):

PR #2286, opened 2023-01-31
Ensure _entities maintains order of representations in result

### LINKED ISSUE #2285: Federated Entities resolver changes the order of results
**Describe the bug**

I'm using federated graph. One of my services receives the list of entities to resolve. It can accept multiple types of the same interface.
The example schema looks like this:
```
type Type1 @key(fields: "id") @extends {
  id: ID! @external
  subgraphField1: Int
}

type Type2 @key(fields: "id") @extends {
  id: ID! @external
  subgraphField2: Int
}
```

So the request to the service would look like this:
```graphql
query SubQuery($representations: [_Any!]!) {
	_entities(representations: $representations) {
		... on Type1 {
                         __typename
			subgraphField1
		}
		... on Type2 {
                         __typename
			subgraphField2
		}
	}
}
```
with the representations:
```json
{
	"representations": [
		{
			"__typename": "Type1",
			"id": "111",
		},
		{
			"__typename": "Type2",
			"id": "222",
		},
                {
			"__typename": "Type1",
			"id": "333",
		},
		{
			"__typename": "Type2",
			"id": "444",
		},
	]
}
```

In Entities resolver representations are grouped by type, making it:
```php
$groupedRepresentations = [
  'Type1' => [ ['id' => 111], ['id' => 333]],
  'Type2' => [ ['id' => 222], ['id' => 444]],
];
```
So after resoving these entites (with or without batch resolver), the result will have entities in the following order:
111, 333, 222, 444, which is different to original order. It leads to the incorrect data to be returned from the federated graph.

We're using apoolo federation v1.

The order of entities in the response should be the same as in the requst.

**Steps to reproduce**

Described above

**Lighthouse Version**

Starting from 5.42.0.

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #2285

apollographql/apollo-federation-subgraph-compatibility#70

**Changes**

Only enabling the failing test so far.

**Breaking changes**

Require implementations of `BatchedEntityResolver` to maintain the keys given in `array $representations`.


spawnia:
@olivernybroe have you given this problem any thought before? Do you have an idea how we can implement it in `Entities`?

spawnia:
Utilizing mutable representation objects would indeed make matching the order quite simple - we can group them, partition them, pass them around, etc. In the end, we just need to map over the original array of representation objects, which are still in order and then extract the possible result.

Another idea I have explored would be to just require batched resolvers to return the same number of results as were given and throw if they do not. We only need to take care of merging things back together the right way after grouping.

olivernybroe:
@spawnia hmm makes sense, the biggest problem I see with that idea is when doing queries the user will have to be very aware of the order they return things from in the query, making it very easy to accidentally break something. 

I feel like mutable objects kinda simplifies this quite a lot, as it removes the whole process of having to remember that the order actually matters and that you have to carefully control this.

spawnia:
@olivernybroe please check the solution I proposed. I think it only affects implementors of `BatchedEntityResolver`.

src/Federation/Resolvers/Entities.php
```diff
@@ -33,27 +28,49 @@ public function __invoke($root, array $args, GraphQLContext $context, ResolveInf
     {
         $results = [];
 
-        /** @var array<string, array<int, array<string, mixed>>> $groupedRepresentations */
+        $representations = $args['representations'];
+        $representationHashes = array_map('serialize', $representations);
```
olivernybroe, line 32: Is it in any way possible for the serialized result to give the same response twice? 🤔

src/Federation/Resolvers/Entities.php
```diff
@@ -33,27 +28,49 @@ public function __invoke($root, array $args, GraphQLContext $context, ResolveInf
...
+
+        $assignResultsByHash = function ($result, string $hash) use ($representationHashes, &$results): void {
+            foreach ($representationHashes as $index => $h) {
+                if ($hash === $h) {
+                    dump($hash, $index);
```
olivernybroe, line 37: Forgotten dump

olivernybroe (approved):
In general this approach looks very good. I would argue that this is not even BC, but just a fix 👀

src/Federation/Resolvers/Entities.php
olivernybroe: Nevermind, idiot question, we actually want that 😂

src/Federation/Resolvers/Entities.php
spawnia:
```suggestion
```

src/Federation/Resolvers/Entities.php
```diff
@@ -33,27 +28,48 @@ public function __invoke($root, array $args, GraphQLContext $context, ResolveInf
...
+            foreach ($representationHashes as $index => $h) {
+                if ($hash === $h) {
+                    $results[$index] = $result;
+                }
+            }
+        };
+
+        /**
+         * Firstly, representations are grouped by typename to allow assigning the correct resolver for each entity.
+         * Secondly, they are deduplicated based on their hash to avoid resolving the same entity twice.
```
kfc, line 44: Should resolving be a concern of a specific entity resolver? For example the same representation can be resolved differently in the same request. A bit of an artificial example, but things like 
```
"representations": [
  {
      "__typename": RandomNumber
      "max": 100
  },
{
      "__typename": RandomNumber
      "max": 100
  }
]
```
will have to return 2 different random numbers.

spawnia: Since we are dealing with *entities* here, I would expect them to be consistent within a single request. For use cases like your example, dedicated custom queries seem more fitting.

Issue #2288, opened 2023-02-01
Cannot use @scope directive in conjunction with @delete


NickBourque:
**Describe the bug**

When using the `@delete` directive to delete a model that has a global scope applied, you cannot use the `@scope` directive to override the global scope.

**Expected behavior/Solution**

You should be able to override global scopes at the same time as deleting a model.

**Steps to reproduce**

1. Create a `Post` model with a global scope that by default omits _archived_ posts.
```
public function apply(Builder $builder, Model $model)
{
    $builder->where('archived', false);
}
```
2. Add an override to your scope that removes the global scope.
```
protected function addWithArchived(Builder $builder)
{
    $builder->macro('withArchived', fn (Builder $builder) => $builder->withoutGlobalScope($this));
}
```
3. Try to delete an `archived` `Post` using `@delete` in conjunction with `@scope`, such as: 
```
deletePost(id: ID! @eq @scope(name: "withArchived")): Post! @delete(model: "App\\Models\\Post")
```

**Output/Logs**

```
Model Not Found exception
```

</details></br>

**Lighthouse Version**

`5.70.3`

spawnia:
That is because `@delete` currently does not use a query builder in the same way as other directives like `@all` do. Instead, it uses `Model::find()` to retrieve the model.

Can scopes and `find()` be combined?

NickBourque:
Yes, scopes and `find()` can be combined. In the above example, the query would look like `Post::withArchived()->find(1);`.

Issue #1145, opened 2020-01-14
Pay attention of @softDeletes in mutation directives


lorado:
**Is your feature proposal related to a problem? Please describe.**

Currently implementation of update/upsert don't use `enchanceBuilder()`, therefore it is not possible to update soft deleted model, because without appending e.g. `withTrashed` model will not be fetched.

**Describe the solution you'd like**

So here is my question: Should we take care of soft deleted models in update/upsert?

In my application I currently disabled in UI the ability to edit entity, if it is soft deleted (it is actually in trash). What do you think?

sweet-greg:
Biased answer, as I'm said user, but it would be great if we could update/upsert with the same trashed options 👍

spawnia:
How would this work when it comes to nested mutations?

sweet-greg:
Is there anything holding us back to do the same thing for update and upsert in a nested mutation that I am missing?
Create, connect and disconnect don't have a findOrFail clause going on (or do they?) so it wouldn't make much sense there I believe.

spawnia:
> Is there anything holding us back to do the same thing for update and upsert in a nested mutation that I am missing?

We haven't implemented filtering of any kind in nested mutations yet. The first step would be to consider how that might work from a client perspective. How would the input arguments be shaped?

sweet-greg:
I drew the following up, curious to your thoughts.

**Schema**
```
type Author {
    id: ID!
    name: String
    email: String!
}

type Article {
    id: ID!
    title: String
    author: Author
}

extend type Mutation {
    updateArticle(input: ArticleInput! @spread): Article @update
}

input ArticleInput {
    id: ID!
    title: String
    author: UpdateAuthorRelation
}

input CreateAuthorInput {
    name: String
    email: String!
}

input UpdateAuthorInput {
    id: ID!
    name: String
    email: String
    trashed: Trashed @trashed
}

input UpdateAuthorRelation {
    connect: ID
    create: CreateAuthorInput
    update: UpdateAuthorInput
    disconnect: Boolean
    delete: Boolean
}
```

mutation example
```
mutation{
  updateArticle(
    input: {
      id: 1
    	title: "Othello"
      author: {
        update: {
          id: 1
          name: "William Shakespeare"
          trashed: WITH
        }
      }
    }    
  ) {
    id
    title
    author {
      id
      name
    	email
    }
  }
}
```

sweet-greg:
@spawnia just following up, do you see a possibility for this functionality?
[...]

PR #2289, opened 2023-02-04
Require filter directives such as `@whereKey` in `@delete`, `@forceDelete` and `@restore`

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nuwave/lighthouse/issues/2288

**Changes**

Prior to this PR, `@delete`, `@forceDelete` and `@restore` relied upon the field they were used upon just defining a single
argument to use as the ID/IDs to find the models to modify.

Now, they will leverage the same mechanisms as other directives like `@all` or `@find` do to find the models:
- scopes
- argument filter directives such as `@eq`
- any other way of influencing the query builder

A new directive `@whereKey` has been added to support the most common use case of finding by primary key.
In addition, this change opens up many new possibilities to leverage those directives, e.g. deleting all models that match
a set of conditions.

TODO: determine if the same change can be made to `@update` and `@upsert`, see https://github.com/nuwave/lighthouse/issues/1145.

**Breaking changes**

- Remove the `globalId` argument from `@delete`, `@forceDelete` and `@restore`
- Require filter directives such as `@whereKey` in `@delete`, `@forceDelete` and `@restore`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Expected resolver arguments in `ResolveInfo::enhanceBuilder()`
 - Pass the path array to `CacheKeyAndTags::key()` https://github.com/nuwave/lighthouse/pull/2176
 - Require implementations of `BatchedEntityResolver` to maintain the keys given in `array $representations` https://github.com/nuwave/lighthouse/pull/2286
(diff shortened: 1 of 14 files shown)


olivernybroe:
I think this is a great idea and in general we should rely more and more on using other directives for manipulating the query builder when a query builder is involved. 

It also removes the ambiguities of when is this an argument to a directive or when is it a new directive.

PR #2297, opened 2023-02-13
Bump class-finder

olivernybroe:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
This bumps the class-finder ot the newest version. According to changelog no BC's have happened, just new features was added in. 
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": "^8",
         "ext-json": "*",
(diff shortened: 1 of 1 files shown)


spawnia:
We can allow both, no need to force an upgrade.

olivernybroe:
@spawnia true :) Just so used to pushing for the higher version always :D

Issue #2303, opened 2023-02-15
@delete without @eq wipes whole table


DaLukasTI:
**Describe the bug**
I had a mutation `deleteProduct(id: ID!): Product @delete @can(ability: "delete")`
I upgraded my project to laravel 10 and this mutation behaves very weird. (I'm on the dev-master)
When using the migration as is, and calling the endpoint it deletes ALL entities in the product table(4000 records in my case).
when I add `@eq` as in `deleteProduct(id: ID! @eq): Product @delete @can(ability: "delete")` it only deletes the one product as expected. This is nowhere to be found in the docs in the docs it is stated as

`
type Mutation {
  deletePost(id: ID!): Post @delete
}` 

Is this intended behaviour or a bug, I'm not sure :sweat_smile: 

</details></br>

**Lighthouse Version**
dev-master

spawnia:
https://lighthouse-php.com/master/api-reference/directives.html#delete mentions using `@whereKey`, where were you looking? 

I could see adding a stronger warning against this error in the docs though, especially in the example with multiple optional filters to apply. We have to make sure that at least one filter is given.

DaLukasTI:
I see where it went wrong, if you are for example on the master docs and use the searchbar on the top, you get booted to the /5/ page of the query you searched, must have missed that. 
Is there a reason to use the `@wherekey` over `@eq` in those queries?

spawnia:
> I see where it went wrong, if you are for example on the master docs and use the searchbar on the top, you get redirected to the /5/ page of the query you searched, must have missed that.

Yeah, that is a known issue in https://github.com/nuwave/lighthouse/issues/1392.

> Is there a reason to use the @wherekey over @eq in those mutations?

`@whereKey` does not care about the name of the argument, it always uses the primary key of the model.

DaLukasTI:
Okay thanks for clearing that out! Should I leave this ticket open for visibility?

spawnia:
@DaLukasTI please review https://github.com/nuwave/lighthouse/pull/2304, it should at least guard against some potential issues.

The upgrade guide mentions the need to use filters in `@delete`, see https://github.com/nuwave/lighthouse/blob/master/UPGRADE.md#use-filters-in-delete-forcedelete-and-restore, but I could see how that can be forgotten. I wonder if we could add additional validation to ensure no mass deletions happen accidentally.

PR #2304, opened 2023-02-16
Require at least one builder enhancement in `@delete`

spawnia:
https://github.com/nuwave/lighthouse/issues/2303


tests/Integration/Schema/Directives/DeleteDirectiveTest.php
```diff
@@ -245,6 +245,27 @@ public function testDeletesMultipleEmptyInput(): void
...
+        }
+
+        type Mutation {
+            deleteUser(id: ID @whereKey, email: String @eq): User @delete
+        }
+        ';
+
+        $this->graphQL(/** @lang GraphQL */ '
+        mutation {
+            deleteUser {
```
DaLukasTI, line 262: correct me if I'm wrong here but here gets checked if u call the mutation and provide an Id to delete. This was also the case in my example. I passed an Id to that function in #2303 but instead of deleting that one id I passed it deleted everything.

DaLukasTI (dismissed):

DaLukasTI (commented):
see notes on DeleteDirectiveTest

tests/Integration/Schema/Directives/DeleteDirectiveTest.php
spawnia: Right, this will not specifically solve what went wrong in your case. Perhaps we can also fail during schema validation if a field with `@delete` has arguments without any `ArgBuilderDirective`?

spawnia: Alright, I have implemented a new solution that dynamically checks if there would be any modifications to the query builder.

pyrou (approved):

Issue #2301, opened 2023-02-15
Fails to install when using Laravel 10


AnthoniG:
When creating a fresh install of Laravel, it now defaults to v10
However when I try to install lighthouse, composer outputs the following :-

```
❯ composer require nuwave/lighthouse
./composer.json has been updated
Running composer update nuwave/lighthouse
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - nuwave/lighthouse[v0.1.0, ..., v0.7.8, v1.0.0] require illuminate/support ^5.2 -> found illuminate/support[v5.2.0, ..., v5.8.36] but these were not loaded, likely because it conflicts with another require.
    - nuwave/lighthouse[v1.0.1, ..., v1.1.0] require illuminate/console ^5.2 || ^5.3 -> found illuminate/console[v5.2.0, ..., v5.8.36] but these were not loaded, likely because it conflicts with another require.
    - nuwave/lighthouse[v2.0.0, ..., v2.2.1] require illuminate/contracts ^5.4 -> found illuminate/contracts[v5.4.0, ..., v5.8.36] but these were not loaded, likely because it conflicts with another require.
    - nuwave/lighthouse[v2.3, ..., v2.6.4] require illuminate/contracts 5.5.*|5.6.*|5.7.* -> found illuminate/contracts[v5.5.0, ..., v5.7.28] but these were not loaded, likely because it conflicts with another require.
    - nuwave/lighthouse[v3.0.0, ..., v3.7.0, v4.0.0, ..., v4.1.1] require illuminate/contracts 5.5.*|5.6.*|5.7.*|5.8.* -> found illuminate/contracts[v5.5.0, ..., v5.8.36] but these were not loaded, likely because it conflicts with another require.
    - nuwave/lighthouse[v4.2.0, ..., v4.9.0] require illuminate/contracts 5.5.* || 5.6.* || 5.7.* || 5.8.* || ^6.0 -> found illuminate/contracts[v5.5.0, ..., v5.8.36, v6.0.0, ..., v6.20.44] but these were not loaded, likely because it conflicts with another require.
[...]

PR #2307, opened 2023-02-17
[6.x] Resolved todo's

wimski:
> There are about 50 code comments marked as `TODO` throughout the codebase, some portion of which specifically mention changes to make before releasing v6 or after dropping older Laravel/PHPStan/PHP versions. Feel free to pick any of those and post a pull request.

https://github.com/nuwave/lighthouse/issues/2301#issuecomment-1434380958

- [X] Added or updated tests
- [x] Documented user facing changes (do any of these changes require documentation updates?)
- [X] Updated CHANGELOG.md

### Changes

I've gone through the todo's and updated the ones that mentioned `v6` or regarded something which is supported with the current requirements.

#### Open

- [x] [`PaginationManipulator:248`](https://github.com/nuwave/lighthouse/blob/ad11e2c4b2cd6596e50ca06d6488d354397e7560/src/Pagination/PaginationManipulator.php#L248): should the `defaultCount` functionality be removed altogether?
```php
        // TODO always add ! in v6
        $definition = 'first: Int'
            . ($defaultCount
                ? " =  {$defaultCount}"
                : '!');
```
- [x] [`ComplexityDirective:62`](https://github.com/nuwave/lighthouse/blob/ad11e2c4b2cd6596e50ca06d6488d354397e7560/src/Schema/Directives/ComplexityDirective.php#L62): what is the expected change exactly?
```php
        return
            // Default complexity for this field itself
            // TODO add this in v6
            // 1 +
            // Scale children complexity by the expected number of results
            $childrenComplexity * $expectedNumberOfChildren;
```
- [x] [`Contracts\GlobalId:8`](https://github.com/nuwave/lighthouse/blob/ad11e2c4b2cd6596e50ca06d6488d354397e7560/src/Support/Contracts/GlobalId.php#L8): move where?
* `Nuwave\Lighthouse\GlobalId`
* `Nuwave\Lighthouse\GlobalId\Contracts`
* `Nuwave\Lighthouse\Support\Contracts\GlobalId`
```php
/**
 * Encode and decode globally unique IDs.
 *
 * TODO move to GlobalId namespace in v6
 */
interface GlobalId
```


src/GraphQL.php
```diff
@@ -26,7 +26,6 @@
 use Nuwave\Lighthouse\Events\StartExecution;
 use Nuwave\Lighthouse\Events\StartOperationOrOperations;
 use Nuwave\Lighthouse\Execution\BatchLoader\BatchLoaderRegistry;
-use Nuwave\Lighthouse\Execution\DataLoader\BatchLoader;
```
spawnia, line 29: This class should be deleted entirely.

tests/Unit/Pagination/PaginationTypeTest.php
```diff
@@ -0,0 +1,38 @@
...
+    public static function invalidPaginationTypes(): array
+    {
+        return [
+            ['paginator'],
+            ['simple'],
+            ['connection'],
+            ['relay'],
+            ['default'],
+            ['foo'],
+        ];
```
spawnia, line 36:
```suggestion
    /**
     * @return iterable<array{string}>
     */
    public static function invalidPaginationTypes(): iterable
    {
        yield ['paginator'];
        yield ['simple'];
        yield ['connection'];
        yield ['relay'];
        yield ['default'];
        yield ['foo'];
```

tests/Unit/Pagination/PaginationTypeTest.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * @dataProvider invalidPaginationTypes
+     *
+     * @param string $type An invalid type
+     */
+    public function testThrowsExceptionForUnsupportedTypes(string $type): void
+    {
+        self::expectException(DefinitionException::class);
+        self::expectExceptionMessage("Found invalid pagination type: {$type}");
```
spawnia, line 19:
```suggestion
        self::expectExceptionObject(new DefinitionException("Found invalid pagination type: {$type}"));
```

spawnia:
Thanks for digging right in.

> do any of these changes require documentation updates?

If significant breakage is expected that is not easily discovered by basic static analysis or requires tricky reworks, the changes should be mentioned in `UPGRADE.md`. A bunch of things are already in there, just have a look through and consider if anything you changed here needs to be added too.

I am solving the open questions you have in https://github.com/nuwave/lighthouse/pull/2308.

src/GraphQL.php
wimski: Also remove these subsequent classes?
* `Tests\Utils\BatchLoaders\UserLoader`
* `Tests\Integration\Execution\DataLoader\RelationBatchLoaderTest`
* `Tests\Integration\Execution\DataLoader\RelationCountBatchLoaderTest`

spawnia: Only `UserLoader`, the tests cover `RelationBatchLoader` which is kept.

wimski: The reason I asked is because `UserLoader` is referenced and used in both those test classes. So what to do about that? Remove the test methods which use the `UserLoader`?

wimski:
> Thanks for digging right in.

I have a fresh Laravel 10 project which needs a GraphQL API asap, so I'm not going sit idly by and wait two months :wink:  

> If significant breakage is expected that is not easily discovered by basic static analysis or requires tricky reworks, the changes should be mentioned in `UPGRADE.md.`

I don't see any changes which match that description, but if I'm wrong about that, let me know.

PR #2320, opened 2023-02-24
Support Laravel 10

stayallive:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

There are some PHPStan changes done, this is not entirely specific to this branch since `5.x` branch with latest `composer update` does the same. PHPStan 1.10 made some big changes and since we are on the bleeding edge and enable almost everything it's being tripped on it. So I decided to remove the bleeding edge and disable the new `treatPhpDocTypesAsCertain` for `5.x`, this might be or not be acceptable but since `v6` is just a few weeks away (where these changes are already made to satisfy PHPStan) I think this makes sense so we can continue to work on `5.x` _if needed_ without doing large swaths of changes to make PHPStan happy.

This is mainly done to keep the Laravel 10 upgraders moving forward and get some pressure of the `v6` release so time can be taken to finish that properly and allow people to upgrade to `v6` (which has a lot of changes) at a later time instead of together with Laravel 10 😄 

**Changes**

Laravel 10.x compatibility.

**Breaking changes**

None.


spawnia:
Thanks for putting in the work, released with https://github.com/nuwave/lighthouse/releases/tag/v5.71.0.

PR #2321, opened 2023-02-24
Remove schema cache v1

thekonz:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Schema cache v1 is dead, long live schema cache v2.

**Breaking changes**

Shouldn't really break things, but it could for some people who have a hard dependency of the schema appearing in their configured cache storage.


spawnia (approved):
Nicely done!

PR #2344, opened 2023-03-09
Cast storage_ttl from env to int

Stevemoretz:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

After you have set env

```env
LIGHTHOUSE_SUBSCRIPTION_STORAGE_TTL=3600
```

And run this:

```bash
php artisan lighthouse:ide-helper
```

you get:

```
  Cannot assign string to property Nuwave\Lighthouse\Subscriptions\Storage\RedisStorageManager::$ttl of type ?int
```

That's because env() function of laravel returns string even if you have a number in your env, it can't understand that's an integer that's why this pull request fixes that bug, any where else in your codebase that you've done the same thing the same error might appear.

**Changes**

A little cast to fix a little bug.

**Breaking changes**

No

### DIFF
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -42,7 +42,7 @@ class RedisStorageManager implements StoresSubscriptions
         if ($ttl) {
             $ttl = (int) $ttl;
         }
-        $this->ttl = (int) $ttl;
+        $this->ttl = $ttl;
     }
 
     public function subscriberByChannel(string $channel): ?Subscriber
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -38,7 +38,11 @@ class RedisStorageManager implements StoresSubscriptions
         $this->connection = $redis->connection(
             $config->get('lighthouse.subscriptions.broadcasters.echo.connection') ?? 'default'
         );
-        $this->ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        $ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        if ($ttl) {
+            $ttl = (int) $ttl;
+        }
+        $this->ttl = (int) $ttl;
     }
 
     public function subscriberByChannel(string $channel): ?Subscriber
@@ -68,7 +72,7 @@ class RedisStorageManager implements StoresSubscriptions
         // This is like using multiple get calls (getSubscriber uses the get command).
         $subscribers = $this->connection->command('mget', [$subscriberIds]);
 
-        return (new Collection($subscribers))
+        return ( new Collection($subscribers) )
             ->filter()
             ->map(static function (?string $subscriber): ?Subscriber {
                 // Some entries may be expired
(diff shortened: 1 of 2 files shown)


src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -38,7 +38,11 @@ public function __construct(ConfigRepository $config, RedisFactory $redis)
         $this->connection = $redis->connection(
             $config->get('lighthouse.subscriptions.broadcasters.echo.connection') ?? 'default'
         );
-        $this->ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        $ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        if (is_string($ttl)) {
+            $ttl = (int) $ttl;
+        }
+        $this->ttl = $ttl;
```
spawnia, line 45: We can directly assign if the value `is_int()`. If it `is_string()`, we should also check it `is_numeric()` before casting, in all other cases we should throw and indicate the value is neither an int or a numeric string.

Stevemoretz: Good idea! Done

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -38,7 +38,14 @@ public function __construct(ConfigRepository $config, RedisFactory $redis)
...
             $config->get('lighthouse.subscriptions.broadcasters.echo.connection') ?? 'default'
         );
-        $this->ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        $ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        if (is_int($ttl)) {
+            $this->ttl = $ttl;
+        } elseif (is_string($ttl) && is_numeric($ttl)) {
+            $this->ttl = (int) $ttl;
+        } else {
+            throw new \ErrorException('ttl is not an integer');
```
spawnia, line 47:
```suggestion
            $notIntOrNumericString = Utils::printSafe($ttl);
            throw new \Exception("Expected config option lighthouse.subscriptions.storage_ttl to be an int or a numeric string, got: {$notIntOrNumericString}.");
```

`GraphQL\Utils\Utils`

Stevemoretz:
@spawnia  What is this about? https://github.com/nuwave/lighthouse/actions/runs/4384947478/jobs/7677039186

spawnia:
> @spawnia What is this about? [nuwave/lighthouse/actions/runs/4384947478/jobs/7677039186](https://github.com/nuwave/lighthouse/actions/runs/4384947478/jobs/7677039186)

Fails somewhat randomly since a couple of weeks ago.

src/Subscriptions/Storage/RedisStorageManager.php
```diff
@@ -38,7 +39,15 @@ public function __construct(ConfigRepository $config, RedisFactory $redis)
...
         );
-        $this->ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        $ttl = $config->get('lighthouse.subscriptions.storage_ttl');
+        if (is_int($ttl) || is_null($ttl)) {
+            $this->ttl = $ttl;
+        } elseif (is_string($ttl) && is_numeric($ttl)) {
+            $this->ttl = (int) $ttl;
+        } else {
+            $notIntOrNumericString = Utils::printSafe($ttl);
+            throw new \Exception("Expected config option lighthouse.subscriptions.storage_ttl to be an int or a numeric string, got: {$notIntOrNumericString}.");
```
spawnia, line 49:
```suggestion
            throw new \Exception("Expected config option lighthouse.subscriptions.storage_ttl to be an int, null or a numeric string, got: {$notIntOrNumericString}.");
```

PR #2345, opened 2023-03-10
Support multiple default guards

pyrou:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md
- [x] Change @auth(guard:) and @whereAuth(guard:) into @auth(guards:) and @whereAuth(guards:)

**Changes**

When using a query or mutation without `@guard` directive, to allow guest users, only one single default guard was attempted for $context->user. Specifying multiple guards with `@guard(with:)` allow to attempt multiples guard, but this would disallow guest users.

This PR propose to change the `lighthouse.guard` configuration to support multiple guards.

**Breaking changes**

- `lighthouse.guard` configuration key has been renamed `lighthouse.guards` and expect now an array.
- `@auth(guard: String)` has been changed to `@auth(guards: [String!])`
- `@whereAuth(guard: String)` has been changed to `@whereAuth(guards: [String!])`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Rename configuration option `schema.register` to `schema_path`
 - Restructure cache related configuration options
 - Register all non-user namespaces through `RegisterDirectiveNamespaces` event https://github.com/nuwave/lighthouse/pull/2337
+- Guard default configuration options in `lighthouse.php` supports multiple guards.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Rename configuration option `schema.register` to `schema_path`
 - Restructure cache related configuration options
 - Register all non-user namespaces through `RegisterDirectiveNamespaces` event https://github.com/nuwave/lighthouse/pull/2337
-- Guard default configuration options in `lighthouse.php` supports multiple guards.
+- Guard default configuration options in `lighthouse.php` supports multiple guards https://github.com/nuwave/lighthouse/pull/2345
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Rename configuration option `schema.register` to `schema_path`
 - Restructure cache related configuration options
 - Register all non-user namespaces through `RegisterDirectiveNamespaces` event https://github.com/nuwave/lighthouse/pull/2337
-- Guard default configuration options in `lighthouse.php` supports multiple guards https://github.com/nuwave/lighthouse/pull/2345
+- Guard default configuration options in `lighthouse.php` allows multiple guards https://github.com/nuwave/lighthouse/pull/2345
+- Make `@auth` and `@whereAuth` support multiple guards https://github.com/nuwave/lighthouse/pull/2345
 
 ### Fixed
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -321,7 +321,7 @@ the default Laravel authentication guard will be used (`auth.defaults.guard`).
 
 ### Update `@auth` and `@whereAuth` directives
 
-The `guard` attribute, on `@auth` and `@whereAuth` directives, has been renamed `guards` and expects an array.
+The `guard` argument on `@auth` and `@whereAuth` directives has been renamed to `guards` and now expects a list instead of a single string.
 
 ```diff
 - @auth(guard: "api")
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -319,6 +319,17 @@ The `lighthouse.guard` configuration key was renamed to `lighthouse.guards` and 
 If `lighthouse.guards` configuration is missing,
 the default Laravel authentication guard will be used (`auth.defaults.guard`).
 
+### Update `@auth` and `@whereAuth` directives
+
+The `guard` attribute, on `@auth` and `@whereAuth` directives, has been renamed `guards` and expects an array.
+
+```diff
+- @auth(guard: "api")
++ @auth(guards: ["api"])
+- @whereAuth(guard: "api")
++ @whereAuth(guards: ["api"])
+```
+
 ## v4 to v5
 
 ### Update PHP, Laravel and PHPUnit
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -307,7 +307,7 @@ in your `app/config.php`, it is no longer registered by default.
 ],
 ```
 
-### Update lighthouse.guard configuration
+### Update `lighthouse.guard` configuration
 
 The `lighthouse.guard` configuration key was renamed to `lighthouse.guards` and expects an array.
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -307,6 +307,18 @@ in your `app/config.php`, it is no longer registered by default.
 ],
 ```
 
+### Update lighthouse.guard configuration
+
+The `lighthouse.guard` configuration key was renamed to `lighthouse.guards` and expects an array.
+
+```diff
+- 'guard' => 'api',
++ 'guards' => ['api'],
+```
+
+If `lighthouse.guards` configuration is missing,
+the default Laravel authentication guard will be used (`auth.defaults.guard`).
+
 ## v4 to v5
 
 ### Update PHP, Laravel and PHPUnit
--- a/src/Auth/AuthDirective.php
+++ b/src/Auth/AuthDirective.php
@@ -47,16 +47,14 @@ GRAPHQL;
      */
     protected function authenticatedUser(array $guards): ?Authenticatable
     {
-        $user = null;
-
         foreach ($guards as $guard) {
-            $user = $this->authFactory->guard($guard)->user();
-
-            if (isset($user)) {
-                break;
+            $user = $this->authFactory->guard($guard)
+                ->user();
+            if ($user !== null) {
+                return $user;
             }
         }
 
-        return $user;
+        return null;
     }
 }
--- a/src/Auth/AuthDirective.php
+++ b/src/Auth/AuthDirective.php
@@ -22,10 +22,10 @@ Return the currently authenticated user as the result of a query.
 """
 directive @auth(
   """
-  Specify which guard to use, e.g. "api".
+  Specify which guards to use, e.g. ["api"].
   When not defined, the default from `lighthouse.php` is used.
   """
-  guard: String
+  guards: [String!]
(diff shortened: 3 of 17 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Rename configuration option `schema.register` to `schema_path`
 - Restructure cache related configuration options
 - Register all non-user namespaces through `RegisterDirectiveNamespaces` event https://github.com/nuwave/lighthouse/pull/2337
+- Guard default configuration options in `lighthouse.php` supports multiple guards.
```
spawnia, line 41:
```suggestion
- Guard default configuration options in `lighthouse.php` supports multiple guards https://github.com/nuwave/lighthouse/pull/2345
```

src/Auth/AuthDirective.php
```diff
@@ -33,7 +33,7 @@ public static function definition(): string
     public function resolveField(FieldValue $fieldValue): callable
     {
         return function (): ?Authenticatable {
-            $guard = $this->directiveArgValue('guard', AuthServiceProvider::guard());
+            $guard = $this->directiveArgValue('guard', current(AuthServiceProvider::guards()) ?: null);
```
spawnia, line 36: What does `current()` do? What is the intent?

src/Auth/AuthServiceProvider.php
```diff
@@ -15,14 +15,21 @@ public function boot(Dispatcher $dispatcher): void
         $dispatcher->listen(RegisterDirectiveNamespaces::class, static fn (): string => __NAMESPACE__);
     }
 
-    public static function guard(): ?string
+    /**
+     * @return string[]
```
spawnia, line 19:
```suggestion
     * @return array<string>
```

src/Auth/AuthServiceProvider.php
```diff
@@ -15,14 +15,21 @@ public function boot(Dispatcher $dispatcher): void
...
-        $lighthouseGuard = $config->get('lighthouse.guard');
         $guards = $config->get('auth.guards');
 
-        return isset($guards[$lighthouseGuard])
-            ? $lighthouseGuard
-            : $config->get('auth.defaults.guard');
+        $lighthouseGuards = array_filter(
+            (array) $config->get('lighthouse.guards', []),
+            static fn (string $guard): bool => isset($guards[$guard]),
+        );
```
spawnia, line 29: I think we should throw if there is a mismatch here

src/Auth/WhereAuthDirective.php
```diff
@@ -47,7 +47,7 @@ public function handleFieldBuilder(QueryBuilder|EloquentBuilder|Relation $builde
             function (object $query): void {
                 assert($query instanceof EloquentBuilder);
 
-                $guard = $this->directiveArgValue('guard', AuthServiceProvider::guard());
+                $guard = $this->directiveArgValue('guard', current(AuthServiceProvider::guards()) ?: null);
```
spawnia, line 50: Since this pattern is repeated, we might want a method for it

src/Schema/Context.php
```diff
@@ -12,15 +12,23 @@ class Context implements GraphQLContext
...
     ) {
-        $this->user = $request->user(AuthServiceProvider::guard());
+        foreach (AuthServiceProvider::guards() as $guard) {
+            $user = $request->user($guard);
+
+            if (null !== $user) {
+                $this->user = $user;
+
+                break;
+            }
```
spawnia, line 30:
```suggestion
            $this->user = $request->user($guard);
            if (isset($this->user)) {
                break;
            }
```

src/Schema/Context.php
```diff
@@ -12,15 +12,23 @@ class Context implements GraphQLContext
     /**
      * An instance of the currently authenticated user.
      */
-    public ?Authenticatable $user;
+    public ?Authenticatable $user = null;
```
spawnia, line 15:
```suggestion
    public ?Authenticatable $user;
```

src/lighthouse.php
```diff
@@ -48,16 +48,18 @@
...
     | Used in directives such as `@guard` or the `AttemptAuthentication` middleware.
-    | Falls back to the Laravel default if the defined guard is either `null` or not found.
+    | Falls back to the Laravel default if the defined guards is `null`.
     |
     */
 
-    'guard' => 'api',
+    'guards' => [
+        'api',
+    ],
```
spawnia, line 62:
```suggestion
    'guards' => ['api'],
```

src/lighthouse.php
```diff
@@ -48,16 +48,18 @@
...
     |--------------------------------------------------------------------------
-    | Authentication Guard
+    | Authentication Guards
     |--------------------------------------------------------------------------
     |
-    | The guard to use for authenticating GraphQL requests, if needed.
+    | The guards to use for authenticating GraphQL requests, if needed.
     | Used in directives such as `@guard` or the `AttemptAuthentication` middleware.
-    | Falls back to the Laravel default if the defined guard is either `null` or not found.
+    | Falls back to the Laravel default if the defined guards is `null`.
```
spawnia, line 56:
```suggestion
    | Falls back to the Laravel default if `null`.
```

tests/Unit/Auth/UserContextTest.php
```diff
@@ -0,0 +1,104 @@
...
+            name: String!
+        }
+
+        type Query {
+            user: User @mock
+        }
+        ';
+
+        $this->mockResolver(function ($_, $args, $context) {
+            return $context->user();
```
spawnia, line 27: use a short closure

tests/Unit/Auth/UserContextTest.php
```diff
@@ -0,0 +1,102 @@
...
+        type User {
+            name: String!
+        }
+
+        type Query {
+            user: User @mock
+        }
+        ';
+
+        $this->mockResolver(static fn ($_, $args, $context) => $context->user());
```
spawnia, line 26:
```suggestion
        $this->mockResolver(static fn ($_, array $args, GraphQLContext $context): ?User => $context->user());
```

src/Schema/Context.php
pyrou: If there is no guard at all, calling $context->user() would result in following error:
> Error : Typed property Nuwave\Lighthouse\Schema\Context::$user must not be accessed before initialization

I now added a $this->user = null; before the foreach loop.

src/Auth/AuthDirective.php
pyrou: Current is used here to get the first item of AuthServiceProvider::guards()

It fallback nicely to `false` on empty array. Whereas AuthServiceProvider::guards()[0] that would throw on empty array.

Unlink `@guard`, the `@auth` directive isn't meant to be mutli-guard currently in LH (should it ?) So I made choice to fallback on first configured guard. 

Other option would be, if no guard attribute is specified on `@auth`, to walk against all guards to find first authenticated. But I found this a bit strange that without attribute multiple guards would be supported, and with attribute only one guard is possible.

Let me now what do you prefer.

src/Auth/AuthServiceProvider.php
pyrou: Nice idea,

I am very bad at naming (and english too obviously) ^^ would this be ok ?
InvalidConfigurationException("Guard ${guard} is not a known guard.")

src/Auth/AuthDirective.php
spawnia: I think we should also allow multiple guards here.

src/Auth/WhereAuthDirective.php
spawnia: Or not, we should probably just allow multiple guards here.

src/Auth/AuthServiceProvider.php
pyrou: This obviously become way more readable :) 

```php
public static function guards(): array
{
    $config = Container::getInstance()->make(ConfigRepository::class);
    
    return $config->get('lighthouse.guards', [$config->get('auth.guards')]);
}
```

src/Auth/AuthDirective.php
pyrou: Just to be sure I clearly understand your proposition,

So we change `@auth(guard: String)` and `@whereAuth(guard: String)` into something like
`@auth(with: [String!])` and `@whereAuth(with: [String!])` to be consistent with `@guard(with:)` ?

This will introduce much more breaking changes than I expected :)

spawnia: I was thinking `@auth(guards: [])`.

> This will introduce much more breaking changes than I expected :)

pyrou: Ok, I'll do this later this week-end, looks like this will involve a lot of changes, tests, documentation, ...
 
I'm glad you are so open to such API changes !

spawnia: Sweet, thanks! I believe cleaning up inconsistencies like this will be beneficial in the long run.

src/Auth/AuthServiceProvider.php
```diff
@@ -15,14 +15,14 @@ public function boot(Dispatcher $dispatcher): void
...
     {
         $config = Container::getInstance()->make(ConfigRepository::class);
-        $lighthouseGuard = $config->get('lighthouse.guard');
-        $guards = $config->get('auth.guards');
 
-        return isset($guards[$lighthouseGuard])
-            ? $lighthouseGuard
-            : $config->get('auth.defaults.guard');
+        return $config->get('lighthouse.guards')
+            ?? [$config->get('auth.default.guard')];
```
pyrou, line 26:
```suggestion
            ?? [$config->get('auth.defaults.guard')];
```

pyrou: I was first surprised existing unit tests was passing with this typo, but in this end, this resulted with a null guard that behaves the same as if it was the default guard.

UPGRADE.md
```diff
@@ -307,6 +307,29 @@ in your `app/config.php`, it is no longer registered by default.
...
+- 'guard' => 'api',
++ 'guards' => ['api'],
+```
+
+If `lighthouse.guards` configuration is missing,
+the default Laravel authentication guard will be used (`auth.defaults.guard`).
+
+### Update `@auth` and `@whereAuth` directives
+
+The `guard` attribute, on `@auth` and `@whereAuth` directives, has been renamed `guards` and expects an array.
```
spawnia, line 324:
```suggestion
The `guard` argument on `@auth` and `@whereAuth` directives has been renamed to `guards` and now expects a list instead of a single string.
```

UPGRADE.md
```diff
@@ -307,6 +307,29 @@ in your `app/config.php`, it is no longer registered by default.
 ],
 ```
 
+### Update lighthouse.guard configuration
```
spawnia, line 310:
```suggestion
### Update `lighthouse.guard` configuration
```

spawnia:
Nicely done @pyrou, thank you!

PR #2346, opened 2023-03-10
chore: cast `schema_cache.enable` to boolean

owenvoke:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

When [setting `LIGHTHOUSE_SCHEMA_CACHE_ENABLE` to `false` in a PHPUnit configuration](https://github.com/nuwave/lighthouse/blob/master/docs/master/performance/query-caching.md#testing-caveats), it seems that this get's parsed as an empty string. This then causes the env to be loaded incorrectly and ends up throwing an error with `Cannot assign string to property Nuwave\\Lighthouse\\Schema\\AST\\ASTCache::$enable of type bool` since the ASTCache was changed to be strictly typed properties.

I'm not sure if this would be considered an issue with PHPUnit (9) not parsing the XML correctly, however this hasn't been an issue until the properties were typed. I guess because before the empty string was treated as `false`. 🤷🏻

**Breaking changes**

### DIFF
--- a/src/Schema/AST/ASTCache.php
+++ b/src/Schema/AST/ASTCache.php
@@ -25,7 +25,7 @@ class ASTCache
     {
         /** @var CacheConfig $cacheConfig */
         $cacheConfig = $config->get('lighthouse.schema_cache');
-        $this->enable = $cacheConfig['enable'];
+        $this->enable = (bool) $cacheConfig['enable'];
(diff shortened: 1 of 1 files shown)


pyrou (approved):

pyrou:
Introduced with https://github.com/nuwave/lighthouse/pull/2316

PR #2352, opened 2023-03-16
remove subscriptions version 1

thekonz:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**(Breaking) Changes**

Removes subscriptions version 1.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -722,3 +722,24 @@ If you use complex where condition directives, such as `@whereConditions`,
 upgrade `mll-lab/graphql-php-scalars` to v4:
 
     composer require mll-lab/graphql-php-scalars:^4
+
+### Subscriptions version 1 removal 
+
+Subscriptions only use version 2 now. That means, the extensions content
+will not contain the `channels` key anymore and the `version` will be `2`.
+
+```diff
+{
+  "data": {...},
+  "extensions": {
+    "lighthouse_subscriptions": {
+-      "version": 1,
++      "version": 2,
+      "channel": "channel-name",
+-      "channels": {
+-        "subscriptionName": "channel-name"
+-      }
+    }
+  }
+}
+```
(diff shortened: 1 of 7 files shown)


UPGRADE.md
```diff
@@ -722,3 +722,24 @@ If you use complex where condition directives, such as `@whereConditions`,
...
+Subscriptions only use version 2 now. That means, the extensions content
+will not contain the `channels` key anymore and the `version` will be `2`.
+
+```diff
+{
+  "data": {...},
+  "extensions": {
+    "lighthouse_subscriptions": {
+-      "version": 1,
++      "version": 2,
```
spawnia, line 737: I am wondering if we should remove this key altogether.

thekonz: was wondering the same, but this already existed before v2... @stayallive you got an idea about this?

stayallive: Since there are no longer multiple versions (probably was done out of future proofing) I'd say remove it for now, if we need to add it in the future we can always add it back 😄

spawnia: Let's get rid of it. Next time maybe we can name the versions `before` and `after` to make it clear this is temporary.

thekonz: removed it

src/Subscriptions/SubscriptionRegistry.php
```diff
@@ -132,31 +132,15 @@ public function handleStartExecution(StartExecution $startExecution): void
...
-            ],
-            2 => [
-                'version' => 2,
-                'channel' => $channel,
-            ],
-            default => throw new DefinitionException("Expected lighthouse.subscriptions.version to be 1 or 2, got: {$version}"),
-        };
-
-        return new ExtensionsResponse('lighthouse_subscriptions', $content);
+        return new ExtensionsResponse('lighthouse_subscriptions', compact('channel'));
```
spawnia, line 143: oh god no, not `compact`

spawnia: I would not consider Laravel a pinnacle of readable code.

PR #2369, opened 2023-03-27
Allow dynamically added manipulator directives

jaulz:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

It would be great to be able to define nested directives that should be applied to a certain field:
```php
  /**
   * Manipulate the AST based on a field definition.
   */
  public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefinitionNode &$fieldDefinition, ObjectTypeDefinitionNode|InterfaceTypeDefinitionNode &$parentType): void
  {
    $fieldDefinition->directives[] = Parser::directive('@paginate');
    $fieldDefinition->directives[] = Parser::directive('@anotherFieldDefinitionManipulator');
  }
```

At the moment, only the initially defined directives are evaluated and hence both of the directives in the example would be ignored (at least their `manipulateFieldDefinition` part). The suggested change looks for any new directives and stops when there are no new ones defined. 

This change should be applied in a similar fashion to the other manipulators as well.

If you think this is fine, I can also implement related tests.

**Breaking changes**

n/a


src/Schema/AST/ASTBuilder.php
```diff
@@ -96,11 +96,14 @@ public function build(): DocumentAST
...
-                as $typeManipulator
-            ) {
+            $executedManipulators = [];
+            while (
+                $typeManipulator = $this->directiveLocator->associatedOfType($typeDefinition, TypeManipulator::class)
+                    ->filter(fn ($typeManipulator) => !in_array(spl_object_hash($typeManipulator->directiveNode), $executedManipulators))
+                    ->first()
+            ) { 
                 $typeManipulator->manipulateTypeDefinition($this->documentAST, $typeDefinition);
+                $executedManipulators[] = spl_object_hash($typeManipulator->directiveNode);
```
spawnia, line 106:
```suggestion
                $executedManipulators[spl_object_id($typeManipulator->directiveNode)] = true;;
```

src/Schema/AST/ASTBuilder.php
```diff
@@ -96,11 +96,14 @@ public function build(): DocumentAST
...
     {
         foreach ($this->documentAST->types as $typeDefinition) {
-            foreach (
-                $this->directiveLocator->associatedOfType($typeDefinition, TypeManipulator::class)
-                as $typeManipulator
-            ) {
+            $executedManipulators = [];
+            while (
+                $typeManipulator = $this->directiveLocator->associatedOfType($typeDefinition, TypeManipulator::class)
+                    ->filter(fn ($typeManipulator) => !in_array(spl_object_hash($typeManipulator->directiveNode), $executedManipulators))
```
spawnia, line 102: missing types

jaulz:
@spawnia great, thanks a lot for your quick review! 😊 I will incorporate your changes and also try to add tests.

.gitpod.yml
spawnia, line 1: You can add this file to `.gitignore` to document that I do not plan to maintain such a config file.

src/Schema/AST/ASTBuilder.php
```diff
@@ -96,11 +97,15 @@ public function build(): DocumentAST
...
         foreach ($this->documentAST->types as $typeDefinition) {
-            foreach (
-                $this->directiveLocator->associatedOfType($typeDefinition, TypeManipulator::class)
-                as $typeManipulator
-            ) {
+            /** @var array<int, bool> $executedManipulators */
+            $executedManipulators = [];
+            while (
+                $typeManipulator = $this->directiveLocator->associatedOfType($typeDefinition, TypeManipulator::class)
+                    ->filter(fn (BaseDirective $typeManipulator) => !isset($executedManipulators[spl_object_id($typeManipulator->directiveNode)]))
```
spawnia, line 104:
```suggestion
                    ->reject(fn (TypeManipulator $typeManipulator): bool => isset($executedManipulators[spl_object_id($typeManipulator->directiveNode)]))
```

jaulz: Unfortunately, that will result in:
```
Access to an undefined property Nuwave\Lighthouse\Support\Contracts\TypeManipulator::$directiveNode
```

jaulz:
@spawnia any good idea where I could add a test? Unfortunately, I could not find anything related to manipulators yet.

src/Schema/AST/ASTBuilder.php
spawnia: I see. Not every directive has to be of type `BaseDirective` though. `associatedOfType` could return a class that just `implements TypeManipulor, Directive`.

In case we do not have access to the directive node, I am not sure what we could use to know if the directive has executed.

spawnia:
> @spawnia any good idea where I could add a test? Unfortunately, I could not find anything related to manipulators yet.

Add to `ASTBuilderTest`.

.gitignore
```diff
@@ -6,6 +6,7 @@
 .idea
 .vscode
 phpunit.xml
+.gitignore
```
spawnia, line 9: `.gitpod.yml`

jaulz: Sorry, seems I am not yet awake 😴

spawnia: Here, have a ☕

jaulz:
Coffee helped, could you please have another look? 😉

src/Schema/AST/ASTBuilder.php
```diff
@@ -225,20 +237,37 @@ protected function applyArgManipulators(): void
...
+    /**
+     * Return unique manipulator identifier.
+     */
+    protected function getManipulatorId(FieldManipulator|ArgManipulator|TypeManipulator $manipulator): string
+    {
+        if ($manipulator instanceof BaseDirective) {
+            return spl_object_hash($manipulator->directiveNode);
+        }
+
+        return get_class($manipulator);
```
spawnia, line 271: The different logic here is actually a huge unexpected gotcha, it basically forbids non-`BaseDirective` manipulators from being used repeatedly. Although unlikely, this is a breaking change.

If we don't find another solution, we can hide the repeated application of manipulators behind a configuration flag to make it opt-in for now. However, I would like to see a plan on how we could resolve this gotcha in an upcoming major version - perhaps by adding a method `directiveNode(): DirectiveNode` to the `Directive` interface?

spawnia: > since this feature would only add extra functionality everything will still work as before

I am saying it would not, so it will not. A schema that has multiple uses of the same non-`BaseDirective` at the same location would behave differently.

jaulz:
@spawnia having a second thought, I think it might be even easier to pass the `$directiveLocator` to `manipulate*Definition` (or hydrate). So instead of:
```php
$fieldDefinition->directives[] = Parser::directive('@cursorPaginate');
```

We could do this:
```php
$fooDirective = $directiveLocator->resolve('foo');
$directiveLocator->create('foo')->hydrate($fooDirective, $fieldDefinition)->manipulateFieldDefinition($documentAST, $fieldDefinition, $parentType);
```

Eventually we could package this also in a utility method like `addDynamicDirective` in `BaseDirective`. This would have the advantage that we can keep the old logic in `ASTBuilder` because the final types are actually already built within the "composite" directive itself. Also, we already see the built node in the `manipulate*Definition` method so (for whatever reason) we could even manipulate the node after a dynamic directive was added.

spawnia:
You can resolve the `DirectiveLocator` from the service container in the `__construct()` method of your directive or inside of your `handle*Definition` method.

``php
    public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefinitionNode &$fieldDefinition, ObjectTypeDefinitionNode|InterfaceTypeDefinitionNode &$parentType): void
    {
        $directiveNode = Parser::directive('@foo(arg: 123)');
        $fieldDefinition->directives []= $directiveNode;

        $directiveLocator = Container::getInstance()->make(DirectiveLocator::class);
        $directiveInstance = $directiveLocator->create($directiveNode->name->value);
        assert($directiveInstance instanceof BaseDirective && $directiveInstance instanceof FieldManipulator);
        $directiveInstance->hydrate($directiveNode, $fieldDefinition);
        $directiveInstance->manipulateFieldDefinition($documentAST, $fieldDefinition, $parentType);
    }
```

Perhaps we can just extract the following utility?

```php
    public static function addDirectiveToNode(string $directiveSource, ScalarTypeDefinitionNode|ScalarTypeExtensionNode|ObjectTypeDefinitionNode|ObjectTypeExtensionNode|InterfaceTypeDefinitionNode|InterfaceTypeExtensionNode|UnionTypeDefinitionNode|UnionTypeExtensionNode|EnumTypeDefinitionNode|EnumTypeExtensionNode|InputObjectTypeDefinitionNode|InputObjectTypeExtensionNode|FieldDefinitionNode|InputValueDefinitionNode|EnumValueDefinitionNode $node): Directive
    {
        $directiveNode = Parser::directive($directiveSource);
        $node->directives []= $directiveNode;

        $directiveLocator = Container::getInstance()->make(DirectiveLocator::class);
        $directiveInstance = $directiveLocator->create($directiveNode->name->value);
        if ($directiveInstance instanceof BaseDirective) {
            $directiveInstance->hydrate($directiveNode, $node);
        }

        return $directiveInstance;
    }
```

src/Schema/AST/ASTBuilder.php
```diff
@@ -241,4 +241,4 @@ protected function applyArgManipulators(): void
             }
         }
     }
-}
+}
```
spawnia, line 244:
```suggestion

```

src/Schema/Directives/BaseDirective.php
spawnia, line 1: Please fix formatting with `make fix`.

src/Schema/Directives/BaseDirective.php
```diff
@@ -292,4 +301,101 @@ protected function loadArgValues(): void
             $this->directiveArgs[$node->name->value] = AST::valueFromASTUntyped($node->value);
         }
     }
+
+    /**
+     * Get an instance of a directive.
+     *
+     * @api
+     */
+    public static function getDirectiveInstance(string $directiveSource, ScalarTypeDefinitionNode|ScalarTypeExtensionNode|ObjectTypeDefinitionNode|ObjectTypeExtensionNode|InterfaceTypeDefinitionNode|InterfaceTypeExtensionNode|UnionTypeDefinitionNode|UnionTypeExtensionNode|EnumTypeDefinitionNode|EnumTypeExtensionNode|InputObjectTypeDefinitionNode|InputObjectTypeExtensionNode|FieldDefinitionNode|InputValueDefinitionNode|EnumValueDefinitionNode $node): Directive
```
spawnia, line 310: I don't like the name. `get...` indicates to me that this only retrieves something that was already there, and certainly not mutate its input. Why not `addDirectiveToNode` as proposed?

Also, this function and the other helpers should better go into `ASTHelper`. Given their static nature, nothing really binds them to `BaseDirective`.

src/Schema/Directives/BaseDirective.php
```diff
@@ -292,4 +301,101 @@ protected function loadArgValues(): void
...
+            || $typeDefinition instanceof InterfaceTypeExtensionNode
+            || $typeDefinition instanceof UnionTypeDefinitionNode
+            || $typeDefinition instanceof UnionTypeExtensionNode
+            || $typeDefinition instanceof EnumTypeDefinitionNode
+            || $typeDefinition instanceof EnumTypeExtensionNode
+            || $typeDefinition instanceof InputObjectTypeDefinitionNode
+            || $typeDefinition instanceof InputObjectTypeExtensionNode
+            || $typeDefinition instanceof FieldDefinitionNode
+            || $typeDefinition instanceof InputValueDefinitionNode
+            || $typeDefinition instanceof EnumValueDefinitionNode)) {
```
spawnia, line 348: Why not use the proper union type for the parameter?

src/Schema/Directives/BaseDirective.php
```diff
@@ -292,4 +301,101 @@ protected function loadArgValues(): void
...
+            throw new DefinitionException("The dynamically added directive [{$directiveSource}] can only be added to nodes that support directives.");
+        }
+
+        $directiveInstance = self::getDirectiveInstance($directiveSource, $typeDefinition);
+
+        if (!($directiveInstance instanceof TypeManipulator)) {
+            throw new DefinitionException("The dynamically added directive [{$directiveSource}] must implement the TypeManipulator interface.");
```
spawnia, line 355: I don't see why this would need to be the case to add a directive to a type definition.

I am suspicious if this method is worth adding, perhaps calling `manipulate*Definition` can just be done ad-hoc.

jaulz: Yep, sorry, you are right I just dropped it.

src/Schema/Directives/BaseDirective.php
jaulz: Done

jaulz:
I just incorporated all the feedback from you, hope it makes more sense now 😊

src/Schema/Directives/BaseDirective.php
spawnia: Some with the other two functions like this one.

jaulz: Sorry, could you please elaborate on this?

src/Schema/AST/ASTHelper.php
```diff
@@ -379,4 +390,64 @@ public static function internalFieldName(FieldDefinitionNode $field): string
...
+        ObjectTypeDefinitionNode|InterfaceTypeDefinitionNode &$parentType,
+    ): void {
+        $directiveInstance = self::addDirectiveToNode($directiveSource, $argDefinition);
+
+        if (!($directiveInstance instanceof ArgManipulator)) {
+            throw new DefinitionException("The dynamically added directive [{$directiveSource}] must implement the ArgManipulator interface.");
+        }
+
+        $directiveInstance->manipulateArgDefinition($documentAST, $argDefinition, $parentField, $parentType);
+    }
```
spawnia, line 452: I would prefer those functions to be removed and done ad-hoc instead.

jaulz: Sure, I can remove them. Would be open to make `ASTHelper` macroable then? Then it wouldn't be necessary to create an own class and keep everything there.

tests/Unit/Schema/AST/ASTHelperTest.php
```diff
@@ -210,4 +221,126 @@ public function testUnderlyingTypeKnowsStandardTypes(): void
...
+            public function manipulateFieldDefinition(
+                DocumentAST &$documentAST,
+                FieldDefinitionNode &$fieldDefinition,
+                ObjectTypeDefinitionNode|InterfaceTypeDefinitionNode &$parentType,
+            ):void {
+                ASTHelper::addDirectiveToFieldDefinition('@foo', $documentAST, $fieldDefinition, $parentType);
+            }
+        };
+
+        Utils::accessProtected($astBuilder, 'directiveLocator')->setResolved('dynamic', $dynamicDirective::class);
```
spawnia, line 262: replace all such uses.

src/Schema/AST/ASTHelper.php
spawnia: I don't see a compelling enough technical reason for a class like `ASTHelper` that is basically just a collection of static functions - which in an ideal world would just be plain old PHP functions. If it had internal state that macros needed access to, I could see it being useful.

jaulz: Okay, fine for me. It's removed 😊

spawnia (approved):

PR #2379, opened 2023-04-09
Add the ability to specify the disk the artisan command `lighthouse:print-schema` will write to via the `--disk` option

Carlwirkus:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

- Add the ability to specify the disk the schema will print to via the `--disk` option OR the config option `lighthouse.print_schema_disk`

**Breaking changes**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+- Add the ability to specify the disk the schema will print too via the `--disk` option OR the config option `lighthouse.print_schema_disk`
+
 ## v6.6.0
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v6.6.0
+
 ### Fixed
 
 - Correctly exclude all built-in types when calling `lighthouse:ide-helper` with `--omit-built-in` flag https://github.com/nuwave/lighthouse/pull/2376
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 ## Unreleased
 
 ### Added
-- Add the ability to specify the disk the schema will print too via the `--disk` option OR the config option `lighthouse.print_schema_disk`
+
+- Add the ability to specify the disk the artisan command `lighthouse:print-schema` will write to via the `--disk` option
 
 ## v6.6.0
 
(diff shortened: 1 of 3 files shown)


Carlwirkus:
Probably doesn't need the config option 🤷

src/lighthouse.php
```diff
@@ -449,4 +449,14 @@
...
+
+    /*
+    |--------------------------------------------------------------------------
+    | Print Schema Disk
+    |--------------------------------------------------------------------------
+    |
+    | Define what disk should be used when printing the schema to file.
+    |
+    */
+    'print_schema_disk' => env('LIGHTHOUSE_PRINT_SCHEMA_DISK', 'local'),
```
spawnia, line 461: Not a fan of this option, just the CLI flag should suffice.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+- Add the ability to specify the disk the schema will print too via the `--disk` option OR the config option `lighthouse.print_schema_disk`
```
spawnia, line 13:
```suggestion

- Add the ability to specify the disk the artisan command `lighthouse:print-schema` will write to via the `--disk` option
```

spawnia (changes requested):
I am not convinced that writing to a disk is useful anyways. How about we allow specifying a path to write to instead?

src/lighthouse.php
Carlwirkus: Sounds good

spawnia:
Have you tested if this still selects the default disk if the option is not specified? What happens when a non-existant disk is specified?

Carlwirkus:
> Have you tested if this still selects the default disk if the option is not specified?
Yes, it just selects the default disk.

> What happens when a non-existant disk is specified?
Not sure, I'll try tonight and see.

I see there's also issues with PHPStan, I'll see if I can get it setup locally and fix the conflicts. I've not worked with it before.

Carlwirkus:
I've tested with the a invalid and missing disk. 
- With invalid disk you'll just get an error 'xyz' disk does not exist. 
- With no disk it will fallback to the default disk.

As for the errors with PHPStan, I'm not very experienced with using it. I've just cast the value to a string or null but let me know if there's a preferred way of dealing with this.

src/Console/PrintSchemaCommand.php
```diff
@@ -52,8 +55,10 @@ public function handle(ASTCache $cache, Filesystem $storage, SchemaBuilder $sche
         }
 
         if ($this->option('write')) {
-            $storage->put($filename, $schemaString);
-            $this->info("Wrote schema to the default file storage (usually storage/app) as {$filename}.");
+            /** @var string|null $disk */
```
spawnia, line 58: This just ignores possible errors - PHPStan is correct that the value here might not be a string or null. Instead, do an actual check and throw if the value is anything else, indicating clearly to the user what was wrong.

Carlwirkus: I've added a check for a bool or array and returned an error to the console

Issue #2380, opened 2023-04-11
Simplify exposing root `Query` type in mutation responses


jaulz:
**What problem does this feature proposal attempt to solve?**

It would be great if there would be a way to resolve the `Query` type again in mutation responses like it's described here:
https://graphql-rules.com/rules/mutation-payload-query

Unfortunately, I did not find a nice way how this could be achieved via a custom directive but maybe I was looking in the wrong spots.

**Which possible solutions should be considered?**

Either there could be a pre-delivered `@query` directive that can be used on fields without any extra effort or otherwise a simple helper would be great that would allow developers to implement it by themselves.

k0ka:
Not sure if it is written somewhere in docs. Query is resolved automatically. Like this:

```graphql
type MutationResult {
    result: Boolean!
    query: Query!
}
```

in resolver:
```php
public function __invoke($_, array $args): array
{
    return [
        'result' => true,
    ];
}
```

spawnia:
There is a solution available, but I think it is quite unintuitive and somewhat cumbersome. Basically what @k0ka is doing with his solution is just provide any non-null value in order to allow the nested field resolution to continue with `query: Query`. Perhaps we can just change the default resolver for all fields that return the root `Query` type to always return `true` and thus allow for this propagation?

One more thing that I could see being useful would actually involve a custom directive: exposing the root `Query` type on mutations that use build-in directives such as `@create`. Perhaps something like this in the schema:

```graphql
type Mutation {
  createFoo(input: CreateFooInput! @spread): Foo! @create @exposeQuery
}
```

The `@exposeQuery` directive would transform the schema definition by introducing an intermediary type and automatically transform the execution result to fit it. The final schema would look like this:

```graphql
type Mutation {
  createFoo(input: CreateFooInput!): CreateFooResult!
}

type CreateFooResult {
  result: Foo!
  query: Query!
}
```

spawnia:
Please review https://github.com/nuwave/lighthouse/pull/2384. Do you think we should document this pattern or is it enough that it *just works™*?

jaulz:
Nice, that would make it much more easier! 

[...]

PR #2384, opened 2023-04-14
Simplify exposing root type `Query` in nested results

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

https://github.com/nuwave/lighthouse/issues/2380

**Changes**

Return any non-null value to continue nested field resolution when the root Query type is returned as part of the result.

**Breaking changes**

None.


k0ka:
Added documentation on how to write mutation result and handle the `Query` resolving

spawnia:
Thanks @k0ka. I moved the docs to `best-practices.md`. Please add docs to `docs/master` first in the future.

PR #2385, opened 2023-04-14
fix: keep order of arguments after renaming

jaulz:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Currently, using the `@rename` directive reorders the arguments because a new key is set and that one will be appended. Thus, you cannot rely anymore on the order of arguments in the array. This fix will remember the order of keys and later reorder the argument set.

**Breaking changes**

It could potentially break existing applications but only if they for some reason rely on an invalid order of arguments.

### DIFF
--- a/src/Schema/Directives/RenameArgsDirective.php
+++ b/src/Schema/Directives/RenameArgsDirective.php
@@ -27,6 +27,9 @@ GRAPHQL;
 
     protected function rename(ArgumentSet &$argumentSet): ArgumentSet
     {
+        /** @var array<int, string> */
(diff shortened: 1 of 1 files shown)


spawnia:
Argument are unordered, see https://spec.graphql.org/draft/#sec-Language.Arguments.Arguments-Are-Unordered. Thus, I see no reason to perform extra work to maintain whatever order they were given in - it can vary from client to client.

jaulz:
@spawnia oh, thanks for the hint, I didn't know that 😊

Issue #2368, opened 2023-03-26
Omit types, queries and mutations in production


jaulz:
**What problem does this feature proposal attempt to solve?**

Sometimes it's helpful to have certain queries or types in development that are not supposed to be in production. Though we can handle it via field middleware, they will still be visible in production. It would be even better if we could remove queries or types completely depending on the environment or other flags.

**Which possible solutions should be considered?**

I think something like `@env( name: "local" )` could be helpful to make queries, manipulations and types only available in certain environments.

spawnia:
It is unclear to me what you are actually asking. Please reformulate your issue in terms of an actual problem and reflect that in the title.

k0ka:
I guess he wanted to expose some fields/queries only for local tests. 
I have something similar in my codebase but not sure if it should be included in the main library tree.

```php
class DebugDirective extends BaseDirective implements FieldMiddleware
{
    private bool $debug;

    public function __construct(Repository $config)
    {
        $this->debug = $config->get('app.debug');
    }

    public static function definition(): string
    {
        return <<<GRAPHQL
"""
Limit field access to debug server
"""
directive @debug on FIELD_DEFINITION
GRAPHQL;
    }

    public function handleField(FieldValue $fieldValue, Closure $next): FieldValue
    {
        $originalResolver = $fieldValue->getResolver();

        return $next(
            $fieldValue->setResolver(
                function ($root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($originalResolver) {
                    if(!$this->debug)
                        return null;

                    return $originalResolver($root, $args, $context, $resolveInfo);
                }
            )
        );
    }
}

```

jaulz:
@k0ka yeah, that's almost what I had in mind 😊  I actually even thought about hiding types or fields completely if that's possible?

k0ka:
Removing fields from AST will be a little harder, but I can do it. 

@spawnia do you think this issue make sense?

LastDragon-ru:
> Sometimes it's helpful to have certain queries or types in development that are not supposed to be in production.

Why not use different location of the schema for specific env?

- `config/lighthouse.php`
  ```
  // ...
  'schema_path' => env('LIGHTHOUSE_SCHEMA`, base_path('graphql/schema.graphql'))
  // ...
  ```
- `graphql/schema.graphql` 
  ```graphql
  type Query {
    a: A!
  }
  ```
- `graphql/schema.dev.graphql`
  ```graphql
[...]

PR #2392, opened 2023-05-10
Add `@show` and `@hide` directives

k0ka:
Resolves https://github.com/nuwave/lighthouse/issues/2368
 
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds `@hide` directive to hide some parts of schema on different environments.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -1792,6 +1792,29 @@ final class Commentable
 }
 ```
 
+## @hide
+
+```graphql
+"""
+Hides field if environment is one of specified
+"""
+directive @hide(
+    """
+    Specify which environments must not use this field, e.g. ["prod"].
+    """
+    env: [String!]!
+) repeatable on FIELD_DEFINITION
+```
+
+This directive lets you exclude some parts of schema in different environments. For example, you might 
+remove debug information in `prod` environment:
+
+```graphql
+type Query {
+    debugInformation: String! @hide(env:["prod"])
+
+```
+
 ## @lazyLoad
 
 ```graphql
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -1796,7 +1796,7 @@ final class Commentable
 
 ```graphql
 """
-Hides field if environment is one of specified
+Excludes the annotated element from the schema conditionally.
 """
 directive @hide(
     """
@@ -3162,6 +3162,30 @@ However, in some situation a custom index might be needed, this can be achieved 
 ```graphql
 type Query {
   posts(search: String @search(within: "my.index")): [Post!]! @paginate
+}
+```
+
+## @show
+
+```graphql
+"""
+Includes the annotated element from the schema conditionally.
+"""
+directive @show(
+    """
+    Specify which environments may use this field, e.g. ["testing"].
+    Compared against the value returned from `app()->environment()`.
+    """
+    env: [String!]!
+) repeatable on FIELD_DEFINITION
+```
+
+This directive lets you include some parts of schema in different environments. For example, you might
+add a test field in `testing` environment:
+
+```graphql
+type Query {
+    testInformation: String! @show(env: ["testing"])
 }
 ```
 
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -1800,7 +1800,8 @@ Hides field if environment is one of specified
 """
 directive @hide(
     """
-    Specify which environments must not use this field, e.g. ["prod"].
+    Specify which environments exclude this element, e.g. `["prod"]`.
+    Compared against the value returned from `app()->environment()`.
     """
     env: [String!]!
 ) repeatable on FIELD_DEFINITION
@@ -1811,8 +1812,8 @@ remove debug information in `prod` environment:
 
 ```graphql
 type Query {
-    debugInformation: String! @hide(env:["prod"])
-
+    debugInformation: String! @hide(env: ["prod"])
+}
 ```
 
 ## @lazyLoad
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -3173,7 +3173,7 @@ Includes the annotated element from the schema conditionally.
 """
 directive @show(
     """
-    Specify which environments may use this field, e.g. ["testing"].
+    Specify which environments include this element, e.g. ["testing"].
     Compared against the value returned from `app()->environment()`.
     """
     env: [String!]!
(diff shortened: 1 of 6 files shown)


k0ka:
Some notes:
1) Should I add and opposite directive - '@show'?
2) Is it ok to use generic terms hide/show for it? I might rename it to `@hideIf`

spawnia:
1. I would like to see `@show` added for symmetry, yes. Their implementation can probably be very similar and share a lot of code, since they are just logical opposites of each other.
2. I think `@show` and `@hide` are fine, they mirror `@skip` and `@include` which also do not use the suffix `If`.

docs/master/api-reference/directives.md
```diff
@@ -1792,6 +1792,29 @@ final class Commentable
...
 
+## @hide
+
+```graphql
+"""
+Hides field if environment is one of specified
+"""
+directive @hide(
+    """
+    Specify which environments must not use this field, e.g. ["prod"].
```
spawnia, line 1803:
```suggestion
    Specify which environments exclude this element, e.g. `["prod"]`.
    Compared against the value returned from `app()->environment()`.
```

docs/master/api-reference/directives.md
```diff
@@ -1792,6 +1792,29 @@ final class Commentable
...
+) repeatable on FIELD_DEFINITION
+```
+
+This directive lets you exclude some parts of schema in different environments. For example, you might 
+remove debug information in `prod` environment:
+
+```graphql
+type Query {
+    debugInformation: String! @hide(env:["prod"])
+
```
spawnia, line 1815:
```suggestion
}
```

src/Schema/Directives/HideDirective.php
```diff
@@ -0,0 +1,53 @@
...
+        return in_array($this->env, $envs);
+    }
+
+    public function manipulateFieldDefinition(DocumentAST &$documentAST, FieldDefinitionNode &$fieldDefinition, ObjectTypeDefinitionNode|InterfaceTypeDefinitionNode &$parentType): void
+    {
+        if (! $this->shouldHide()) {
+            return;
+        }
+
+        $index = collect($parentType->fields)->search($fieldDefinition);
```
spawnia, line 49: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#laravel-feature-usage

`new Collection` - or just a loop would be fine too.

src/Schema/Directives/HideDirective.php
```diff
@@ -0,0 +1,53 @@
...
+directive @hide(
+  """
+  Specify which environments must not use this field, e.g. ["prod"].
+  """
+  env: [String!]!
+) repeatable on FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    /** Checks if the current field must be hidden. Can be overridden. */
```
spawnia, line 35:
```suggestion
    /** Should the annotated element be excluded from the schema? */
```

tests/Unit/Schema/Directives/HideDirectiveTest.php
```diff
@@ -0,0 +1,92 @@
+<?php declare(strict_types=1);
+
+namespace Tests\Unit\Schema\Directives;
+
+use Tests\TestCase;
+
+class HideDirectiveTest extends TestCase
```
spawnia, line 7:
```suggestion
final class HideDirectiveTest extends TestCase
```

tests/Unit/Schema/Directives/HideDirectiveTest.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    public function testHiddenOnTestingEnv(): void
+    {
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            shownField: String! @mock
+            hiddenField: String! @mock @hide(env:["testing"])
```
spawnia, line 14:
```suggestion
            hiddenField: String! @mock @hide(env: ["testing"])
```

everywhere

docs/master/api-reference/directives.md
```diff
@@ -1792,6 +1792,29 @@ final class Commentable
...
+    env: [String!]!
+) repeatable on FIELD_DEFINITION
+```
+
+This directive lets you exclude some parts of schema in different environments. For example, you might 
+remove debug information in `prod` environment:
+
+```graphql
+type Query {
+    debugInformation: String! @hide(env:["prod"])
```
spawnia, line 1814:
```suggestion
    debugInformation: String! @hide(env: ["prod"])
```

tests/Unit/Schema/Directives/HideDirectiveTest.php
```diff
@@ -0,0 +1,92 @@
...
+        }
+        ';
+
+        $introspectionQuery = /** @lang GraphQL */ '
+        query  {
```
spawnia, line 19:
```suggestion
        {
```

Generally omit this, since it is default.

docs/master/api-reference/directives.md
```diff
@@ -3141,6 +3165,30 @@ type Query {
...
 
+## @show
+
+```graphql
+"""
+Includes the annotated element from the schema conditionally.
+"""
+directive @show(
+    """
+    Specify which environments may use this field, e.g. ["testing"].
```
spawnia, line 3176:
```suggestion
    Specify which environments include this element, e.g. ["testing"].
```

k0ka:
>We don't have to do this in the first iteration, but I could see this directive being useful for other schema elements. Let's keep the description generic already so it does not have to be changed when we support types, arguments, etc. 

I thought about it. But "hiding" types means we have to remove all fields which returns it. This might be confusing.

spawnia:
> > We don't have to do this in the first iteration, but I could see this directive being useful for other schema elements. Let's keep the description generic already so it does not have to be changed when we support types, arguments, etc.
> 
> I thought about it. But "hiding" type means we have to remove all fields which returns it. This might be confusing.

Right, that seems tough. It is easier for arguments, input fields and enum values though.

pyrou:
> > > We don't have to do this in the first iteration, but I could see this directive being useful for other schema elements. Let's keep the description generic already so it does not have to be changed when we support types, arguments, etc.
> > 
> > 
> > I thought about it. But "hiding" type means we have to remove all fields which returns it. This might be confusing.
> 
> Right, that seems tough. It is easier for arguments, input fields and enum values though.

@spawnia I have a use case with a multi-tenant GraphQL API. I use `@feature` (extending `@hide`) to hide some queries depending on context. But on introspection, I would love to also be able to hide some unused Types. In addition, for some specific tenant, I added some specific args. On same manner, I'd like to be able to hide some args with a `@feature` directive.

I made a little proof of concept on [this branch](https://github.com/nuwave/lighthouse/commit/d0290cb). Let me know If that would be ok in a PR.

spawnia:
Yeah @pyrou, I could see that being useful.

PR #2409, opened 2023-06-05
Use PHPUnit 10 for testing

spawnia:


composer.json
```diff
@@ -45,7 +45,7 @@
     "require-dev": {
         "algolia/algoliasearch-client-php": "^3 || ^4",
         "bensampo/laravel-enum": "^5 || ^6",
-        "dms/phpunit-arraysubset-asserts": "^0.4",
+        "dms/phpunit-arraysubset-asserts": "dev-master",
```
spawnia, line 48: Await release of https://github.com/rdohms/phpunit-arraysubset-asserts/pull/70.

tests/Integration/Schema/Types/UnionTest.php
```diff
@@ -190,18 +190,16 @@ public static function fetchResults(): EloquentCollection
...
-        return [
-            // This uses the default type resolver
-            $this->schemaAndQuery(false),
-            // This scenario requires a custom resolver, since the types User and Post do not match
-            $this->schemaAndQuery(true),
-        ];
+        // This uses the default type resolver
+        yield self::schemaAndQuery(false);
+        // This scenario requires a custom resolver, since the types User and Post do not match
+        yield self::schemaAndQuery(true);
```
pyrou, line 198: For better output on failing tests, datasets can be "named"

```suggestion
        yield "default type resolver" => self::schemaAndQuery(false);
        // This scenario requires a custom resolver, since the types User and Post do not match
        yield "custom type resolver" => self::schemaAndQuery(true);
```

On failing test, instead of 
```
There were 5 errors:

1) Tests\Integration\Schema\Types\UnionTest::testResolveUnionTypes with data set #1
```

You will get 
```
There were 5 errors:

1) Tests\Integration\Schema\Types\UnionTest::testResolveUnionTypes with data set "custom type resolver"
```

spawnia:
The remaining issue is this:

```
Failed asserting that exception of type "Illuminate\Contracts\Container\BindingResolutionException" matches expected exception "Nuwave\Lighthouse\Exceptions\DefinitionException". Message was: "Unresolvable dependency resolving [Parameter #0 [ <required> string $name ]] in class PHPUnit\Framework\TestCase" at
/home/runner/work/lighthouse/lighthouse/vendor/laravel/framework/src/Illuminate/Container/Container.php:1143
...
/home/runner/work/lighthouse/lighthouse/tests/Integration/Schema/Types/UnionTest.php:170
```

Essentially, the problem lies within the use of `$this->qualifyTestResolver()` - whereas `PHPUnit\Framework\TestCase` required no constructor arguments prior to PHPUnit 10, it now needs one - meaning the container cannot instantiate it. The only solution I can think of is to move the test resolvers to separate files, which will complicate things a bit.

pyrou:
> The remaining issue is this:
> 
> ```
> Failed asserting that exception of type "Illuminate\Contracts\Container\BindingResolutionException" matches expected exception "Nuwave\Lighthouse\Exceptions\DefinitionException". Message was: "Unresolvable dependency resolving [Parameter #0 [ <required> string $name ]] in class PHPUnit\Framework\TestCase" at
> /home/runner/work/lighthouse/lighthouse/vendor/laravel/framework/src/Illuminate/Container/Container.php:1143
> ...
> /home/runner/work/lighthouse/lighthouse/tests/Integration/Schema/Types/UnionTest.php:170
> ```
> 
> Essentially, the problem lies within the use of `$this->qualifyTestResolver()` - whereas `PHPUnit\Framework\TestCase` required no constructor arguments prior to PHPUnit 10, it now needs one - meaning the container cannot instantiate it. The only solution I can think of is to move the test resolvers to separate files, which will complicate things a bit.

This is a possible (dirty) workaround

https://laravel.com/docs/10.x/container#binding-primitives

```php
$this->app->when(TestCase::class)
          ->needs('$name')
          ->give("TestName");
```

spawnia:
Thanks @pyrou, unfortunately I could not get it to work. Could you check out the branch and see if you can make it work?

pyrou:
Looks like because its called to early. Somehow it is reset somewhere

When moving code in TestCase::setUp() it works fine

```php
// tests/TestCase.php
protected function setUp(): void
{
    parent::setUp();

    // This default is only valid for testing Lighthouse itself and thus
    // is not defined in the reusable test trait.
    $this->schema ??= self::PLACEHOLDER_QUERY;
    $this->setUpTestSchema();

    $this->app->when($this::class)
        ->needs('$name')
        ->give('TestName');

}
```

PR #2413, opened 2023-06-17
Add `PaginatorInfo`, `SimplePaginatorInfo`, `PageInfo` only when required

### LINKED ISSUE #2388: Add Paginator types only if required

Not really high priority but similar to https://github.com/nuwave/lighthouse/pull/2387 it would be great if the `PageInfo`, `SimplePaginatorInfo` and `PaginatorInfo` types would only be registered if they are actually required and used.

n/a

csetepeter:
Resolves #2388
 
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Currently pagination-related types are registered in the document ast even if they're never used. After this PR, these types will only be registered if they're actually needed.

**Breaking changes**

None as far as I can tell.

### DIFF
--- a/src/Pagination/PaginationManipulator.php
+++ b/src/Pagination/PaginationManipulator.php
@@ -72,6 +72,10 @@ class PaginationManipulator
         int $maxCount = null,
         ObjectTypeDefinitionNode $edgeType = null,
     ): void {
+        if (! isset($this->documentAST->types[self::pageInfo()->getName()->value])) {
+            $this->documentAST->setTypeDefinition(self::pageInfo());
+        }
+
         $fieldTypeName = ASTHelper::getUnderlyingTypeName($fieldDefinition);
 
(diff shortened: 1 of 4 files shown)


csetepeter:
This is my first time contributing to this project, let me know if I've missed anything.

src/Pagination/PaginationManipulator.php
```diff
@@ -72,6 +72,10 @@ protected function registerConnection(
         int $maxCount = null,
         ObjectTypeDefinitionNode $edgeType = null,
     ): void {
+        if (! isset($this->documentAST->types[self::pageInfo()->getName()->value])) {
+            $this->documentAST->setTypeDefinition(self::pageInfo());
```
spawnia, line 76: Extract the duplicated call to `self::pageInfo()`, same with the other places like this.

PR #2426, opened 2023-07-28
Upsert fields directive

GavG:
- [x ] Added or updated tests
- [ x] Updated CHANGELOG.md

**Changes**

Adds optional identifyingColumns arg to upsert directive

**Breaking changes**

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add support for identifyingColumns on upserts
+
 ## v6.15.0
 
 ### Added
@@ -524,7 +528,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Fixed
 
-- Distinguish between client-safe and non-client-safe errors in `TestResponse::assertGraphQLError()` 
+- Distinguish between client-safe and non-client-safe errors in `TestResponse::assertGraphQLError()`
 
 ## v5.46.0
 
@@ -831,7 +835,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Fixed
 
-- Avoid PHP 8.1 deprecation warning by implementing `__serialize()` and `__unserialize()` https://github.com/nuwave/lighthouse/pull/1987 
+- Avoid PHP 8.1 deprecation warning by implementing `__serialize()` and `__unserialize()` https://github.com/nuwave/lighthouse/pull/1987
 
 ### Deprecated
 
@@ -1107,7 +1111,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Changed
 
-- Improve performance through [`graphql-php` lazy field definitions](https://github.com/webonyx/graphql-php/pull/861) https://github.com/nuwave/lighthouse/pull/1851 
+- Improve performance through [`graphql-php` lazy field definitions](https://github.com/webonyx/graphql-php/pull/861) https://github.com/nuwave/lighthouse/pull/1851
 - Load individual subscription fields lazily instead of loading them all eagerly https://github.com/nuwave/lighthouse/pull/1851
 - Require `webonyx/graphql-php:^14.7` https://github.com/nuwave/lighthouse/pull/1851
 
--- a/src/Execution/Arguments/UpsertModel.php
+++ b/src/Execution/Arguments/UpsertModel.php
@@ -9,10 +9,14 @@ class UpsertModel implements ArgResolver
     /** @var callable|\Nuwave\Lighthouse\Support\Contracts\ArgResolver */
     protected $previous;
 
+    /** @var array<string> */
+    protected array $identifyingColumns;
+
     /** @param  callable|\Nuwave\Lighthouse\Support\Contracts\ArgResolver  $previous */
-    public function __construct(callable $previous)
+    public function __construct(callable $previous, ?array $identifyingColumns)
     {
         $this->previous = $previous;
+        $this->identifyingColumns = $identifyingColumns ?? [];
     }
 
     /**
@@ -22,17 +26,36 @@ class UpsertModel implements ArgResolver
     public function __invoke($model, $args): mixed
     {
         // TODO consider Laravel native ->upsert(), available from 8.10
-        $id = $args->arguments['id']
-            ?? $args->arguments[$model->getKeyName()]
-            ?? null;
+        $existingModel = null;
 
-        if ($id !== null) {
+        if (!empty($this->identifyingColumns)) {
             $existingModel = $model
                 ->newQuery()
-                ->find($id->value);
+                ->firstWhere(
+                array_intersect_key(
+                        $args->toArray(),
+                        array_flip($this->identifyingColumns)
+                    )
+                );
 
             if ($existingModel !== null) {
                 $model = $existingModel;
+            }
+        }
+
+        if ($existingModel === null) {
+            $id = $args->arguments['id']
+                ?? $args->arguments[$model->getKeyName()]
+                ?? null;
+
+            if ($id !== null) {
+                $existingModel = $model
+                    ->newQuery()
+                    ->find($id->value);
+
+                if ($existingModel !== null) {
+                    $model = $existingModel;
+                }
             }
         }
 
--- a/src/Schema/Directives/UpsertDirective.php
+++ b/src/Schema/Directives/UpsertDirective.php
@@ -22,6 +22,12 @@ directive @upsert(
   model: String
 
   """
+  Specify the columns by which to upsert the model.
+  This is optional, defaults to the ID or model Key.
+  """
+  identifyingColumns: [String!] = []
+
+  """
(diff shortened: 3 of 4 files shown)


src/Schema/Directives/UpsertDirective.php
```diff
@@ -21,6 +21,12 @@ public static function definition(): string
   """
   model: String
 
+  """
+  Specify the columns by which to upsert the model.
+  This is optional, defaults to the ID or model Key.
+  """
+  identifyingColumns: [String!] = []
```
spawnia, line 28:
```suggestion
  identifyingColumns: [String!]
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add support for identifyingColumns on upserts
```
```suggestion
- Specify identifying columns on nested mutation upserts with `@upsert(identifyingColumns: ["foo", "bar"])` https://github.com/nuwave/lighthouse/pull/2426
```

src/Execution/Arguments/UpsertModel.php
```diff
@@ -9,10 +9,14 @@ class UpsertModel implements ArgResolver
     /** @var callable|\Nuwave\Lighthouse\Support\Contracts\ArgResolver */
     protected $previous;
 
+    /** @var array<string> */
+    protected array $identifyingColumns;
```
spawnia, line 13:
```suggestion
    /** @var array<string>|null */
    protected ?array $identifyingColumns;
```

src/Execution/Arguments/UpsertModel.php
```diff
@@ -9,10 +9,14 @@ class UpsertModel implements ArgResolver
...
+
     /** @param  callable|\Nuwave\Lighthouse\Support\Contracts\ArgResolver  $previous */
-    public function __construct(callable $previous)
+    public function __construct(callable $previous, ?array $identifyingColumns)
     {
         $this->previous = $previous;
+        $this->identifyingColumns = $identifyingColumns ?? [];
```
spawnia, line 19:
```suggestion
        $this->identifyingColumns = $identifyingColumns;
```

src/Execution/Arguments/UpsertModel.php
```diff
@@ -22,20 +26,39 @@ public function __construct(callable $previous)
     public function __invoke($model, $args): mixed
     {
         // TODO consider Laravel native ->upsert(), available from 8.10
```
spawnia, line 28: I wonder if Laravel upsert() does what we need already and we unnecessarily diverge?

src/Execution/Arguments/UpsertModel.php
```diff
@@ -22,20 +26,39 @@ public function __construct(callable $previous)
...
-        $id = $args->arguments['id']
-            ?? $args->arguments[$model->getKeyName()]
-            ?? null;
+        $existingModel = null;
 
-        if ($id !== null) {
+        if (! empty($this->identifyingColumns)) {
```
spawnia, line 31:
```suggestion
        if (isset($this->identifyingColumns)) {
```

The directive should implement `manipulateArgDefinition` and `manipulateFieldDefinition` to validate the given array is not empty and throw a `DefinitionException`.

src/Schema/Directives/UpsertDirective.php
```diff
@@ -21,6 +21,12 @@ public static function definition(): string
   """
   model: String
 
+  """
+  Specify the columns by which to upsert the model.
+  This is optional, defaults to the ID or model Key.
```
spawnia, line 26:
```suggestion
  Optional, by default `id` or the primary key of the model are used.
```

It is not like this argument itself defaults to those values, the mechanics behind it behave that way.

src/Schema/Directives/UpsertDirective.php
spawnia, line 1: Update the directive definition in `docs/master/api-reference/directives.md` with the final definition.

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -198,6 +199,150 @@ interface IUser
...
+                    'email' => 'foo@te.st',
+                    'name' => 'bar',
+                ],
+            ],
+        ]);
+
+        $user = User::firstOrFail();
+
+        $this->assertSame('bar', $user->name);
+        $this->assertSame('foo@te.st', $user->email);
```
spawnia, line 238: Use variables to show the connection between the passed values.

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -198,6 +199,150 @@ interface IUser
...
+    }
+
+    public function testDirectUpsertByIdentifyingColumns(): void
+    {
+        $company = factory(Company::class)->create(['id' => 1]);
+
+        $this->schema
+            /** @lang GraphQL */
+            .= '
```
spawnia, line 271:
```suggestion
        $this->schema .= /** @lang GraphQL */ '
```

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -198,6 +199,150 @@ interface IUser
...
+        }
+
+        type Mutation {
+            upsertUser(name: String!, email: String!, company_id:ID!): User @upsert(identifyingColumns: ["name", "company_id"])
+        }
+        ';
+
+        $this->graphQL(
+            /** @lang GraphQL */
+            '
```
spawnia, line 286:
```suggestion
        $this->graphQL(/** @lang GraphQL */ '
```

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -198,6 +199,150 @@ interface IUser
...
+
+        $user = User::firstOrFail();
+
+        $this->assertSame('bar', $user->name);
+        $this->assertSame('foo@te.st', $user->email);
+        $this->assertSame(1, $user->company_id);
+
+        $this->graphQL(
+            /** @lang GraphQL */
+            '
```
spawnia, line 317:
```suggestion
        $this->graphQL(/** @lang GraphQL */ '
```

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -198,6 +199,150 @@ interface IUser
...
+
+        $user->refresh();
+
+        $this->assertSame('foo', $user->name);
+        $this->assertSame('foo@te.st', $user->email);
+    }
+
+    public function testDirectUpsertByIdentifyingColumns(): void
+    {
+        $company = factory(Company::class)->create(['id' => 1]);
```
spawnia, line 267: https://github.com/nuwave/lighthouse/blob/master/CONTRIBUTING.md#test-data-setup

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -198,6 +199,150 @@ interface IUser
...
+
+        $this->schema
+            /** @lang GraphQL */
+            .= '
+        type User {
+            id: ID!
+            email: String!
+            name: String!
+            company_id: ID!
```
spawnia, line 276: No raw foreign keys in GraphQL, not even as a test case. Often, those test cases serve as an example, so I don't want them to implement anti-patterns.

spawnia (changes requested):
Implementation and test make sense so far, after some tweaks I am fine with adding this.

src/Execution/Arguments/UpsertModel.php
GavG: I looked into this, unfortunately not, Laravel's upsert simply bulk inserts raw data into the DB bypassing the model events, casts, setters etc. Probably not what most users want here. I think the Laravel 'upsert' is more inline with the database layer concept of upsert and not the graphql mutation upsert concept.

GavG:
Thanks for the review @spawnia. Sorry I've been a bit quiet on this, recently had our first child so haven't had any free time 🙃

I've been giving the solution a bit more thought, I think supporting nested relationship upsert logic is a must, so I'll try and tackle that when I come to making the suggest changes.

I've also been considering my own use case again, where more complex upsert logic was required, and I think a cleaner solution to this might be to permit the specifying of a custom query class that would receive the input params in place of the package's query logic. I may have a go at this in a new fork later.

tests/Integration/Schema/Directives/UpsertDirectiveTest.php
```diff
@@ -4,20 +4,26 @@
...
 
-        $task = factory(Task::class)->create();
-        $this->assertInstanceOf(Task::class, $task);
+        $task = factory(Task::class)->make();
         $task->id = 1;
+        $task->user()->associate($user);
+        $task->save();
+        $this->assertInstanceOf(Task::class, $task);
         $task->name = 'old';
         $task->save();
```
spawnia: avoid double save

spawnia:
Released with https://github.com/nuwave/lighthouse/releases/tag/v6.65.0, please consider sponsoring.

PR #2428, opened 2023-07-31
docs(federation): to categorize information about resolvers

rprybudko:
The package provides great functionality for working with federations, but some nuances of its work may not be sufficiently demonstrated. This lack of clarity could discourage users from working with the package as it increases the complexity of getting familiar with it.

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

- The documentation has been updated in the section about federations.
- Examples of resolvers writing have been added.
- A base batched reference resolver class has been created, which provides helper methods for forming responses.

**Breaking changes**


docs/master/federation/entities.md
```diff
@@ -2,98 +2,10 @@
 
 A core component of a federation capable GraphQL service is the `_entities` field.
 For a given `__typename` in the given `$representations`, Lighthouse will look for
-a resolver to return the full `_Entity`.
+a reference resolver to return the full `_Entity`.
 
-## Class Based Resolvers
+### Extends
```
spawnia, line 7: Straight to h3, should be h2 or no heading at all

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,85 @@
...
+A representation always consists of:
+- A `__typename` field;
+- Values for the entity's primary key fields.
+
+In this way, the resolver should return an array consisting of the `__typename` field, which corresponds to the entity's
+name, and the primary key that can identify the entity.
+
+### Example 1
+
+In this example, the `review-service` has an entity called `Review`, which in turn has an entity called `Product`
```
spawnia, line 16: What is the `review-service`? I think it would be clearer to simply put review service without the quotes.

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,85 @@
...
+defined in a separate `product-service`. The relationship between the review and the product is one-to-one.
```
spawnia, line 17:
```suggestion
defined in a separate product service. The relationship between `Review` and `Product` one-to-one.
```

Aside from formatting, one-to-one does not really make sense here. I would assume a product can have multiple reviews.

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,85 @@
...
+    score: Int!
+    product: Product! @field(resolver: "App\\GraphQL\\Resolvers\\Product")
+}
+
+type Product @key(fields: "uuid") @extends{
+    uuid: String! @external
+}
+```
+
+The resolver for the product in the `review-service` returns an array consisting of:
```
spawnia, line 30:
```suggestion
The resolver for `Product` in the review service returns an array consisting of:
```

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,85 @@
...
+```php
+namespace App\GraphQL\Resolvers;
+
+final class Product
+{
+    public function __invoke($model): array
+    {
+        return [
+            '__typename' => 'Product',
+            'uuid'       => $model->product_id,
```
spawnia, line 43:
```suggestion
            'uuid' => $model->product_id,
```

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,85 @@
...
+```php
+namespace App\GraphQL\Resolvers;
+
+final class Products
+{
+    public function __invoke($model): array
+    {
+        $products = [];
+
+        $model->products()->each(function ($product) use (&$products){
```
spawnia, line 78: use `->map()`

docs/master/federation/getting-started.md
```diff
@@ -8,7 +8,7 @@ Read more about the core concepts and motivation in the [Apollo Federation docs]
 Lighthouse can act as a federation capable service as described in the [Apollo Federation specification](https://www.apollographql.com/docs/federation/federation-spec).
 It can not serve as a [federation gateway](https://www.apollographql.com/docs/federation/gateway).
 
-## Setup
+### Setup
```
spawnia, line 11: Again, wrong heading structure

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+when needed. Read more about the reference resolvers in
+the [Apollo Federation docs](https://www.apollographql.com/docs/federation/v1/entities#reference-resolvers).
+
+Lighthouse will look for a class which name is equivalent to `__typename` in the
+namespace configured in `lighthouse.federation.entities_resolver_namespace`.
+
+When you need to retrieve information from subgraphs, the gateway automatically generates a request to the corresponding
+endpoint of the subgraph. More details about this can be found in
+section [Query._entities of the Apollo Federation docs](https://www.apollographql.com/docs/federation/building-supergraphs/subgraphs-overview#query_entities)
+.
```
spawnia, line 14:
```suggestion
section [Query._entities of the Apollo Federation docs](https://www.apollographql.com/docs/federation/building-supergraphs/subgraphs-overview#query_entities).
```

docs/master/federation/entities.md
```diff
@@ -2,98 +2,10 @@
 
 A core component of a federation capable GraphQL service is the `_entities` field.
 For a given `__typename` in the given `$representations`, Lighthouse will look for
-a resolver to return the full `_Entity`.
+a reference resolver to return the full `_Entity`.
```
spawnia, line 5:
```suggestion
a [reference resolver](reference-resolvers.md) to return the full `_Entity`.
```

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+        ... on Foo {
+            id
+        }
+    }
+}
+```
+
+## Single Entity Resolvers
+
+After validating the type `Foo` exists, Lighthouse will look for a resolver class in
```
spawnia, line 30:
```suggestion
After validating that type `Foo` exists, Lighthouse will look for a resolver class in
```

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+config `lighthouse.federation.entities_resolver_namespace` namespace. The resolver class is expected to contain a
```
spawnia, line 31:
```suggestion
the namespace configured in `lighthouse.federation.entities_resolver_namespace`. The resolver class is expected to contain a
```

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+method `__invoke()` which takes a single argument: the array form of the representation.
+
+```php
+namespace App\GraphQL\ReferenceResolvers;
+
+final class Foo
+{
+    /**
+     * @param  array{__typename: string, id: int}  $representation
+     */
```
spawnia, line 41:
```suggestion
    /** @param  array{__typename: string, id: int}  $representation */
```

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+    public function __invoke(array $representation)
+    {
+        // TODO return a value that matches type Foo
+    }
+}
+```
+
+The method should only return an object that has the same name as the entity, but the namespace of the object may be
+different.
```
spawnia, line 50: This is not a requirement, any value could work - object, array, even something like an identifier - given the resolvers of nested fields know how to deal with it.

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+        $foo = SomeFooRepository::findBySomeQuery($id);
+
+        return FooEntity::fromArray($foo);
+    }
+}
+```
+
+It is also possible to return an object of the current class, but in such case, you need to either create a new object
+or return its copy. This is necessary because when returning more than one entity, all entities will be assigned the
+value of the last entity since the resolver's object for all entities of the same type is the same.
```
spawnia, line 73: I would not recommend this approach or describe its caveats. Of course modifying and returning the same object repeatedly won't work.

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+        $ids = Arr::pluck($representation, 'id');
+        $products = SomeProductRepository::getSomeProductsCollection($ids);
+
+        foreach ($representation as $hash => $representation) {
+            $id = Arr::get($representation, 'id');
+
+            $result[$hash] = $products->where('id', $id)->first(); // returned iterable _must_ have the same keys as the given `array $representations`
+        }
+
+        return $result;
```
spawnia, line 147:
```suggestion
use App\Repositories\ProductRepository;
use Illuminate\Support\Arr;
use Nuwave\Lighthouse\Federation\BatchedEntityResolver;

final class Product implements BatchedEntityResolver
{
    public function __invoke(array $representations): iterable
    {
        $products = ProductRepository::byIDs(Arr::pluck($representation, 'id'));

        $result = [];
        foreach ($representations as $key => $representation) {
            $result[$key] = $products->firstWhere('id', $representation['id']);
        }

        return $result;
```

src/Federation/Resolvers/BaseBatchedReferenceResolver.php
```diff
@@ -0,0 +1,93 @@
+<?php declare(strict_types=1);
+
+namespace Nuwave\Lighthouse\Federation\Resolvers;
+
+use GraphQL\Type\Introspection;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
+use Nuwave\Lighthouse\Federation\BatchedEntityResolver;
+
+abstract class BaseBatchedReferenceResolver implements BatchedEntityResolver
```
spawnia, line 10: Let's hold off on this for this pull request. Feel free to propose it separately.

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+```
+
+### Base Batched Entity Resolvers
```
spawnia, line 152: Maybe in another PR, let's make the other improvements first.

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+called and expected to return a single result.
+
+```php
+$results = App\Models\Foo::query()
+    ->where('bar', 'asdf')
+    ->where('baz', 42)
+    ->get();
+
+if ($results->count() > 1) {
+    throw new Error('The query returned more than one result.');
```
spawnia, line 203:
```suggestion
    throw new GraphQL\Error\Error('The query returned more than one result.');
```

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+}
+
+return $results->first();
+```
+
+For model resolvers, it is typical that they make one query to the database for each entity, which can lead to the "n+1"
+problem. Therefore, for a large number of entities, it is worth
+considering [Batched Entity Resolves](reference-resolvers.md#batched-entity-resolves) to avoid this issue.
```
spawnia, line 211:
```suggestion
considering [Batched Entity Resolvers](reference-resolvers.md#batched-entity-resolvers) to avoid this issue.
```

docs/master/federation/reference-resolvers.md
```diff
@@ -0,0 +1,211 @@
...
+
+if ($results->count() > 1) {
+    throw new Error('The query returned more than one result.');
+}
+
+return $results->first();
+```
+
+For model resolvers, it is typical that they make one query to the database for each entity, which can lead to the "n+1"
+problem. Therefore, for a large number of entities, it is worth
```
spawnia, line 210:
```suggestion
The default model resolver makes one database query for each entity.
Therefore, for a large number of entities, it is worth
```

There is no question about it, they definitely do. Also, this is not a case of n+1, it is just n.

spawnia (changes requested):

docs/master/federation/reference-resolvers.md
rprybudko: Oh yes, thank you, I indeed missed this code https://github.com/nuwave/lighthouse/blob/master/src/Schema/TypeRegistry.php#L504

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,84 @@
...
+- A `__typename` field;
+- Values for the entity's primary key fields.
+
+In this way, the resolver should return an array consisting of the `__typename` field, which corresponds to the entity's
+name, and the primary key that can identify the entity.
+
+## Example 1
+
+In this example, subgraph for order service has an entity called `Order`, which in turn has an entity called `Receipt`
+defined in a separate subgraph for payment service. The relationship between `Order` and `Receipt` is one-to-one.
```
spawnia, line 18: Good example.

docs/master/federation/entity-representations.md
```diff
@@ -0,0 +1,84 @@
...
+defined in a separate subgraph for product service. The relationship between `Order` and `Product` is one-to-many.
+
+```graphql
+type Order {
+    sum: Int!
+    products: [Product!]! @field(resolver: "App\\GraphQL\\Resolvers\\Products")
+}
+
+type Product @key(fields: "uuid") @extends{
+    uuid: String! @external
```
spawnia, line 62: I think for this, no custom resolver would really be necessary if defined like this:

```suggestion
    products: [Product!]! @hasMany
}

type Product @key(fields: "id") @extends{
    id: String! @external
```

rprybudko: In this case, it is necessary to define the `products` relationship in Order model. Perhaps it's worth keeping the current example for the case when there is no relationship and adding a new example for the one you described?

spawnia: I would not mind having an example for a non-Eloquent non-relation use-case, but I think it is misleading to have a custom resolver for Eloquent relations. It is probably a good idea to have both to show both approaches and the appropriate implementation for each.

rprybudko: Alright, I will provide examples for both cases, I believe it will indeed be helpful.

spawnia:
Thanks, I made some final tweaks and merged.

PR #2430, opened 2023-08-02
Reset cached values of arguments in `BaseDirective::hydrate()`.

LastDragon-ru:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

`BaseDirective::hydrate()` doesn't reset cached args. Not sure when/where it can be critical, just found in my notes....

**Breaking changes**

Nope.


LastDragon-ru:
One test failed, but it seems is not related to PR itself

image

Issue #2444, opened 2023-09-08
`@in` directive breaks if client sends null


MissaelAnda:
The `@in` directive doesn't handle the possible case where the field is nullable and the client sends `null`.

The solution could be adding a null coalesce in `InDirective.php` from lines 30 - 33:
```php
        return $builder->whereIn(
            $this->directiveArgValue('key', $this->nodeName()),
            $value ?? [],
        );
```

Error when sending null: `TypeError  count(): Argument #1 ($value) must be of type Countable|array, null given.`

Lighthouse Version: v6.18.0

spawnia:
I do not agree with the proposed solution, passing `null` should be equivalent to not sending the argument value at all. Fixing this in https://github.com/nuwave/lighthouse/pull/2445.

Please consider sponsoring.

PR #2446, opened 2023-09-15
Prevent crashes with explicit nulls on arg directives

MissaelAnda:
Same as #2444 for more ArgDirectives
 
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

- Explicit nulls are ignored in directives `@notIn` `@whereBetween` and `@whereNotBetween`.
- `@where` directive has now the `ignoreNull` argument so the user can explicitly tell when to ignore null values.

**Breaking changes**
No

### DIFF
--- a/docs/6/api-reference/directives.md
+++ b/docs/6/api-reference/directives.md
@@ -3677,6 +3677,12 @@ directive @where(
   Exclusively required when this directive is used on a field.
   """
   value: WhereValue
+
+  """
+  Wether null values should be ignored.
+  If set to true null values will be treated as if the argument is not present in the request.
+  """
+  ignoreNull: Boolean! = false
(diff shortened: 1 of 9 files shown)


docs/6/api-reference/directives.md
spawnia, line 1: Please extract the changes to `@where` to a separate merge request. Also, modify `docs/master` instead.

spawnia (changes requested):
Please extract the changes to `@where` to a separate merge request so we can discuss it there. I am fine with the rest of the changes.

spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.18.2.

PR #2447, opened 2023-09-15
`@where` ignoreNull attribute

MissaelAnda:
Treat null values the same as if they're not present, useful to prevent InvalidArgumentException if the operator is not on of `=`, `<>` or `!=`.
 
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
Added the optional `ignoreNull` attribute to `@where` directive.

**Breaking changes**
No

### DIFF
--- a/docs/6/api-reference/directives.md
+++ b/docs/6/api-reference/directives.md
@@ -3677,6 +3677,12 @@ directive @where(
   Exclusively required when this directive is used on a field.
   """
   value: WhereValue
+
+  """
+  Wether null values should be ignored.
+  If set to true null values will be treated as if the argument is not present in the request.
+  """
+  ignoreNull: Boolean! = false
 ) repeatable on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | FIELD_DEFINITION
 
 """
@@ -3709,6 +3715,14 @@ When used on a field, you must define `key` and `value`:
 ```graphql
 type Query {
   importantPosts: [Post!]! @all @where(key: "priority", operator: ">", value: 5)
+}
+```
+
+If you want to prevent null values to be passed to the query you can set `ignoreNull` to true:
+
+```graphql
+type Query {
+    posts(before: DateTime @where(operator: "<", ignoreNull: true)): [Post!]! @all
 }
 ```
 
--- a/docs/6/api-reference/directives.md
+++ b/docs/6/api-reference/directives.md
@@ -3677,12 +3677,6 @@ directive @where(
   Exclusively required when this directive is used on a field.
   """
   value: WhereValue
-
-  """
-  Wether null values should be ignored.
-  If set to true null values will be treated as if the argument is not present in the request.
-  """
-  ignoreNull: Boolean! = false
 ) repeatable on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | FIELD_DEFINITION
 
 """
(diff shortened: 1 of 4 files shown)


docs/6/api-reference/directives.md
spawnia, line 1: Modify `docs/master` instead.

src/Schema/Directives/WhereDirective.php
```diff
@@ -41,6 +41,12 @@ public static function definition(): string
   Exclusively required when this directive is used on a field.
   """
   value: WhereValue
+
+  """
+  Wether null values should be ignored.
+  If set to true null values will be treated as if the argument is not present in the request.
```
spawnia, line 47:
```suggestion
  Treat explicit `null` as if the argument is not present in the request?
```

spawnia (commented):
Naming and implementation of this are fine. My only gripe with it is that it might create inconcistency for the client, who don't know how an argument would treat `null`. I understand the usefulness of this in combination with using variables, but I wonder if there is a better way.

spawnia:
> Couldn't come up with a way to do this client-aware so I'm thinking in refactoring to only be null-safe with operators that don't support null comparissons.

That seems difficult to get right. Perhaps we can encourage documenting the behaviour for clients in the argument description by providing an example in the docs.

docs/master/api-reference/directives.md
```diff
@@ -3677,6 +3677,11 @@ directive @where(
   Exclusively required when this directive is used on a field.
   """
   value: WhereValue
+
+  """
+  Treat explicit `null` as if the argument is not present in the request.
```
spawnia, line 3682:
```suggestion
  Treat explicit `null` as if the argument is not present in the request?
```

docs/master/api-reference/directives.md
```diff
@@ -3712,6 +3717,30 @@ type Query {
 }
 ```
 
+If you want to prevent explicit `null` values to be passed to the query you can use `ignoreNull`:
+
+```graphql
+type Post {
+    id: ID!
+    # There are not null titles
```
spawnia, line 3725:
```suggestion
    # Never null
```

src/Schema/Directives/WhereDirective.php
```diff
@@ -52,12 +57,27 @@ public static function definition(): string
 
     public function handleBuilder(QueryBuilder|EloquentBuilder|Relation $builder, $value): QueryBuilder|EloquentBuilder|Relation
     {
+        if ($value === null && $this->directiveArgValue('ignoreNull', false)) {
+            return $builder;
+        }
+
+        $operator = $this->directiveArgValue('operator', '=');
+
+        // If the client sends an illegal value to the used operator it would crash the query
```
spawnia, line 66: Just doing nothing seems very unexpected and hard to debug.

This change also seems unnecessary for what this pull request is trying to do. Conflating distinct changes in a single pull request slows the review process - I would merge now if this change were not added on top.

MissaelAnda: Sure I'll remove it from this PR but I think that the illegal operator and value combination exception should be catched instead of throwing it to the client, how should it be done?

spawnia: I don't think so, errors should definitely be surfaced to the client. Swallowing the error and defaulting to a no-op is just about the worst possible course of action. A possible improvement would be to provide a better error message.

PR #2449, opened 2023-09-20
Allow customizing the definition of the root `Query` field `node` added by `@node`

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This change allows customizing the behaviour of the automatically added `Query.node`.

**Breaking changes**

Custom definitions of field `Query.node` are no longer overwritten, I consider this more of a bugfix though.

### DIFF
--- a/docs/master/api-reference/directives.md
+++ b/docs/master/api-reference/directives.md
@@ -2366,7 +2366,7 @@ If you want to customize its description, change the resolver or add middleware,
 ```graphql
 type Query {
     "This description is up to you."
-    node(id: ID! @globalId): Node @field(resolver: "Nuwave\\Lighthouse\GlobalId\\NodeRegistry@resolve")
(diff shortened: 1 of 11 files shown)


docs/master/api-reference/directives.md
```diff
@@ -2360,6 +2360,18 @@ directive @node(
 ) on OBJECT
 ```
 
+When you use `@node` on a type, Lighthouse will add a field `node` to the root Query type.
+If you want to customize its description, change the resolver or add middleware, you can add it yourself like this:
+
+```graphql
+type Query {
+    "This description is up to you."
+    node(id: ID! @globalId): Node @field(resolver: "Nuwave\\Lighthouse\GlobalId\\NodeRegistry@resolve")
```
exodusanto, line 2369: Missing one escape `"Nuwave\\Lighthouse\\GlobalId\\NodeRegistry@resolve"`

PR #2452, opened 2023-10-02
Allow checking root model in @can directive

awarrenlove:
No issue for this that I'm aware of, but it's a feature I've built into a custom directive I thought would be useful in the upstream as it came up recently in the community Slack.

- [X] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR allows users to check the root value against a Policy so a field can be controlled by its parent model Policy.

**Breaking changes**

None

### DIFF
--- a/docs/6/api-reference/directives.md
+++ b/docs/6/api-reference/directives.md
@@ -621,7 +621,7 @@ directive @can(
   Check the policy against the model instances returned by the field resolver.
   Only use this if the field does not mutate data, it is run before checking.
 
-  Mutually exclusive with `query` and `find`.
+  Mutually exclusive with `query`, `find`, and `rootModel`.
   """
   resolved: Boolean! = false
 
@@ -648,7 +648,7 @@ directive @can(
   Query for specific model instances to check the policy against, using arguments
   with directives that add constraints to the query builder, such as `@eq`.
 
-  Mutually exclusive with `resolved` and `find`.
+  Mutually exclusive with `resolved`, `find`, and `rootModel`.
   """
   query: Boolean! = false
 
@@ -663,7 +663,7 @@ directive @can(
 
   You may pass the string in dot notation to use nested inputs.
 
-  Mutually exclusive with `resolved` and `query`.
+  Mutually exclusive with `resolved`, `query`, and `rootModel`.
   """
   find: String
 
@@ -671,6 +671,13 @@ directive @can(
   Should the query fail when the models of `find` were not found?
   """
   findOrFail: Boolean! = true
+
+  """
+  If your policy should check against the root value.
+
+  Mutually exclusive with `resolved`, `query`, and `find`.
+  """
+  rootModel: Boolean! = false
 ) repeatable on FIELD_DEFINITION
 
 """
--- a/docs/6/api-reference/directives.md
+++ b/docs/6/api-reference/directives.md
@@ -621,7 +621,7 @@ directive @can(
   Check the policy against the model instances returned by the field resolver.
   Only use this if the field does not mutate data, it is run before checking.
 
-  Mutually exclusive with `query`, `find`, and `rootModel`.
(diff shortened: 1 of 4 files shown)


docs/6/api-reference/directives.md
spawnia, line 1: Modify `docs/master` instead

src/Auth/CanDirective.php
```diff
@@ -95,14 +95,21 @@ public static function definition(): string
...
   Should the query fail when the models of `find` were not found?
   """
   findOrFail: Boolean! = true
+
+  """
+  If your policy should check against the root value.
+
+  Mutually exclusive with `resolved`, `query`, and `find`.
+  """
+  rootModel: Boolean! = false
```
spawnia, line 112: Just `root` should be sufficiently clear.

tests/Unit/Auth/CanDirectiveTest.php
```diff
@@ -281,6 +281,40 @@ public function testInjectArgsPassesClientArgumentToPolicy(): void
...
+        });
+
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            user(foo: String): User!
+                @mock
+        }
+
+        type User {
+            name: String @can(ability:"view", rootModel: true)
```
spawnia, line 299:
```suggestion
            name: String @can(ability: "view", rootModel: true)
```

tests/Unit/Auth/CanDirectiveTest.php
```diff
@@ -281,6 +281,40 @@ public function testInjectArgsPassesClientArgumentToPolicy(): void
...
+        $this->be(new User());
+
+        $this->mockResolver(function (): User {
+            return $this->resolveUser();
+        });
+
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            user(foo: String): User!
+                @mock
```
spawnia, line 295:
```suggestion
            user(foo: String): User! @mock
```

tests/Unit/Auth/CanDirectiveTest.php
```diff
@@ -281,6 +281,40 @@ public function testInjectArgsPassesClientArgumentToPolicy(): void
         ]);
     }
 
+    public function testChecksAgainstRootModel(): void
+    {
+        $this->be(new User());
+
+        $this->mockResolver(function (): User {
+            return $this->resolveUser();
+        });
```
spawnia, line 290:
```suggestion
        $this->mockResolver(fn (): User => $this->resolveUser());
```

tests/Unit/Auth/CanDirectiveTest.php
```diff
@@ -281,6 +281,40 @@ public function testInjectArgsPassesClientArgumentToPolicy(): void
...
+        }
+
+        type User {
+            name: String @can(ability:"view", rootModel: true)
+        }
+        ';
+
+        $this->graphQL(/** @lang GraphQL */ '
+        {
+            user(foo: "bar"){
```
spawnia, line 305:
```suggestion
            user(foo: "bar") {
```

tests/Unit/Auth/CanDirectiveTest.php
```diff
@@ -281,6 +281,40 @@ public function testInjectArgsPassesClientArgumentToPolicy(): void
...
+
+        $this->schema = /** @lang GraphQL */ '
+        type Query {
+            user(foo: String): User!
+                @mock
+        }
+
+        type User {
+            name: String @can(ability:"view", rootModel: true)
+        }
```
spawnia, line 300: Can you add another field with `@can(root: true)` that fails?

spawnia (changes requested):

spawnia:
Thanks, released with https://github.com/nuwave/lighthouse/releases/tag/v6.22.0. Please consider sponsoring.

PR #2460, opened 2023-10-30
Apollo federation v2

### LINKED ISSUE #2160: Support Apollo Federation 2
Currently Lighthouse has support for Apollo Federation, however recently Apollo Federation 2 came out of beta and is now completely released. 
Lighthouse does not currently support Apollo Federation 2, so this issue is about adding support for it and updating our compatibility on the subgraph repository to support it also.

https://github.com/apollographql/apollo-federation-subgraph-compatibility#:~:text=PHP%20(Laravel)-,Lighthouse,-_service

cappuc:
This PR adds support for [Apollo Federation v2](https://www.apollographql.com/docs/federation/subgraph-spec/) (except for federated tracing).

I've tested this changes with the apollo compatibility test suite and I will make a PR with the required changes to the test app ([here](https://github.com/cappuc/apollo-federation-subgraph-compatibility/tree/lighthouse-v2) you can see the changes for the compatibility test). 

Resolves #2160 
 
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

In order to support Federation v2 the subgraph needs to extend the `schema` with the link to the spec version and import the used directives:

```graphql
extend schema
  @link(
    url: "https://specs.apollo.dev/federation/v2.3",
    import: [
      "@composeDirective",
      "@extends",
      "@external",
      "@inaccessible",
      "@interfaceObject",
      "@key",
      "@override",
      "@provides",
      "@requires",
      "@shareable",
      "@tag"
    ]
  )
```

With this implementation the schema extension must be manually added by the user to the graphql schema.

I've added the new directives of v2 but right now there is no check on usage of this new directives without the schema extension.

The other required changes are on the `FederationPrinter` because we need to print in the service sdl the schema extension and the definition of directives passed to [`@composeDirective`](https://www.apollographql.com/docs/federation/federated-types/federated-directives#composedirective).

**Possible improvements**

The apollo spec allow to use federation directives without the import, adding the spec prefix:

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3")

type Product @federation__key(fields: "id") {
 id: ID!
}
```

or with alias: 

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", as: "fed")

type Product @fed__key(fields: "id") {
 id: ID!
}
```

and also renaming imported directives: 

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", import: [{ name: "@key", as: "@uniqueKey"}])

type Product @uniqueKey(fields: "id") {
 id: ID!
}
```

This options are not supported right now.

**Breaking changes**

It should be backward compatible with v1 because the schema extension required for v2 must be manually added by the user.

### DIFF
--- a/tests/Integration/Federation/FederationSchemaTest.php
+++ b/tests/Integration/Federation/FederationSchemaTest.php
@@ -18,7 +18,8 @@ final class FederationSchemaTest extends TestCase
 
     public function testServiceQueryShouldReturnValidSdl(): void
     {
-        $foo = /** @lang GraphQL */ <<<'GRAPHQL'
+        $foo /** @lang GraphQL */
+            = <<<'GRAPHQL'
 type Foo @key(fields: "id") {
   id: ID! @external
   foo: String!
@@ -26,7 +27,8 @@ type Foo @key(fields: "id") {
 
 GRAPHQL;
 
-        $query = /** @lang GraphQL */ <<<'GRAPHQL'
+        $query /** @lang GraphQL */
+            = <<<'GRAPHQL'
 type Query {
   foo: Int!
 }
@@ -43,7 +45,8 @@ GRAPHQL;
 
     public function testServiceQueryShouldReturnValidSdlWithoutQuery(): void
     {
-        $foo = /** @lang GraphQL */ <<<'GRAPHQL'
+        $foo /** @lang GraphQL */
+            = <<<'GRAPHQL'
 type Foo @key(fields: "id") {
   id: ID! @external
   foo: String!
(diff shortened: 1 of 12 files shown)


tests/Integration/Federation/FederationSchemaTest.php
```diff
@@ -18,15 +18,17 @@ protected function getPackageProviders($app): array
 
     public function testServiceQueryShouldReturnValidSdl(): void
     {
-        $foo = /** @lang GraphQL */ <<<'GRAPHQL'
+        $foo /** @lang GraphQL */
+            = <<<'GRAPHQL'
```
spawnia, line 22: I liked it better in one line

tests/Integration/Federation/FederationSchemaTest.php
```diff
@@ -86,6 +89,43 @@ public function testFederatedSchemaShouldContainCorrectEntityUnion(): void
         $this->assertSame('Bar', $types[1]->name);
     }
 
+    public function testServiceQueryShouldReturnFederationV2SchemaExtension(): void
+    {
+        $schemaExtension = 'extend schema @link(url: "https:\/\/specs.apollo.dev\/federation\/v2.3", import: ["@composeDirective", "@extends", "@external", "@inaccessible", "@interfaceObject", "@key", "@override", "@provides", "@requires", "@shareable", "@tag"])';
```
spawnia, line 94: Use `/** @lang GraphQL */ <<<'GRAPHQL'`

spawnia (changes requested):
Nicely done, happy to include this.

spawnia:
Do the docs need additional updates? Any gotchas regarding how Lighthouse implements the spec?

cappuc:
I've update the docs with some info of how to enabled v2 support and the unsupported features regarding directive imports

stayallive:
@cappuc I'm playing around with this trying to combine multiple Lighthouse servers into a supergraph and running into that the built-in types are colliding, like the `PaginatorInfo` type. They should be marked `@shareable` I guess (which I manually did to make it work).

Did you encounter that too or are you not using this with multiple Lighthouse servers yet?

cappuc:
I think that always adding the `@shareable` will break federation v1.
But maybe it con works if we detect what version of federation is used and drop the v2 only directives when printing the schema.

stayallive:
Ah yes. That makes sense, could also be a configuration setting to make it easier. But forgot that it's backwards compatible of course.

cappuc:
@stayallive I've opened a PR #2485 to mark pagination types as `@shareable`.
Did you find other types that should be marked as `@shareable`?

stayallive:
Not yet, but very nice, I don't have a complicated graph but pagination is something I use heavily so maybe there is more, I'm sure someone will open an issue if needed 💪

Issue #2462, opened 2023-11-03
Extend union support


jsvdvis:
**What problem does this feature proposal attempt to solve?**
As far as I can see, the package currently does not support extending unions. The syntax is parsed but silently ignored near https://github.com/nuwave/lighthouse/blob/master/src/Schema/AST/ASTBuilder.php#L122.

Am I missing something here? Is there a specific reason this is the case?

**Which possible solutions should be considered?**

If there is no specific reason that this is ignored, I am willing to write a PR for this, though this might take me a while. If there are specific things I should keep in mind when doing this, please let me know.

PR #2468, opened 2023-11-13
Extend union support

jsvdvis:
Resolves https://github.com/nuwave/lighthouse/issues/2462

- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This MR adds support for extending unions. The `extend union` syntax was already supported by the parser, but was not yet implemented in the `ASTBuilder`.

**Breaking changes**

Code that already uses the syntax might expect it to silently be ignored, which is no longer the case with this MR.


jsvdvis:
The failing phpstan check is likely due to a recent phpstan package update and unrelated to my changes.

spawnia (approved):

spawnia:
Thank you, could you please resolve conflicts so this can be merged?

Issue #2474, opened 2023-11-21
New InputManipulator directive


jaulz:
**What problem does this feature proposal attempt to solve?**

It would be great if there was a possibility to define directives that can manipulate the input value. There is already `ArgManipulator` but that one can only manipulate root level arguments as far as I could debug.

**Which possible solutions should be considered?**

We could add a new method `applyInputManipulators` in `src/Schema/AST/ASTBuilder.php` to apply this new manipulator on `InputValueDefinitionNode`:
```php


    /** Apply directives on input that can manipulate the AST. */
    protected function applyInputManipulators(): void
    {
        foreach ($this->documentAST->types as $typeDefinition) {
            if ($typeDefinition instanceof InputObjectTypeDefinitionNode) {
                foreach ($typeDefinition->fields as $fieldDefinition) {
                    foreach (
                        $this->directiveLocator->associatedOfType($fieldDefinition, InputManipulator::class) as $inputManipulator
                    ) {
                        $inputManipulator->manipulateInputDefinition(
                            $this->documentAST,
                            $fieldDefinition,
                            $typeDefinition,
                        );
                    }
                }
            }
        }
    }
```

 The corresponding interface could look like this:
```php
<?php declare(strict_types=1);

namespace Nuwave\Lighthouse\Support\Contracts;

use GraphQL\Language\AST\InputObjectTypeDefinitionNode;
use GraphQL\Language\AST\InputValueDefinitionNode;
use Nuwave\Lighthouse\Schema\AST\DocumentAST;

interface InputManipulator extends Directive
{
    /** Manipulate the AST. */
    public function manipulateInputDefinition(
        DocumentAST &$documentAST,
        InputValueDefinitionNode &$inputDefinition,
        InputObjectTypeDefinitionNode &$parentType,
    ): void;
}

```

spawnia:
Thank you @jaulz, released with https://github.com/nuwave/lighthouse/releases/tag/v6.31.0.

PR #2476, opened 2023-11-22
feat: add `InputManipulator` directive

jaulz:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md

**Changes**

As mentioned in https://github.com/nuwave/lighthouse/issues/2474 it would be great if there would be a way to manipulate nested input values. Unfortunately, we cannot change the `ArgManipulator` in a non-breakable way because the `$parentType` is different and also there is no `$parentField`. Thus, I think we need a new directive interface that allows users to manipulate values in a similar way like it's possible with arguments today. What do you think?

**Breaking changes**

No breaking changes. The failing tests are unrelated as far as I can see.


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+- Add `InputManipulator` directive https://github.com/nuwave/lighthouse/pull/2476
```
```suggestion
### Added

- Add `InputManipulator` directive interface https://github.com/nuwave/lighthouse/pull/2476
```

src/Schema/AST/ASTBuilder.php
```diff
@@ -222,5 +224,25 @@ protected function applyArgManipulators(): void
                 }
             }
         }
+    }    
+    
+    /** Apply directives on input that can manipulate the AST. */
```
spawnia, line 229:
```suggestion
    /** Apply directives on input fields that can manipulate the AST. */
```

src/Support/Contracts/InputManipulator.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
+use GraphQL\Language\AST\InputObjectTypeDefinitionNode;
+use GraphQL\Language\AST\InputValueDefinitionNode;
+use Nuwave\Lighthouse\Schema\AST\DocumentAST;
+
+interface InputManipulator extends Directive
```
spawnia, line 9: I think this should be called `InputFieldManipulator`

spawnia:
I am not quite sure where to add this in the docs. `argument-directives.md` seems misleading, perhaps we need to rename it to `input-value-directives.md` to be more general, or split it into `argument-directives.md` and `input-field-directives.md`. Some directives can be both, so I think we will need `input-value-directives.md` anyways.

jaulz:
Sure, I will work on this and the other PR soon since I am quite busy right now but I didn't forget it ;)

jaulz:
@spawnia I gave it a try, could you please have a look?

jaulz:
@spawnia thanks!

PR #2477, opened 2023-11-28
fix federation schema printer

cappuc:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Fix a bug in the federation printer introduced in #2460.

It was always printing `extend schema` even when there are no extensions.

### DIFF
--- a/tests/Console/PrintFederationSchemaCommandTest.php
+++ b/tests/Console/PrintFederationSchemaCommandTest.php
@@ -38,9 +38,11 @@ GRAPHQL;
         $tester = $this->commandTester(new PrintSchemaCommand());
         $tester->execute(['--federation' => true]);
 
-        $this->assertStringContainsString(self::SCHEMA_TYPE, $tester->getDisplay());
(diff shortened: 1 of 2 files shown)


tests/Console/PrintFederationSchemaCommandTest.php
```diff
@@ -40,6 +40,7 @@ public function testPrintsSchemaAsGraphQLSDL(): void
 
         $this->assertStringContainsString(self::SCHEMA_TYPE, $tester->getDisplay());
         $this->assertStringContainsString(self::SCHEMA_QUERY, $tester->getDisplay());
+        $this->assertStringNotContainsString('extend schema', $tester->getDisplay());
```
spawnia, line 43: Let's deduplicate those repeated calls.

spawnia (changes requested):

spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.23.2.

PR #2479, opened 2023-11-29
Federated tracing

### LINKED ISSUE #2053: Federated trace data

Full compliance with the [Apollo Federation Subgraph compatibility test suite](https://github.com/apollographql/apollo-federation-subgraph-compatibility).

https://www.apollographql.com/docs/federation/metrics

### LINKED ISSUE #2068: Protocol buffer tracing standard
Apollo has deprecated it's pre-version 3 [tracing format](https://github.com/apollographql/apollo-tracing#response-format) in favour of a [protobuff implementation](https://www.apollographql.com/docs/studio/metrics/usage-reporting/#tracing-format) based on the operation level. 

Investigate and implement buffered tracing solution as per the [reference implementation](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-reporting-protobuf/src/reports.proto)

cappuc:
Resolves #2053 
Resolves #2068 
 
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR adds support for [Apollo Federated tracing](https://www.apollographql.com/docs/federation/metrics/).

I used the [apollo server](https://github.com/apollographql/apollo-server/blob/main/packages/server/src/plugin/inlineTrace/index.ts) implementation as reference.

**Error reporting**

Federated tracing spec supports errors reporting but we need to decide how to implement this feature in order to prevent exposing sensitive data (without user control), **right now there is no setting for this**.

Apollo server has this options:
- mask error (default)
- include the full error
- transform the error with a callback

Here the [docs](https://www.apollographql.com/docs/apollo-server/api/plugin/inline-trace#options) and [implementation](https://github.com/apollographql/apollo-server/blob/c1105b3b566b293c849922c7f600cc6b4f5f8ecb/packages/server/src/plugin/traceTreeBuilder.ts#L43C1-L53C51)

**Breaking changes**

None. the old tracing implementation is enabled by default.
This new implementation must be manually enabled through config.

**Some notes**

The new tracing spec uses `protobuf` and the updated of the proto file is manual:
- copy the `reports.proto` file from the [apollo-server repo](https://github.com/apollographql/apollo-server/blob/main/packages/usage-reporting-protobuf/src/reports.proto)
- remove the js specific directives (for example: `[(js_use_toArray)=true]` and `[(js_preEncoded)=true]`
- add the php options:
   ```proto
   option php_namespace = "Nuwave\\Lighthouse\\Tracing\\FederatedTracing\\Proto";
   option php_metadata_namespace = "Nuwave\\Lighthouse\\Tracing\\FederatedTracing\\Proto\\Metadata";
   ```
Then run `make:proto` and `make:fix` to generate (and fix sytle) the protobuf classes


composer.json
```diff
@@ -38,6 +38,7 @@
         "illuminate/routing": "^9 || ^10",
         "illuminate/support": "^9 || ^10",
         "illuminate/validation": "^9 || ^10",
+        "google/protobuf": "^3.21",
```
spawnia, line 41: Let's add that to the dev-dependencies and to `suggest` as well, adding new deps is a breaking change.

src/Events/BuildExtensionsResponse.php
```diff
@@ -2,10 +2,20 @@
...
  * Listeners may return a @see \Nuwave\Lighthouse\Execution\ExtensionsResponse
  * to include in the response.
  */
-class BuildExtensionsResponse {}
+class BuildExtensionsResponse
+{
+    public function __construct(
+        /**
+         * The result of resolving a single operation.
+         */
```
spawnia, line 18:
```suggestion
        /** The result of resolving a single operation. */
```

src/Tracing/ApolloTracing/ApolloTracing.php
```diff
@@ -0,0 +1,78 @@
...
+use Nuwave\Lighthouse\Events\StartExecution;
+use Nuwave\Lighthouse\Events\StartRequest;
+use Nuwave\Lighthouse\Execution\ExtensionsResponse;
+use Nuwave\Lighthouse\Execution\ResolveInfo;
+use Nuwave\Lighthouse\Tracing\Tracing;
+use Nuwave\Lighthouse\Tracing\TracingUtilities;
+
+/**
+ * See https://github.com/apollographql/apollo-tracing#response-format.
+ */
```
spawnia, line 16:
```suggestion
/** See https://github.com/apollographql/apollo-tracing#response-format. */
```

src/Tracing/FederatedTracing/FederatedTracing.php
```diff
@@ -0,0 +1,194 @@
...
+use Nuwave\Lighthouse\Federation\FederationServiceProvider;
+use Nuwave\Lighthouse\Tracing\FederatedTracing\Proto\Trace;
+use Nuwave\Lighthouse\Tracing\FederatedTracing\Proto\Trace\Location;
+use Nuwave\Lighthouse\Tracing\FederatedTracing\Proto\Trace\Node;
+use Nuwave\Lighthouse\Tracing\Tracing;
+use Nuwave\Lighthouse\Tracing\TracingUtilities;
+
+/**
+ * See https://www.apollographql.com/docs/federation/metrics/.
+ */
```
spawnia, line 25:
```suggestion
/** See https://www.apollographql.com/docs/federation/metrics. */
```

src/Tracing/FederatedTracing/FederatedTracing.php
```diff
@@ -0,0 +1,194 @@
...
+{
+    use TracingUtilities;
+
+    protected bool $isSubgraph;
+
+    protected bool $enabled = true;
+
+    protected Trace $trace;
+
+    protected float|int $requestStartPrecise;
```
spawnia, line 36: Unit?

src/Tracing/FederatedTracing/FederatedTracing.php
```diff
@@ -0,0 +1,194 @@
...
+        $this->trace->setDurationNs($this->diffTimeInNanoseconds($this->requestStartPrecise, $requestEndPrecise));
+
+        foreach ($buildExtensionsResponse->result->errors as $resultError) {
+            $this->recordError($resultError);
+        }
+
+        assert($this->trace !== null);
+
+        return new ExtensionsResponse(
+            'ftv1',
```
spawnia, line 101: duplicated magic value

src/Tracing/TracingServiceProvider.php
```diff
@@ -25,7 +36,8 @@ public function boot(Dispatcher $dispatcher): void
             $manipulateAST->documentAST,
             Parser::constDirective('@tracing'),
         ));
-        $dispatcher->listen(StartExecution::class, Tracing::class . '@handleStartExecution');
-        $dispatcher->listen(BuildExtensionsResponse::class, Tracing::class . '@handleBuildExtensionsResponse');
```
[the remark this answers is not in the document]
cappuc: I had problems when running with laravel octane using the shortcuts

cappuc:
I've extracted the `ftv1` string to a constant and replaced it inside the driver and in the headers value of the tests.

I'm not sure it is a good idea to replace it in the test assertions too because it should be replaced event in calls like `$response->json('extensions.ftv1')` if we want to always use the constant and I think it is worst for readability.

I also renamed the value of the config to `federated-tracing` so it is not confused with the header/extension value.

composer.json
```diff
@@ -77,7 +78,9 @@
         "laravel/scout": "Required for the @search directive",
         "mll-lab/graphql-php-scalars": "Useful scalar types, required for @whereConditions",
         "mll-lab/laravel-graphiql": "A graphical interactive in-browser GraphQL IDE - integrated with Laravel",
-        "pusher/pusher-php-server": "Required when using the Pusher Subscriptions driver"
+        "pusher/pusher-php-server": "Required when using the Pusher Subscriptions driver",
+        "google/protobuf": "Required when using ftv1 tracing driver",
```
spawnia, line 82:
```suggestion
        "google/protobuf": "Required when using the tracing driver federated-tracing",
```

src/Tracing/TracingServiceProvider.php
```diff
@@ -25,7 +36,8 @@ public function boot(Dispatcher $dispatcher): void
...
+        $dispatcher->listen(StartRequest::class, static fn (StartRequest $event) => Container::getInstance()->make(Tracing::class)->handleStartRequest($event));
```
spawnia, line 39:
```suggestion
        $dispatcher->listen(StartRequest::class, static fn (StartRequest $event) => Container::getInstance()
            ->make(Tracing::class)
            ->handleStartRequest($event));
```

src/lighthouse.php
```diff
@@ -453,4 +453,24 @@
...
+
+    'tracing' => [
+        /*
+         * Driver used for tracing.
+         *
+         * Available options: 'apollo-tracing', 'federated-tracing'
+         *
+         * Will be removed in Lighthouse v7 and implicitly default to 'federated-tracing'.
+         */
+        'driver' => env('LIGHTHOUSE_TRACING', \Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::NAME),
```
spawnia, line 474:
```suggestion
        'driver' => env('LIGHTHOUSE_TRACING', Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::NAME),
```

spawnia: I am not sure an `env` variable is needed, I don't see this changing per environment? We could just use the class names instead then, that would also allow the addition of custom classes because we would no longer need to hardcode driver name to classes.

tests/Integration/Tracing/FederatedTracingExtensionTest.php
```diff
@@ -0,0 +1,164 @@
...
+            parent::getPackageProviders($app),
+            [FederationServiceProvider::class, TracingServiceProvider::class],
+        );
+    }
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        config()->set('lighthouse.tracing.driver', FederatedTracing::NAME);
```
spawnia, line 34:
```suggestion
        $config = $this->app->make(ConfigRepository::class);
        $config->set('lighthouse.tracing.driver', FederatedTracing::NAME);
```

src/lighthouse.php
cappuc: Changing the tracing driver per environment I don't think is too useful but it can be useful for example to enable/disable tracing if we add a `null` driver that does nothing. This can be useful for testing/local environment where tracing is only an overhead. If we remove use the class name directly, this can be implemented with a simple boolean flag.

Another thing to consider is keep the "driver" architecture to allow other drivers to be supported in the future (for example `opentelemetry`)

cappuc: The new federated tracing is enabled only when there is the `apollo-federation-include-trace` header, the old apollo tracing is always enabled.

Anyway conditionally registering the service provider is a good solution for enabling/disabling the tracing.

buf.gen.yaml
```diff
@@ -0,0 +1,7 @@
+# buf.gen.yaml
```
spawnia, line 1:
```suggestion
```

cappuc:
The workflow should work but I couldn't get it to work on pull request (the main problem is the commit part).
I enabled it only on push like the `format` workflow

buf.gen.yaml
```diff
@@ -0,0 +1,6 @@
+version: v1
+managed:
+  enabled: true
+plugins:
+  - plugin: buf.build/protocolbuffers/php
```
cappuc, line 5: @spawnia Do you want to pin a specific protobuf version or always use the latest?

CONTRIBUTING.md
```diff
@@ -113,6 +113,14 @@ $user = User::create([
 ]);
 ```
 
+## Working with proto files
+
+Lighthouse uses [protobuf](https://developers.google.com/protocol-buffers) files for federated tracing schema `src/Tracing/FederatedTracing/Schema`.
+When updating the proto files, PHP classes needs to be regenerated.
+The generation is done with [buf](https://buf.build/docs/generate/overview).
+The `make proto` command generates the new PHP classes and replace the old ones.
+
```
spawnia, line 122:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -113,6 +113,14 @@ $user = User::create([
 ]);
 ```
 
+## Working with proto files
+
+Lighthouse uses [protobuf](https://developers.google.com/protocol-buffers) files for federated tracing schema `src/Tracing/FederatedTracing/Schema`.
```
spawnia, line 118:
```suggestion
Lighthouse uses [protobuf](https://developers.google.com/protocol-buffers) files for [federated tracing](src/Tracing/FederatedTracing/Schema).
```

CONTRIBUTING.md
```diff
@@ -113,6 +113,14 @@ $user = User::create([
...
+When updating the proto files, PHP classes needs to be regenerated.
```
spawnia, line 119:
```suggestion
When updating the proto files, the PHP classes need to be regenerated.
```

buf.gen.yaml
spawnia: Can we pin a major version to avoid incidental breaking changes but still allow for bugfixes?

src/Tracing/TracingServiceProvider.php
```diff
@@ -3,19 +3,30 @@
...
     public function register(): void
     {
-        $this->app->singleton(Tracing::class);
+        $this->app->scoped(Tracing::class, static function (Application $app): Tracing {
+            if (config('lighthouse.tracing.driver') === FederatedTracing::class) {
+                return $app->make(FederatedTracing::class);
+            }
+
+            return $app->make(ApolloTracing::class);
+        });
```
spawnia, line 29:
```suggestion
        $this->app->scoped(Tracing::class, static fn (Application $app): Tracing => $app->make(config('lighthouse.tracing.driver')));
```

src/lighthouse.php
```diff
@@ -453,4 +453,24 @@
...
+    |
+    | Configuration for tracing support.
+    |
+    */
+
+    'tracing' => [
+        /*
+         * Driver used for tracing.
+         *
+         * Available options: '\Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::class', '\Nuwave\Lighthouse\Tracing\FederatedTracing\FederatedTracing::class'
```
spawnia, line 470:
```suggestion
         * Accepts the fully qualified class name of a class that implements Nuwave\Lighthouse\Tracing\Tracing.
         * Lighthouse provides:
         * - Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::class
         * - Nuwave\Lighthouse\Tracing\FederatedTracing\FederatedTracing::class
```

src/lighthouse.php
```diff
@@ -453,4 +453,24 @@
...
+         *
+         * Will be removed in Lighthouse v7 and implicitly default to 'Nuwave\Lighthouse\Tracing\FederatedTracing\FederatedTracing::class'.
+         */
+        'driver' => \Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::class,
```
spawnia, line 474:
```suggestion
        'driver' => Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::class,
```

src/lighthouse.php
```diff
@@ -453,4 +453,24 @@
...
+    |
+    */
+
+    'tracing' => [
+        /*
+         * Driver used for tracing.
+         *
+         * Available options: '\Nuwave\Lighthouse\Tracing\ApolloTracing\ApolloTracing::class', '\Nuwave\Lighthouse\Tracing\FederatedTracing\FederatedTracing::class'
+         *
+         * Will be removed in Lighthouse v7 and implicitly default to 'Nuwave\Lighthouse\Tracing\FederatedTracing\FederatedTracing::class'.
```
spawnia, line 472: I guess we can still allow overwriting the binding of the scoped singleton for customization then?

buf.gen.yaml
cappuc: The plugin follows the protobuf version number (and there is a new "major" version every few months but I think they don't follow semver) but usually there are no breaking changes in generated PHP code. 
In the last 4/5 releases that I used they added some phpdoc to improve types

Here are the release pages:
https://buf.build/protocolbuffers/php
https://github.com/protocolbuffers/protobuf/releases

src/lighthouse.php
cappuc: Yes we can simply change the default in v7 but keeping the driver structure so it allows customization

cappuc:
I've added a workflow to automatically update the `reports.proto` file from apollo. I set it to run weekly

buf.gen.yaml
```diff
@@ -0,0 +1,6 @@
+version: v1
+managed:
+  enabled: true
+plugins:
+  - plugin: buf.build/protocolbuffers/php:v25.1
```
spawnia, line 5:
```suggestion
  # For new releases, see https://github.com/protocolbuffers/protobuf/releases
  - plugin: buf.build/protocolbuffers/php:v25.1
```

.github/workflows/update-reports-proto.yml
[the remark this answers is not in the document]
cappuc: There is this option to push to protected branches:
https://github.com/stefanzweifel/git-auto-commit-action#push-to-protected-branches

spawnia:
Apart from adding the changelog entry and updating the docs to say that federated tracing is supported, I am fine to go ahead and merge this. Excellent work!

Do you plan on updating the Apollo compatibility repository as well? I think we should just need to change `tracing.driver` in the configuration to make the example project there work?

cappuc:
We should take a look at error reporting (I wrote some notes in the first comment).

Apart from that and changelog/docs update it, everything should be fine.

Yes I think the only required change for apollo compatibility is the change of `tracing.driver`. I will make a PR to update apollo compatibility when this PR is merged.

spawnia:
I am not sure how error reporting in tracing is/should be different from error reporting for clients. Does https://lighthouse-php.com/master/digging-deeper/error-handling.html apply? It would be nice to keep things simple and just utilize the same mechanisms. See `\Nuwave\Lighthouse\GraphQL::errorsHandler()`.

cappuc:
Yes it apply all handlers.
I just changed the message to be "Internal server error" when error is not client safe and removed the json data, so we provide only message and location

docs/master/federation/getting-started.md
```diff
@@ -50,6 +50,13 @@ extend schema @link(
 )
 ```
 
+## Federated tracing
+
+In order to use federated tracing, you need to enabled [tracing](../performance/tracing.md)
+and set the driver to `Nuwave\Lighthouse\Tracing\FederatedTracing\FederatedTracing::class` in your `config/lighthouse.php`:
```
spawnia, line 56: Were you planning on adding example code here, or should we end this with `.`?

docs/master/performance/tracing.md
```diff
@@ -12,3 +11,17 @@ Add the service provider to your `config/app.php`
     \Nuwave\Lighthouse\Tracing\TracingServiceProvider::class,
 ],
 ```
+
+## Drivers
+
+Lighthouse tracing is implemented though drivers, this allows to easily support different tracing formats.
```
spawnia, line 17:
```suggestion
Lighthouse tracing is implemented though drivers, this allows supporting different tracing formats.
```

https://justsimply.dev 😉

spawnia (changes requested):

spawnia:
Ready to merge and release?

cappuc:
Yes, I think it's ready

Maxwell2022:
We need to update the doc too: https://lighthouse-php.com/6/federation/getting-started.html#federated-tracing

> Experimental: not enabled by default, not guaranteed to be stable.

Is this still experimental?

PR #2485, opened 2023-12-11
Federation v2 `@shareable` types

cappuc:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR marks shared types (types that can be shared by multiple services) with the `@shareable` directive when federation v2 is enabled.

Right now I've updated the pagination types. @spawnia do you think there are other types that makes sense to mark as `@shareable`?

**Breaking changes**

None.
 When using federation v2, types declared in multiple services must be marked as `@shareable` so right now you must override this types in order to use federation v2

### DIFF
--- a/src/Federation/FederationHelper.php
+++ b/src/Federation/FederationHelper.php
@@ -5,6 +5,7 @@ namespace Nuwave\Lighthouse\Federation;
 use GraphQL\Language\AST\DirectiveNode;
 use GraphQL\Type\Schema;
 use Nuwave\Lighthouse\Schema\AST\ASTHelper;
+use Nuwave\Lighthouse\Schema\AST\DocumentAST;
 
 class FederationHelper
 {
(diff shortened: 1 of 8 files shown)


src/Federation/FederationHelper.php
```diff
@@ -43,4 +44,23 @@ public static function schemaExtensionDirectives(Schema $schema): array
 
         return $schemaDirectives;
     }
+
+    public static function isUsingFederationV2(Schema|DocumentAST $schemaOrDocument): bool
+    {
+        $schemaExtensionNodes = $schemaOrDocument instanceof Schema
+            ? $schemaOrDocument->extensionASTNodes
```
spawnia, line 51: Missing tests, does this ever happen?

src/Federation/FederationHelper.php
```diff
@@ -43,4 +44,23 @@ public static function schemaExtensionDirectives(Schema $schema): array
...
+            : $schemaOrDocument->schemaExtensions;
+
+        foreach ($schemaExtensionNodes as $extension) {
+            foreach (ASTHelper::directiveDefinitions($extension, 'link') as $directive) {
+                $url = ASTHelper::directiveArgValue($directive, 'url');
+
+                if (str_starts_with($url, 'https://specs.apollo.dev/federation/')) {
```
spawnia, line 58:
```suggestion
                if (str_starts_with($url, 'https://specs.apollo.dev/federation/v2')) {
```

tests/Integration/Federation/FederationSchemaTest.php
```diff
@@ -8,6 +8,10 @@
 
 final class FederationSchemaTest extends TestCase
 {
+    protected string $federationV2SchemaExtension = /** @lang GraphQL */ <<<'GRAPHQL'
```
spawnia, line 11: `const`

spawnia (changes requested):

src/Federation/FederationHelper.php
spawnia: It is not called there though. Let's not add features we do not actually need.

tests/Integration/Federation/FederationSchemaTest.php
cappuc: I've changed to `const` and concatenated with the rest of the schema definition

spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.28.0

PR #2492, opened 2023-12-25
Fix guard type in stateful authentication example

lykbl:
Logout sanctum example is missing a Arr::first() call, which causes an illegal offset exception


spawnia:
Thank you

PR #2511, opened 2024-02-19
Remove Request from the ThrottleDirective constructor

k0ka:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The PR fixes a case of "persistent" directive usage, when it is ran several times for different requests. This might happen in test or `Laravel Octane` environments. We can't get `Request` in the constructor in these cases. But must get it inside resolver from `GraphQLContext`

### DIFF
--- a/src/Schema/Directives/ThrottleDirective.php
+++ b/src/Schema/Directives/ThrottleDirective.php
@@ -63,7 +63,7 @@ GRAPHQL;
 
         $fieldValue->wrapResolver(fn (callable $resolver): \Closure => function (mixed $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver, $name, $prefix, $maxAttempts, $decayMinutes): mixed {
             $request = $context->request();
-            if ($request == null){
+            if ($request == null) {
                 return $resolver($root, $args, $context, $resolveInfo);
             }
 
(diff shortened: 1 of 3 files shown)


src/Schema/Directives/ThrottleDirective.php
```diff
@@ -58,45 +56,44 @@ public static function definition(): string
...
-                    'maxAttempts' => $limit->maxAttempts,
-                    'decayMinutes' => $limit->decayMinutes,
-                ];
+        $prefix = $this->directiveArgValue('prefix', '');
+        $maxAttempts = $this->directiveArgValue('maxAttempts', 60);
+        $decayMinutes = $this->directiveArgValue('decayMinutes', 1.0);
+
+        $fieldValue->wrapResolver(fn (callable $resolver): \Closure => function (mixed $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($resolver, $name, $prefix, $maxAttempts, $decayMinutes): mixed {
+            $request = $context->request();
+            if ($request == null) {
```
spawnia, line 66:
```suggestion
            if ($request === null) {
```

src/Schema/Directives/ThrottleDirective.php
spawnia, line 1: Can you add tests for the uncovered cases?

src/Schema/Directives/ThrottleDirective.php
k0ka: updated it

src/Schema/Directives/ThrottleDirective.php
k0ka: Added one test for `GraphQLContext` with null `Request`. 
Another line wasn't necessary as we had this check during schema validation.

spawnia:
Thanks!

Issue #2509, opened 2024-02-19
`extend scalar X` directives are lost


LastDragon-ru:
**Describe the bug**

The resulting directives for `Scalar`

```graphql
scalar Scalar 
@scalar(class: "GraphQL\\Type\\Definition\\StringType")
@directive(location: "scalar")

extend scalar Scalar 
@directive(location: "extend")
```

will be 

```graphql
scalar Scalar 
@scalar(class: "GraphQL\\Type\\Definition\\StringType")
@directive(location: "scalar")
```

Meanwhile the graphql-php gives both (as it should be):

```graphql
scalar Scalar
@scalar(class: "GraphQL\\Type\\Definition\\StringType")
@directive(
    location: "scalar"
)
@directive(
    location: "extend"
)
```

**Expected behavior/Solution**

`Scalar` should have all directives from extension nodes.

**Steps to reproduce**

1. Create the schema with `extend scalar X @directive`
2. Print it

**Lighthouse Version**

v6.33.0

PR #2512, opened 2024-02-19
Directives from Scalar/Enum/Input/Type/Interface extensions

LastDragon-ru:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md
- [x] Closes: #2509  

**Changes**

Directives defined in Scalar/Enum/Input/Type/Interface type extensions will be merged into target. The following schema will create type `A` with two directives `@foo` and `@bar` (previously only with `@foo`).

```graphql
type A @foo {
  field: String
}

extend type A @bar
```

Custom scalars seems fine, but for standard (`Int`/`String`/etc) the AST node is required. And I'm not sure how it should work.

Technically, it is possible to add `scalar Int`, but it will not work without `@scalar`[^1]. Adding `@scalar(class: "GraphQL\Type\Definition\IntType")` for standard scalars looks superfluous, IMHO. Moreover despite `scalar Int @scalar(class: "My\Custom\Class")` is fully valid, but seems useless, because `Type::getStandardTypes()` has a bigger priority and thus `My\Custom\Class` will not be used (at least in `Schema::getType()`).

Another issue - to get directives we need AST node, standard types don't have it, so we probably need to assign it somewhere. 

My proposal is to extend `TypeRegistry::resolveScalarType()` to

* Bind AST node to standard type if defined (required to add directives to standard types via `extend Int @mydirective`)[^2].
* Allow standard types without `@scalar`

What do you think?

Also, would be nice if someone can clarify usage of `scalar Int @scalar(class: "My\Custom\Class")` (allowed/should work or not).

**Breaking changes**

* Signature of `ASTBuilder::missingBaseDefinition()` and `ASTBuilder::assertExtensionMatchesDefinition()` changed to accept `ScalarTypeDefinitionNode`/`ScalarTypeExtensionNode` 
* if the `extend scalar X` is used alone (= no scalar X defined it the schema) there will be an error that was not present before. But I think this is ok, because it works the same as for extending other types.

[^1]: Failed to find class Int extends GraphQL\\Type\\Definition\\ScalarType in namespaces [App\\GraphQL\\Scalars] for the scalar Int.
[^2]: I'm not sure that is will work in all cases, will try to test it tomorrow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
-- Merge directives from Scalar/Enum/Type/Input/Interface extension node into target node.
+### Added
+
+- Merge directives from Scalar/Enum/Type/Input/Interface extension node into target node https://github.com/nuwave/lighthouse/pull/2512
 
 ## v6.33.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+- Merge directives from Scalar/Enum/Type/Input/Interface extension node into target node.
+
 ## v6.33.4
 
 ### Fixed
--- a/src/Schema/AST/ASTBuilder.php
+++ b/src/Schema/AST/ASTBuilder.php
@@ -10,6 +10,8 @@ use GraphQL\Language\AST\InterfaceTypeDefinitionNode;
 use GraphQL\Language\AST\InterfaceTypeExtensionNode;
 use GraphQL\Language\AST\ObjectTypeDefinitionNode;
 use GraphQL\Language\AST\ObjectTypeExtensionNode;
+use GraphQL\Language\AST\ScalarTypeDefinitionNode;
+use GraphQL\Language\AST\ScalarTypeExtensionNode;
 use GraphQL\Language\AST\UnionTypeDefinitionNode;
 use GraphQL\Language\AST\UnionTypeExtensionNode;
 use GraphQL\Language\Parser;
@@ -33,6 +35,7 @@ class ASTBuilder
         ObjectTypeExtensionNode::class => ObjectTypeDefinitionNode::class,
         InputObjectTypeExtensionNode::class => InputObjectTypeDefinitionNode::class,
         InterfaceTypeExtensionNode::class => InterfaceTypeDefinitionNode::class,
+        ScalarTypeExtensionNode::class => ScalarTypeDefinitionNode::class,
         EnumTypeExtensionNode::class => EnumTypeDefinitionNode::class,
         UnionTypeExtensionNode::class => UnionTypeDefinitionNode::class,
     ];
@@ -122,6 +125,8 @@ class ASTBuilder
                     || $typeExtension instanceof InterfaceTypeExtensionNode
                 ) {
                     $this->extendObjectLikeType($typeName, $typeExtension);
+                } elseif ($typeExtension instanceof ScalarTypeExtensionNode) {
+                    $this->extendScalarType($typeName, $typeExtension);
                 } elseif ($typeExtension instanceof EnumTypeExtensionNode) {
                     $this->extendEnumType($typeName, $typeExtension);
                 } elseif ($typeExtension instanceof UnionTypeExtensionNode) {
@@ -166,6 +171,17 @@ class ASTBuilder
         }
     }
 
+    protected function extendScalarType(string $typeName, ScalarTypeExtensionNode $typeExtension): void
+    {
+        $extendedScalar = $this->documentAST->types[$typeName]
+            ?? throw new DefinitionException($this->missingBaseDefinition($typeName, $typeExtension));
+        assert($extendedScalar instanceof ScalarTypeDefinitionNode);
+
+        $this->assertExtensionMatchesDefinition($typeExtension, $extendedScalar);
+
+        $extendedScalar->directives = $extendedScalar->directives->merge($typeExtension->directives);
+    }
+
     protected function extendEnumType(string $typeName, EnumTypeExtensionNode $typeExtension): void
     {
         $extendedEnum = $this->documentAST->types[$typeName]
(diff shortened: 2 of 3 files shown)


LastDragon-ru:
> I'm not sure that is will work in all cases, will try to test it tomorrow

Seems it will not work. So, I don't know how to extend directives for standard types and thus they will be processing like custom scalars. This is means that the following graphql will do nothing (the same as before the PR).

```graphql
scalar Int @scalar(class: "xxx")
extend scalar Int @directive(location: "extend")
```

Also, another possible BC: if the `extend scalar X` is used alone (= no scalar X defined it the schema) there will be an error that was not present before. But I think this is ok, because it works the same as for extending other types.

LastDragon-ru:
> Error: Child process error (exit code 1): ErrorException thrown in /home/runner/work/lighthouse/lighthouse/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php on line 335 while loading bootstrap file /home/runner/work/lighthouse/lighthouse/vendor/larastan/larastan/bootstrap.php: symlink(): File exists

Looks like failed tests are not releated to the PR

k0ka:
Yep, these tests are randomly failing. Just re-run them.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+- Merge directives from Scalar/Enum/Type/Input/Interface extension node into target node.
```
```suggestion
### Added

- Merge directives from Scalar/Enum/Type/Input/Interface extension node into target node https://github.com/nuwave/lighthouse/pull/2512
```

src/Schema/AST/ASTBuilder.php
```diff
@@ -194,12 +212,12 @@ protected function extendUnionType(string $typeName, UnionTypeExtensionNode $typ
...
+    protected function missingBaseDefinition(string $typeName, ObjectTypeExtensionNode|InputObjectTypeExtensionNode|InterfaceTypeExtensionNode|ScalarTypeExtensionNode|EnumTypeExtensionNode|UnionTypeExtensionNode $typeExtension): string
     {
         return "Could not find a base definition {$typeName} of kind {$typeExtension->kind} to extend.";
     }
 
-    protected function assertExtensionMatchesDefinition(ObjectTypeExtensionNode|InputObjectTypeExtensionNode|InterfaceTypeExtensionNode|EnumTypeExtensionNode|UnionTypeExtensionNode $extension, ObjectTypeDefinitionNode|InputObjectTypeDefinitionNode|InterfaceTypeDefinitionNode|EnumTypeDefinitionNode|UnionTypeDefinitionNode $definition): void
+    protected function assertExtensionMatchesDefinition(ObjectTypeExtensionNode|InputObjectTypeExtensionNode|InterfaceTypeExtensionNode|ScalarTypeExtensionNode|EnumTypeExtensionNode|UnionTypeExtensionNode $extension, ObjectTypeDefinitionNode|InputObjectTypeDefinitionNode|InterfaceTypeDefinitionNode|ScalarTypeDefinitionNode|EnumTypeDefinitionNode|UnionTypeDefinitionNode $definition): void
```
spawnia, line 220:
```suggestion
    protected function assertExtensionMatchesDefinition(
        ObjectTypeExtensionNode|InputObjectTypeExtensionNode|InterfaceTypeExtensionNode|ScalarTypeExtensionNode|EnumTypeExtensionNode|UnionTypeExtensionNode $extension,
        ObjectTypeDefinitionNode|InputObjectTypeDefinitionNode|InterfaceTypeDefinitionNode|ScalarTypeDefinitionNode|EnumTypeDefinitionNode|UnionTypeDefinitionNode $definition,
    ): void
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -7,14 +7,19 @@
...
 use Illuminate\Support\Collection;
 use Nuwave\Lighthouse\Exceptions\DefinitionException;
 use Nuwave\Lighthouse\Schema\AST\ASTBuilder;
 use Nuwave\Lighthouse\Schema\AST\ASTHelper;
+use Nuwave\Lighthouse\Schema\DirectiveLocator;
+use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
 use Nuwave\Lighthouse\Schema\RootType;
 use Tests\TestCase;
 
+use function assert;
```
spawnia, line 21:
```suggestion
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -49,6 +54,35 @@ public function testMergeTypeExtensionFields(): void
         $this->assertCount(3, $queryType->fields);
     }
 
+    public function testMergeTypeExtensionDirectives(): void
+    {
+        $directive = new class() extends BaseDirective {
+            public static function definition(): string
+            {
+                return /** @lang GraphQL */ 'directive @foo on OBJECT';
```
spawnia, line 62:
```suggestion
                return /** @lang GraphQL */ 'directive @foo repeatable on OBJECT';
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -105,6 +139,35 @@ public function testMergeInputExtensionFields(): void
         $this->assertCount(3, $inputs->fields);
     }
 
+    public function testMergeInputExtensionDirectives(): void
+    {
+        $directive = new class() extends BaseDirective {
+            public static function definition(): string
+            {
+                return /** @lang GraphQL */ 'directive @foo on INPUT_OBJECT';
```
spawnia, line 147:
```suggestion
                return /** @lang GraphQL */ 'directive @foo repeatable on INPUT_OBJECT';
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -128,6 +191,62 @@ interface Named {
         $this->assertCount(3, $named->fields);
     }
 
+    public function testMergeInterfaceExtensionDirectives(): void
+    {
+        $directive = new class() extends BaseDirective {
+            public static function definition(): string
+            {
+                return /** @lang GraphQL */ 'directive @foo on INTERFACE';
```
spawnia, line 199:
```suggestion
                return /** @lang GraphQL */ 'directive @foo repeatable on INTERFACE';
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -128,6 +191,62 @@ interface Named {
...
+
+        $this->assertCount(2, $myInterface->directives);
+    }
+
+    public function testMergeScalarExtensionDirectives(): void
+    {
+        $directive = new class() extends BaseDirective {
+            public static function definition(): string
+            {
+                return /** @lang GraphQL */ 'directive @foo on SCALAR';
```
spawnia, line 228:
```suggestion
                return /** @lang GraphQL */ 'directive @foo repeatable on SCALAR';
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -152,6 +271,36 @@ enum MyEnum {
         $this->assertCount(4, $myEnum->values);
     }
 
+    public function testMergeEnumExtensionDirectives(): void
+    {
+        $directive = new class() extends BaseDirective {
+            public static function definition(): string
+            {
+                return /** @lang GraphQL */ 'directive @foo on ENUM';
```
spawnia, line 279:
```suggestion
                return /** @lang GraphQL */ 'directive @foo repeatable on ENUM';
```

tests/Unit/Schema/AST/ASTBuilderTest.php
```diff
@@ -173,6 +322,26 @@ public function testMergeUnionExtensionFields(): void
         $this->assertCount(3, $myUnion->types);
     }
 
+    public function testDoesNotAllowExtendingUndefinedScalar(): void
+    {
+        $directive = new class() extends BaseDirective {
+            public static function definition(): string
+            {
+                return /** @lang GraphQL */ 'directive @foo on SCALAR';
```
spawnia, line 330:
```suggestion
                return /** @lang GraphQL */ 'directive @foo repeatable on SCALAR';
```

spawnia:
Do you have a use cases for changing the standard types? If not, I would rather not spend effort on it.

LastDragon-ru:
> Do you have a use cases for changing the standard types?

Yep, for `@searchBy` may be useful to `extend scalar Int @appComparisonOperator` (will add a new operator to `Int`). But I've added a workaround for it 🤗 (it just move operators into custom scalar).

> I would rather not spend effort on it.

Me too. Maybe it will be easy to implement it after https://github.com/webonyx/graphql-php/pull/1426 (but I didn't dig into pr)

PR #2522, opened 2024-03-07
Introduce ClientSafeModelNotFoundException for `@can` and `@canFind` directives

ChaplyginWezom:
Resolves #2521 
 
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Added custom `ClientSafeModelNotFoundException` to wrap around `ModelNotFoundException` in `@can` and `@canFind` directives.
New exception is linked to previous exception, allowing for usage in custom error handlers. With this change developers could be able to modify original exception how they want, formatting or replacing conditionally

**Breaking changes**

Probably none, as new exception being safely catched into same `GraphQL\Error` and maintain original message

### DIFF
--- a/src/Exceptions/ClientSafeModelNotFoundException.php
+++ b/src/Exceptions/ClientSafeModelNotFoundException.php
@@ -3,11 +3,26 @@
 namespace Nuwave\Lighthouse\Exceptions;
 
 use GraphQL\Error\ClientAware;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 
-class ClientSafeModelNotFoundException extends \Exception implements ClientAware
+/**
+ * @template TModel of \Illuminate\Database\Eloquent\Model
+ * @extends ModelNotFoundException<TModel>
+ */
+class ClientSafeModelNotFoundException extends ModelNotFoundException implements ClientAware
 {
     public function isClientSafe(): bool
     {
         return true;
+    }
+
+    /**
+     * @param  ModelNotFoundException<TModel>  $laravelException
+     * @return self<TModel>
+     */
+    public static function fromLaravel(ModelNotFoundException $laravelException): self
+    {
+        return (new static($laravelException->getMessage(), $laravelException->getCode()))
+            ->setModel($laravelException->getModel(), $laravelException->getIds());
     }
 }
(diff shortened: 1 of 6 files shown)


src/Exceptions/ClientSafeModelNotFoundException.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types=1);
+
+namespace Nuwave\Lighthouse\Exceptions;
+
+use GraphQL\Error\ClientAware;
+
+class ClientSafeModelNotFoundException extends \Exception implements ClientAware
```
spawnia, line 7: Why not this?

```suggestion
class ClientSafeModelNotFoundException extends ModelNotFoundException implements ClientAware
```

I think it would be better for code that checks for `instanceof ModelNotFoundException`.

tests/Integration/Auth/CanDirectiveDBTest.php
```diff
@@ -91,6 +93,43 @@ public function testFailsToFindSpecificModel(): void
...
+            $this->graphQL(/** @lang GraphQL */ '
+            {
+                user(id: "not-present") {
+                    name
+                }
+            }
+            ');
+        } catch (Error $error) {
+            $this->assertNotEmpty($error->getPrevious());
+            $this->assertEquals(ClientSafeModelNotFoundException::class, get_class($error->getPrevious()));
```
spawnia, line 130:
```suggestion
            $previous = $error->getPrevious();
            $this->assertNotNull($previous);
            $this->assertSame(ClientSafeModelNotFoundException::class, get_class($previous));

            $previousPrevious = $previous->getPrevious();
            $this->assertNotNull($previousPrevious);
            $this->assertSame(ModelNotFoundException::class, get_class($previousPrevious));
```

tests/Integration/Auth/CanFindDirectiveDBTest.php
```diff
@@ -128,6 +130,44 @@ public function testFailsToFindSpecificModel(): void
...
+            $this->graphQL(/** @lang GraphQL */ '
+            {
+                user(id: "not-present") {
+                    name
+                }
+            }
+            ');
+        } catch (Error $error) {
+            $this->assertNotEmpty($error->getPrevious());
+            $this->assertEquals(ClientSafeModelNotFoundException::class, get_class($error->getPrevious()));
```
spawnia, line 167: As in the other test.

ChaplyginWezom:
Ok, so I've extended ModelNotFoundException and now mention of previous exception is redundant

spawnia:
Thank you

Issue #2222, opened 2022-10-31
`@with` directive fails to eager-load multiple nested relations on the same parent relation


cradigan:
I've come across what i think is a bug involving the use of multiple @with directives

Here is an example schema:

```
type User{
   viewerSummary: ViewerSummary
     @with(relation: "shows.producers")
     @with(relation: "shows.actors")
     @with(relation: "films.producers")
     @with(relation: "films.actors")
     @method(name: "getViewerSummary")
}
```

In the given example, after loading a user with viewerSummary, actors are eager-loaded but producers are not.

**Expected behavior/Solution**

I would expect both relationships (producers & actors) to be eager loaded 

**Lighthouse Version**

v5.63

spawnia:
This issue was uncovered in https://github.com/nuwave/lighthouse/pull/1915 and has not been fixed yet, see https://github.com/nuwave/lighthouse/blob/26c25e9eadd0b5aecdd637a84354b82d93e2e9bd/tests/Integration/Schema/Directives/WithDirectiveTest.php#L306-L308.

https://github.com/nuwave/lighthouse/issues/1873 is related, it is probably a good idea to work on both issues at once.

spawnia:
Duplicate of https://github.com/nuwave/lighthouse/issues/1132.

Issue #1873, opened 2021-06-09
Deduplicate nested relations using dot notation with equivalent fields


spawnia:
**What problem does this feature proposal attempt to solve?**

See https://github.com/nuwave/lighthouse/pull/1871. Database queries are unnecessarily duplicated.

**Which possible solutions should be considered?**

When using dot notation to batch load a relation, e.g. `@with(relation: "foo.bar.baz")`, we would have to pick apart the individual relation segments and resolve a batch loader for each one separately.

Rough draft of how that might look:

```php
        /** @var array<int, \Nuwave\Lighthouse\Execution\BatchLoader\RelationBatchLoader> $relationBatchLoader */
        $loaders = [];

        // Dot notation may be used to eager load nested relations
        $parts = explode('.', $this->relation());

        // Includes the field we are loading the relation for
        $path = $resolveInfo->path;

        // In case we have no args, we can combine eager loads that are the same
        if ($args === []) {
            array_pop($path);
        }

        $lastRelation = array_pop($parts);

        foreach ($parts as $intermediaryRelation) {
            $path []= $intermediaryRelation;
            $loaders []= BatchLoaderRegistry::instance(
                $intermediaryRelation,
                function () use ($resolveInfo, $intermediaryRelation): RelationBatchLoader {
                    return new RelationBatchLoader($this->relationLoader($resolveInfo, $intermediaryRelation));
                }
            );
        }

        $path []= $lastRelation;
        $path = array_merge($path, $this->scopes());

        $loaders []= BatchLoaderRegistry::instance(
            $path,
            function () use ($resolveInfo, $lastRelation): RelationBatchLoader {
                return new RelationBatchLoader($this->relationLoader($resolveInfo, $lastRelation));
            }
        );

        return new Deferred(static function () use ($loaders, $parent) {
            /** @var RelationBatchLoader $loader */
            foreach ($loaders as $loader) {
                $loader->load($parent);
            }
        });
```

Scopes and additional constraints only apply to the last relation query, so the intermediary queries can safely be batched together with other unconstrained queries on the same relation:

![image](https://user-images.githubusercontent.com/12158000/121381455-085ae380-c946-11eb-9041-2d8ce6441c84.png)

vgedzius:
Eloquent provides a method to load relations only when they are not loaded already `loadMissing`. Could it be utilised here?

spawnia:
[...]

PR #2525, opened 2024-03-14
Eager load multiple nested relations using @with

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/2222, probably also https://github.com/nuwave/lighthouse/issues/1873
 
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**

### DIFF
--- a/src/Exceptions/ClientSafeModelNotFoundException.php
+++ b/src/Exceptions/ClientSafeModelNotFoundException.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\ModelNotFoundException;
 
 /**
  * @template TModel of \Illuminate\Database\Eloquent\Model
+ *
  * @extends ModelNotFoundException<TModel>
  */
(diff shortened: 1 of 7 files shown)


robsontenorio:
In this exactly old version `5.70.3`, the nested relationship eager loading  works .
But not on latest state version 6.x.

```graphql
type Bar { 
      city: City  @method(name: "handleCity") @with(relation: "city.state")
}
````

robsontenorio:
By the way , this PR works fine and solve the problem. If you need some help to merge it let me know.

spawnia:
> By the way , I have tested this PR and works fine, solve the problem. If you need some help to merge it let me know.

The changes I made fix eager loading of multiple `@with`, but break other scenarios - see https://github.com/nuwave/lighthouse/actions/runs/9380728096/job/25828392738. Thus, it can not be merged as is, the final solution will have to look different.

PR #2540, opened 2024-04-09
Compare raw attributes only when ensuring uniqueness of paginated related models

jaulz:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Right now, `toJson` is used which "is recursive, so all attributes and relations will be converted to JSON" (see https://laravel.com/docs/11.x/eloquent-serialization#serializing-to-json). This means that all accessors are executed as well which can affect the performance of an application. The new approach simply compares the raw attributes of the models which should be enough in almost all cases as far as I can see.

**Breaking changes**

There might be a tiny chance that something doesn't work as before because accessors are not compared anymore.

### DIFF
--- a/src/Execution/ModelsLoader/PaginatedModelsLoader.php
+++ b/src/Execution/ModelsLoader/PaginatedModelsLoader.php
@@ -84,7 +84,9 @@ class PaginatedModelsLoader implements ModelsLoader
         return $relatedModels->unique(
             // Compare all attributes because there might not be a unique primary key
             // or there could be differing pivot attributes.
-            static fn (Model $relatedModel): string => $relatedModel->toJson(),
+            static function (Model $relatedModel): string {
+                return json_encode($relatedModel->getOriginal());
+            },
         );
     }
 
--- a/src/Execution/ModelsLoader/PaginatedModelsLoader.php
+++ b/src/Execution/ModelsLoader/PaginatedModelsLoader.php
@@ -85,9 +85,7 @@ class PaginatedModelsLoader implements ModelsLoader
         return $relatedModels->unique(
             // Compare all attributes because there might not be a unique primary key
             // or there could be differing pivot attributes.
-            static function (Model $relatedModel): string {
-                return json_encode($relatedModel->getOriginal());
-            },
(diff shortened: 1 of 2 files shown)


src/Execution/ModelsLoader/PaginatedModelsLoader.php
```diff
@@ -84,7 +85,9 @@ protected function loadRelatedModels(EloquentCollection $parents): EloquentColle
         return $relatedModels->unique(
             // Compare all attributes because there might not be a unique primary key
             // or there could be differing pivot attributes.
-            static fn (Model $relatedModel): string => $relatedModel->toJson(),
+            static function (Model $relatedModel): string {
+                return json_encode($relatedModel->getOriginal());
+            },
```
spawnia, line 90: I don't see the need for a full closure, a short fn is simpler here

src/Execution/ModelsLoader/PaginatedModelsLoader.php
```diff
@@ -84,7 +85,9 @@ protected function loadRelatedModels(EloquentCollection $parents): EloquentColle
         return $relatedModels->unique(
             // Compare all attributes because there might not be a unique primary key
             // or there could be differing pivot attributes.
```
spawnia, line 87: Are pivot attributes also included in `getOriginal()`?

jaulz: Yes, they are:
image

src/Execution/ModelsLoader/PaginatedModelsLoader.php
jaulz: Fixed

skaesdorf:
@spawnia Hi, this breaks when Models have native Enum casted attributes. `$model->getOriginal()` casts the attributes. `json_encode($relatedModel->getOriginal())` will then fail because enums are not stringable.

jaulz:
Hm, the best would actually be to pick `$original` from the model but it's protected.

skaesdorf:
That would be `$model->getRawOriginal()`

jaulz:
@skaesdorf great, can you raise a PR for that?

PR #2542, opened 2024-04-17
Register missing WhereConditionsServiceProvider

xaviapa:
WhereConditionsServiceProvider is missing to be registered in composer.json. Found it because when trying to use the directive @whereHasConditions, it was telling me that it was not found.


spawnia:
As mentioned in the docs, it is experimental and not added by default.

Issue #2520, opened 2024-03-03
An error occurs in the SubscriptionRegistry when sending a subscription event via Subscription::broadcast with Laravel Octane (Swoole)


danbka33:
**Describe the bug**

An error occurs in the SubscriptionRegistry when sending a subscription event. 

```
"debugMessage": "Undefined array key \"scheduleUpdated\"",
"file": "/var/www/app/vendor/nuwave/lighthouse/src/Subscriptions/SubscriptionRegistry.php",
"line": 77,
```

The event is dispatched from model events.

```php

protected static function boot() {
    self::created(fn (PatientAppointment $appointment) => $appointment->subscriptionBroadcast('created'));
    self::updated(fn (PatientAppointment $appointment) => $appointment->subscriptionBroadcast('updated'));
}

/**
     * Broadcast appointment change to subscription.
     * @param string $action
     * @return void
     */
    private function subscriptionBroadcast(string $action): void
    {
        Subscription::broadcast('scheduleUpdated', [
            [
                'appointment' => $this,
                'action' => $action,
            ],
        ]);
    }
```

Definition of subscription

```graphql
extend type Subscription {
    scheduleUpdated(
        date: Date
        statuses: [AppointmentStatus!]
    ): [AppointmentChange!]
}

```

**Expected behavior/Solution**

**Steps to reproduce**

1. Use Laravel Octane (Swoole)
1. Create any subscription
2. Fire subscription from Subscription::broadcast
3. Got error

Sometimes it works, sometimes it doesn't... I was unable to accurately reproduce the problem in 100% of cases.

**Output/Logs**

<details><summary>Click to expand</summary>

```
{
	"errors": [
		{
			"message": "Internal server error",
			"locations": [
				{
					"line": 2,
					"column": 3
				}
			],
			"path": [
				"patientAppointmentCreate"
			],
			"extensions": {
				"debugMessage": "Undefined array key \"scheduleUpdated\"",
				"file": "/var/www/app/vendor/nuwave/lighthouse/src/Subscriptions/SubscriptionRegistry.php",
				"line": 77,
				"trace": [
					{
						"file": "/var/www/app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php",
						"line": 255,
[...]

PR #2575, opened 2024-06-28
Document adding `SubscriptionRegistry` to `octane.warm`

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/2520

Thank you @danbka33 for firguring this out, does this look right to you?


danbka33 (approved):
Looks good.

PR #2579, opened 2024-07-03
feat: make `loadPersistedQuery` public

jaulz:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR will remember persisted queries locally. The idea is related to https://github.com/nuwave/lighthouse/discussions/2561. Right now, I use the `StartOperationOrOperations` event to determine whether the query is allowed or not. This works fine but previously defined queries will not work anymore as soon as I release another version. Hence, I would like to check first if the query was persisted before:
```php
$query = rescue(fn () => $graphql->loadPersistedQuery($operationId), null, false);

if ($query) {
  return true;
}
```

This PR makes `loadPersistedQuery` public and also remembers persisted queries locally to avoid hitting the cache twice.

Please let me know what you think and I can try to implement tests as well.

**Breaking changes**

n/a

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -49,13 +49,6 @@ class GraphQL
      */
     protected $errorsHandler;
 
-    /**
-     * Previously persisted queries.
-     *
-     * @var array<string, DocumentNode>
-     */
-    protected array $persistedQueries = [];
-
     public function __construct(
         protected SchemaBuilder $schemaBuilder,
         protected Pipeline $pipeline,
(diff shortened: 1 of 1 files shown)


src/GraphQL.php
```diff
@@ -286,7 +293,7 @@ public function toSerializableArray(ExecutionResult $result): array
     }
 
     /** Loads persisted query from the query cache. */
```
spawnia, line 295:
```suggestion
    /**
     * Loads persisted query from the query cache.
     *
     * @api
     */
```

src/GraphQL.php
```diff
@@ -306,10 +313,19 @@ protected function loadPersistedQuery(string $sha256hash): DocumentNode
...
+        if (in_array($sha256hash, $this->persistedQueries)) {
+            return $this->persistedQueries[$sha256hash];
+        }
+
         $cacheFactory = Container::getInstance()->make(CacheFactory::class);
         $store = $cacheFactory->store($cacheConfig['store']);
+        $persistedQuery = $store->get("lighthouse:query:{$sha256hash}");
+
+        if ($persistedQuery) {
+            $this->persistedQueries[$sha256hash] = $persistedQuery;
```
spawnia, line 325: I think this will leak memory when used with Laravel Octane. The cache should be pretty fast, and the method expresses that it is reaching out to something external with the verb `load` in its name, so users should be aware.

spawnia:
Can we just make the method public for now?

jaulz:
@spawnia sure, fine for me 😊  I anyway hope that there will be something built-in to allow pre-defined queries and mutations in the future 😉

spawnia:
Thank you @jaulz, released with https://github.com/nuwave/lighthouse/releases/tag/v6.40.0.

PR #2588, opened 2024-07-18
Trigger model events on nested mutation delete of belongsTo relationships

skaesdorf:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

When deleting a model using a nested belongs to relation, the model will now be queried and then deleted. This ensures model events to be triggered.

**Breaking changes**

None.


spawnia:
This would probably also apply to `@delete` when used as an `ArgResolver`. It also calls `$relation->delete()` or `$related::destroy()`.

spawnia:
Would it be simpler to let Laravel handle the fetching by using `destroy()` in all cases?

skaesdorf:
Sure, will do. Should I merge the cases or just use `destroy` in the `BelongsTo`/`HasOne` case?

skaesdorf:
I just stumbled across an edgecase pitfall of `destroy()` that maybe should be considered.

`destroy()` itself is a static method, therefore only global scopes of the model may be respected. When a relation has a scope attached to it like this ...
```php
    public function verifiedUsers () : BelongsTo
    {
        return $this->belongsTo( User::class )->onlyVerified();
    }
```

... it will not be respected since it is lost when calling `$related = $relation->make();` Therefore `destroy()` might not be a good idea at all?

skaesdorf:
@spawnia Any thoughts? I would like to solve this.

skaesdorf:
Hi, can we please merge this?

spawnia:
> Hi, can we please merge this?

I let Copilot review - I think the comments at least show some gaps in the test cases. I would like to have the behavior nailed down.

PR #2595, opened 2024-07-31
Update batch reference resolvers example code snippet

kfc:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

The suggested code snippet for batch reference resolvers uses the `firstWhere` method on Collection of results. This means that for every representation it searches the value with the ID in the collection (O(n) complexity). It's not a problem on small amounts of entities, however when a number of representations is quite large, it becomes slower and slower. In the project I was using it, it was 2x slower (40s vs 20s)

With the suggested change the collection is mapped by key (ID) and then uses simple array access by key to find the value (O(1) complexity), or if not found, `get` method will return `NULL` by default. 

**Breaking changes**

### DIFF
--- a/docs/6/federation/reference-resolvers.md
+++ b/docs/6/federation/reference-resolvers.md
@@ -101,11 +101,11 @@ final class Product implements BatchedEntityResolver
 {
     public function __invoke(array $representations): iterable
     {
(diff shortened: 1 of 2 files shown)


spawnia:
Thank you, good suggestion

PR #2601, opened 2024-08-19
Remove expired members from subscription topic stored in Redis set map

kirills-morozovs:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

When lighthouse subscriptions storage TTL is set every new subscription to channel will create a redis key with channel serializzed information and add the channel id to a map set related to specific topic and both keys will have TTL, but when channel is expired or closed without a webhook triggered the map set will keep filling with "dead" channels, `subscribersByTopic` method will already filter out the channel ids that expired, so this PR is going to collect those missing channel keys and then run SREM command to remove them from set map

### DIFF
--- a/src/Subscriptions/Storage/RedisStorageManager.php
+++ b/src/Subscriptions/Storage/RedisStorageManager.php
@@ -65,6 +65,9 @@ class RedisStorageManager implements StoresSubscriptions
             return new Collection();
         }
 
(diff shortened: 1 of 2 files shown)


spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.43.0.

Issue #2018, opened 2022-01-03
Query validation caching


k0ka:
**Are you willing to provide a PR for this issue or aid in developing it?**

Yes.

**What problem does this feature proposal attempt to solve?**

Query validation takes a lot of time and resources (about **30-40% of processing time**). However, a typical website uses a limited amount of queries that are compiled on the client. So we may cache the result of query validation, and identical queries will be processed much faster.

**Which possible solutions should be considered?**

```There are only two hard things in Computer Science: cache invalidation and naming things. --Phil Karlton```

In order to cache things, we have to figure out when to invalidate this cache. The validation is performed in the `webonyx/graphql-php` [GraphQL::promiseToExecute()](https://github.com/webonyx/graphql-php/blob/master/src/GraphQL.php#L148)

Here is what it depends on:
1. Schema 
2. Query
3. Validation rules

The first two are presented as simple AST. The schema is already cached, and I opened https://github.com/nuwave/lighthouse/pull/2017 to cache queries. So we can easily realize when to invalidate them. 

The latter is more difficult. Rules are php classes that can't be easily tracked. They can also depend on any other data or code. For example, the `QueryComplexity` validator depends on all passed query variables. 

I guess we can add a function `getHash()` to the `ProvidesValidationRules` interface. It should return the hash of all the data it depends on. So if anything is changed, we will be able to check it and invalidate the cache. For the default lighthouse `ValidationRulesProvider` it should return the hash of corresponding config values + query variables if the `QueryComplexity` validator is enabled. 

Given these 3 hashes, we can store the result of validation and be assured it has not been changed. 

I guess we should also extract the validation method in the `webonyx/graphql-php` `GraphQL::promiseToExecute()` to execute it independently. 

Do you have any objections or suggestions? This may improve Lighthouse performance significantly.

spawnia:
Unless you have somebody DDOS'ing your server, I would expect invalid queries not to be repeated often. Thus, we could just cache queries that were known to be valid and then just skip validation entirely.

k0ka:
I agree that in normal circumstances there should not be many queries with validation errors. However it's easy to implement (we'll store an array of validation errors and it'll be empty for validated queries).

Also I can imagine where this might be helpful. For example, old clients might spam old requests that are not validated against a new schema. Also a lot of DoS attacks are very dumb and reuse the same query.

LastDragon-ru:
Today I'm trying to optimize tests and found that query validation takes ~30% of the time :( 

[...]

PR #2603, opened 2024-08-22
Feature: cache validation rules

k0ka:
resolves [#2018](https://github.com/nuwave/lighthouse/issues/2018)
 
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Added possibility to cache query validation successful result. So the 2nd time the query would appear, most validator won't run against it.

Notes:
- added `hash` to `\Nuwave\Lighthouse\Schema\AST\DocumentAST`
- created interface `ProvidesCacheableValidationRules` which extends `ProvidesValidationRules` for backward compatibility. In the next release we might remove `ProvidesValidationRules`. 
- some execution paths doesn't provide query hash, so they won't use this feature
  - `Defer::executeDeferred`
  - `SubscriptionBroadcaster::broadcast`
  - `AsyncMutation::handle`
- validation cache is disabled by default for now. 
- it only caches successful results as `Errors` might fail to serialize.
- passing hashes with schema and query is quite ugly https://github.com/nuwave/lighthouse/blob/0d8ea1665f5ded77ad7ba6dde2aec8c703e6b368/src/GraphQL.php#L393-L397 but these objects are created by `webonyx/graphql-php`, so there is no easy way to extend them and store hashes inside objects.

I also added different benchmarks. `HugeRequestBench` creates requests with a lot of simple fragments: 1, 10 and 100 for `benchmark1`, `benchmark10` and `benchmark100` respectively. Enabling validation cache improves execution time:
- `benchmark1` from 4.222ms to 2.886ms 
- `benchmark10` from 14.984ms  to 4.429ms 
- `benchmark100` from 189.641ms to 19.783ms


k0ka:
Hello @spawnia,

do you have time to check this PR? 
It significantly improves speed and is disabled by default. So it won't break backward compatibility and we can gradually test it on real loads.

src/GraphQL.php
```diff
@@ -373,4 +384,51 @@ protected function parseQuery(string $query): DocumentNode
             'noLocation' => ! $this->configRepository->get('lighthouse.parse_source_location'),
         ]);
     }
+
+    /**
+     * Execute the validation rules that are cacheable.
+     *
+     * @return array<Error>
+     */
+    protected function executeAndCacheValidationRules(
```
spawnia, line 393: This sounds like it caches the validation rules themselves.

src/LighthouseServiceProvider.php
```diff
@@ -100,7 +101,8 @@ public function provideSubscriptionResolver(FieldValue $fieldValue): \Closure
             }
         });
 
-        $this->app->bind(ProvidesValidationRules::class, ValidationRulesProvider::class);
+        // $this->app->bind(ProvidesValidationRules::class, ValidationRulesProvider::class);
```
spawnia, line 104:
```suggestion
```

src/Schema/AST/DocumentAST.php
```diff
@@ -88,6 +91,9 @@ class DocumentAST implements Arrayable
     /** @var array<int,SchemaExtensionNode> */
     public array $schemaExtensions = [];
 
+    /** A hash of the schema. */
+    public ?string $hash = null;
```
spawnia, line 95: Why is this nullable?

src/Support/Contracts/ProvidesCacheableValidationRules.php
```diff
@@ -0,0 +1,27 @@
+<?php declare(strict_types=1);
+
+namespace Nuwave\Lighthouse\Support\Contracts;
+
+interface ProvidesCacheableValidationRules extends ProvidesValidationRules
+{
+    /**
+     * A set of rules for the first query validation step.
+     *
+     * These rules would be executed first and their result would be cached.
```
spawnia, line 10:
```suggestion
     * These rules are executed first and their result is cached.
```

src/Support/Contracts/ProvidesCacheableValidationRules.php
```diff
@@ -0,0 +1,27 @@
...
+     *
+     * @return array<string, \GraphQL\Validator\Rules\ValidationRule>
+     */
+    public function cacheableValidationRules(): array;
+
+    /**
+     * A set of rules for the second query validation step.
+     *
+     * These rules would always be executed and not cached.
```
spawnia, line 19:
```suggestion
     * These rules are always be executed and not cached.
```

tests/Integration/ValidationCachingTest.php
```diff
@@ -0,0 +1,250 @@
...
+use Illuminate\Cache\Events\CacheHit;
+use Illuminate\Cache\Events\CacheMissed;
+use Illuminate\Cache\Events\KeyWritten;
+use Illuminate\Contracts\Cache\Factory as CacheFactory;
+use Illuminate\Contracts\Config\Repository as ConfigRepository;
+use Illuminate\Support\Facades\Event;
+use Tests\TestCase;
+use Tests\Utils\Queries\Foo;
+
+class ValidationCachingTest extends TestCase
```
spawnia, line 12:
```suggestion
final class ValidationCachingTest extends TestCase
```

tests/Integration/ValidationCachingTest.php
```diff
@@ -0,0 +1,250 @@
...
+        Event::assertDispatchedTimes(KeyWritten::class, 2);
+    }
+
+    public function testSameSchemaAndSameQueryHaveSameKeys(): void
+    {
+        $config = $this->app->make(ConfigRepository::class);
+        $config->set('lighthouse.query_cache.enable', false);
+        $config->set('lighthouse.validation_cache.enable', true);
+
+        $event = Event::fake();
```
spawnia, line 157: This is how I would like the other tests to use this.

tests/Integration/ValidationCachingTest.php
```diff
@@ -0,0 +1,250 @@
...
+        $this->refreshApplication();
+        $this->app->instance(CacheFactory::class, $cacheFactory);
+
+        $this->setUpTestSchema();
+
+        $config = $this->app->make(ConfigRepository::class);
+        $config->set('lighthouse.query_cache.enable', false);
+        $config->set('lighthouse.validation_cache.enable', true);
+
+        Event::fake();
```
spawnia, line 184: why fake twice?

tests/Unit/Schema/AST/DocumentASTTest.php
```diff
@@ -111,5 +113,7 @@ public function testBeSerialized(): void
         $schemaExtension = $reserialized->schemaExtensions[0];
         $this->assertInstanceOf(SchemaExtensionNode::class, $schemaExtension);
         $this->assertInstanceOf(DirectiveNode::class, $schemaExtension->directives[0]);
+
+        $this->assertEquals($documentAST->hash, $reserialized->hash);
```
spawnia, line 117:
```suggestion
        $this->assertSame($documentAST->hash, $reserialized->hash);
```

src/Schema/AST/DocumentAST.php
k0ka: For backward compatibility. In case someone is creating DocumentAST on their own, without filling the field.

src/GraphQL.php
k0ka: renames to `validateCacheableRules`

tests/Integration/ValidationCachingTest.php
k0ka: changed `Event::` assertions to `$event->`

tests/Integration/ValidationCachingTest.php
k0ka: To override the default event handler after application refresh. Changed it to `$this->app->instance(EventsDispatcher::class, $event);` so the new application would get the same event dispatcher.

src/Schema/AST/DocumentAST.php
spawnia: Not necessary, see https://github.com/nuwave/lighthouse?tab=readme-ov-file#versioning:

> However, only code elements marked with @api will remain compatible - all other code in Lighthouse is considered internal and is thus subject to change.

src/Support/Contracts/ProvidesCacheableValidationRules.php
```diff
@@ -0,0 +1,27 @@
...
+     * These rules are executed first and their result is cached.
+     *
+     * @return array<string, \GraphQL\Validator\Rules\ValidationRule>
+     */
+    public function cacheableValidationRules(): array;
+
+    /**
+     * A set of rules for the second query validation step.
+     *
+     * These rules are always be executed and not cached.
```
spawnia, line 19:
```suggestion
     * These rules are always executed and not cached.
```

tests/Unit/Schema/AST/DocumentASTTest.php
```diff
@@ -28,6 +28,8 @@ public function testParsesSimpleSchema(): void
             ObjectTypeDefinitionNode::class,
             $documentAST->types[RootType::QUERY],
         );
+
+        $this->assertNotNull($documentAST->hash);
```
spawnia, line 32: hashing should be consistent, so we can assert the actual value.

src/Schema/AST/DocumentAST.php
k0ka: ok, made it non-nullable

src/GraphQL.php
```diff
@@ -373,4 +390,54 @@ protected function parseQuery(string $query): DocumentNode
...
+        }
+
+        $result = DocumentValidator::validate($schema, $query, $validationRules);
+        if ($result !== []) {
+            return $result;
+        }
+
+        $store->put($cacheKey, self::CACHEABLE_RULES_ERROR_FREE_RESULT, $cacheConfig['ttl']);
+
+        return [];
```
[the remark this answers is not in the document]
k0ka: In this case we will have to serialize errors. I'm not sure all errors can be properly serialized. 

`GraphQL\Error\Error` does implement the `JsonSerializable` interface, but it doesn't mean it would be properly serialized/unserialized during cache handling.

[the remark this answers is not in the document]
spawnia: Thanks, I added an explanatory comment. Maybe we can improve this in the future.

spawnia (approved):
Nice work @k0ka!

spawnia:
Released with https://github.com/nuwave/lighthouse/releases/tag/v6.45.0.

PR #2612, opened 2024-09-05
Use pusher driver for laravel reverb

### LINKED ISSUE #2564: GraphQL subscription with Laravel Reverb
When a subscription request is received by Reverb, it throws the following error:

```php
TypeError Laravel\Reverb\Protocols\Pusher\EventHandler::subscribe(): Argument #4 ($data) must be of type ?string, array given, called in /var/www/DemoGraphQL/vendor/laravel/reverb/src/Protocols/Pusher/EventHandler.php on line 29 at vendor/laravel/reverb/src/Protocols/Pusher/EventHandler.php:56
```

The error is caused by the payload data being an array instead of a string, which is expected by the subscribe method in the EventHandler class.

The payload that triggers the error looks like this:
```php
[
  "channel_data" => [
    "user_id" => "363c7a56eee7c386bf2c4ac980b661e6"
    "user_info" => []
  ]
  "channel" => "private-lighthouse-vEibwD5lS5FA0xmHHrMuH3Z5A8nVSvbC-1717741629"
]
```

The `authorized`Method in the `EchoBroadcaster` Class should send an string for 'channel_data' instead an array.

**Steps to reproduce**

1. Starting laravel reverb
2. Add an Subscription to the Frontend

**Output/Logs**

<details><summary>Laravel Exception</summary>

```
  TypeError 

  Laravel\Reverb\Protocols\Pusher\EventHandler::subscribe(): Argument #4 ($data) must be of type ?string, array given, called in /Users/dev/Herd/lv-ep-gw-api/vendor/laravel/reverb/src/Protocols/Pusher/EventHandler.php on line 29

  at vendor/laravel/reverb/src/Protocols/Pusher/EventHandler.php:56
     52▕ 
     53▕     /**
     54▕      * Subscribe to the given channel.
     55▕      */
  ➜  56▕     public function subscribe(Connection $connection, string $channel, ?string $auth = null, ?string $data = null): void
     57▕     {
     58▕         $channel = $this->channels
     59▕             ->for($connection->app())
     60▕             ->findOrCreate($channel);
```

</details></br>

**Lighthouse Version**

6.36.3

dstemberger:
Resolves #2564 
 
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Allow reverb broadcast to reuse the pusher driver

**Breaking changes**

### DIFF
--- a/src/Subscriptions/BroadcastManager.php
+++ b/src/Subscriptions/BroadcastManager.php
@@ -40,7 +40,7 @@ class BroadcastManager extends DriverManager
         $connection = $config['connection'] ?? 'pusher';
         $driverConfig = config("broadcasting.connections.{$connection}");
 
-        if (empty($driverConfig) || $driverConfig['driver'] !== 'pusher') {
+        if (empty($driverConfig)) {
             throw new \RuntimeException("Could not initialize Pusher broadcast driver for connection: {$connection}.");
         }
 
--- a/src/lighthouse.php
+++ b/src/lighthouse.php
@@ -419,6 +419,11 @@ return [
                 'connection' => env('LIGHTHOUSE_SUBSCRIPTION_REDIS_CONNECTION', 'default'),
                 'routes' => Nuwave\Lighthouse\Subscriptions\SubscriptionRouter::class . '@echoRoutes',
             ],
+            'reverb' => [
+                'driver' => 'pusher',
+                'routes' => Nuwave\Lighthouse\Subscriptions\SubscriptionRouter::class . '@pusher',
+                'connection' => 'reverb'
+            ],
         ],
 
         /*


thekonz (approved):
looks like it should work, but i have to tested it

spawnia:
> looks like it should work, but i have to tested it

So should I wait or merge now?

stayallive:
My 2 cents:

I don't really feel like this is really needed... since there is essentially just a single change which is the connection name vs. the current "pusher" driver.

This might be better fitted as a documentation blurb? Otherwise we might need to consider creating one for every Pusher compatible project like Reverb, Socketi, Socket Chief etc.

Also Reverb doesn't support web hooks so technically that route isn't needed and cleanup is not happening which is also nice if this is knows. Same issue Laravel WebSockets has.

Not against this change but also not really enthusiastic about it personally.

spawnia:
> Also Reverb doesn't support web hooks so technically that route isn't needed

So we could add another method to `SubscriptionRouter` that does the minimal amount of work for it?

I don't mind adding support for Reverb, as it is an official Laravel package.

stayallive:
> So we could add another method to SubscriptionRouter that does the minimal amount of work for it?

The route should be removed completely for Reverb, since it will never be called (because Reverb does not have web hooks)

> I don't mind adding support for Reverb, as it is an official Laravel package.

Agreed!

spawnia:
I made a bit of a larger overhaul in https://github.com/nuwave/lighthouse/pull/2639, please review that.

spawnia:
I released https://github.com/nuwave/lighthouse/pull/2639 with https://github.com/nuwave/lighthouse/releases/tag/v6.47.0.

Issue #2629, opened 2024-11-06
Enum returns incorrect case in Lighthouse GraphQL API


martio:
**Description:** When querying an enum in the Lighthouse GraphQL API, the enum values are returned with only the first letter capitalized, which does not align with the expected uppercase format. For instance, when querying the UserStatus enum, values are returned as "Active", "Inactive", etc., instead of "ACTIVE", "INACTIVE", etc.

**Expected Behavior:** The enum values should be returned in uppercase, matching the convention in GraphQL specifications and our application requirements.

**Steps to Reproduce:**

Define a native PHP enum with lowercase values, as follows:

```php
enum StatusEnum: string {
    case INACTIVE = 'inactive';
    case ACTIVE = 'active';
    case BANNED = 'banned';
}
```

Configure the enum in Lighthouse:

```graphql
enum UserStatus {
    INACTIVE @enum(value: "inactive")
    ACTIVE @enum(value: "active")
    BANNED @enum(value: "banned")
}
```

Execute a GraphQL query that retrieves this enum:

```graphql
query {
  user(id: 1) {
    id
    status
  }
}
```

Observe the format of the returned enum value.

Current Response:

```json
{
  "data": {
    "user": {
      "id": 1,
      "status": "Active"
    }
  }
}
```

Expected Response:

```json
{
  "data": {
    "user": {
      "id": 1,
      "status": "ACTIVE"
    }
  }
}
```

spawnia:
I added a test to show how `@enum` is supposed to work in https://github.com/nuwave/lighthouse/pull/2630. I can not explain the behaviour you describe, perhaps you are defining the enum in mulitple places, such as registering it programmatically? Or perhaps the `status` field uses a different type?

martio:
@spawnia The issue concerns enum serialization. If the keys in the native enum are formatted as "Example" instead of "EXAMPLE," the value returned by GraphQL will also be in the format "Example" instead of the expected "EXAMPLE."

```php
<?php

declare(strict_types=1);

namespace App\GraphQL\Types;

use BackedEnum;
use GraphQL\Error\SerializationError;
use GraphQL\Type\Definition\PhpEnumType;
use GraphQL\Utils\Utils;
use Override;
use TypeError;
use ValueError;

/**
 * GraphQL.
 */
class BackedEnumType extends PhpEnumType
{
    #[Override]
    public function serialize($value): string
    {
        if ($value instanceof $this->enumClass) {
[...]

PR #2630, opened 2024-11-07
Add explicit test for `@enum` directive

spawnia:
Resolves https://github.com/nuwave/lighthouse/issues/2629

- [x] Added or updated tests


spawnia:
The PHPStan failures are unrelated to this added test, I am merging regardless for now and plan to fix them later.

PR #2645, opened 2024-12-27
Add `@bind` directive as a GraphQL analogue for Laravel's Route Model Binding

remipelhate:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This PR aims to bring (in my opinion) one of Laravel's most convenient features to Lighthouse's resolver system: [Route Model Binding](https://laravel.com/docs/11.x/routing#route-model-binding). 

The code is based on an implementation I did a couple of months ago for a different project and as it has been really useful, I thought it maybe could be useful for others as well.

The general idea is to allow developers to define bindings for resolver arguments through the schema using a new `@bind` directive, which resolves a scalar identifier to an (Eloquent) model instance:
```graphql
type Mutation {
    addUserToCompany(
        user: ID! @bind(class: "App\\Models\\User")
        company: ID! @bind(class: "App\\Models\\Company")
    ): Boolean!
}
```

These bindings are resolved before the query or mutation resolver is invoked, meaning that we have the instances available in the resolver without having to query for it in the resolver itself:
```php
<?php

namespace App\Http\GraphQL\Mutations;

class AddUserToCompany
{
    /**
     * @param array{user: \App\Models\User, company: \App\Models\Company} $args
     */
    public function __invoke(mixed $root, array $args): bool
    {
        return $args['user']->associate($args['company'])->save();
    }
}
```

I added more examples and usages to the docs. Let me know if you think this would be a good addition to Lighthouse ;) 

**Breaking changes**

This PR does not introduce any breaking changes.


tests/Integration/Bind/BindDirectiveTest.php
```diff
@@ -0,0 +1,1363 @@
...
+            'data' => [
+                'addUserToCompany' => true,
+            ],
+        ]);
+        $resolver->assertArgs(function (array $args) use ($user, $company): void {
+            $this->assertArrayHasKey('user', $args);
+            $this->assertTrue($user->is($args['user']));
+            $this->assertArrayHasKey('company', $args);
+            $this->assertTrue($company->is($args['company']));
+        });
```
spawnia, line 1347:
```suggestion
        $this->mockResolver(fn ($_, array $args) => $args);
        $this->schema =  /* @lang GraphQL */ <<<'GRAPHQL'
            type User {
                id: ID!
            }

            type Company {
                id: ID!
            }

           type UserAndCompany {
             user: User!
             company: Company!
           }

            type Mutation {
                addUserToCompany(
                    user: ID! @bind(class: "Tests\\Utils\\Models\\User")
                    company: ID! @bind(class: "Tests\\Utils\\Models\\Company")
                ): UserAndCompany! @mock
            }
            GRAPHQL . self::PLACEHOLDER_QUERY;

        $response = $this->graphQL(/* @lang GraphQL */ <<<'GRAPHQL'
            mutation ($user: ID!, $company: ID!) {
                addUserToCompany(user: $user, company: $company) {
                  user {
                    id
                  }
                  company {
                    id
                  }
            }
            GRAPHQL,
            [
                'user' => $user->getKey(),
                'company' => $company->getKey(),
            ],
        );

        $response->assertGraphQLErrorFree();
        $response->assertJson([
            'data' => [
                'addUserToCompany' => [
                    'user' => [
                        'id' => $user->id,
                    ],
                    'company' => [
                        'id' => $company->id,
                    ],
                ],
            ],
        ]);
```

I feel like something like this is easier to understand. I get the concept of a spy, but I also think it introduces an API and a new way of doing tests that is not really necessary.

Alternatively, you can just do something like this:

```php
$resolverArgs = null;
$this->mockResolver(function ($_, array $args) use (&$resolverArgs): bool {
    $resolverArgs = $args;
    return true;
});
```

This approach has worked fine in the past, and it saves me from having to read and understand what `SpyResolver` and `assertArgs` do in order to understand if my test is correct.

spawnia (commented):
I like the concept!

docs/master/api-reference/directives.md
```diff
@@ -460,6 +460,153 @@ type RoleEdge {
...
+```php
+<?php
+
+namespace App\Http\GraphQL\Mutations;
+
+class RemoveUsers
+{
+    /**
+     * @param array{
+     *     users: \Illuminate\Support\Collection<\App\Models\User>,
```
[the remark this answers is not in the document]
remipelhate: You are absolutely correct.

docs/master/api-reference/directives.md
```diff
@@ -460,6 +460,153 @@ type RoleEdge {
...
+     *     company: \App\Models\Company
+     * } $args
+     */
+    public function __invoke(mixed $root, array $args): bool
+    {
+        foreach ($args['users'] as $user) {
+            if (! $user->associate($args['company'])->save()) {
+                return false;
+            }
```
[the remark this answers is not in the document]
remipelhate: It felt like a quick and simple example, but it isn't a good one indeed. I actually wouldn't do this myself to be honest, so I'll change the example.

src/Bind/BindDefinition.php
```diff
@@ -0,0 +1,95 @@
...
+
+    public function validate(
+        InputValueDefinitionNode $definitionNode,
+        FieldDefinitionNode|InputObjectTypeDefinitionNode $parentNode,
+    ): void {
+        $nodeName = $definitionNode->name->value;
+        $parentNodeName = $parentNode->name->value;
+        $valueType = $this->valueType($definitionNode->type);
+
+        if (! in_array($valueType, self::SUPPORTED_VALUE_TYPES, true)) {
```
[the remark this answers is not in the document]
remipelhate: Hmm, true... I mainly did so to prevent any other value types from being passed to `Nuwave\Lighthouse\Bind\ModelBinding`, which expects an int, string or a list of either of these. But perhaps we should move these value type assertions (not the GQL value type) to `Nuwave\Lighthouse\Bind\ModelBinding` itself. WDYT?

src/Bind/ModelBinding.php
```diff
@@ -0,0 +1,67 @@
...
+     * @return \Illuminate\Database\Eloquent\Collection<string, \Illuminate\Database\Eloquent\Model>
+     */
+    private function modelCollection(
+        EloquentCollection $results,
+        IlluminateCollection $values,
+        BindDefinition $definition,
+    ): EloquentCollection {
+        if ($results->count() > $values->unique()->count()) {
+            throw new MultipleRecordsFoundException($results->count());
+        }
```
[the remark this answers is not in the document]
remipelhate: Correct, there could be too few results as well, but this is covered on the `BindDirective` itself through the `BindingExists` rule. 

A binding can be defined as either required or optional:
| Binding | Required | Expected behaviour | Test (`Tests\Integration\Bind\BindDirectiveTest`) |
|---------|----------|---------------------|-----|
| Instance | `true` | Validation error when the binding does not exist (by applying the `BindingExists` rule) | `testMissingModelBindingOnFieldArgument` |
| Instance | `false` | Resolve the binding to `null` when it does not exist | `testMissingOptionalModelBindingOnFieldArgument` |
| Collection | `required` | Validation error for all bindings in the collection that do not exist (by applying the `BindingExists` rule) | `testMissingModelCollectionBindingOnFieldArgument` |
| Collection | `false` | Filter bindings that do not exist from the collection | `testMissingOptionalModelCollectionBindingOnFieldArgument` |

All of this could have been done within `Nuwave\Lighthouse\Bind\ModelBinding`, but as the implementation also allows to resolve bindings through callable classes, I wanted the behaviour to be consistent regardless of which binding resolver is used. This is why I put this validation logic in the directive level itself.

The reason why "too many" results is not covered the same way is because I do not consider this a validation error. If you have too many results for a binding, it means you're not using a unique identifier to resolve the binding, which to me is more of a configuration error. Therefore, I preferred to have this case throw a non-client-safe error.

Does this make sense? It's fully open for discussion :)

tests/Integration/Bind/BindDirectiveTest.php
remipelhate: Fair point. I'm using spies quite frequently, so it made sense to me to solve it like this. But indeed, I haven't come across a similar setup in any of the test cases in Lighthouse, so it's better to adhere to patterns that are already used throughout the project.

src/Bind/BindDefinition.php
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        if (! class_exists($this->class)) {
+            throw new DefinitionException(
+                "@bind argument `class` defined on `$parentNodeName.$nodeName` " .
+                "must be an existing class, received `$this->class`.",
+            );
+        }
+
+        if ($this->isModelBinding()) {
```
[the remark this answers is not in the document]
remipelhate: In order to resolve an instance, there must be a unique identifier which consists of a single value. It doesn't have to be a PK, any other column on the model can be used through the column argument of `@bind` (see `Tests\Integration\Bind\BindDirectiveTest::testModelBindingByColumnOnFieldArgument`). 

As for composite primary keys, they are [not natively supported](https://github.com/laravel/framework/discussions/48152) by Eloquent, so I don't think we should worry about it.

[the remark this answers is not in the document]
spawnia: Alright. If you can think of a validation we could do before runtime to ensure the configuration is sound, feel free to put it in.

src/Bind/ModelBinding.php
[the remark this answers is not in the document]
spawnia: > The reason why "too many" results is not covered the same way is because I do not consider this a validation error. If you have too many results for a binding, it means you're not using a unique identifier to resolve the binding, which to me is more of a configuration error. Therefore, I preferred to have this case throw a non-client-safe error.

Got it, that makes total sense! An explanatory comment for this in the code would be great.

docs/master/api-reference/directives.md
```diff
@@ -460,6 +460,160 @@ type RoleEdge {
...
+    /**
+     * @param array{
+     *     users: \Illuminate\Database\Eloquent\Collection<\App\Models\User>,
+     *     company: \App\Models\Company,
+     * } $args
+     * @return \Illuminate\Database\Eloquent\Collection<\App\Models\User>
+     */
+    public function __invoke(mixed $root, array $args): Collection
+    {
+        return Collection::make($args['users'])
```
spawnia, line 602:
```suggestion
        return $args['users']
```

docs/master/api-reference/directives.md
```diff
@@ -460,6 +460,160 @@ type RoleEdge {
...
+            ->map(function (User $user) use ($args): ?User {
+                $user->associate($args['company']);
+
+                if (! $user->save()) {
+                    return null;
+                }
+
+                return $user;
```
spawnia, line 610:
```suggestion
                return $user->save()
                    ? $user
                    : null;
```

src/Bind/BindDefinition.php
```diff
@@ -0,0 +1,75 @@
...
+use Illuminate\Database\Eloquent\Model;
+use Nuwave\Lighthouse\Exceptions\DefinitionException;
+use Nuwave\Lighthouse\Schema\AST\ASTHelper;
+
+/**
+ * @template-covariant TClass of object
+ * @property-read class-string<TClass> $class
+ * @property-read string $column
+ * @property-read array<string> $with
+ * @property-read bool $optional
```
spawnia, line 18: I don't see why those annotations would be necessary - the last one seems wrong even.

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+use Nuwave\Lighthouse\Support\Contracts\ArgManipulator;
+use Nuwave\Lighthouse\Support\Contracts\ArgTransformerDirective;
+use Nuwave\Lighthouse\Support\Contracts\ArgumentValidation;
+use Nuwave\Lighthouse\Support\Contracts\InputFieldManipulator;
+
+class BindDirective extends BaseDirective implements ArgumentValidation, ArgTransformerDirective, ArgDirectiveForArray, ArgManipulator, InputFieldManipulator
+{
+    /**
+     * @var \Nuwave\Lighthouse\Bind\BindDefinition<object>|null
+     */
```
spawnia, line 24:
```suggestion
    /** @var \Nuwave\Lighthouse\Bind\BindDefinition<object>|null */
```

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+    private ?BindDefinition $definition = null;
+    private mixed $binding;
```
spawnia, line 26:
```suggestion
    protected ?BindDefinition $definition = null;

    protected mixed $binding;
```

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+
+    public function __construct(
+        private Container $container,
```
spawnia, line 29:
```suggestion
        protected Container $container,
```

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+    as null or, when the argument is an array, any item in the argument value will be filtered out of the collection.
+    """
+    required: Boolean! = true
+) on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    /**
+     * @return \Nuwave\Lighthouse\Bind\BindDefinition<object>
+     */
```
spawnia, line 74:
```suggestion
    /** @return \Nuwave\Lighthouse\Bind\BindDefinition<object> */
```

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+    private function bindDefinition(): BindDefinition
+    {
+        return $this->definition ??= new BindDefinition(
+            $this->directiveArgValue('class'),
+            $this->directiveArgValue('column', 'id'),
+            $this->directiveArgValue('with', []),
+            $this->directiveArgValue('required', true),
```
spawnia, line 81:
```suggestion
            class: $this->directiveArgValue('class'),
            column: $this->directiveArgValue('column', 'id'),
            with: $this->directiveArgValue('with', []),
            required: $this->directiveArgValue('required', true),
```

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+    """
+    required: Boolean! = true
+) on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
+GRAPHQL;
+    }
+
+    /**
+     * @return \Nuwave\Lighthouse\Bind\BindDefinition<object>
+     */
+    private function bindDefinition(): BindDefinition
```
spawnia, line 75: Never use `private`, at least `protected`. Please check the rest of the code too.

src/Bind/BindDirective.php
```diff
@@ -0,0 +1,138 @@
...
+        if (! $this->binding instanceof PendingBinding) {
+            return $this->binding;
+        }
+
+        $definition = $this->bindDefinition();
+
+        $bind = match ($definition->isModelBinding()) {
+            true => $this->container->make(ModelBinding::class),
+            false => $this->container->make($definition->class),
+        };
```
spawnia, line 134:
```suggestion
        $bind = $definition->isModelBinding()
            ? $this->container->make(ModelBinding::class)
            : $this->container->make($definition->class);
```

src/Bind/ModelBinding.php
```diff
@@ -0,0 +1,65 @@
...
+
+        return $results->keyby($definition->column);
```
spawnia, line 63:
```suggestion
        return $results->keyBy($definition->column);
```

tests/Utils/Bind/SpyCallableClassBinding.php
```diff
@@ -0,0 +1,45 @@
...
+use PHPUnit\Framework\Assert;
+
+/** @template TReturn */
+final class SpyCallableClassBinding
+{
+    private mixed $value = null;
+
+    /**
+     * @var \Nuwave\Lighthouse\Bind\BindDefinition<object>|null
+     */
```
spawnia, line 15:
```suggestion
    /** @var \Nuwave\Lighthouse\Bind\BindDefinition<object>|null */
```

tests/Utils/Bind/SpyCallableClassBinding.php
```diff
@@ -0,0 +1,45 @@
...
+    private ?BindDefinition $definition = null;
+
+    /**
+     * @param TReturn $return
+     */
+    public function __construct(
+        private mixed $return = null,
```
spawnia, line 22:
```suggestion
    public function __construct(
        /** @var TReturn */
        private mixed $return = null,
```

tests/Utils/Bind/SpyCallableClassBinding.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->value = $value;
+        $this->definition = $definition;
+
+        return $this->return;
+    }
+
+    /**
+     * @param \Nuwave\Lighthouse\Bind\BindDefinition<object> $definition
+     */
```
spawnia, line 39:
```suggestion
    /** @param \Nuwave\Lighthouse\Bind\BindDefinition<object> $definition */
```

spawnia:
Released with https://github.com/nuwave/lighthouse/releases/tag/v6.48.0.

PR #2646, opened 2024-12-28
Force `int` cast on ModelsLoader count

robsontenorio:
When dealing with **legacy databases**  the count for this related model returns a string like `"88"`, not `88`.

```
comments: [Comment] @hasMany(type: PAGINATOR)
```

After upgrading to PHP 8.4 this does not work anymore. And it throws a error 

_"Nuwave\Lighthouse\Execution\ModelsLoader\CountModelsLoader::extractCount(): Return value must be of type int, string returned"_

**Changes**

Always cast to integer. Because, this count must be an integer, always.

**Breaking changes**

No.

### DIFF
--- a/src/Execution/ModelsLoader/CountModelsLoader.php
+++ b/src/Execution/ModelsLoader/CountModelsLoader.php
@@ -32,9 +32,12 @@ class CountModelsLoader implements ModelsLoader
          */
         $countAttributeName = Str::snake("{$relationName}_count");
 
-        $count = (int) $model->getAttribute($countAttributeName);
-        assert(is_int($count), 'avoid runtime check in production since the return type validates this anyway');
+        $count = $model->getAttribute($countAttributeName);
+        if (! is_numeric($count)) {
+            $nonNumericCount = Utils::printSafe($count);
+            throw new \Exception('Expected numeric count, got: {$nonNumericCount}.");
+        }
 
-        return $count;
+        return (int) $count;
     }
 }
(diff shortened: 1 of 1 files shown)


src/Execution/ModelsLoader/CountModelsLoader.php
```diff
@@ -32,7 +32,7 @@ public static function extractCount(Model $model, string $relationName): int
          */
         $countAttributeName = Str::snake("{$relationName}_count");
 
-        $count = $model->getAttribute($countAttributeName);
+        $count = (int) $model->getAttribute($countAttributeName);
         assert(is_int($count), 'avoid runtime check in production since the return type validates this anyway');
 
         return $count;
```
spawnia, line 38:
```suggestion
        $count = $model->getAttribute($countAttributeName);
        if (! is_numeric($count)) {
            $nonNumericCount = Utils::printSafe($count);
            throw new \Exception('Expected numeric count, got: {$nonNumericCount}.");
        }

        return (int) $count;
```

I am always wary of unchecked casts, since PHP will do the most senseless cast and just give a wrong result instead of throwing.

spawnia (changes requested):

robsontenorio:
@spawnia your  suggested approach didn’t work. 

I also fixed the quote issue, but no luck.

spawnia:
How is it not working? It is still casting to int.

robsontenorio:
@spawnia the pipeline is failing :(

robsontenorio:
@spawnia i forced an dummy commit but I don’t get what is wrong with GitHub actions. Now phpstan complains about other things.

spawnia:
Released with https://github.com/nuwave/lighthouse/releases/tag/v6.47.1, thank you @robsontenorio

PR #2647, opened 2024-12-29
Make reporting of client-safe errors configurable

remipelhate:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

While it makes perfect sense to not pass client safe errors to Laravel's exception handler by default, in some cases, it can be useful to be able to do so. 

Laravel allows to define a [minimum log level](https://laravel.com/docs/11.x/errors#exception-log-levels) at which each individual log channel is invoked. For example, `sentry` can be set to only report errors with a minimum level of `warning`, whereas `slack` only gets triggered as of `critical`:
```php
// config/logging.php

return [
    'channels' => [
        'slack' => [
            'driver' => 'slack',
            'url' => env('LOG_SLACK_WEBHOOK_URL'),
            'username' => 'Laravel Log',
            'emoji' => ':boom:',
            'level' => env('LOG_LEVEL', Psr\Log\LogLevel::CRITICAL),
        ],
        'sentry' => [
            'driver' => 'sentry',
            'level' => env('LOG_LEVEL', Psr\Log\LogLevel::WARNING),
            'bubble' => true,
        ],
    ],
];
```

I've used this setup for a project where the team wanted to be able to track all authentication and authorization exceptions in Prometheus, as those can give insights on potential ACL issues or misuse. The problem is that Lighthouse marks these errors as client safe, causing them not to be passed to Laravel's exception handler.

This PR introduces a new `Nuwave\Lighthouse\Execution\AlwaysReportingErrorHandler` which can be used instead of `Nuwave\Lighthouse\Execution\ReportingErrorHandler` in the `lighthouse.php` config:
```diff
'error_handlers' => [
-    Nuwave\Lighthouse\Execution\ReportingErrorHandler::class,
+    Nuwave\Lighthouse\Execution\AlwaysReportingErrorHandler::class,
],
```

When using `Nuwave\Lighthouse\Execution\AlwaysReportingErrorHandler`, client-safe exceptions will be passed to the 
default Laravel exception handler, allowing you to configure appropriate error reporting outside of Lighthouse.

**Breaking changes**

No

### DIFF
--- a/src/lighthouse.php
+++ b/src/lighthouse.php
@@ -289,20 +289,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Report Client-safe Errors
-    |--------------------------------------------------------------------------
-    |
-    | By default, the error handler Nuwave\Lighthouse\Execution\ReportingErrorHandler only reports non-client-safe
-    | errors through the default Laravel exception handler. However, as Laravel allows to define a minimum log
-    | level at which each individual log channel is invoked, it can sometimes be useful to push client-safe
-    | errors to the Laravel exception handler so it can determine which errors to report or not.
-    |
-    */
-
-    'report_client_safe_errors' => env('LIGHTHOUSE_REPORT_CLIENT_SAFE_ERRORS', false),
-
-    /*
-    |--------------------------------------------------------------------------
     | Field Middleware
     |--------------------------------------------------------------------------
     |
--- a/src/lighthouse.php
+++ b/src/lighthouse.php
@@ -289,6 +289,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Report Client-safe Errors
+    |--------------------------------------------------------------------------
+    |
+    | By default, the error handler Nuwave\Lighthouse\Execution\ReportingErrorHandler only reports non-client-safe
+    | errors through the default Laravel exception handler. However, as Laravel allows to define a minimum log
+    | level at which each individual log channel is invoked, it can sometimes be useful to push client-safe
+    | errors to the Laravel exception handler so it can determine which errors to report or not.
+    |
+    */
+
+    'report_client_safe_errors' => env('LIGHTHOUSE_REPORT_CLIENT_SAFE_ERRORS', false),
+
+    /*
+    |--------------------------------------------------------------------------
     | Field Middleware
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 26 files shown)


tests/Integration/Execution/ReportingErrorHandlerTest.php
```diff
@@ -0,0 +1,126 @@
...
+        $error = new Error(previous: $previousError);
+        $next = fn (Error $error): array => \compact('error');
+
+        $result = (new ReportingErrorHandler($this->handler, $config))($error, $next);
+
+        $this->assertSame(\compact('error'), $result);
+        $this->handler->assertReported($previousError);
+    }
+
+    public static function clientSafeErrors(): \Generator
```
spawnia, line 75:
```suggestion
    /** @return iterable<array{\Exception|null}> */
    public static function clientSafeErrors(): iterable
```

src/lighthouse.php
```diff
@@ -287,6 +287,20 @@
         Nuwave\Lighthouse\Execution\ReportingErrorHandler::class,
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Report Client-safe Errors
```
spawnia, line 292:
```suggestion
    | Report Client-Safe Errors
```

remipelhate:
> I am not fully convinced that a configuration option at that level is the best option. The problem I see with this approach is that its value is only used from within `ReportingErrorHandler`, which is configurable in itself. Thus, for users that omit or replace `ReportingErrorHandler`, the option `report_client_safe_errors` has no effect.

That's a valid concern. Would you perhaps prefer to solve this on the `ReportingErrorHandler` itself? I noticed first-party Laravel packages are using static config callbacks more often lately, so perhaps that could be an option? Something along the lines of:
```php
// app/Providers/GraphQLServiceProvider.php

public function boot(): void
{
    ReportingErrorHandler::reportClientSafeErrors();
}
```
I'm not always keen on static config, but it does fit Laravel's way of working.

remipelhate:
Or perhaps we could solve this through a named constructor on the `ReportingErrorHandler` which could be referenced in the config?
```php
// config/lighthouse.php
return [
    'error_handlers' => [
        [Nuwave\Lighthouse\Execution\ReportingErrorHandler::class, 'includeClientSafeErrors'],
        // or...
        Nuwave\Lighthouse\Execution\ReportingErrorHandler::class . '@includeClientSafeErrors',
    ],
];
```

spawnia:
Another option would be to add a new class to replace `ReportingErrorHandler`, perhaps called `AlwaysReportingErrorHandler`?

remipelhate:
> Another option would be to add a new class to replace `ReportingErrorHandler`, perhaps called `AlwaysReportingErrorHandler`?

Sure, that might work too. On it.

spawnia:
Thank you @remipelhate, released with https://github.com/nuwave/lighthouse/releases/tag/v6.49.0.

Issue #2653, opened 2025-01-15
`n+1` problem with eager loading models in different databases (connections)


wojo1206:
**What problem does this feature proposal attempt to solve?**

I found `n+1` problem with eager loading models in different databases (connections). Currently, the code will not eager load entitles that live on two different connections. I debug and tracked the problem to this specific line at `src/Schema/Directives/RelationDirective.php`:

```
if (
    $this->lighthouseConfig['batchload_relations']
    // Batch loading joins across both models, thus only works if they are on the same connection
    && $this->isSameConnection($relation) //  <-- Right here.
) {
```

**Which possible solutions should be considered?**

I don't know what is the rationale for this limitation but that should be OK by default to eager load models on different connections. To solve this problem we can introduce new config variable that allows to override this check at `$this->isSameConnection($relation)`, something like:

```
    /*
    |--------------------------------------------------------------------------
    | Batchload Relations
    |--------------------------------------------------------------------------
    |
    | If set to true, relations marked with directives like @hasMany or @belongsTo
    | will be optimized by combining the queries through the BatchLoader.
    |
    */

    'batchload_relations' => true,
    'batchload_relations_only_on_same_connections' => true,
```

spawnia:
The reason for this is right there in the comment:

> Batch loading joins across both models

By `joins` i mean an actual SQL `JOIN` statement. A single database statement involves the tables of both models.

wojo1206:
Hmm, joins across multiple databases are fine as long as the databases are on the same server instance. The concept of multi-database system in Laravel means that you have set multiple connections even if your databases are on the same instance.

PR #2654, opened 2025-01-15
feat: provide override for eager loading models on different connections

wojo1206:
[](https://github.com/nuwave/lighthouse/issues/2653)
 
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Introduced config var that can override `isSameConnection()` for batchload relations. This fixes `n+1` problem for models that exists on different connections.
 
**Breaking changes**

None


spawnia:
Have you actually tried this? The check to avoid batch loading if separate connections are involved was added because the resulting database queries for batch loading uses tables of both models, which would not work when they are on different databases. In what instance would this not hold true?

src/Schema/Directives/RelationDirective.php
```diff
@@ -72,10 +72,10 @@ public function resolveField(FieldValue $fieldValue): callable
             ) {
                 $relationBatchLoader = BatchLoaderRegistry::instance(
                     $this->qualifyPath($args, $resolveInfo),
-                    static fn (): RelationBatchLoader => new RelationBatchLoader(
+                    static fn(): RelationBatchLoader => new RelationBatchLoader(
                         $paginationArgs === null
-                            ? new SimpleModelsLoader($relationName, $decorateBuilder)
-                            : new PaginatedModelsLoader($relationName, $decorateBuilder, $paginationArgs),
+                        ? new SimpleModelsLoader($relationName, $decorateBuilder)
+                        : new PaginatedModelsLoader($relationName, $decorateBuilder, $paginationArgs),
```
spawnia, line 78: Please revert reformatting.

wojo1206:
Yes, I use this fork on my production already. Works like a charm. I think you are suggesting that the check to avoid batch loading on separate connections creates possible conflicts due to resulting queries not being database prefixed? I don't know what is the default Eloquant behavior nowadays. It is good practice to include databases prefixes in queries. I believe in the past the Eloquant didn't do that. I have multi database setup and I override `getTable()` on all models to ensure that my queries are database prefixed. 

I think, this config var is valuable addition to the project because it isn't braking change and provides option to the users. In future, this functionality should better align to what the recent Eloquant behavior is.

spawnia:
> I think that original logic in lighthouse is flawed and doesn't match Laravel in this matter.

How do expect me to respond to that? Lighthouse does different things than Laravel. What exactly is flawed? What needs to change?

> but I think that by default Eloquent relationships don't come with this assumption.

What assumption?

> they combine joins on different connections into single query just fine

Eager loading in Laravel uses a `WHERE IN` clause, thus it does not require joins.

I am actually not quite sure in what scenario Lighthouse would construct a query that spans multiple tables during batch loading. I could not find an example in https://github.com/nuwave/lighthouse/pull/1979 or https://github.com/nuwave/lighthouse/pull/1945. I am sure there must have been a reason, I just don't remember or know what exactly.

wojo1206:
@spawnia Please ignore my comment. I just recalled that by default Laravel doesn’t automatically qualify tables across connections. My models are setup to produce fully qualified table names. With fully qualified table names you can do multi-database queries given same host and permission are ok. I will just keep using my fork until farther development.

PR #2665, opened 2025-03-08
Allow Laravel 12

stayallive:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Allow Laravel 12 and adjust tests and dependencies where needed.

**Breaking changes**

None. Laravel 12 is also a super simple release for us from what it seems (in the BC department at least).


stayallive:
This is shaking out some nasty phpstan stuff, a small part is probably a bug on their side (for php 8.4; phpstan/phpstan-src#3855) but others seems to be related to larastan maybe not correctly detecting the testing mixing. Not sure how to fix this. Tests are passing so it's all working but it looks like phpstan/larastan might not understand some stuff all of a sudden. Quick search on both repo's did not yield any red flags. It might be on their side since only the "highest" dependency phpstan runs are failing (and 8.4 runs because of linked bug).

Appreciate a quick look @spawnia since you might know what is going on here or maybe we should just have a little patience and this will resolve itself in a few days 🫣

src/Console/FieldCommand.php
```diff
@@ -36,7 +36,7 @@ protected function getDefaultNamespace($rootNamespace): string
     /** @return array{string, string} */
     protected function nameParts(): array
     {
-        $name = $this->argument('name');
+        $name = (string) $this->argument('name');
```
spawnia, line 39: Looks like Larastan/PHPStan is going back and forth on what type this call is supposed to return. You can check the change history of this file, i think we used to check for `string`.

src/Schema/Types/Scalars/DateTimeUtc.php
```diff
@@ -17,6 +17,6 @@ protected function format(Carbon $carbon): string
     protected function parse(mixed $value): Carbon
```
spawnia, line 17: Side-note about the comment in this class - it works with Carbon 2 and above right?

src/Schema/Types/Scalars/DateTimeUtc.php
```diff
@@ -17,6 +17,6 @@ protected function format(Carbon $carbon): string
...
     {
         // @phpstan-ignore-next-line We know the format to be good, so this can never return `false`
-        return Carbon::createFromIsoFormat('YYYY-MM-DDTHH:mm:ss.SSSSSSZ', $value);
+        return Carbon::createFromIsoFormat('YYYY-MM-DDTHH:mm:ss.SSSSSSZ', (string) $value);
```
spawnia, line 20: I think we should be able to change the interface for `parse` to accept `string` and validate the values beforehand in `DateScalar::tryParsingDate`.

spawnia (commented):
Thank you @stayallive for tackling this. Regarding the points you brought up:

- Requiring a fork of `dms/phpunit-arraysubset-asserts` is fine for now
- I have also encountered the detection of mixins to be broken, as seen in https://github.com/mll-lab/laravel-utils/pull/29. Perhaps it will resolve itself once when PHPStan releases https://github.com/phpstan/phpstan-src/pull/3855?

src/Schema/Types/Scalars/DateTimeUtc.php
stayallive: Yes. It does, updated!

src/Console/FieldCommand.php
stayallive: Locally this issue seems to be gone, might be caused by the referenced PRs messing something up.

stayallive:
Locally I'm left with 2 errors on PHP 8.4, test suite does not seem to agree with me yet though 🙈

spawnia:
Released with https://github.com/nuwave/lighthouse/releases/tag/v6.54.0.

stayallive:
Thanks @spawnia for taking this over the finish line 🧡

PR #2666, opened 2025-03-14
Fix shortcut foreign key selection when including __typename

Ommetaphobia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

When selecting the foreign key of a belongs to relation and including  `__typename` in the related models field selection no queries will be fired.

**Breaking changes**

None.


spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.51.0.

Issue #2668, opened 2025-03-18
Array to string conversion exception in foreign key selection shortcut


LVMade01:
The Version 6.51.1 throws the error
```
[2025-03-18 14:12:57] production.ERROR: Array to string conversion {"userId":18,"exception":"[object] (ErrorException(code: 0): Array to string conversion at /var/www/app/vendor/nuwave/lighthouse/src/Schema/Directives/RelationDirective.php:57)
```
when "$resolveInfo->getFieldSelection()" is multidimensional (see https://bugs.php.net/bug.php?id=60198) and "shortcut_foreign_key_selection" is enabled. I'm using PHP 8.4 and Laravel 11.

A possible solution: at this code-line we only care about the keys. So a 
`(array_diff(array_keys($resolveInfo->getFieldSelection()), ['id', '__typename']) === [])` should work.

**Lighthouse Version**

 6.51.1

spawnia:
I tried to add a failing test case with https://github.com/nuwave/lighthouse/pull/2669/files#diff-03fc86ec33bc8899cb9e8fb5414b1f58ad727a87566d46448a086cde533e2924, but apparently it does not trigger the error. Can you provide a test case that reproduces the bug?

LVMade01:
In the config if you set "shortcut_foreign_key_selection" to true then you will get 5 errors in the unit-tests. Without this flag the new code from 6.51.0 is not triggered.

spawnia:
Fix released with https://github.com/nuwave/lighthouse/releases/tag/v6.51.2.

PR #2670, opened 2025-03-19
Fix morph to shortcuts foreign key selection when relationship is morph to

Ommetaphobia:
Resolves [#2668](https://github.com/nuwave/lighthouse/issues/2668)
 
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

When querying a morphto relationship using a union or interface with inline fragments, shortcut foreign key logic will throw an exception when comparing a multidimensional array (https://bugs.php.net/bug.php?id=60198).

Instead of using `array_diff_assoc` when checking if the selection set matches the criteria for shortcut foreign key selection this change will now check if the keys of the selection set match the specific criteria.

In addition, if the relationship is a `morphTo` the model will be returned with it's `id` property filled so that `TypeRegistry::typeResolverFallback` can resolve the correct type when using unions or interfaces.

**Breaking changes**

None.


tests/Integration/Schema/Directives/MorphToDirectiveTest.php
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3: Please revert this and other reformattings.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Use array keys when comparing shortcut foreign key field selection and return model instance when relation is morphTo https://github.com/nuwave/lighthouse/pull/2670
```
spawnia, line 14: Please reformulate this in terms of the issues that are solved. You may split into multiple entries if applicable.

src/Schema/Directives/RelationDirective.php
```diff
@@ -54,15 +55,16 @@ public function resolveField(FieldValue $fieldValue): callable
...
+                && (array_diff(array_keys($resolveInfo->getFieldSelection()), ['id', '__typename']) === [])
                 && $relation instanceof BelongsTo
                 && $args === []
             ) {
                 $id = $parent->getAttribute($relation->getForeignKeyName());
 
-                return $id === null
-                    ? null
-                    : ['id' => $id];
+                if ($id === null) return null;
```
spawnia, line 64: Let's use braces.

spawnia, line 64: Let's use braces.

src/Schema/Directives/RelationDirective.php
```diff
@@ -54,15 +55,16 @@ public function resolveField(FieldValue $fieldValue): callable
             // that we know to be present on the parent model.
             if (
                 $this->lighthouseConfig['shortcut_foreign_key_selection']
-                && (array_diff_assoc($resolveInfo->getFieldSelection(), ['id' => true, '__typename' => true]) === [])
+                && (array_diff(array_keys($resolveInfo->getFieldSelection()), ['id', '__typename']) === [])
```
spawnia, line 58: Can you show me which test case is failing when reverting this line, and how?

spawnia: Got it from https://github.com/nuwave/lighthouse/actions/runs/13946583856/job/39035112800, thanks.

spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.51.2.

PR #2671, opened 2025-03-21
Fix using incorrect root for decorator when paginating

daannet:
In the PaginatedModelsLoader, the decorateBuilder closure has an optional second parameter to pass a specific model:
https://github.com/nuwave/lighthouse/blob/master/src/Execution/ModelsLoader/PaginatedModelsLoader.php#L58C17-L58C18 

Currently, this option does not work as expected. The builder always uses the first model that created the batch loader instance, instead of the provided model.

- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
Updated the decorateBuilder closure to correctly use the specific model when provided.

**Breaking changes**
None

### DIFF
--- a/src/Schema/Directives/RelationDirectiveHelpers.php
+++ b/src/Schema/Directives/RelationDirectiveHelpers.php
@@ -42,6 +42,11 @@ trait RelationDirectiveHelpers
             $resolveInfo->enhanceBuilder(
                 $builder,
                 $this->scopes(),
+                /**
+                 * Sometimes overridden to use a different model than the usual root.
+                 * @see \Nuwave\Lighthouse\Execution\ModelsLoader\PaginatedModelsLoader::loadRelatedModels
+                 * @see \Tests\Integration\Schema\Directives\BuilderDirectiveTest::testCallsCustomBuilderMethodOnFieldWithSpecificModel
+                 */
                 $specificRoot ?? $root,
                 $args,
                 $context,
--- a/src/Schema/Directives/RelationDirectiveHelpers.php
+++ b/src/Schema/Directives/RelationDirectiveHelpers.php
@@ -33,8 +33,6 @@ trait RelationDirectiveHelpers
     protected function makeBuilderDecorator(mixed $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): \Closure
     {
         return function (object $builder, mixed $specificRoot = null) use ($root, $args, $context, $resolveInfo): void {
-            $effectiveRoot = $specificRoot ?? $root;
-
             if ($builder instanceof Relation) {
                 $builder = $builder->getQuery();
             }
@@ -44,7 +42,7 @@ trait RelationDirectiveHelpers
             $resolveInfo->enhanceBuilder(
                 $builder,
                 $this->scopes(),
-                $effectiveRoot,
+                $specificRoot ?? $root,
                 $args,
                 $context,
                 $resolveInfo,
--- a/src/Schema/Directives/RelationDirectiveHelpers.php
+++ b/src/Schema/Directives/RelationDirectiveHelpers.php
@@ -28,11 +28,13 @@ trait RelationDirectiveHelpers
     /**
      * @param  array<string, mixed>  $args
      *
-     * @return \Closure(QueryBuilder|\Illuminate\Database\Eloquent\Builder<\Illuminate\Database\Eloquent\Model>|\Illuminate\Database\Eloquent\Relations\Relation<\Illuminate\Database\Eloquent\Model>): void
+     * @return \Closure(QueryBuilder|\Illuminate\Database\Eloquent\Builder<\Illuminate\Database\Eloquent\Model>|\Illuminate\Database\Eloquent\Relations\Relation<\Illuminate\Database\Eloquent\Model>, ?mixed = null): void
      */
     protected function makeBuilderDecorator(mixed $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): \Closure
     {
-        return function (object $builder) use ($root, $args, $context, $resolveInfo): void {
+        return function (object $builder, mixed $specificRoot = null) use ($root, $args, $context, $resolveInfo): void {
+            $effectiveRoot = $specificRoot ?? $root;
+
             if ($builder instanceof Relation) {
                 $builder = $builder->getQuery();
             }
@@ -42,7 +44,7 @@ trait RelationDirectiveHelpers
             $resolveInfo->enhanceBuilder(
                 $builder,
                 $this->scopes(),
-                $root,
+                $effectiveRoot,
                 $args,
                 $context,
                 $resolveInfo,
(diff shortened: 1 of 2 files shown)


src/Schema/Directives/RelationDirectiveHelpers.php
```diff
@@ -28,11 +28,13 @@ protected function relation(): string
...
      *
-     * @return \Closure(QueryBuilder|\Illuminate\Database\Eloquent\Builder<\Illuminate\Database\Eloquent\Model>|\Illuminate\Database\Eloquent\Relations\Relation<\Illuminate\Database\Eloquent\Model>): void
+     * @return \Closure(QueryBuilder|\Illuminate\Database\Eloquent\Builder<\Illuminate\Database\Eloquent\Model>|\Illuminate\Database\Eloquent\Relations\Relation<\Illuminate\Database\Eloquent\Model>, ?mixed = null): void
      */
     protected function makeBuilderDecorator(mixed $root, array $args, GraphQLContext $context, ResolveInfo $resolveInfo): \Closure
     {
-        return function (object $builder) use ($root, $args, $context, $resolveInfo): void {
+        return function (object $builder, mixed $specificRoot = null) use ($root, $args, $context, $resolveInfo): void {
+            $effectiveRoot = $specificRoot ?? $root;
+
```
spawnia, line 37:
```suggestion
```

src/Schema/Directives/RelationDirectiveHelpers.php
```diff
@@ -42,7 +44,7 @@ protected function makeBuilderDecorator(mixed $root, array $args, GraphQLContext
             $resolveInfo->enhanceBuilder(
                 $builder,
                 $this->scopes(),
-                $root,
+                $effectiveRoot,
```
spawnia, line 47:
```suggestion
                /**
                 * Sometimes overridden to use a different model than the usual root.
                 * @see \Nuwave\Lighthouse\Execution\ModelsLoader\PaginatedModelsLoader::loadRelatedModels
                 * @see \Tests\Integration\Schema\Directives\BuilderDirectiveTest::testCallsCustomBuilderMethodOnFieldWithSpecificModel
                 */
                $specificRoot ?? $root,
```

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -155,6 +156,47 @@ public function testCallsCustomBuilderMethodOnFieldWithoutValue(): void
...
+        }
+        GRAPHQL;
+
+        $users = factory(User::class, 2)->create();
+
+        foreach ($users as $user) {
+            $tasks = factory(Task::class, 3)->make();
+            $tasks[0]->name = $user->name;
+            $user->tasks()->saveMany($tasks);
+        }
```
spawnia, line 182:
```suggestion
        $users = factory(User::class, 2)->create();

        foreach ($users as $user) {
            assert($user instanceof User);

            $taskWithSameName = factory(Task::class)->make();
            assert($taskWithSameName instanceof Task);
            $taskWithSameName->name = $user->name;
            $taskWithSameName->user()->associate($user);
            $taskWithSameName->save();

            $taskWithOtherName = factory(Task::class)->make();
            assert($taskWithOtherName instanceof Task);
            $taskWithOtherName->name = "Different from {$user->name}";
            $taskWithOtherName->user()->associate($user);
            $taskWithOtherName->save();
        }
```

Could be flaky otherwise.

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -164,4 +206,17 @@ public static function limit(QueryBuilder|EloquentBuilder $builder, ?int $value)
...
         return $builder->limit($value ?? 2);
     }
+
+    /**
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder<\Tests\Utils\Models\User>  $builder
+     *
+     * @return \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder<\Tests\Utils\Models\User>
+     */
+    public static function specificModel(
+        QueryBuilder|EloquentBuilder $builder,
```
spawnia, line 216: Don't we know exactly what we are getting here?

spawnia (commented):
Apparently the second parameter was already given in `\Nuwave\Lighthouse\Execution\ModelsLoader\PaginatedModelsLoader::loadRelatedModels`. Can you add think of a test case for that that also requires your fix?

daannet:
> Apparently the second parameter was already given in `\Nuwave\Lighthouse\Execution\ModelsLoader\PaginatedModelsLoader::loadRelatedModels`. Can you add think of a test case for that that also requires your fix?

@spawnia  The added test case catches this problem. Without the fix, this case will fail. 

Currently:
For example, if you use @builder, you get the root/model of the first item in the collection. This was the first item the decorateBuilder initialised before the batch loader was created. 

The calls to the decorateBuilder in the PaginatedModelsLoader uses the first closure created by the RelationDirective and given to the first batch loader instance. This way, the decorateBuilder always used the first root.

Now:
The root was already passed to the closure, but the parameter was not defined/used. Now it is and passes the correct model/root to builders. 

Now that the correct root has been passed in, in the builder the query can be adjusted on an item-by-item basis based on the correct root, rather than the first one.

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
daannet: In this simple case we apply a filter based on the current user name. This wouldn't work when the decorator uses the same (first) model.
(The purpose is that the query is customized per parent record, using the correct model/root)

spawnia: I am specifically talking about the type. It should be clear what kind of Builder we receive (and return) here.

tests/Integration/Schema/Directives/BuilderDirectiveTest.php
```diff
@@ -155,6 +156,57 @@ public function testCallsCustomBuilderMethodOnFieldWithoutValue(): void
...
+
+            $taskWithSameName = factory(Task::class)->make();
+            assert($taskWithSameName instanceof Task);
+            $taskWithSameName->name = $user->name;
+            $taskWithSameName->user()->associate($user);
+            $taskWithSameName->save();
+
+            $taskWithOtherName = factory(Task::class)->make();
+            assert($taskWithOtherName instanceof Task);
+            $taskWithOtherName->name = "Different from {$user->name}";
```
spawnia, line 189:
```suggestion
            assert($user instanceof User);

            $userName = $user->name;
            assert(is_string($userName), 'set by UserFactory');

            $taskWithSameName = factory(Task::class)->make();
            assert($taskWithSameName instanceof Task);
            $taskWithSameName->name = $userName;
            $taskWithSameName->user()->associate($user);
            $taskWithSameName->save();

            $taskWithOtherName = factory(Task::class)->make();
            assert($taskWithOtherName instanceof Task);
            $taskWithOtherName->name = "Different from {$userName}";
```

spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.52.1.

PR #2684, opened 2025-05-19
ci: update deprecated action

owenvoke:
**Changes**

This updates from `bufbuild/buf-setup-action` to `bufbuild/buf-action` as the `buf-setup-action` has been deprecated.

See: https://github.com/bufbuild/buf-action/blob/main/MIGRATION.md#buf-setup-action

**Breaking changes**


spawnia:
Thank you @owenvoke

PR #2686, opened 2025-05-29
Don't conceal errors in resolvers for canFind, canModel and canQuery directives

MarcEspiard:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Context**

While switching from `@can` to `@canFind` recently, we noticed that exceptions thrown from within a mutation class were caught and rethrown as generic `AuthorizationException` by the `action: EXCEPTION_NOT_AUTHORIZED` argument on can* directives.

This is an issue for our use case as we perform sanity checks (business logic) in the mutation code and throw specific exception / error messages back to the user in some cases. 
We do want to use `action: EXCEPTION_NOT_AUTHORIZED` to hide `ModelNotFoundException` from end users, but we still need to return specific exception / message / error codes to the user manually (when the `@can*` directive actually passed) from the mutation class.

**Changes**

With this PR, I moved resolving the field to outside of the authorization try-catch block, when applicable.
I kept the changes non breaking.

Note (unchanged behavior):
1. Exceptions thrown in @canResolved are always caught / modified as resolving happens during the authorization.
2.Exceptions throwing in @canRoot are never caught / modified as resolving happens before the authorization, outside of the try-catch

Unsure if this needs to be documented, happy to update the docs if needed?

**Breaking changes**

None

### DIFF
--- a/tests/Integration/Auth/CanRootDirectiveDBTest.php
+++ b/tests/Integration/Auth/CanRootDirectiveDBTest.php
@@ -1,0 +1,36 @@
+<?php declare(strict_types=1);
+
+namespace Integration\Auth;
+
+use Tests\DBTestCase;
+use Tests\Utils\Models\User;
+use Tests\Utils\Mutations\ThrowWhenInvoked;
+use Tests\Utils\Policies\UserPolicy;
+
+final class CanRootDirectiveDBTest extends DBTestCase
+{
+    public function testDoesntConcealResolverException(): void
+    {
+        $admin = new User();
+        $admin->name = UserPolicy::ADMIN;
+        $this->be($admin);
+
+        $this->schema = /** @lang GraphQL */ '
+        type Mutation {
+            throwWhenInvoked: Task
+        }
+
+        type Task {
+            name: String! @canRoot(ability: "adminOnly", action: EXCEPTION_NOT_AUTHORIZED)
+        }
+        ' . self::PLACEHOLDER_QUERY;
+
+        $this->graphQL(/** @lang GraphQL */ '
+        mutation {
+            throwWhenInvoked {
+                name
+            }
+        }
+        ')->assertGraphQLErrorMessage(ThrowWhenInvoked::ERROR_MESSAGE);
+    }
+}


src/Auth/BaseCanDirective.php
```diff
@@ -88,7 +88,10 @@ public function handleField(FieldValue $fieldValue): void
             $authorizeModel = fn (mixed $model) => $this->authorizeModel($gate, $ability, $model, $checkArguments);
 
             try {
-                return $this->authorizeRequest($root, $args, $context, $resolveInfo, $resolver, $authorizeModel);
+                $resolved = $this->authorizeRequest($root, $args, $context, $resolveInfo, $resolver, $authorizeModel);
+                if ($resolved) {
+                    return $resolved;
```
k0ka, line 93: If the `$resolved` value is null or false for `CanResolvedDirective`, the resolver would be executed twice. I guess it's better to split the resolved value and whether it was resolved into two different variables.

tests/Integration/Auth/CanFindDirectiveDBTest.php
```diff
@@ -218,7 +256,8 @@ public function testFailsToFindSpecificModelWithFindOrFailFalse(): void
 
         $this->mockResolver(null);
 
-        $this->schema = /** @lang GraphQL */ '
+        $this->schema /** @lang GraphQL */
```
k0ka, line 259: It's better not to change coding style in unrelated places.

k0ka (dismissed):

src/Auth/BaseCanDirective.php
k0ka: As I know, only methods with `@api` phpdoc must be preserved. So you may change the `authorizeRequest` signature (but it's better to wait for @spawnia response to make sure). 

Another approach is to add another method like `authorizeRequestWithResolving` to `BaseCanDirective`. The base method should just execute `authorizeRequest` of the child class - so if someone overrides this function outside of the library, it would still work. Other our `Can*Directive` should override `authorizeRequestWithResolving`

tests/Integration/Auth/CanFindDirectiveDBTest.php
```diff
@@ -23,7 +24,8 @@ public function testQueriesForSpecificModel(): void
         $user = factory(User::class)->create();
         assert($user instanceof User);
 
-        $this->schema = /** @lang GraphQL */ '
+        $this->schema /** @lang GraphQL */
+            = '
```
spawnia, line 28: Please revert the reformatting.

tests/Integration/Auth/CanFindDirectiveDBTest.php
spawnia: Please revert that part.

tests/Integration/Auth/CanModelDirectiveDBTest.php
```diff
@@ -0,0 +1,38 @@
...
+final class CanModelDirectiveDBTest extends DBTestCase
+{
+    public function testDoesntConcealResolverException(): void
+    {
+        $admin = new User();
+        $admin->name = UserPolicy::ADMIN;
+        $this->be($admin);
+
+        $this->schema /** @lang GraphQL */
+            = '
```
spawnia, line 19:
```suggestion
        $this->schema = /** @lang GraphQL */ '
```

tests/Integration/Auth/CanQueryDirectiveDBTest.php
```diff
@@ -83,6 +84,38 @@ public function testFailsToFindSpecificModelWithQuery(): void
...
+    {
+        $admin = new User();
+        $admin->name = UserPolicy::ADMIN;
+        $this->be($admin);
+
+        $user = factory(User::class)->create();
+        assert($user instanceof User);
+
+        $this->schema /** @lang GraphQL */
+            = '
```
spawnia, line 97:
```suggestion
        $this->schema = /** @lang GraphQL */ '
```

src/Auth/BaseCanDirective.php
spawnia: Feel free to break the internal API of BaseCanDirective to simplify things and avoid double resolving.

spawnia (changes requested):

src/Auth/BaseCanDirective.php
MarcEspiard: @spawnia Updated my approach to simply keep track of if the resolver was called and return accordingly. It's value independent so should work much more consistently.
No API changes needed.
Let me know your thoughts.

spawnia: Great idea!

spawnia (commented):
Nicely done. What about adding tests for `@canRoot`? There is an existing `CanRootDirectiveTest`.

MarcEspiard:
Errors should be fixed and added an extra test for CanRoot

spawnia (approved):

MarcEspiard:
Thanks for the quick release!

Issue #2687, opened 2025-06-04
Type based on Laravel model ignores regular php properties


tminich:
**Describe the bug**

```php
class Test extends Model
{
    public string $test = 'test';
}
```

```graphql
type Test {
    test: String
}
```

This will always return null for `test`, it never reads the property.

**Expected behavior/Solution**

It should read the property value. 

We are moving away from Laravel get___Attribute getters (which work) as the native php 8.4 virtual property syntax has much better IDE integration and static analysis, so this is a bug that could lead to easily missed mistakes. I guess the implementation uses Laravel array access to retrieve the values, which is imo fine, but it should fall back to trying regular php object properties in case that fails.

Note: A current workaround is using `@rename` to 'rename' the attribute to itself.

**Lighthouse Version**

6.57.1

PR #2688, opened 2025-06-06
Test for PHP property access

tminich:
Adds a (currently failing) test for accessing php properties on Laravel models as requested in #2687


spawnia:
Thank you @tminich for the test and your feedback. I merged a modified version of the tests to ensure the current behavior and opened https://github.com/nuwave/lighthouse/pull/2693 to potentially change the default resolver.

PR #2694, opened 2025-06-15
feat(ast): add path() getter method to ASTCache class

as-yakovenko:
- [ ] Added or updated tests (not applicable for this change)
- [x] Updated CHANGELOG.md

**Changes**

Added a public `path()` getter method to the `ASTCache` class to improve API consistency and make the class easier to extend. The method returns the current schema cache file path and allows clean external access to override the cache path without relying on direct property access.

This change:
- Exposes the internal `$path` property via a consistent public API
- Improves extensibility for use cases with dynamic schema caching (e.g. multi-tenant or per-request schemas)
- Aligns with object-oriented encapsulation best practices

**Breaking changes**

None. This is a fully backward-compatible addition that does not affect existing behavior or usage.

### DIFF
--- a/src/Schema/AST/ASTCache.php
+++ b/src/Schema/AST/ASTCache.php
@@ -61,10 +61,11 @@ class ASTCache
     /** @param  callable(): DocumentAST  $build */
     public function fromCacheOrBuild(callable $build): DocumentAST
     {
-        if ($this->filesystem()->exists($this->path())) {
-            $ast = require $this->path();
+        $path = $this->path();
+        if ($this->filesystem()->exists($path)) {
+            $ast = require $path;
             if (! is_array($ast)) {
-                throw new InvalidSchemaCacheContentsException($this->path(), $ast);
+                throw new InvalidSchemaCacheContentsException($path, $ast);
             }
 
             /** @var SerializableDocumentAST $ast */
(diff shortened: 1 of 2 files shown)


src/Schema/AST/ASTCache.php
```diff
@@ -42,24 +47,24 @@ public function set(DocumentAST $documentAST): void
...
     /** @param  callable(): DocumentAST  $build */
     public function fromCacheOrBuild(callable $build): DocumentAST
     {
-        if ($this->filesystem()->exists($this->path)) {
-            $ast = require $this->path;
+        if ($this->filesystem()->exists($this->path())) {
+            $ast = require $this->path();
             if (! is_array($ast)) {
-                throw new InvalidSchemaCacheContentsException($this->path, $ast);
+                throw new InvalidSchemaCacheContentsException($this->path(), $ast);
```
spawnia, line 67: Avoid calling the method repeatedly by extracting to a variable.

spawnia (changes requested):
Please add a changelog entry.

src/Schema/AST/ASTCache.php
as-yakovenko: > Avoid calling the method repeatedly by extracting to a variable.

@spawnia Sorry, I missed that initially. Thanks for pointing it out!

PR #1867, opened 2021-06-02
Fix combining `@lazyLoad` with paginated relations

hosmelq:
- [x] Added or updated tests
- [ ] Documented user facing changes
- [x] Updated CHANGELOG.md

**Changes**

Allow `LengthAwarePaginator` as a parameter in `LazyLoadDirective` directive.

**Breaking changes**

None.

PR #2695, opened 2025-06-19
Allow combining `@lazyLoad` and `@paginate`

Nextra:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Previously #1867 removed the `Collection` type-hint from the `LazyLoadDirective` field `resultHandler`. This had the primary intent of accepting `LengthAwarePaginator` (as can be seen in the added PHPdoc), but would implicitly handle all Laravel paginators. When the PHPdoc got converted back into a type-hint utilizing DNF types in #1706, this accidentally stopped working.

I believe the type-hint can be widened to `AbstractPaginator` without issue, allowing `@lazyLoad` to combine with `@paginate` once more.

**Breaking changes**

Since this widens the accepted parameters, no compatibility break expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+
+- Allow combining `@lazyLoad` and `@paginate` https://github.com/nuwave/lighthouse/pull/2695
+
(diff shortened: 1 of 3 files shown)


spawnia:
Thank you, released with https://github.com/nuwave/lighthouse/releases/tag/v6.60.0.

PR #2696, opened 2025-06-23
Add `--sort` flag for `lighthouse:print-schema` console command

loganswartz:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds a `--sort` flag for the `lighthouse:print-schema` console command.

The main motivation behind this is to allow for stronger normalization of the final compiled schema, which can be useful for ex: compiled schemas tracked by source control. Currently, the order in which items are defined in the raw schema determines their order in the output, which means that moving items around in the raw schema can result in the compiled schema changing, even when it is logically identical. Using this flag, any identical raw schema should always compile down to the exact same output.

**Breaking changes**

None, since this is opt-in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add `--sort` option to `lighthouse:print-schema` command to sort the compiled
+  schema before printing.
+
 ## v6.60.0
 
 ### Changed
--- a/src/Console/PrintSchemaCommand.php
+++ b/src/Console/PrintSchemaCommand.php
@@ -24,6 +24,7 @@ lighthouse:print-schema
 {--D|disk= : The disk to write the file to}
 {--json : Output JSON instead of GraphQL SDL}
 {--federation : Include federation directives and exclude federation spec additions, like _service.sdl}
+{--sort: Sort all types, fields, and arguments in alphabetical order, when possible}
 SIGNATURE;
 
     protected $description = 'Compile the GraphQL schema and print the result.';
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add `--sort` option to `lighthouse:print-schema` command to sort the compiled
+  schema before printing.
```
spawnia, line 15:
```suggestion
- Add `--sort` option to `lighthouse:print-schema` command to sort the compiled schema before printing https://github.com/nuwave/lighthouse/pull/2696
```

src/Console/PrintSchemaCommand.php
```diff
@@ -24,6 +24,7 @@ class PrintSchemaCommand extends Command
 {--D|disk= : The disk to write the file to}
 {--json : Output JSON instead of GraphQL SDL}
 {--federation : Include federation directives and exclude federation spec additions, like _service.sdl}
+{--sort: Sort all types, fields, and arguments in alphabetical order, when possible}
```
spawnia, line 27: What do you mean `when possible`?

src/Console/PrintSchemaCommand.php
```diff
@@ -42,14 +43,37 @@ public function handle(ASTCache $cache, FilesystemManager $filesystemManager, Sc
...
+            $options = [];
+            if ($this->option('sort')) {
+                $options = [
+                    'sortArguments' => true,
+                    'sortEnumValues' => true,
+                    'sortFields' => true,
+                    'sortInputFields' => true,
+                    'sortTypes' => true,
+                ];
+            }
```
spawnia, line 73:
```suggestion
            $options = $this->option('sort')
                ? [
                    'sortArguments' => true,
                    'sortEnumValues' => true,
                    'sortFields' => true,
                    'sortInputFields' => true,
                    'sortTypes' => true,
                ]
                : [];
```

spawnia (changes requested):
I like it!

src/Console/PrintSchemaCommand.php
loganswartz: I initially added that `when possible` before I had added the checks to throw on an invalid combination of flags, so it was intended to communicate that `--sort` is incompatible with `--json` and `--federation`.

I see how that could be confusing now, so I'll delete it.

src/Console/PrintSchemaCommand.php
```diff
@@ -24,6 +24,7 @@ class PrintSchemaCommand extends Command
 {--D|disk= : The disk to write the file to}
 {--json : Output JSON instead of GraphQL SDL}
 {--federation : Include federation directives and exclude federation spec additions, like _service.sdl}
+{--sort: Sort all types, fields, and arguments in alphabetical order}
```
spawnia, line 27:
```suggestion
{--sort: Sort all possible schema elements in alphabetical order}
```

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add `--sort` option to `lighthouse:print-schema` command to sort the compiled
+  schema before printing https://github.com/nuwave/lighthouse/pull/2696
```
spawnia, line 15:
```suggestion
- Add option `--sort` to command `lighthouse:print-schema` https://github.com/nuwave/lighthouse/pull/2696
```

spawnia:
Released with https://github.com/nuwave/lighthouse/releases/tag/v6.61.0. Please consider sponsoring.

PR #2698, opened 2025-06-27
POC: Create a failing test to demo filtering BETWEEN with variables

Luke-Shepp:
This PR exists as a demo for an issue: https://github.com/nuwave/lighthouse/issues/2699

### DIFF
--- a/tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
+++ b/tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
@@ -20,6 +20,7 @@ final class WhereConditionsDirectiveTest extends DBTestCase
         id: ID!
         name: String
         email: String
+        date_of_birth: String
     }
 
     type Post {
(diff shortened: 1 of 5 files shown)


tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
```diff
@@ -54,6 +56,116 @@ protected function getPackageProviders($app): array
...
+                'users' => [
+                    [
+                        'id' => "{$user2->id}",
+                    ],
+                ],
+            ],
+        ]);
+    }
+
+    public function testBetweenWithOneVariable(): void
```
spawnia, line 97: I have added this test case, which works. So the part that does not work is more closely related to using variables in a list literal of a non-list type.

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
```diff
@@ -20,6 +21,7 @@ final class WhereConditionsDirectiveTest extends DBTestCase
         id: ID!
         name: String
         email: String
+        date_of_birth: String
```
spawnia, line 24:
```suggestion
```

Seems unnecessary for the test at hand.

tests/database/factories/UserFactory.php
```diff
@@ -12,4 +12,5 @@
     'name' => $faker->name,
     'email' => $faker->unique()->safeEmail,
     'password' => '$2y$10$TKh8H1.PfQx37YgCzwiKb.KjNyWgaHb9cbcoQgdIVFlYg7B77UdFm', // secret
+    'date_of_birth' => $faker->dateTimeBetween,
```
spawnia, line 15:
```suggestion
```

Not really necessary, the attribute can just stay null.

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
Luke-Shepp: This was not a PR to be merged and therefore not refined in anyway - it accurately reflected my failing case, to demonstrate for the purpose of creating an issue - hence the date for use with the BETWEEN - this was my use-case, so the test reflected that for the issue creation


Feel free to close or update to better reflect the package standards.

spawnia: I generally like to make fixes right in the PR that adds the failing test case. It keeps things together and credits you as the author.

tests/Integration/WhereConditions/WhereConditionsDirectiveTest.php
spawnia: Found the bug, see https://github.com/mll-lab/graphql-php-scalars/pull/35.

PR #2706, opened 2025-07-23
Simplify cache configuration access

spawnia:


tests/Integration/ValidationCachingTest.php
```diff
@@ -74,29 +73,6 @@ public function testDisabled(): void
         $event->assertDispatchedTimes(KeyWritten::class, 0);
     }
 
-    public function testConfigMissing(): void
-    {
-        $config = $this->app->make(ConfigRepository::class);
-        $config->set('lighthouse.query_cache.enable', false);
-        $config->set('lighthouse.validation_cache', null);
```
spawnia, line 81: Realistically, this never happens due to the configuration merging.

Issue #2723, opened 2025-10-27
Define input enums in WHERE HAS conditions


DrPowerSlam:
When defining a WHERE HAS query with a condition in complex queries, Lighthouse will fetch the column enums from the parent type, not the relation, if the `where` definition uses any sort of column enumeration.

```graphql
{
  drivers(
    where: {
      OR: [
        {HAS: {relation: "employee", condition: {column: NAME, value: "Peter"}}}, // `column: DriverColumnEnum` instead of `column: EmployeeColumnEnum`
        {column: ID, operator: EQ, value: "6217"} // `column: DriverColumnEnum`
      ]
    }
  ) {
    data {
      id
      name
      employee {
        name
      }
    }
  }
}
```

I think it would make sense if you could define the relations in a where, the same way we do in `orderBy`:

```
extend type Query {
    "List multiple drivers."
    drivers(
        orderBy: _ @orderBy(
            columnsEnum: "DriverColumn",
            relations: [{relation: "employee", columns: ["name"]}]
        )
        where: _ @whereConditions(
            columnsEnum: "DriverColumn",
            relations: [{relation: "employee", columns: ["name"]}] // or columnsEnum.
        )
    ): [Driver!]! @paginate(defaultCount: 10)
}
```
This would both enable stronger schema generation by having enums for the relation name (right now it's currently just a string), make `orderBy` and `where` more similar in how they are written in the schema definitions, and of course fix the issue described in the discussion link below.
One of the huge benefits of using enum types in the schema, is that the schema documentation can explain exactly what a API consumer can use, so I think this would be the absolute best solution for usability. Otherwise a API consumer will need intricate knowledge of the database columns available. 
I think it would also make sense that the `relations.column` type will default to `String` if column enumerations are missing, instead of trying to resovle the column enum type of the parent.

Thanks for taking your time to consider this. :)

Original Q&A discussion this is based on: https://github.com/nuwave/lighthouse/discussions/2720#discussioncomment-14779266

spawnia:
> I think it would also make sense that the relations.column type will default to String if column enumerations are missing, instead of trying to resovle the column enum type of the parent.

Perhaps this can be implemented as a first step, can you post a pull request?

PR #2725, opened 2025-10-29
Change HAS condition type to always use String in @whereConditions

DrPowerSlam:
https://github.com/nuwave/lighthouse/issues/2723
 
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Changes the condition type of `HAS` arguments in `@whereConditions` to always use `String` type, so the `column` type isn't inherited from the parent node. 

**Breaking changes**

Technically this is a breaking change, since users that currently use the unupdated code will go from an Enum type to a String type.
However, I would argue that it's almost impossible to correctly use the current functionality, since you need a perfect overlap between columns in a node and all its relations to make it work.

### DIFF
--- a/src/WhereConditions/WhereConditionsServiceProvider.php
+++ b/src/WhereConditions/WhereConditionsServiceProvider.php
@@ -107,10 +107,40 @@ GRAPHQL
 
     public static function createHasConditionInputType(string $name, string $description, string $columnType): InputObjectTypeDefinitionNode
     {
-        return self::createWhereConditionsInputType(
-            $name . self::DEFAULT_HAS_CONDITION,
-            $description,
-            $columnType,
+        $hasRelationInputName = $name . self::DEFAULT_WHERE_RELATION_CONDITIONS;
+        $name .= self::DEFAULT_HAS_CONDITION;
+
+        $operator = Container::getInstance()->make(Operator::class);
+
+        $operatorName = Parser::enumTypeDefinition(
+            $operator->enumDefinition(),
+        )
+            ->name
+            ->value;
+        $operatorDefault = $operator->default();
+
+        return Parser::inputObjectTypeDefinition(/** @lang GraphQL */ <<<GRAPHQL
+            "{$description}"
+            input {$name} {
+                "The column that is used for the condition."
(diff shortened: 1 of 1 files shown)


spawnia:
@oortasertecfarma How did this break your project? I was inclined to agree with @DrPowerSlam assessment:

> it's almost impossible to correctly use the current functionality, since you need a perfect overlap between columns in a node and all its relation

Saarnaki:
> @oortasertecfarma How did this break your project? I was inclined to agree with @DrPowerSlam assessment:
> 
> > it's almost impossible to correctly use the current functionality, since you need a perfect overlap between columns in a node and all its relation

Can't speak for @oortasertecfarma , but this was a breaking change for me as well. This increased the cached schema file size from 17mb to 21mb, which caused memory exhaustion errors in ASTCache.php fromCacheOrBuild method, line 64, while reading the cache file.

oortasertecfarma:
After looking into it, the issue that occurred in my project after upgrading was caused by a custom override of manipulateArgDefinition in WhereConditionsBaseDirective, which already allowed the use of String in HAS conditions and covered a slightly broader use case.
Once I aligned the override with the new explicit behavior introduced in this PR, all tests are passing again.

Regarding the schema size mentioned above, I also checked it on my side.
In my case, the increase was relatively small, going from ~13.94 MB to ~13.96 MB in the cached schema.

spawnia:
@trckster Can you please elaborate? So far, nobody has articulated how and why this was a breaking change.

DrPowerSlam:
@trckster It's not backwards compatible, but at the same time it's so hard to use correctly that the fix will impact a minimum of people. I'm of course sorry it has affected you, but I think you would have hit a roadblock if you continued (mis)using this functionality.

I can't decide on which which version this fix should be bundled with (and after 2 weeks we've only had 1 legitimate use-case where it broke something), I'm just a contributor, not a maintainer.

On your last point, I think you would have to create a custom directive that checks if an API consumer is inputting something that isn't allowed. The schema can't really validate an input string, so you would have to do it in runtime.
At least until someone has created a PR to make sure that the `column` input field can use an enum type.

Issue #1449, opened 2020-06-24
Artisan command `lighthouse:ide-helper` is very slow


spawnia:
Starting from https://github.com/nuwave/lighthouse/releases/tag/v4.14.0, running

    php artisan lighthouse:ide-helper

has become very slow. Some profiling should help to uncover what is taking so long.

spawnia:
Trying https://www.christianroessler.net/tech/2015/php-xdebug-profile-single-cli-application-without-enabling-globally.html for starters, other options:

- https://github.com/NoiseByNorthwest/php-spx
- https://blackfire.io/docs/up-and-running/installation

mathieu-bour:
What do you mean by "very slow"? In my project, it's something like two seconds for something like 50 queries, mutations and types.

spawnia:
It takes about 2 minutes.

I am running in WSL2 with the project on a Windows filesystem, which probably explains the difference. The bottleneck definitely seems to be filesystem access.

We use `haydenpierce/class-finder` to look for schema directive classes. It accesses the filesystem a whole lot, probably more than necessary. See https://gitlab.com/hpierce1102/ClassFinder/-/issues/14.

mathieu-bour:
I just remembered that I also had severe issues while exploring filesystem on WSL 2. I'll definitely join the discussion on ClassFinder.

mathieu-bour:
@spawnia I cannot reproduce the slowness on WSL2, even with a "quite big" schema...
Could you detail your PHP setup and the slow schema?

spawnia:
- WSL2 running Ubuntu 16.04
- Using docker-desktop
- Run PHP 7.3 inside a container
- Project is located within windows filesystem
- F-Secure antivirus is running, might have an impact?

Anyways, while profiling i found that `ClassFinder` generates a lot of filesystem accesses (functions such as `scandir`, `file_exists`, `file_get_contents`). The slowness in my system is definitely related to that, but I am sure it can be optimized.

Jofairden:
If it is of any help, the command took me 16.22 seconds running in a Vagrant box, I have 1 schema, 41 types along with 1-2 queries per type and at least one enum (sometimes more). I setup my box to use NFS sync which should improve speed, I don't actually know if it does though.

mathieu-bour:
Ok, I guess that my native setup PHP in WSL2 has definitely a perform impact. But is it really a problem? I mean, class finder can be optimized but I guess that the slowness is due to the setup itself.

[...]

PR #2768, opened 2026-05-22
Discover directive classes via Composer's autoload maps

Gregory-Gerard:
Hi!

First time contributing to Lighthouse, so happy to take any feedback on the approach.

On a personal Laravel app I noticed `php artisan lighthouse:ide-helper` was taking ~18s, which felt long for what the command actually does.

Dug a bit with the Xdebug profiler, ended up on this old issue (#1449) that already describes the problem.
A quick rewrite of the directive discovery dropped the same command to ~1s on my project, so I figured it was worth a PR.

- [x] Added or updated tests
- [x] Updated CHANGELOG.md (skip for docs-only changes)

**Changes**

`IdeHelperCommand::scanForDirectives` and `DirectiveLocator::classes()` both call `HaydenPierce\ClassFinder::getClassesInNamespace`, which rescans the entire vendor PSR-4 tree from scratch on every call (no caching, recurses into every subdir of every package).

This PR replaces those calls with a small `ComposerClassFinder` that reads Composer's already-built maps (`getClassMap()` + `getPrefixesPsr4()`).

It is meant as a drop-in for `ClassFinder::STANDARD_MODE`: same inputs, same outputs, same implementation details.
Notably we keep the strict leading-backslash behavior of the old lib (a FQN like `\Foo\Bar` returns an empty set instead of being normalized), so existing callers cannot break.
This is locked in by unit test `testReturnsEmptyArrayForLeadingBackslash`.

The `haydenpierce/class-finder` dependency is dropped.

**Profiling (Xdebug profile mode, on a real Laravel 12 + Lighthouse 6.66 app)**

| function                                      | inclusive ms | self ms | calls |
|-----------------------------------------------| --- | --- | --- |
| `IdeHelperCommand::handle`                    | 24943 | 0.1 | 1 |
| - `scanForDirectives`                         | **23405** | 0.7 | 1 |
| -- `ClassFinder::getClassesInNamespace`       | 23403 | 0.3 | 12 |
| --- `PSR4NamespaceFactory::getPSR4Namespaces` | 21312 | 11.6 | **23** |
| ---- `PSR4NamespaceFactory::createNamespace`  | 55228 | 684 | **87 906** |
| `php::scandir`                                | 9936 | 9936 | **88 170** |
| `php::is_dir`                                 | 1569 | 1569 | **429 633** |

For 12 directive namespaces, `getPSR4Namespaces()` runs 23 times (twice per `getClassesInNamespace`, once in `isNamespaceKnown` and once in `findBestPSR4Namespace`).
Each run walks the whole PSR-4 tree of every vendor package.

**`php artisan lighthouse:ide-helper` benchmark on the same app**

|                 | Before | After          |
|-----------------| --- |----------------|
| Total time      | ~18 s | ~1.1 s         |
| Speedup         | | ~16x           |
| Generated files | | identical |

3 sample runs after the change: 1.43 s / 1.07 s / 1.10 s.

**New implementation vs the old lib**

I ran both implementations side by side on representative namespaces (flat ns, ns with sub-namespaces, autoload-dev ns, unknown ns, leading/trailing backslash variants).
After switching `trim` to `rtrim` to match the old strict behavior on leading `\`, both return identical sets on every case.

The unit test pins the behavior we replicate: concrete classes only (interfaces and traits filtered via `class_exists`), no recursion into sub-namespaces, trailing `\` tolerated, leading `\` returns empty (matches old lib), discovery from PSR-4 prefixes and from the optimized classmap.

**Breaking changes**

None.
The new finder is an internal detail and the generated IDE helper files are unchanged.

### DIFF
--- a/tests/Unit/Support/ComposerClassFinderTest.php
+++ b/tests/Unit/Support/ComposerClassFinderTest.php
@@ -1,0 +1,84 @@
+<?php declare(strict_types=1);
+
+namespace Tests\Unit\Support;
+
+use Nuwave\Lighthouse\Cache\CacheDirective;
+use Nuwave\Lighthouse\Cache\CacheKeyAndTags;
+use Nuwave\Lighthouse\Cache\CacheServiceProvider;
+use Nuwave\Lighthouse\Support\AppVersion;
+use Nuwave\Lighthouse\Support\ComposerClassFinder;
+use Nuwave\Lighthouse\Support\DriverManager;
+use Nuwave\Lighthouse\Support\Utils;
+use Tests\Console\UnionDirective;
+use Tests\TestCase;
+
+final class ComposerClassFinderTest extends TestCase
+{
+    public function testReturnsConcreteClassesInNamespace(): void
+    {
+        $classes = ComposerClassFinder::directClassesInNamespace('Nuwave\\Lighthouse\\Cache');
+
+        $this->assertContains(CacheDirective::class, $classes);
+        $this->assertContains(CacheServiceProvider::class, $classes);
+    }
+
+    public function testFiltersInterfacesAndTraits(): void
+    {
+        $classes = ComposerClassFinder::directClassesInNamespace('Nuwave\\Lighthouse\\Cache');
+
+        $this->assertNotContains(CacheKeyAndTags::class, $classes, 'Interfaces must not be returned.');
+    }
+
+    public function testDoesNotRecurseIntoSubNamespaces(): void
+    {
+        $classes = ComposerClassFinder::directClassesInNamespace('Nuwave\\Lighthouse\\Support');
+
+        $this->assertContains(ComposerClassFinder::class, $classes);
+        $this->assertContains(Utils::class, $classes);
+        $this->assertContains(AppVersion::class, $classes);
+        $this->assertContains(DriverManager::class, $classes);
+
+        foreach ($classes as $class) {
+            $this->assertStringNotContainsString(
+                'Nuwave\\Lighthouse\\Support\\Contracts\\',
+                $class,
+                'Sub-namespace Contracts must not be recursed into.',
+            );
+            $this->assertStringNotContainsString(
+                'Nuwave\\Lighthouse\\Support\\Traits\\',
+                $class,
+                'Sub-namespace Traits must not be recursed into.',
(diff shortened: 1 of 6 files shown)


tests/Unit/Support/ComposerClassFinderTest.php
```diff
@@ -0,0 +1,84 @@
...
+use Nuwave\Lighthouse\Support\DriverManager;
+use Nuwave\Lighthouse\Support\Utils;
+use Tests\Console\UnionDirective;
+use Tests\TestCase;
+
+final class ComposerClassFinderTest extends TestCase
+{
+    public function testReturnsConcreteClassesInNamespace(): void
+    {
+        $classes = ComposerClassFinder::directClassesInNamespace('Nuwave\\Lighthouse\\Cache');
```
Gregory-Gerard, line 19: Should we use fixtures here instead of real classes so we don't rely on implementation details?

spawnia: Yeah, should make the tests more stable - seeing them fail while working on something related to cache would be weird.

spawnia:
Have you looked into https://gitlab.com/hpierce1102/ClassFinder/-/work_items/14? https://gitlab.com/hpierce1102/ClassFinder seems to be actively maintained, so if we can keep Lighthouse lean and improve the upstream package instead, that would generally be my preference.

Gregory-Gerard:
> Have you looked into https://gitlab.com/hpierce1102/ClassFinder/-/work_items/14? https://gitlab.com/hpierce1102/ClassFinder seems to be actively maintained, so if we can keep Lighthouse lean and improve the upstream package instead, that would generally be my preference.

@spawnia I'll take a look. I get the reasoning and I'm fine going that way. My only instinct was that a single small file felt a bit leaner. I'll try to push a PR upstream on ClassFinder, and/or come back here with an alternative that keeps the dependency while still cutting down the discovery cost 😁

Gregory-Gerard:
Pushed an MR upstream on ClassFinder: https://gitlab.com/hpierce1102/ClassFinder/-/merge_requests/33

I'll keep this PR in draft for now. If the GitLab MR gets merged, I'll update this PR to just bump the version constraint. Thanks!

PR #2772, opened 2026-06-11
Pass scalar overrides explicitly to keep lazy type loading lazy on graphql-php >= 15.31

### LINKED ISSUE #2771: Lazy schema becomes eager on every request with graphql-php >= 15.31: built-in scalar lookup resolves possibleTypes()
### Summary

On `webonyx/graphql-php >= 15.31.0`, Lighthouse's lazy schema silently becomes eager on essentially every request. The executor's first built-in scalar lookup (`Schema::getType('Boolean')` etc. — this happens for virtually any operation, including `{ __typename }`) now triggers scalar-override discovery inside graphql-php, which resolves the schema's `types` callable. In Lighthouse that callable is `TypeRegistry::possibleTypes()`, which builds **every type in the schema from the AST**. Since the `Schema` is rebuilt per request under PHP-FPM, the documented lazy design is defeated on each request.

### Mechanism

`SchemaBuilder::build()` configures:

```php
// Use lazy type loading to prevent unnecessary work
$config->setTypeLoader(
    fn (string $name): ?Type => $this->typeRegistry->search($name),
);

// Enables introspection to list all types in the schema
$config->setTypes(
    /** @return array<string, \GraphQL\Type\Definition\Type> */
    fn (): array => $this->typeRegistry->possibleTypes(),
);
```

graphql-php introduced scalar overrides in 15.31.0 (webonyx/graphql-php#1869) and in 15.31.2 moved override discovery exclusively to scanning `types` (webonyx/graphql-php#1884, see also webonyx/graphql-php#1874). With a lazy `types` callable, that scan forces full resolution the first time any built-in scalar is looked up.

### Measured impact

Production app on Lighthouse 6.66, upgrading graphql-php 15.30.2 → 15.32.3:

- every operation 1.5–3.4× slower wall-clock
- ~6× more PHP work (xdebug cachegrind: 9.1 MB → 57 MB for an authenticated `{ __typename }`)
- ablation: patching Lighthouse to `setTypes(fn () => [])` restores the 15.30.2 baseline

Standalone repro and benchmark table: webonyx/graphql-php#1927.

Affected users currently have only bad options: pin graphql-php to 15.30.2 (re-exposes GHSA-r7cg-qjjm-xhqq, GHSA-fc86-6rv6-2jpm, CVE-2026-40476) or accept the per-request cost.

### Proposed change

webonyx/graphql-php#1927 (draft) adds `SchemaConfig::setScalarOverrides(?array)`: `null` (default) keeps scan-based discovery; an explicit array — including `[]` — skips the scan entirely.

Lighthouse can determine the overrides cheaply from the AST without building any types: a scalar override is a user-defined type named after a built-in scalar, and `$documentAST->types` is keyed by name. Sketch (in `build()`, after `setTypes`):

```php
// TODO remove the guard when requiring the graphql-php release that ships setScalarOverrides
if (method_exists($config, 'setScalarOverrides')) {
    $scalarOverrides = [];
    foreach (Type::BUILT_IN_SCALAR_NAMES as $name) {
        if (isset($documentAST->types[$name])) {
            $type = $this->typeRegistry->get($name);
            assert($type instanceof ScalarType);
            $scalarOverrides[] = $type;
        }
    }
    $config->setScalarOverrides($scalarOverrides);
}
```

The `method_exists` guard keeps the current `webonyx/graphql-php: ^15` constraint intact and follows the same pattern as the query-complexity guard added in #2637; it can be dropped whenever the constraint is eventually bumped past the release containing the new API.

For the common case (no built-in scalar names redefined in the SDL) this is five `isset()` checks and an explicit empty array — the hot path stays fully lazy. The loop is not dead code: `TypeRegistry::search()` consults `$documentAST->types` before `Type::getStandardTypes()`, so a user SDL *can* redefine e.g. `String`.

I'm happy to submit a PR with this change plus a regression test asserting that a simple query execution does not invoke `possibleTypes()`, once the upstream API lands.

tmoitie:
Resolves #2771

- [x] Added or updated tests
- [x] Updated CHANGELOG.md (skip for docs-only changes)

**Changes**

On `webonyx/graphql-php >= 15.31.0`, the executor's first built-in scalar lookup triggers scalar-override discovery, which resolves the lazy `types` callable (`TypeRegistry::possibleTypes()`) and eagerly builds every type in the schema from the AST — on every request under PHP-FPM. Details and measurements in #2771.

webonyx/graphql-php#1927 adds `SchemaConfig::setScalarOverrides(?array)`: when overrides are passed explicitly (including an empty array), the scan of `types` is skipped entirely and lazy loading stays lazy.

This PR makes `SchemaBuilder::build()` determine the overrides cheaply from the document AST — a scalar override is a user-defined type named after a built-in scalar, and `$documentAST->types` is keyed by name — and pass them explicitly. For the common case (no built-in scalar names redefined in the SDL) this is five `isset()` checks and an explicit empty array.

The call is guarded with `method_exists($config, 'setScalarOverrides')`, following the pattern of the query complexity guard from #2637, so the `webonyx/graphql-php: ^15` constraint is unchanged. The guard (and the corresponding test skips) can be removed once the minimum graphql-php version includes the method.

Tests added:

- explicit empty overrides are set for a plain schema
- a redefined built-in scalar (`scalar String @scalar(class: "Email")`) is passed as an override
- regression: looking up a built-in scalar on the schema does not resolve unrelated types (this test fails without the `SchemaBuilder` change when running against graphql-php with #1927 applied)

All three tests are skipped on graphql-php versions without `setScalarOverrides`; I verified them locally against the #1927 branch (15 tests, 41 assertions, 0 skipped) and confirmed the regression test fails without the fix.

**Draft status**: marked draft until webonyx/graphql-php#1927 is merged and released.

**Breaking changes**

None. Behavior is unchanged on graphql-php versions without `setScalarOverrides`; on versions with it, schemas behave identically but built-in scalar lookups no longer force full type resolution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Schema/SchemaBuilder.php
```diff
@@ -70,6 +71,23 @@ protected function build(DocumentAST $documentAST): Schema
...
         );
 
+        // Passing scalar overrides explicitly prevents the first lookup of a built-in scalar
+        // from discovering them by resolving the lazy types callable, which would eagerly
+        // build every type in the schema, see https://github.com/nuwave/lighthouse/issues/2771.
+        // TODO remove this check when the minimum version of webonyx/graphql-php includes the method
+        if (method_exists($config, 'setScalarOverrides')) {
+            $scalarOverrides = [];
+            foreach (Type::BUILT_IN_SCALAR_NAMES as $name) {
+                if (isset($documentAST->types[$name])) {
```
spawnia: Skips scalar overrides from just SDL right? Not sure if that would even work, but I am thinking of a case like:

```graphql
# custom description
"UUID v4."
scalar ID
```

tests/Unit/Schema/SchemaBuilderTest.php
```diff
@@ -209,6 +213,57 @@ public function testExtendTypes(): void
...
+        $config = $schema->getConfig();
+        if (! method_exists($config, 'getScalarOverrides')) {
+            $this->markTestSkipped('Requires a version of webonyx/graphql-php that supports SchemaConfig::setScalarOverrides.');
+        }
+
+        $scalarOverrides = $config->getScalarOverrides();
+        $this->assertNotNull($scalarOverrides);
+        $this->assertCount(1, $scalarOverrides);
+
+        $stringOverride = reset($scalarOverrides);
```
spawnia:
```suggestion
        $stringOverride = $scalarOverrides[Type::STRING];
```

src/Schema/SchemaBuilder.php
tmoitie: Good catch — as written it only looked at the SDL, but the `types` scan also discovers overrides registered programmatically via `TypeRegistry::overwrite()`/`overwriteLazy()` (they end up in `possibleTypes()`). I moved the detection into a new `TypeRegistry::scalarOverrides()` that checks all three sources (`$types`, `$documentAST->types`, `$lazyTypes`), and added a test covering the programmatic case. Plain `register()` can't produce one — it throws "tried to register present type" for built-in names because `has()` falls back to the standard types.

Regarding the description-only case:

```graphql
"UUID v4."
scalar ID
```

That doesn't work today either: `TypeRegistry::resolveScalarType()` requires a `ScalarType` class (without `@scalar` it looks for a class named `ID` in the scalar namespaces) and throws `DefinitionException: Failed to find class ID extends GraphQL\Type\Definition\ScalarType ...`. I verified empirically on master with graphql-php 15.32.3: the scan resolves it on the first query, so it throws there; with this PR the same exception is thrown at schema build time instead — same failure, just earlier and deterministic rather than per-request.

tests/Unit/Schema/SchemaBuilderTest.php
tmoitie: Applied, thanks — and used the same keyed access in the new programmatic-override test.

spawnia (approved):

PR #2781, opened 2026-07-15
Support pagination with unlimited results

### LINKED ISSUE #2734: Support pagination with unlimited results

I currently have a field where pagination is useful in many, but not all, cases.  In some cases I'd like to be able to specify `first: null` or `first: -1` and receive all results.  My current workaround is to simply provide a large limit (e.g., 1000000), but doing so creates inefficient queries, particularly when accessing nested paginated records.

It would be great to be able to make the `first` argument nullable, thus dropping all limit clauses for a field.

williamjallen:
Closes https://github.com/nuwave/lighthouse/issues/2734.
 
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md (skip for docs-only changes)

**Changes**

Users can now specify `first: null` for paginated relationships, which returns the full set of records more efficiently than `first: <large number>` for nested relationships.

**Breaking changes**

None.  This is a purely additive change.

### DIFF
--- a/src/Execution/ModelsLoader/PaginatedModelsLoader.php
+++ b/src/Execution/ModelsLoader/PaginatedModelsLoader.php
@@ -48,6 +48,20 @@ class PaginatedModelsLoader implements ModelsLoader
      */
     protected function loadRelatedModels(EloquentCollection $parents): EloquentCollection
     {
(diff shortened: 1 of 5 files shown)


williamjallen:
@spawnia Any chance you could take a quick look at this?

PR #2777, opened 2026-06-29
Allow nested arg resolvers to run before parent model is saved

spawnia:
## Summary

Adds `SaveAwareArgResolver` interface (`@api`) that lets ArgResolver directives declare whether they run before or after the parent model is saved during mutation execution.

## Motivation

For BelongsTo relations, the FK lives on the parent model and must be set before `$model->save()`.
The existing implicit detection in `SaveModel` handles this for known relation methods, but there was no way for custom directives to participate in the pre-save phase.

`SaveAwareArgResolver` is the extension point: any directive implementing it can control its execution timing relative to `$model->save()`.

## Design decisions

`NestDirective` becomes a pure marker — `ResolveNested` handles recursion directly, creating a fresh instance without `SaveModel` to prevent double-saves at each nesting level. Pre-save resolvers inside `@nest` are recursively lifted to the top level by `ArgPartitioner::liftPreSaveResolversFromNest()`.

`@nest` now validates at schema build time that it is used on a non-list input object type (via `ArgManipulator` and `InputFieldManipulator`).

`runBeforeSave(Model $model)` is class-based, not instance-dependent — the model may not yet be hydrated when this is called. Implementations must base the decision on the model class and its relations.

When `runBeforeSave()` returns true, `__invoke()` may receive null as `$value` if the client sends null for a nullable input field.

## Documentation

- New `SaveAwareArgResolver` section in `docs/master/custom-directives/input-value-directives.md` with usage example
- `@nest` directive reference notes the type constraint
- `ArgResolver` interface PHPDoc improved with description and `@see` reference

🤖 Generated with Claude Code

PR #2764, opened 2026-03-29
Add Laravel 13 support

sk4rp:
**Changes**

Add Laravel 13 and PHP 8.5 to the supported version matrix:
  
- Extend all `illuminate/*` package constraints to include `^13`
- Extend `laravel/framework` dev constraint to include `^13`
- Extend `orchestra/testbench` dev constraint to include `^11.0`
- Add `^13` to the `laravel-version` matrix in `static-analysis` and `tests` CI jobs, with PHP 8.0 and 8.1 excluded (Laravel 13 requires PHP 8.2+)
- Add PHP 8.5 to the `php-version` matrix
- Remove `bensampo/laravel-enum` when installing with Laravel 13, as it does not support Laravel 13
- Use `laravel/framework:^13` instead of `illuminate/contracts:^13` for the Composer install step, as in Laravel 13 the standalone `illuminate/contracts` package conflicts with `laravel/framework`
- Update `coverage` and `benchmarks` CI jobs to run against PHP 8.5 + Laravel 13

**Breaking changes**

None.

PR #2784, opened 2026-07-28
Fix SaveAwareArgResolver ignored in nested mutations

spawnia:
Fixes a regression from https://github.com/nuwave/lighthouse/pull/2777 (v6.68.0) where `SaveAwareArgResolver::runBeforeSave()` was only honoured at the top level of a mutation. One level down inside nested HasMany/HasOne/ManyToMany/BelongsTo mutations, the pre-save resolver ran AFTER the child model was already persisted — silently discarding its attribute changes.

The root cause: `ResolveNested` defaulted to a partitioner that unconditionally routed all resolver-bearing args into the post-save set. Only `ModelMutationDirective::executeMutation()` overrode this with the pre-save-aware variant; all 12 nested resolver call sites used the broken default.

The fix collapses `nestedArgResolversWithoutPreSave` into `nestedArgResolvers` so the correct behavior is the default everywhere. When root is not a Model, behavior is unchanged. When root is a Model, `SaveAwareArgResolver` instances with `runBeforeSave=true` stay in the regular set so `SaveModel` executes them before persisting. The now-redundant explicit partitioner in `ModelMutationDirective` is removed.

Coverage spans the `create`, `update` and `upsert` branches of nested relation mutations, since all three build `ResolveNested` with the same default partitioner. Each new test fails without the fix, leaving the attributes null.

## Second fix: silent data loss when a `@nest` child collides with a sibling

Making `nestedArgResolvers` the default widened the reach of a separate, pre-existing bug (from https://github.com/nuwave/lighthouse/pull/2764, v6.68.0): lifting pre-save resolvers out of `@nest` wrote them into the flat regular set keyed by the child's bare field name. Field names are unique per input type, not across an input tree, so a name coincidence silently discarded an argument. Three ways to lose data:

1. A plain sibling `difficulty: Int` next to a `@nest` child `difficulty: LocationInput @geocode` — different columns, no semantic conflict, yet `difficulty` persisted as null.
2. Same-named children of two sibling `@nest` blocks — one resolver silently never ran.
3. A `@nest` child named `id` clobbering the primary key that `UpdateModel` and `UpsertModel` consume to identify the row.

Widening the default also turned case 3 into a hard error where the colliding name matched a relation operation namespace (`create`/`update`/…), since `NestedOneToMany` and friends then iterate an `ArgumentSet` where they expect a list.

Resolvers are not name-addressed — `SaveModel` iterates them by value and discards the keys — so lifted resolvers now travel as an ordered list rather than a name-keyed map, handed to the saver through the new `PreSaveArgumentsAware` interface. Lifting also no longer mutates the nest it reads from, which matters because `FieldValue` caches transformed argument sets per path and `@spread` shares `Argument` objects across rebuilt sets.

The capability check is what keeps this non-regressive: `UpdateModel` and `UpsertModel` forward it to their own `previous` (via the shared `DelegatesPreSaveArguments` trait), and a saver that can not consume the list gets nothing lifted, so its children run post-save exactly as on master. Both declining paths are covered by unit tests: a `previous` that does not implement the interface, and one that implements it but returns `null`. Third-party savers built as `new ResolveNested($closure)` — the pattern documented in `docs/master/concepts/arg-resolvers.md` — fall into the first case.

All colliding resolvers now run in schema order, so the last write wins where two of them target the same column. The interface is deliberately not marked `@api`.

Still open, and out of scope here: `@nest` inside a relation operation namespace has its lifted pre-save resolvers ignored entirely, because `NestedOneToMany` and friends never consume them. The hard error is gone; whether `@nest` should be legal there at all is a separate design question.

One unrelated commit names the `strict` flag on the existing `in_array()` calls in `src/`, which is purely cosmetic and easy to drop if you would rather keep it out.

🤖 Generated with Claude Code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Honour `SaveAwareArgResolver::runBeforeSave()` in nested mutations instead of only at the top level https://github.com/nuwave/lighthouse/pull/2784
+
 ## v6.69.1
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 ### Fixed
 
 - Honour `SaveAwareArgResolver::runBeforeSave()` in nested mutations instead of only at the top level https://github.com/nuwave/lighthouse/pull/2784
+- Run all pre-save arg resolvers lifted out of `@nest` instead of silently discarding those whose field name collides with a sibling https://github.com/nuwave/lighthouse/pull/2784
 
 ## v6.69.1
 
--- a/src/Cache/QueryCache.php
+++ b/src/Cache/QueryCache.php
@@ -43,14 +43,14 @@ class QueryCache
 
     public function clear(?int $opcacheTTLHours, bool $opcacheOnly): void
     {
-        if (in_array($this->mode, ['store', 'hybrid'], true)
+        if (in_array($this->mode, ['store', 'hybrid'], strict: true)
             && ! $opcacheOnly
         ) {
             $store = $this->makeCacheStore();
             $store->clear();
         }
 
-        if (in_array($this->mode, ['opcache', 'hybrid'], true)) {
+        if (in_array($this->mode, ['opcache', 'hybrid'], strict: true)) {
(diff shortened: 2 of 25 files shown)


spawnia:
Self-reviewed at 18882fb6 (thorough, full)

🤖 Posted by Claude Code

spawnia:
Self-reviewed at 89a56a2 (standard, full)

🤖 Posted by Claude Code

PR #2787, opened 2026-09-02
Let explicitly resolved directive classes win in classes()

spawnia:
DirectiveLocator::setResolved() already took precedence in resolve(), but classes() only scanned the namespaces, so definitions() could print a directive definition that differs from the directive that actually executes. This also makes it possible to pin a single directive name across namespaces, e.g. to keep a built-in directive that a plugin shadows.

The branch history starts from an attempt to also add DirectiveLocator::disable(), which turned out to be unsafe and was dropped: Lighthouse resolves the fields generated by @paginate, @node and federation through @field, so disabling @field broke those schemas. Users cannot know which built-ins are used internally, so a ban has to be enforced by whoever parses the hand-written schema instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Added
+
+- Add `DirectiveLocator::disable()` to make a directive unavailable, leaving its definition out of the schema https://github.com/nuwave/lighthouse/pull/2787
+
+### Changed
+
+- Let classes passed to `DirectiveLocator::setResolved()` take precedence in `DirectiveLocator::classes()` and thus `DirectiveLocator::definitions()` https://github.com/nuwave/lighthouse/pull/2787
+
 ## v6.69.2
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,6 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
-### Added
-
-- Add `DirectiveLocator::disable()` to make a directive unavailable, leaving its definition out of the schema https://github.com/nuwave/lighthouse/pull/2787
-
 ### Changed
 
 - Let classes passed to `DirectiveLocator::setResolved()` take precedence in `DirectiveLocator::classes()` and thus `DirectiveLocator::definitions()` https://github.com/nuwave/lighthouse/pull/2787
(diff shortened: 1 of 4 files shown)


src/Schema/DirectiveLocator.php
```diff
@@ -77,7 +79,7 @@ public function namespaces(): array
      */
     public function classes(): array
     {
-        $directives = [];
+        $directives = $this->resolvedClassnames;
```
spawnia: Why not inline and directly mutate/return `$this->resolvedClassnames`?

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+
+- Let classes passed to `DirectiveLocator::setResolved()` take precedence in `DirectiveLocator::classes()` and thus `DirectiveLocator::definitions()` https://github.com/nuwave/lighthouse/pull/2787
```
```suggestion
- Classes passed to `DirectiveLocator::setResolved()` take precedence in `DirectiveLocator::classes()` and `DirectiveLocator::definitions()` https://github.com/nuwave/lighthouse/pull/2787
```

src/Schema/DirectiveLocator.php
spawnia: Addressed in [Address review](https://github.com/nuwave/lighthouse/pull/2787/commits/d7c128382d9c59f441709d65da76afe4273e526a). The scan result is now cached in the map, so resolve() no longer walks the namespaces for anything classes() already found.

CHANGELOG.md
spawnia: Addressed in [Address review](https://github.com/nuwave/lighthouse/pull/2787/commits/d7c128382d9c59f441709d65da76afe4273e526a)

src/Schema/DirectiveLocator.php
```diff
@@ -94,11 +94,11 @@ public function classes(): array
                 }
 
                 // Only add the first directive that was found
-                $directives[self::directiveName($class)] ??= $class;
+                $this->resolvedClassnames[self::directiveName($class)] ??= $class;
```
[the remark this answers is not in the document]
spawnia: Addressed in [Pin a directive class that declares the name it is pinned to](https://github.com/nuwave/lighthouse/pull/2787/commits/ceb6b6f01ba69421422a4f04ebf20e30809e3c75) by fixing the test, not the implementation. The duplicate only appears when a name is pinned to a class that declares a different directive name, which was already broken before this change: resolve() returned a directive whose definition did not match the name it was resolved for. A well-formed override has directiveName($class) equal to the key, so the scan cannot add the class a second time.

